PR #1, opened 2020-09-12
Allocator tests

Przemog1:


devshgraphicsprogramming:
merge other examples into ex 10

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+using namespace core;
 
-/*
-	Problems with GeneralPurposeallocator:
-		1. GeneralPurposeallocator is not able to allocate as much address space as it is expected to
-		2. Program doesn't compile, when I try to call safe_shrink_size
-		3. "Assertion failed: false, file *\include\irr\core\alloc\GeneralpurposeAddressAllocator.h, line 368"
+/*
+	If traits say address allocator supports arbitrary order free,
+	then ETAT_CHOOSE_RANDOMLY is valid. Otherwise ETAT_CHOOSE_MOST_RECENTLY_ALLOCATED
```
devshgraphicsprogramming, line 11: stack allocator is "most recently allocated" but the linear allocator has `free_addr` as a NO-OP, basically the whole allocator needs to be `reset` to actually reclaim the memory.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+constexpr size_t maxAlignOffset = 64;
+constexpr size_t maxBlockSize = 600;
+
+//TODO: full static
+class RandomNumberGenerator
+{
+public:
+	inline uint32_t getRndAllocCnt()    { return  allocsPerFrameRange(mt);  }
+	inline uint32_t getRndSize()        { return  sizePerFrameRange(mt); }
+	inline uint32_t getRndMaxAlign()    { return  (1u << maxAlignmentExpPerFrameRange(mt)); } //128 is max
```
devshgraphicsprogramming, line 43: max alignment and size do not really apply to the Pool Address allocator, because there the allocations must be a fixed size (block size) so the allocation and size will be both equal to that.

You can however, test different block sizes for multiple instances of Pool Address Allocators.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+struct AllocationCreationParameters
+{	
+	size_t multiAllocCnt;              //! Specifies amount of adress to be allocated with certain choosen allocator
+	size_t adressesToDeallocateCnt;    //! Specifies amount of adress to be deallocated with certain choosen allocator. Must be less than all allocated and must pick number less than `traits::max_multi_free`, but we don't have `max_multi_free`
+};
+
+constexpr size_t minAllocs = 10000u;
+constexpr size_t maxAllocs = 20000u;
+constexpr size_t maxSizeForSquaring = 1024u;
+constexpr size_t maxAlignmentExp = 6u;
```
devshgraphicsprogramming, line 30: oooh, alignment can be muuuch bigger than that (at least 4096)

examples_tests/43.AllocatorTest/main.cpp
devshgraphicsprogramming: Also I don't think there's really a requirement for my Address Allocators to only service PoT alignments, in-fact the alignments can be any non-zero positive integer.... this is why GeneralPurposeAddressAllocator can be used to manage any format mesh vertices ;)

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+constexpr size_t minByteSizeForReservingMemory = 2048;			// 2kB
+constexpr size_t maxByteSizeForReservingMemory = 2147483648;	// 2GB
```
devshgraphicsprogramming, line 32: the address allocator's `reserved_size` tells you exactly how much reserved memory you need.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+{
+	ETAT_CHOOSE_RANDOMLY,
+	ETAT_CHOOSE_MOST_RECENTLY_ALLOCATED,
+	ETAT_COUNT
+};
+
+struct AllocationCreationParameters
+{	
+	size_t multiAllocCnt;              //! Specifies amount of adress to be allocated with certain choosen allocator
+	size_t adressesToDeallocateCnt;    //! Specifies amount of adress to be deallocated with certain choosen allocator. Must be less than all allocated and must pick number less than `traits::max_multi_free`, but we don't have `max_multi_free`
```
devshgraphicsprogramming, line 24: the number of multi_allocs and how many addresses each multi_alloc should allocate should be random.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
-				os::Printer::print(std::to_string(addr));
+	}
+	
+	//struct for holding temporary data (inputs of `multi_alloc_addr` i `multi_free_addr`)
+	struct EntryForFrameData
+	{
+		uint32_t outAddr[maxAllocs] = { ~0u };
+		uint32_t sizes[maxAllocs] = { 0u };
+		uint32_t alignments[maxAllocs] = { 0u };
+	} perFrameData;
```
devshgraphicsprogramming, line 84: correctness, not performance is the focus here, so you have my blessing to use `std::vector`

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
-			if (addr == alctr.invalid_address)
-				os::Printer::print("invalid adress");
-			else
-				os::Printer::print(std::to_string(addr));
+	}
+	
+	//struct for holding temporary data (inputs of `multi_alloc_addr` i `multi_free_addr`)
+	struct EntryForFrameData
+	{
+		uint32_t outAddr[maxAllocs] = { ~0u };
```
devshgraphicsprogramming, line 81: initialize all with `invalid_address`, the value ~0u is not that, `multi_alloc` wont work is the output array of addresses is not valid.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+		uint32_t sizes[maxAllocs] = { 0u };
+		uint32_t alignments[maxAllocs] = { 0u };
+	} perFrameData;
+	
+	void executeAllocatorTest()
+	{
+		os::Printer::log("Executing Pool Allocator Test!", ELL_INFORMATION);
+	
+		for (size_t i = 0; i < creationParameters.multiAllocCnt; ++i)
+			executeForFrame();
```
devshgraphicsprogramming, line 91: you need 3 scopes:
- random number of iteration loops (times `reset` of an allocator is performed)
- random number of multi-alloc calls
- random number of allocations in the multi-alloc

You need to handle failure to allocate gracefully.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+
+		// randomly decide the number of iterations of allocation and reset
+			// declare `core::vector` to hold allocated addresses and sizes
+
+			// randomly decide how many `multi_allocs` to do
+				// randomly decide how many allocs in a `multi_alloc` NOTE: must pick number less than `traits::max_multi_alloc`
+					// randomly decide sizes (but always less than `address_allocator_traits::max_size`)
+				// record all successful alloc addresses to the `core::vector`
+
+				// run random dealloc function
```
devshgraphicsprogramming, line 110: its important to throw in some random multi frees between every multi-alloc, this helps us test fragmentation and bookeeping of free memory.

Simply doing all allocations followed by all frees would leave bug loopholes.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+				// reset
+			// ELSE
+				// free everything with a series of multi_free
+	void executeForFrame()
+	{
+		uint32_t randMaxAlign = rng.getRndMaxAlign();
+		uint32_t randAddressSpaceSize = rng.getRndBuffSize();
+		uint32_t randAlignOffset = rng.getRandomNumber(0u, randMaxAlign - 1u);
+		uint32_t randOffset = rng.getRandomNumber(0u, randAddressSpaceSize - 1u); //?
+		uint32_t randBlockSz = rng.getRandomNumber(0u, randAddressSpaceSize - 1u); //?
```
devshgraphicsprogramming, line 125: looks good, except that I'd make the block size between 1u and `(randAddressSpaceSize-randOffset)/2u`

devshgraphicsprogramming: and assertthe end result is >0

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+
+		uint32_t allocationAmountForSingeFrameTest = rng.getRndAllocCnt();
+		
+		const auto reservedSize = PoolAddressAllocator::reserved_size(randMaxAlign, randAddressSpaceSize, randBlockSz); // 3rd parameter onward is custom for each address alloc type
+		void* reservedSpace = _IRR_ALIGNED_MALLOC(reservedSize, _IRR_SIMD_ALIGNMENT);
+		auto poolAlctr = PoolAddressAllocator(reservedSpace, randOffset, randAlignOffset, randMaxAlign, randAddressSpaceSize, randBlockSz);
+
+		// randomly decide how many `multi_allocs` to do
+		const uint32_t multiAllocCnt = rng.getRandomNumber(1u, 5u); //range?
```
devshgraphicsprogramming, line 134: as many as reasonable time-wise for the test, probably in the thousands or millions though.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+		for (uint32_t i = 0u; i < multiAllocCnt; i++)
+		{
+			// randomly decide how many allocs in a `multi_alloc` NOTE: must pick number less than `traits::max_multi_alloc`
+			const uint32_t allocCntInMultiAlloc = rng.getRandomNumber(1u, 4096u); //?
```
devshgraphicsprogramming, line 138: yeah, all good, just use the max multi alloc trait

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+
+			for (size_t i = 0u; i < allocCntInMultiAlloc; ++i)
+			{
+				// randomly decide sizes (but always less than `address_allocator_traits::max_size`)
+				perFrameData.sizes[i] = rng.getRandomNumber(1u, Traits::max_size(poolAlctr));
+				perFrameData.alignments[i] = rng.getRndMaxAlign();
```
devshgraphicsprogramming, line 144: pool allocator cant do random sizes and alignments, they always need to be block sized... other allocators can though.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+		_IRR_ALIGNED_FREE(reservedSpace);
+	}
+
+	// random dealloc function
+		// randomly decide how many calls to `multi_alloc` | ??
+			// randomly how many addresses we should deallocate (but obvs less than all allocated) NOTE: must pick number less than `traits::max_multi_free`
+				// if traits say address allocator supports arbitrary order free, then choose randomly, else choose most recently allocated
+	void randFreeAllocatedAddresses(PoolAddressAllocator& alctr)
+	{
+		//here I'm freeing previously generated addresses, is it ok?
```
devshgraphicsprogramming, line 178: yes ofc.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+			// free everything with a series of multi_free
+			while (results.size() != 0u)
+				randFreeAllocatedAddresses(poolAlctr);
+		}
+		 
+		_IRR_ALIGNED_FREE(reservedSpace);
+	}
+
+	// random dealloc function
+		// randomly decide how many calls to `multi_alloc` | ??
```
devshgraphicsprogramming, line 173: typo.. I meant multi-free

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+		const uint32_t addressesToFreeCnt = rng.getRandomNumber(0u, results.size());
+
+		for (uint32_t i = 0u; i < addressesToFreeCnt; i++)
+		{
+			perFrameData.outAddr[i] = results[i].outAddr;
+			perFrameData.sizes[i] = results[i].size;
 		}
-		_IRR_ALIGNED_FREE(resSpc);
+
+		Traits::multi_free_addr(alctr, addressesToFreeCnt, perFrameData.outAddr, perFrameData.sizes);
```
devshgraphicsprogramming, line 193: you want a random number of calls to this.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+				// if traits say address allocator supports arbitrary order free, then choose randomly, else choose most recently allocated
+	void randFreeAllocatedAddresses(PoolAddressAllocator& alctr)
+	{
+		//here I'm freeing previously generated addresses, is it ok?
+
+		if(results.size() == 0u)
+			return;
+
+		// randomly how many addresses we should deallocate (but obvs less than all allocated) NOTE: must pick number less than `traits::max_multi_free`
+		//TODO: random interval
```
devshgraphicsprogramming, line 184: you mean the traits max multi free

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+		const uint32_t addressesToFreeCnt = rng.getRandomNumber(0u, results.size());
+
+		for (uint32_t i = 0u; i < addressesToFreeCnt; i++)
+		{
+			perFrameData.outAddr[i] = results[i].outAddr;
+			perFrameData.sizes[i] = results[i].size;
```
devshgraphicsprogramming, line 190: erm actually....

You either want to:
- For a Stack allocator, add the results in reverse (FILO) and remove them from the vector.
- For all other allocator types randomly shuffle the `results`, and then do as above (right now you're always freeing in the same order they were allocated, which is not a good test). Or do some other method to **randomly choose** N items from a set of K.

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
 		}
-		_IRR_ALIGNED_FREE(resSpc);
+
+		Traits::multi_free_addr(alctr, addressesToFreeCnt, perFrameData.outAddr, perFrameData.sizes);
+		results.erase(results.begin(), results.begin() + addressesToFreeCnt);
```
devshgraphicsprogramming, line 194: The O(N^2) nature of this `erase` call will cripple us.

devshgraphicsprogramming: You need:
- a dummy (NOOP) container for linear allocator
- a FILO container for stack allocator
- a container that supports random deletion and in-order insert in O(logn) or O(1)

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -1,53 +1,219 @@
...
+
+			// record all successful alloc addresses to the `core::vector`
+			if (perFrameData.outAddr[i] != PoolAddressAllocator::invalid_address)
+				results.push_back({ perFrameData.outAddr[i], perFrameData.sizes[i], perFrameData.alignments[i] });
+
+			randFreeAllocatedAddresses(poolAlctr);
+		}
+
+		// randomly choose between reset and freeing all `core::vector` elements
+		bool reset = static_cast<bool>(rng.getRandomNumber(0u, 1u));
```
devshgraphicsprogramming, line 157: for a linear allocator, it would always need to be reset (since free is a noop)

examples_tests/43.AllocatorTest/main.cpp
Przemog1: > Also I don't think there's really a requirement for my Address Allocators to only service PoT alignments
https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/irr/core/alloc/AddressAllocatorBase.h#L73

devshgraphicsprogramming: that's max requestable that needs to be PoT, not the individuals.

examples_tests/43.AllocatorTest/main.cpp
devshgraphicsprogramming: also you need to choose the rand alignment differently than the max for the allocator
https://github.com/Devsh-Graphics-Programming/Nabla/pull/1/files/0078e583bc8f69dd8697b9a6d5eb5e35f0aaa769#r496261998

examples_tests/43.AllocatorTest/main.cpp
Przemog1: buffer size of a virtual memory comes from this range: 
<`minByteSizeForReservingMemory`, `maxByteSizeForReservingMemory`]>

names of those variables may be misleading

examples_tests/43.AllocatorTest/main.cpp
Przemog1: I can't find this function :/

examples_tests/43.AllocatorTest/main.cpp
devshgraphicsprogramming: ah ok, change the names then.

examples_tests/43.AllocatorTest/main.cpp
devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/irr/core/alloc/address_allocator_traits.h#L163

its a constexpr.

examples_tests/43.AllocatorTest/main.cpp
Przemog1: > The O(N^2) nature of this erase call will cripple us.

isn't it linear? https://en.cppreference.com/w/cpp/container/vector/erase

devshgraphicsprogramming: for a single removal its linear O(N), now do N removals on a vector of N items will give you N^2 runtime

examples_tests/43.AllocatorTest/main.cpp
```diff
@@ -269,4 +276,202 @@ int main()
...
+				os::Printer::print(std::to_string(addr));
+		}
+		_IRR_ALIGNED_FREE(resSpc);
+	}
+
+	//2
+	{
+		void* resSpc = _IRR_ALIGNED_MALLOC(core::GeneralpurposeAddressAllocator<uint32_t>::reserved_size(alignof(uint32_t), 16u, 1u), _IRR_SIMD_ALIGNMENT);
+		auto alctr = core::GeneralpurposeAddressAllocator<uint32_t>(resSpc, 0u, 0u, alignof(uint32_t), 16u, 1u);
+		alctr.safe_shrink_size(8u);
```
devshgraphicsprogramming: this code you can nuke, yours replaces it entirely.

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+#define ALLOCATOR_TEST
+//#define ALLOCATOR_TEST_OLD
+//#define ADDRESS_ALLOCATOR_TRAITS_TEST
+//#define ALLOC_PREF_TEST
+
+#ifdef ALLOCATOR_TEST
+
+struct AllocationCreationParameters
+{	
+	size_t multiAllocCnt;              //! Specifies amount of adress to be allocated with certain choosen allocator
```
devshgraphicsprogramming, line 18: call this something else its confusing, you have two multi alloc counts.

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+		if constexpr (std::is_same<AlctrType, core::LinearAddressAllocator<uint32_t>>::value)
+		{
+			alctr = AlctrType(nullptr, randAllocParams.offset, randAllocParams.alignOffset, randAllocParams.maxAlign, randAllocParams.addressSpaceSize);
+		}
+		else
+		{
+			const auto reservedSize = AlctrType::reserved_size(randAllocParams.maxAlign, randAllocParams.addressSpaceSize, randAllocParams.blockSz);
+			reservedSpace = _IRR_ALIGNED_MALLOC(reservedSize, _IRR_SIMD_ALIGNMENT);
+			alctr = AlctrType(reservedSpace, randAllocParams.offset, randAllocParams.alignOffset, randAllocParams.maxAlign, randAllocParams.addressSpaceSize, randAllocParams.blockSz);
```
devshgraphicsprogramming, line 110: for pool allocator the block size is basically a constant allocation size (can never be different)

But for the Generalpurpose allocator, its the minimum allocation size.

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+		}
+
+		// randomly decide how many `multi_allocs` to do
+		const uint32_t multiAllocCnt = rng.getRandomNumber(1u, 5u); //TODO: will change it later
+		for (uint32_t i = 0u; i < multiAllocCnt; i++)
+		{
+			outAddresses.clear();
+			sizes.clear();
+			alignments.clear();
```
devshgraphicsprogramming, line 119: this is wrong, it means you deallocate all allocations from each multi_alloc (which means you'll never multi_alloc on a fragmented allocator)

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+			// randomly decide how many allocs in a `multi_alloc` NOTE: must pick number less than `traits::max_multi_alloc`
+			constexpr uint32_t upperBound = core::address_allocator_traits<AlctrType>::maxMultiOps;
+			const uint32_t allocCntInMultiAlloc = rng.getRandomNumber(1u, upperBound);
+
+			for (size_t i = 0u; i < allocCntInMultiAlloc; ++i)
+			{
+				// randomly decide sizes (but always less than `address_allocator_traits::max_size`)
+				outAddresses.emplace_back(AlctrType::invalid_address);
```
devshgraphicsprogramming, line 128: resize + std::fill would be a good choice here

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+				if constexpr (std::is_same<AlctrType, core::PoolAddressAllocator<uint32_t>>::value)
+				{
+					sizes.emplace_back(randAllocParams.blockSz);
+					alignments.emplace_back(randAllocParams.blockSz);
+				}
+				else
+				{
+					sizes.emplace_back(rng.getRandomNumber(1u, Traits::max_size(alctr)));
+					alignments.emplace_back(rng.getRndMaxAlign());
```
devshgraphicsprogramming, line 138: for the generalpurpose allocator the size needs to be >= minimum size (block size)

For Pool allocator both size and alignment should be equal.

I dont recall any restrictions on alignment for generalpurpose allocator though.

devshgraphicsprogramming: its probably why generalpurpose crashes during the test.

examples_tests/10.AllocatorTest/main.cpp
devshgraphicsprogramming: my bad, it got this confused with `results`

examples_tests/10.AllocatorTest/main.cpp
devshgraphicsprogramming: actually you could resize all arrays to `core::address_allocator_traits<AlctrType>::maxMultiOps` in the constructor up-front (or even use a std:array)

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+	{
+		if(results.size() == 0u)
+			return;
+
+		outAddresses.clear();
+		sizes.clear();
+		alignments.clear();
+
+		// randomly decide how many calls to `multi_free`
+		const uint32_t multiFreeCnt = rng.getRandomNumber(1u, 10000u); //TODO
```
devshgraphicsprogramming, line 189: 1 to `results.size()` ?

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+		//TODO:
+		//shuffle results
+
+		for (uint32_t i = 0u; i < multiFreeCnt; i++)
+		{
+			// randomly how many addresses we should deallocate (but obvs less than all allocated) NOTE: must pick number less than `traits::max_multi_free`
+			const uint32_t addressesToFreeCnt = rng.getRandomNumber(0u, results.size()); //that should be restrained somehow, I think... so it it less likely to free all of allocated addresses
```
devshgraphicsprogramming, line 197: upper bound should be `min(core::address_allocator_traits<AlctrType>::maxMultiOps,1)`

devshgraphicsprogramming: dont restrain any more, we need to be able to hit mutli_free everything every once in a while, this could be an important edge case to test.

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+		outAddresses.clear();
+		sizes.clear();
+		alignments.clear();
+
+		// randomly decide how many calls to `multi_free`
+		const uint32_t multiFreeCnt = rng.getRandomNumber(1u, 10000u); //TODO
+
+		//TODO:
+		//shuffle results
```
devshgraphicsprogramming, line 192: only for generalpurpose.

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+			const uint32_t addressesToFreeCnt = rng.getRandomNumber(0u, results.size()); //that should be restrained somehow, I think... so it it less likely to free all of allocated addresses
+
+			for (uint32_t i = 0u; i < addressesToFreeCnt; i++)
+			{
+				outAddresses.push_back(results[i].outAddr);
+				sizes.push_back(results[i].size);
+			}
+
+			Traits::multi_free_addr(alctr, addressesToFreeCnt, outAddresses.data(), sizes.data());
+			results.erase(results.begin(), results.begin() + addressesToFreeCnt);
```
devshgraphicsprogramming, line 206: I'd rather you free the `results.end()-addressesToFreeCnt,results.end()` range because then you wont get the O(n^2) removal.

devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/1#discussion_r496272644

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,438 @@
...
+				alctr.reset();
+				results.clear();
+			}
+			else
+			{
+				// free everything with a series of multi_free
+				while (results.size() != 0u)
+					randFreeAllocatedAddresses(alctr);
+			}
+		}
```
devshgraphicsprogramming, line 138: https://github.com/Devsh-Graphics-Programming/Nabla/pull/1#discussion_r496264837

> random number of iteration loops (times reset of an allocator is performed)
random number of multi-alloc calls with random number of allocations in the multi-alloc
random number of multi-free calls (not every multi-allocation is paired 1:1 with multi-free)

examples_tests/10.AllocatorTest/main.cpp
devshgraphicsprogramming: STILL NOT DONE

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+		if constexpr (std::is_same<AlctrType, core::LinearAddressAllocator<uint32_t>>::value)
+		{
+			alctr = AlctrType(nullptr, randAllocParams.offset, randAllocParams.alignOffset, randAllocParams.maxAlign, randAllocParams.addressSpaceSize);
+		}
+		else
+		{
+			const auto reservedSize = AlctrType::reserved_size(randAllocParams.maxAlign, randAllocParams.addressSpaceSize, randAllocParams.blockSz);
+			reservedSpace = _NBL_ALIGNED_MALLOC(reservedSize, _NBL_SIMD_ALIGNMENT);
+			alctr = AlctrType(reservedSpace, randAllocParams.offset, randAllocParams.alignOffset, randAllocParams.maxAlign, randAllocParams.addressSpaceSize, randAllocParams.blockSz);
+		}
```
devshgraphicsprogramming, line 130: this should move to `executeAllocatorTest` 

because right now you're only testing `reset` at most once per allocator creation (you dont test reuse of allocator after reset)

examples_tests/43.AllocatorTest/main.cpp
devshgraphicsprogramming: > random number of times you try different parameters (buffer size, etc.) to construct a new allocator

this would be the `executeAllocatorTest` being invoked multiple times! Not done!

devshgraphicsprogramming: > random number of iteration loops (times reset of an allocator is performed)

Not done because in the function that does the allocs + reset you create a new address allocator, but not far off

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+				alctr.reset();
+				results.clear();
+			}
+			else
+			{
+				// free everything with a series of multi_free
+				while (results.size() != 0u)
+					randFreeAllocatedAddresses(alctr);
+			}
+		}
```
devshgraphicsprogramming, line 168: there's no reset or free when you use linear address allocator ....

examples_tests/10.AllocatorTest/main.cpp
```diff
@@ -0,0 +1,420 @@
...
+		 
+		if constexpr (!std::is_same<AlctrType, core::LinearAddressAllocator<uint32_t>>::value)
+			_NBL_ALIGNED_FREE(reservedSpace);
```
devshgraphicsprogramming, line 171: move that out to `executeAllocatorTest`

examples_tests/43.AllocatorTest/main.cpp
devshgraphicsprogramming: > random number of multi-alloc calls with random number of allocations in the multi-alloc

Looks good and proper.

> random number of multi-free calls (not every multi-allocation is paired 1:1 with multi-free)

Looks good and proper

examples_tests/10.AllocatorTest/main.cpp
Przemog1: it is always reseted here: https://github.com/Devsh-Graphics-Programming/Nabla/pull/1/files/e575e6a4745f77f2be5657dc667808d1a775d997#diff-fd4cbad7bd04514c211c9432df7dfbbf19bdb575c02570581349ac24ad31d2d8R150

devshgraphicsprogramming: `randFreeAllocatedAddresses` should not reset an allocator, also from what I see.. it only calls `multi_free` which for a linear allocator is a NOOP

Przemog1: there is template specialization for a linear allocator:
https://github.com/Devsh-Graphics-Programming/Nabla/pull/1/files/e575e6a4745f77f2be5657dc667808d1a775d997#diff-fd4cbad7bd04514c211c9432df7dfbbf19bdb575c02570581349ac24ad31d2d8R285

devshgraphicsprogramming: `randFreeAllocatedAddresses` should only allocate a subset of addresses... so for linear allocator that would be a "random" `reset` (with really low probability)

the true reset should happen at the end... here

PR #4, opened 2020-09-21
Added `.clang-format` flie

Przemog1:
## Formating settings

 - Access modifiers (public/protected/private) will be aligned to the beginning of the line
 - column limit of 0, which means there is no limitation on how many characters may reside on the same line (if coulum limit would be set to N, then clang format would break line (if possible), when character count in that line exceeds N)
 - no tabs
 - indent: 4 spaces
 - short ifs on a single line are not allowed, so for example:
    `if(true) return 0;`
    will be changed to:
    ```
    if(true)
        return 0;
    ```
    same for short loops
 - Allow short lambdas on a single line
 - Do not force break after template declarations
 - Always break before empty braces, so for example:
    ```
    int foo(){
        //..
    }
    ```
    will be changed to:
    ```
    int foo()
    {

    }
    ```
 - constructor initializer list may reside on a single line, no matter how many characters there are
 - always indent `case` labels (4 spaces)
 - remove consecutive empty lines, for example:
    ```
    int a = 0;


    int b = 1;
    ```
    will be changed to:
    ```
    int a = 0;

    int b = 0;
    ```
 - remove namespace indentation
 - pointers and references are always aligned to the type, for example: `T* Ptr`
 - use only LF as a line ending

**Alignment:**
 - Don't align: 
    - params after an open bracket
    - consecutive macros
    - consecutive assignemts
    - consecutive declarations
    - operands
    - consecutive comments
 - Align:
    - escaped lines to the left (multiple line macros)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,92 @@
+---
+Language: Cpp
+#BasedOnStyle: ??
+AccessModifierOffset: -4
+AlignConsecutiveMacros: false
+AlignAfterOpenBracket: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
+AlignOperands: false
+AlignTrailingComments: false
+AllowAllArgumentsOnNextLine: false
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLambdasOnASingleLine: All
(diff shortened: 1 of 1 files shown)


devshgraphicsprogramming:
> Access modifiers (public/protected/private) will be aligned to the beginning of the line
I prefer them indendented, and then anything under them indented even more.

> column limit of 0, which means there is no limitation on how many characters may reside on the same line (if coulum limit would be set to N, then clang format would break line (if possible), when character count in that line exceeds N)
128 or enough characters to fill a 1080p screen.

> Do not force break after template declarations
Please do break after templates.

> Don't align:

    params after an open bracket
    consecutive macros
    consecutive assignemts
    consecutive declarations
    operands
    consecutive comments


Explain?

devshgraphicsprogramming:
update the branch to include armageddo, address the comments above, then apply the clang format on yet another branch so we can see what it produces

AnastaZIuk:

PR #24, opened 2020-12-11
Ballot test

Hazardu:
## Description
Added a test for ballot bit count functions from `nbl/builtin/glsl/workgroup/ballot.glsl` to example 48.ArithmeticUnitTest
Additionally, after merging I encountered problems with CMeshManipulator.cpp, the template parameter count changed. Should this change be undone or is this ok?

## TODO list:
- [x] `#include <nbl/builtin/glsl/subgroup/arithmetic_portability_impl.glsl>` is causing compilation errors, despite the file being where it should be.
- [x] last time, bitCount functions were all returning 0, I need to confirm what the correct, expected output should be. I previously understood it should either return 0 or 1, as the value to be counted was a boolean (which also take values of that range), and it should sum the results of previous invocations within the workgroup.

### DIFF
--- a/examples_tests/48.ArithmeticUnitTest/main.cpp
+++ b/examples_tests/48.ArithmeticUnitTest/main.cpp
@@ -18,7 +18,7 @@ struct and
 	_NBL_STATIC_INLINE_CONSTEXPR T IdentityElement = ~0ull; // this should be a reinterpret cast
 
 	inline T operator()(T left, T right) { return left & right; }
-
+	_NBL_STATIC_INLINE_CONSTEXPR bool runOPonFirst = false;
 	_NBL_STATIC_INLINE_CONSTEXPR const char* name = "and";
 };
 template<typename T>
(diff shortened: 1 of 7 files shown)


src/nbl/asset/CMeshManipulator.cpp
```diff
@@ -1535,21 +1535,21 @@ bool CMeshManipulator::calcMaxQuantizationError(const SAttribTypeChoice& _srcTyp
 	return true;
 }
 
-core::smart_refctd_ptr<ICPUBuffer> IMeshManipulator::idxBufferFromTriangleStripsToTriangles(const void* _input, size_t _idxCount, E_INDEX_TYPE _idxType)
+core::smart_refctd_ptr<ICPUBuffer> IMeshManipulator::idxBufferFromTriangleStripsToTriangles(const void* _input, size_t _idxCount, E_INDEX_TYPE _idxType, E_INDEX_TYPE _outIndexType)
```
Crisspl, line 1538: what's going on? Why output param is pass-by-value and whatsmore its not even used??

devshgraphicsprogramming: my code, we've went over it

devshgraphicsprogramming:
> is causing compilation errors, despite the file being where it should be.

Well, no shit, because all the macros and fuctions there are prefixed with `irr` and not `nbl`

devshgraphicsprogramming:
does it pass tests?

devshgraphicsprogramming:
ok I need to fix the ballot for WG>32

PR #31, opened 2020-12-22
Start the process of fixing code for GCC

devshgraphicsprogramming:
## Description
GCC compiler is anal about syntax and less forgiving than MSVC.

## Testing 
Build with MinGW on Windows or GCC on Linux.

## TODO list:
Every error Ninja reports when building!


devshgraphicsprogramming:
@Hazardu also try to compile the library and examples with clang ;)

include/IDriverFence.h
```diff
@@ -133,7 +133,7 @@ class GPUEventWrapper : public core::Uncopyable
             return *this;
         }
 
-        template<class Clock=std::chrono::steady_clock, class Duration=typename Clock::duration>
+        template<class Clock=std::chrono::high_resolution_clock, class Duration=typename Clock::duration>
```
devshgraphicsprogramming, line 136: wont that break the MSVC build then?

include/nbl/asset/IMeshPacker.h
```diff
@@ -217,7 +217,7 @@ class IMeshPacker : public MeshPackerBase
                 MeshPackerBase::MeshPackerConfigParams<MeshIterator> configParams
                 {
                     currMeshVtxInputParams,
-                    SRange<void, MeshIterator>(sortedMeshBuffersOut, sortedMeshBuffersOut)
+                    nbl::core::SRange<void, MeshIterator>(sortedMeshBuffersOut, sortedMeshBuffersOut)
```
devshgraphicsprogramming, line 220: should only need `core::` code has the `nbl::` namespace in common

include/nbl/core/alloc/PoolAddressAllocator.h
```diff
@@ -47,9 +47,7 @@ class PoolAddressAllocator : public AddressAllocatorBase<PoolAddressAllocator<_s
 
         static constexpr bool supportsNullBuffer = true;
 
-        #define DUMMY_DEFAULT_CONSTRUCTOR PoolAddressAllocator() : blockSize(1u), blockCount(0u) {}
-        GCC_CONSTRUCTOR_INHERITANCE_BUG_WORKAROUND(DUMMY_DEFAULT_CONSTRUCTOR)
-        #undef DUMMY_DEFAULT_CONSTRUCTOR
```
devshgraphicsprogramming, line 52: this sort of fix was better, because its an actual GCC bug

include/nbl/video/CPropertyPoolHandler.h
```diff
@@ -46,7 +46,7 @@ class CPropertyPoolHandler final : public core::IReferenceCounted, public core::
 			core::SRange<uint32_t> outIndices;
 			const void* const* data; 
 		};
-		transfer_result_t addProperties(const AllocationRequest* requestsBegin, const AllocationRequest* requestsEnd, const std::chrono::steady_clock::time_point& maxWaitPoint=GPUEventWrapper::default_wait());
+		transfer_result_t addProperties(const AllocationRequest* requestsBegin, const AllocationRequest* requestsEnd, const std::chrono::high_resolution_clock::time_point& maxWaitPoint=GPUEventWrapper::default_wait());
```
devshgraphicsprogramming, line 49: will changing the type break the MSVC build?

include/nbl/core/alloc/PoolAddressAllocator.h
Crisspl: it was my idea to get rid of it. I dont get this fix, what it does is defining default ctor but only in gcc versions below 7.2 (and in fact msvc as well because the `#if` is poorly written XD -- thats why msvc build even worked), while default ctor is expected to be there always, regardless of platform/compiler/config

devshgraphicsprogramming: ah ok.. so it turns out every compiler needs it?

Crisspl: yeah, tbh i dont know what this fix is for

devshgraphicsprogramming: it seemed that MSVC can do without the default ctor

Crisspl: nah, it's just that `#if` was poorly written and it was actually creating this ctor when `__GNUC__` was not defined as well

devshgraphicsprogramming: ok

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
```diff
@@ -247,12 +247,11 @@ class CMaterialCompilerGLSLBackendCommon
...
 	#include "nbl/nblpack.h"
 		struct STextureData {
-			STextureData() = default;
 
-			VTID vtid = VTID::invalid();
+			VTID vtid;
 			union {
 				uint32_t prefetch_reg;//uint32
-				uint32_t scale = 0u;//float
+				uint32_t scale;//float
```
devshgraphicsprogramming, line 254: removing initializers will fuck it up... now you need to make a proper constructor for this struct.

include/nbl/asset/material_compiler/IR.h
```diff
@@ -299,7 +299,7 @@ class IR : public core::IReferenceCounted
                 if (found != (array+count))
                 {
                     if (ix)
-                        ix[0] = found-array;
+                        ix[0] = reinterpret_cast<size_t>(found)-reinterpret_cast<size_t>(array);
```
devshgraphicsprogramming, line 302: you probably want to use `std::distance` here

devshgraphicsprogramming:
@Hazardu waiting for lastest commits

include/nbl/asset/material_compiler/IR.h
devshgraphicsprogramming: `reinterpret_cast` IS WRONG

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
```diff
@@ -475,27 +476,28 @@ class CMaterialCompilerGLSLBackendCommon
...
 
-				union {
+				union{
 					uint64_t qword[2];
 					uint32_t dword[4];
-					struct {
+					struct{
 						STextureData tex_data;
 						uint32_t dword4;
-					} PACK_STRUCT;
```
devshgraphicsprogramming, line 498: @Crisspl wont this break the sizeof the struct?

include/matrix3x4SIMD_impl.h
```diff
@@ -217,9 +217,9 @@ inline matrix3x4SIMD& matrix3x4SIMD::setScale(const core::vectorSIMDf& _scale)
 	vectorSIMDu32& rowAlias0 = reinterpret_cast<vectorSIMDu32&>(rows[0]);
 	vectorSIMDu32& rowAlias1 = reinterpret_cast<vectorSIMDu32&>(rows[1]);
 	vectorSIMDu32& rowAlias2 = reinterpret_cast<vectorSIMDu32&>(rows[2]);
-	rowAlias0 = (scaleAlias & BUILD_MASKF(1, 0, 0, 0)) | (rowAlias0 & mask0001);
-	rowAlias1 = (scaleAlias & BUILD_MASKF(0, 1, 0, 0)) | (rowAlias1 & mask0001);
-	rowAlias2 = (scaleAlias & BUILD_MASKF(0, 0, 1, 0)) | (rowAlias2 & mask0001);
+	rowAlias0 = (scaleAlias & static_cast<vectorSIMDf>(BUILD_MASKF(1, 0, 0, 0))) | (rowAlias0 & static_cast<vectorSIMDf>(mask0001));
+	rowAlias1 = (scaleAlias & static_cast<vectorSIMDf>(BUILD_MASKF(0, 1, 0, 0))) | (rowAlias1 & static_cast<vectorSIMDf>(mask0001));
+	rowAlias2 = (scaleAlias & static_cast<vectorSIMDf>(BUILD_MASKF(0, 0, 1, 0))) | (rowAlias2 & static_cast<vectorSIMDf>(mask0001));
```
devshgraphicsprogramming, line 222: vectorSIMDf constructor from __m128 is explicit, so use explicit constructor call `vectorSIMDf(m128_value)` instead of `static_cast<vectorSIMDf>(m128_value)`.

`static_cast` is only what you want to see for PoD to PoD conversions and pointer conversions between related classes.

Explicit constructor invocation `T(u)` is not the same as a C-style cast `(T)u`

include/nbl/core/containers/refctd_dynamic_array.h
```diff
@@ -51,7 +51,9 @@ class NBL_FORCE_EBO refctd_dynamic_array : public IReferenceCounted, public dyna
 		static_assert(sizeof(base_t) == sizeof(meta_base_t), "non-CRTP and CRTP base class definitions differ in size");
 		static_assert(sizeof(meta_base_t) == sizeof(impl::dynamic_array_base<allocator,T,OverAlignmentTypes...>), "memory has been added to dynamic_array"); // TODO: fix
 
-		class NBL_FORCE_EBO fake_size_class : public IReferenceCounted, meta_base_t {};
+		class NBL_FORCE_EBO fake_size_class : public IReferenceCounted, meta_base_t {
+			using meta_base_t::operator delete;
```
devshgraphicsprogramming, line 55: nice trick, didnt know such a thing existed

include/nbl/video/alloc/ResizableBufferingAllocator.h
```diff
@@ -17,7 +17,7 @@ namespace video
 
 template<class BasicAddressAllocator, class CPUAllocator=core::allocator<uint8_t>, bool onlySwapRangesMarkedDirty = false, class CustomDeferredFreeFunctor=void >
 class ResizableBufferingAllocatorST : public core::MultiBufferingAllocatorBase<BasicAddressAllocator,onlySwapRangesMarkedDirty>,
-                                                            protected SubAllocatedDataBuffer<core::ResizableHeterogenousMemoryAllocator<core::HeterogenousMemoryAddressAllocatorAdaptor<BasicAddressAllocator,HostDeviceMirrorBufferAllocator<>,CPUAllocator> >,CustomDeferredFreeFunctor>,
+                                                            public SubAllocatedDataBuffer<core::ResizableHeterogenousMemoryAllocator<core::HeterogenousMemoryAddressAllocatorAdaptor<BasicAddressAllocator,HostDeviceMirrorBufferAllocator<>,CPUAllocator> >,CustomDeferredFreeFunctor>,
```
devshgraphicsprogramming, line 20: it was protected inheritance for a reason

include/nbl/video/alloc/StreamingTransientDataBuffer.h
```diff
@@ -20,7 +20,7 @@ namespace video
 
 
 template< typename _size_type=uint32_t, class CPUAllocator=core::allocator<uint8_t>, class CustomDeferredFreeFunctor=void >
-class StreamingTransientDataBufferST : protected SubAllocatedDataBuffer<core::HeterogenousMemoryAddressAllocatorAdaptor<core::GeneralpurposeAddressAllocator<_size_type>,StreamingGPUBufferAllocator,CPUAllocator>,CustomDeferredFreeFunctor>,
+class StreamingTransientDataBufferST : public SubAllocatedDataBuffer<core::HeterogenousMemoryAddressAllocatorAdaptor<core::GeneralpurposeAddressAllocator<_size_type>,StreamingGPUBufferAllocator,CPUAllocator>,CustomDeferredFreeFunctor>,
```
devshgraphicsprogramming, line 23: protected inheritance for a reason

include/nbl/video/alloc/StreamingTransientDataBuffer.h
```diff
@@ -97,7 +97,7 @@ class StreamingTransientDataBufferST : protected SubAllocatedDataBuffer<core::He
 
 
 template< typename _size_type=uint32_t, class CPUAllocator=core::allocator<uint8_t>, class CustomDeferredFreeFunctor=void, class RecursiveLockable=std::recursive_mutex>
-class StreamingTransientDataBufferMT : protected StreamingTransientDataBufferST<_size_type,CPUAllocator,CustomDeferredFreeFunctor>, public virtual core::IReferenceCounted
+class StreamingTransientDataBufferMT : public StreamingTransientDataBufferST<_size_type,CPUAllocator,CustomDeferredFreeFunctor>, public virtual core::IReferenceCounted
```
devshgraphicsprogramming, line 100: protected inheritance for a reason

Hazardu: the reason for changing protected to public was because a constructor from the protected class was used externally. Clang marked it as a compile time error

include/nbl/core/containers/refctd_dynamic_array.h
Hazardu: me neither

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
Hazardu: there was an assert statement few lines below, checking if the size of the struct is equal to some value; It did compile, so probably its okay

devshgraphicsprogramming: assert or static assert?

Also does it check the size of both structs?

include/nbl/core/containers/refctd_dynamic_array.h
devshgraphicsprogramming: does it work for any parent's base method?

devshgraphicsprogramming: ah it does
https://en.cppreference.com/w/cpp/language/using_declaration

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
Hazardu: `static_assert(sizeof(prefetch_instr_t) == sizeof_uvec4);`

include/nbl/video/alloc/StreamingTransientDataBuffer.h
devshgraphicsprogramming: ah yes, the `using Base::Base;` is a problem when the base is not public (when you do `using Base::method` you inherit the private/protected/public scoping)

In that case you'll need to write a perfect forwarding constructor
```cpp
template<typename... Args>
StreamingTransientDataBufferMT::StreamingTransientDataBufferMT(Args&&... args) : Base(std::forward<Args>(args)...) {}
```

and keep the Base protected as before

Hazardu: oh, i was wrong, it was not the constructor being inaccessible, it was the `operator new` of `nbl::core::impl::ResolveAlignment<nbl::core::AlignedBase<16> >` which was inherited from protected base class. The new operator is being used in `make_smart_refctd_ptr`

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
Crisspl: > @Crisspl wont this break the sizeof the struct?

it might, compiler and options dependent

devshgraphicsprogramming: @Crisspl just give us the correct asserts for size and alignment then

include/matrix3x4SIMD_impl.h
```diff
@@ -210,16 +210,16 @@ inline vectorSIMDf matrix3x4SIMD::getTranslation3D() const
...
 
 	vectorSIMDu32& rowAlias0 = reinterpret_cast<vectorSIMDu32&>(rows[0]);
 	vectorSIMDu32& rowAlias1 = reinterpret_cast<vectorSIMDu32&>(rows[1]);
 	vectorSIMDu32& rowAlias2 = reinterpret_cast<vectorSIMDu32&>(rows[2]);
-	rowAlias0 = (scaleAlias & vectorSIMDf(BUILD_MASKF(1, 0, 0, 0))) | (rowAlias0 & vectorSIMDf(mask0001));
-	rowAlias1 = (scaleAlias & vectorSIMDf(BUILD_MASKF(0, 1, 0, 0))) | (rowAlias1 & vectorSIMDf(mask0001));
-	rowAlias2 = (scaleAlias & vectorSIMDf(BUILD_MASKF(0, 0, 1, 0))) | (rowAlias2 & vectorSIMDf(mask0001));
+	rowAlias0 = (scaleAlias & vectorSIMDf(vectorSIMDu32(BUILD_MASKF(1, 0, 0, 0)))) | (rowAlias0 & vectorSIMDf(mask0001));
+	rowAlias1 = (scaleAlias & vectorSIMDf(vectorSIMDu32(BUILD_MASKF(0, 1, 0, 0)))) | (rowAlias1 & vectorSIMDf(mask0001));
+	rowAlias2 = (scaleAlias & vectorSIMDf(vectorSIMDu32(BUILD_MASKF(0, 0, 1, 0)))) | (rowAlias2 & vectorSIMDf(mask0001));
```
devshgraphicsprogramming, line 222: this should be a `reinterpret_Cast` because as it stands it will convert the integer values to floats (not reinterpret the bits)

include/nbl/video/alloc/StreamingTransientDataBuffer.h
```diff
@@ -20,7 +27,7 @@ namespace video
 
 
 template< typename _size_type=uint32_t, class CPUAllocator=core::allocator<uint8_t>, class CustomDeferredFreeFunctor=void >
-class StreamingTransientDataBufferST : public SubAllocatedDataBuffer<core::HeterogenousMemoryAddressAllocatorAdaptor<core::GeneralpurposeAddressAllocator<_size_type>,StreamingGPUBufferAllocator,CPUAllocator>,CustomDeferredFreeFunctor>,
+class StreamingTransientDataBufferST : protected SubAllocatedDataBuffer<core::HeterogenousMemoryAddressAllocatorAdaptor<core::GeneralpurposeAddressAllocator<_size_type>,StreamingGPUBufferAllocator,CPUAllocator>,CustomDeferredFreeFunctor>,
```
devshgraphicsprogramming: why is the inhericance type changed?

Hazardu: It used to be like this initially, this was me reverting it

devshgraphicsprogramming: ah ok

include/nbl/video/alloc/StreamingTransientDataBuffer.h
devshgraphicsprogramming: solved with the passhtrough

include/matrix3x4SIMD_impl.h
```diff
@@ -217,9 +217,9 @@ inline matrix3x4SIMD& matrix3x4SIMD::setScale(const core::vectorSIMDf& _scale)
 	vectorSIMDu32& rowAlias0 = reinterpret_cast<vectorSIMDu32&>(rows[0]);
 	vectorSIMDu32& rowAlias1 = reinterpret_cast<vectorSIMDu32&>(rows[1]);
 	vectorSIMDu32& rowAlias2 = reinterpret_cast<vectorSIMDu32&>(rows[2]);
-	rowAlias0 = (scaleAlias & vectorSIMDf(vectorSIMDu32(BUILD_MASKF(1, 0, 0, 0)))) | (rowAlias0 & vectorSIMDf(mask0001));
-	rowAlias1 = (scaleAlias & vectorSIMDf(vectorSIMDu32(BUILD_MASKF(0, 1, 0, 0)))) | (rowAlias1 & vectorSIMDf(mask0001));
-	rowAlias2 = (scaleAlias & vectorSIMDf(vectorSIMDu32(BUILD_MASKF(0, 0, 1, 0)))) | (rowAlias2 & vectorSIMDf(mask0001));
+	rowAlias0 = (scaleAlias & reinterpret_cast<vectorSIMDf>(vectorSIMDu32(BUILD_MASKF(1, 0, 0, 0)))) | (rowAlias0 & reinterpret_cast<vectorSIMDf>(mask0001));
+	rowAlias1 = (scaleAlias & reinterpret_cast<vectorSIMDf>(vectorSIMDu32(BUILD_MASKF(0, 1, 0, 0)))) | (rowAlias1 & reinterpret_cast<vectorSIMDf>(mask0001));
+	rowAlias2 = (scaleAlias & reinterpret_cast<vectorSIMDf>(vectorSIMDu32(BUILD_MASKF(0, 0, 1, 0)))) | (rowAlias2 & reinterpret_cast<vectorSIMDf>(mask0001));
```
devshgraphicsprogramming, line 222: when you want to cast an l-value to another type it gets a little complicated

```cpp
T something(initVal);
reinterpret_cast<const vectorSIMDf&>(something)
```

with reinterpret_cast you can only cast pointers and references

devshgraphicsprogramming:
ok just merge the lastest `master` and we can call it a day.

src/nbl/builtin/builtinHeaderGen.py
```diff
@@ -1,39 +0,0 @@
-# Creates a header file for builtin resources
```
devshgraphicsprogramming, line 1: what happened to this file?

devshgraphicsprogramming:
@Crisspl are you merging this or should I close it and delete the branch?

wrenchonline:
I failed to compile with ubuntu18.04 gcc7.5.0 ,About syntax errors.

Crisspl:
> @Crisspl are you merging this or should I close it and delete the branch?

already merged, but don't delete yet. You can close though

devshgraphicsprogramming:
> 
> 
> I failed to compile with ubuntu18.04 gcc7.5.0 ,About syntax errors. The above changes do not solve the problem.

@Hazardu sorry to drag you back from the green pastures of C# webdev, what GCC version and Ubuntu version did you use when you did the fixes? Probably something newer than 7.5.0?

@wrenchonline you tested this PR as-is or did you merge it into the current `master` and then try?

Hazardu:
> > I failed to compile with ubuntu18.04 gcc7.5.0 ,About syntax errors. The above changes do not solve the problem.
> 
> @Hazardu sorry to drag you back from the green pastures of C# webdev, what GCC version and Ubuntu version did you use when you did the fixes? Probably something newer than 7.5.0?
> 
> @wrenchonline you tested this PR as-is or did you merge it into the current `master` and then try?

yes, i think i was using gcc version 9

PR #43, opened 2021-01-18
Fixes for ex 22

Crisspl:
## Description
Merge whenever you want, however GL enum color functions need some more touch


devshgraphicsprogramming:
> however GL enum color functions need some more touch

What's with those?

PR #49, opened 2021-01-27
README update

AnastaZIuk:


devshgraphicsprogramming:
throw out some of the noisy screenshots of the Ditt office flythrough

also the original staircase screenshot (bad materials, shadows and lights)

devshgraphicsprogramming:
staircase is upside down, and the noisy flythrough gifs that converge slowly can also be removed

PR #54, opened 2021-02-04
STL/PLY corrections and fixes

AnastaZIuk:
We need to correct and fix STL & PLY loaders/writers


src/nbl/asset/IAssetManager.cpp
```diff
@@ -400,4 +423,13 @@ void IAssetManager::insertBuiltinAssets()
...
             addBuiltInToCaches(pipelineLayout, path);
     }
+
+    /*
+        Pipelines
+    */
+
+    //! Okay, an explaination is needed 
+    {
+
```
AnastaZIuk, line 433: well, refering to https://github.com/Devsh-Graphics-Programming/Nabla/blob/animation/src/nbl/asset/interchange/CPLYMeshFileLoader.cpp#L674

if I have to builtin the pipeline and insert it into cache I need to know what to do with constructor input parameters, since vertex, blend inputs (etc) are inmutable after pipeline creation. 

and if it comes to https://github.com/Devsh-Graphics-Programming/Nabla/blob/animation/src/nbl/asset/interchange/CPLYMeshFileLoader.cpp#L644

`// this metadata should probably go into pipeline layout's asset bundle (@Crisspl TODO: review)`

okay, API has changed and now I'm able to insert entire metadata upon the bundle being return, so
actually since the descriptor set layout is cached anyways I would move

```
constexpr IRenderpassIndependentPipelineMetadata::E_COMMON_SHADER_INPUT types[DS1_METADATA_ENTRY_CNT] =
		{
			IRenderpassIndependentPipelineMetadata::ECSI_WORLD_VIEW_PROJ,
			IRenderpassIndependentPipelineMetadata::ECSI_WORLD_VIEW,
			IRenderpassIndependentPipelineMetadata::ECSI_WORLD_VIEW_INVERSE_TRANSPOSE
		};
		constexpr uint32_t sizes[DS1_METADATA_ENTRY_CNT] =
		{
			sizeof(SBasicViewParameters::MVP),
			sizeof(SBasicViewParameters::MV),
			sizeof(SBasicViewParameters::NormalMat)
		};
		constexpr uint32_t relOffsets[DS1_METADATA_ENTRY_CNT] =
		{
			offsetof(SBasicViewParameters,MVP),
			offsetof(SBasicViewParameters,MV),
			offsetof(SBasicViewParameters,NormalMat)
		};
		for (uint32_t i = 0u; i < DS1_METADATA_ENTRY_CNT; ++i)
		{
			auto& semantic = (shaderInputsMetadata->end() - i - 1u)[0];
			semantic.type = types[i];
			semantic.descriptorSection.type = IRenderpassIndependentPipelineMetadata::ShaderInput::ET_UNIFORM_BUFFER;
			semantic.descriptorSection.uniformBufferObject.binding = ds1layout->getBindings().begin()[0].binding;
			semantic.descriptorSection.uniformBufferObject.set = 1u;
			semantic.descriptorSection.uniformBufferObject.relByteoffset = relOffsets[i];
			semantic.descriptorSection.uniformBufferObject.bytesize = sizes[i];
			semantic.descriptorSection.shaderAccessFlags = ICPUSpecializedShader::ESS_VERTEX;
		}
```

to **IAssetManager.cpp** and execute `placeMeta`

src/nbl/asset/interchange/CPLYMeshWriter.cpp
```diff
@@ -152,6 +152,8 @@ bool CPLYMeshWriter::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
 
     const void* ind = meshBuffer->getIndices();
     auto idxCnt = meshBuffer->getIndexCount();  // when you convert triangle Fan or triangle strip to triangle list, the index count changes, and thats what you should derive your face count from
+    // github comment
```
AnastaZIuk, line 155: todo

AnastaZIuk: as I can see in someone's code index count changes following formula `const size_t outputSize = _idxCount = (_idxCount - 2) * 3;` according to vulkan docs and it creates new buffer for Triangle Lists primitive topology. Either `idxBufferFromTrianglesFanToTriangles` and `idxBufferFromTrianglesFanToTriangles` use above formula to move vertices to a new buffer being created and `faceCount` variable is initialized as `faceCount = buf->getSize() / (idxtype == asset::EIT_16BIT ? 2u : 4u) / 3u;`, so I can see no reference to old `idxCnt`

Crisspl: `faceCount = buf->getSize() / (idxtype == asset::EIT_16BIT ? 2u : 4u) / 3u;`
seems ok to me
https://github.com/Devsh-Graphics-Programming/Nabla/pull/54/files#diff-31af0c63cf74f1d0b3003ba1c3fa9f8b57801e5ab5dd68e0450cb342bfdf7676R174

examples_tests/50A.LoaderFixes/main.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include <cstdio>
+#include <nabla.h>
+
+#include "../common/QToQuitEventReceiver.h"
+#include "nbl/ext/ScreenShot/ScreenShot.h"
+
+using namespace nbl;
+using namespace core;
+
+// #define NBL_USE_STL_LOADER // Uncomment it to use STL loader, otherwise PLY will be in use
```
devshgraphicsprogramming, line 16: use both in same same sample

Also use  number 27 for this example

examples_tests/media/ply/license.txt
```diff
@@ -0,0 +1,49 @@
+This 3D model by Artec Group inc. (www.artec3d.com) is licensed under the Creative Commons Attribution 3.0 Unported License. To view this license, visit http://creativecommons.org/licenses/by/3.0/ or send a letter to Creative Commons, 444 Castro Street, Suite 900, Mountain View, California, 94041, USA.
```
devshgraphicsprogramming, line 1: avoid spaces in filenames... call it `industrial_compressor.ply`

src/nbl/asset/IAssetManager.cpp
```diff
@@ -218,27 +218,50 @@ void IAssetManager::insertBuiltinAssets()
...
+			{   
+                "nbl/builtin/material/debug/vertex_uv/specialized_shader",
+                "nbl/builtin/material/debug/vertex_uv/specialized_shader.frag"
+            });
 		buildInGLSLShader(fileSystem->loadBuiltinData<NBL_CORE_UNIQUE_STRING_LITERAL_TYPE("nbl/builtin/material/debug/vertex_normal/specialized_shader.frag")>(),
 			asset::ISpecializedShader::ESS_FRAGMENT,
-			{"nbl/builtin/material/debug/vertex_normal/specialized_shader.frag","nbl/builtin/material/debug/vertex_color/specialized_shader.frag"});
+			{
+                "nbl/builtin/material/debug/vertex_normal/specialized_shader",
+                "nbl/builtin/material/debug/vertex_color/specialized_shader",
```
devshgraphicsprogramming, line 261: do not add cache key aliases without file extensions for specialized shaders

src/nbl/asset/IAssetManager.cpp
devshgraphicsprogramming: > if I have to builtin the pipeline and insert it into cache I need to know what to do with constructor input parameters, since vertex, blend inputs (etc) are inmutable after pipeline creation.

make several versions, only thing that can change are vertex inputs and there's a limited number of permutations.

PLY doesn't have any blend params that change depending on model.

devshgraphicsprogramming: Nothing goes into `IAssetManager.cpp`

if something is a built-in for a specific loader it either gets added to builtins from its constructor... or the `initialize` method

devshgraphicsprogramming: > okay, API has changed and now I'm able to insert entire metadata upon the bundle being returned, so
actually since the descriptor set layout is cached anyways I would move

I already moved it
https://github.com/Devsh-Graphics-Programming/Nabla/blob/animation/src/nbl/asset/interchange/IRenderpassIndependentPipelineLoader.cpp

src/nbl/asset/interchange/CPLYMeshFileLoader.cpp
```diff
@@ -606,9 +604,9 @@ bool CPLYMeshFileLoader::genVertBuffersForMBuffer(
 	{
 		constexpr std::array<std::pair<uint8_t, std::string_view>, 3> avaiableOptionsForShaders
 		{ 
-			std::make_pair(E_COL, "nbl/builtin/material/debug/vertex_color_debug_shader/specialized_shader"),
-			std::make_pair(E_UV, "nbl/builtin/material/debug/uv_debug_shader/specialized_shader"),
-			std::make_pair(E_NORM, "nbl/builtin/material/debug/normal_debug_shader/specialized_shader")  // TODO: `normal_debug` is a rather bad name
+			std::make_pair(E_COL, "nbl/builtin/material/debug/vertex_color/specialized_shader"),
+			std::make_pair(E_UV, "nbl/builtin/material/debug/vertex_uv/specialized_shader"),
+			std::make_pair(E_NORM, "nbl/builtin/material/debug/vertex_normal/specialized_shader")
```
devshgraphicsprogramming, line 609: `normal_debug_shader` -> `vertex_normal` needs matching change all around the source code and folders

src/nbl/asset/interchange/CPLYMeshWriter.cpp
devshgraphicsprogramming: input mesh can be any topology

this code doesnt account for triangle strip, triangle fan, etc.

The plethora of static methods in `IMeshManipulator` for getting polycounts and getting triangles exists for a reason!

devshgraphicsprogramming: > when you convert triangle Fan or triangle strip to triangle list, the index count changes, and thats what you should derive your face count from

src/nbl/asset/IAssetManager.cpp
AnastaZIuk: sure

src/nbl/asset/IAssetManager.cpp
AnastaZIuk: hmm, actually I will do the same like glTF does

src/nbl/asset/interchange/CPLYMeshWriter.cpp
AnastaZIuk: https://github.com/Devsh-Graphics-Programming/Nabla/pull/54/files/e09503d06963541ab68393939d21d8c44cd750a2#r571410313 and that's why it is being converted I guess, but the line with `faceCount` uses buffer with Triangle Lists topoloy for sure to calculate itself. I will use the method anyways, I didn't know about it before

Crisspl: > this code doesnt account for triangle strip, triangle fan, etc.

it does, first idx buffer is converted to triangle and from that face count is computed

devshgraphicsprogramming: the "shitshow" actually starts here
https://github.com/Devsh-Graphics-Programming/Nabla/blob/e09503d06963541ab68393939d21d8c44cd750a2/src/nbl/asset/interchange/CPLYMeshWriter.cpp#L97

basically from this point onwards its wrong because assuming triangle list topology

Crisspl: oh ok, didnt notice this, i thought we're talking about line 174

AnastaZIuk: yeah, it's a piece of shit, but it then gets updated and is fine

AnastaZIuk: anyway, I will use the method

src/nbl/asset/interchange/CPLYMeshFileLoader.cpp
AnastaZIuk: there is only one STL loader using it, but it will get updated soon when I;m done with PLY

examples_tests/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ add_subdirectory(23.Autoexposure EXCLUDE_FROM_ALL)
 #add_subdirectory(24.BufferSubDataAlignTest EXCLUDE_FROM_ALL) # TODO: @Anastazluk or @Hazardu
 #add_subdirectory(25.Blur EXCLUDE_FROM_ALL)
 add_subdirectory(26.MultidrawIndirectVSCPUCull EXCLUDE_FROM_ALL)
+add_subdirectory(27A.LoaderFixes EXCLUDE_FROM_ALL)
```
devshgraphicsprogramming, line 39: just use number 24, we dont need buffer sub align test

src/nbl/asset/interchange/CPLYMeshWriter.cpp
```diff
@@ -85,37 +85,84 @@ bool CPLYMeshWriter::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
```
devshgraphicsprogramming, line 85: either you should just save the first meshbuffer and ignore the rest, or if you're gonna keep on quitting if meshbuffer count >1 then at least log an error for the user about it.

src/nbl/asset/interchange/CPLYMeshWriter.cpp
```diff
@@ -85,37 +85,84 @@ bool CPLYMeshWriter::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
...
+        else 
+            return nullptr;
+    };
+
+    const auto cpuConvertedMeshBuffer = getConvertedMeshBufferWithNewIndexBuffer();
+    const asset::ICPUMeshBuffer* rawConvertedMeshBuffer = cpuConvertedMeshBuffer.get() ? cpuConvertedMeshBuffer.get() : *meshbuffers.begin();
+    const auto primitiveType = rawConvertedMeshBuffer->getPipeline()->getPrimitiveAssemblyParams().primitiveType;
+    const auto indexType = rawConvertedMeshBuffer->getIndexType();
+    uint32_t faceCount = {}; 
+    size_t vertexCount = {};
```
devshgraphicsprogramming, line 128: all of this can be done with 
```cpp
auto copyMeshBuffer = core::smart_refctd_ptr_static_cast<asset::ICPUMeshBuffer>(inputMeshBuffer->clone());
IMeshManipulator::homogenizePrimitiveTypeAndIndices(&copyMeshBuffer,&copyMeshBuffer+1,asset::EPT_TRIANGLE_LIST,asset::EIT_32BIT);
```

src/nbl/asset/interchange/CPLYMeshWriter.cpp
```diff
@@ -85,37 +85,84 @@ bool CPLYMeshWriter::writeAsset(io::IWriteFile* _file, const SAssetWriteParams&
...
+
+    void* indices = nullptr;
+    {
+        auto indexCount = rawConvertedMeshBuffer->getIndexCount();
+
+        indices = _NBL_ALIGNED_MALLOC(indexCount * (indexType == asset::EIT_16BIT ? 2u : 4u), _NBL_SIMD_ALIGNMENT);
+        memcpy(indices, rawConvertedMeshBuffer->getIndices(), indexCount * (indexType == asset::EIT_16BIT ? 2u : 4u));
```
devshgraphicsprogramming, line 135: why are you copying the indices?

src/nbl/asset/IAssetManager.cpp
AnastaZIuk: corrected in following commit

examples_tests/CMakeLists.txt
AnastaZIuk: corrected in following commit

src/nbl/asset/interchange/CPLYMeshWriter.cpp
AnastaZIuk: corrected in following commit

src/nbl/asset/interchange/CPLYMeshWriter.cpp
AnastaZIuk: corrected in following commit

src/nbl/asset/interchange/CPLYMeshWriter.cpp
AnastaZIuk: I wanted to just raw point to it, but the writing function takes non const pointer somehow and it modifies the indicies under `_forceFaces` bool condition, I will leave it as it is I think

AnastaZIuk: not my implementation, but it is needed probably

AnastaZIuk: that's why it was copied before I assume

src/nbl/asset/interchange/CPLYMeshWriter.cpp
AnastaZIuk: okay, after looking closer at the code, **not entire** code can be replaced with this, but this should be used inside, corrected in following commit

devshgraphicsprogramming: why not entire code?

AnastaZIuk: because it won't magically create index buffer if there is no such, and it will crash bellow upon 

```C++
if (primitiveType == asset::EPT_POINT_LIST)
        faceCount = 0u;
    else if (indices && indexType != asset::EIT_UNKNOWN)
    {
        header += "element face ";
        header += std::to_string(faceCount) + '\n';
        idxT = indexType;
        const std::string idxTypeStr = idxT == asset::EIT_32BIT ? "uint32" : "uint16";
        header += "property list uchar " + idxTypeStr + " vertex_indices\n";
    }
    else if (primitiveType == asset::EPT_TRIANGLE_LIST)
    {
        forceFaces = true;

        header += "element face ";
        header += std::to_string(faceCount) + '\n';
        idxT = vertexCount <= ((1u<<16) - 1) ? asset::EIT_16BIT : asset::EIT_32BIT;
        const std::string idxTypeStr = idxT == asset::EIT_32BIT ? "uint32" : "uint16";
        header += "property list uchar " + idxTypeStr + " vertex_indices\n";
    }
    else
        faceCount = 0u;
    header += "end_header\n";
```

It will convert index buffer, verticies or indicies, but in that case copying to `indicies`  pointer **each time** is wrong and will be leading to leaks/crashes

devshgraphicsprogramming: `homogenizePrimitiveTypeAndIndices` should "magically create an index buffer" if the input meshbuffer's index type is EIT_UNKNOWN and the index buffer binding is nullptr

AnastaZIuk: yeah
```C++
const bool iota = indexType==EIT_UNKNOWN||!cpumb->getIndexBufferBinding().buffer;

if (iota)
	correctlyOffsetIndexBufferPtr = iotaUint32Buffer->getPointer();
else
	correctlyOffsetIndexBufferPtr = cpumb->getIndices();

  /// ...
  /// conversion using correctlyOffsetIndexBufferPtr
```

devshgraphicsprogramming: so you'll always have an index buffer and it wont crash

are you even using the return value from `homogenize` ?

AnastaZIuk: it is a void call

AnastaZIuk: I mean, there will be `index` buffer, but the code I pasted above should not check `indicies` pointer after the changes with `homogenize` but index buffer binding as I can see, otherwise the code doesn't make any sense in my opinion

devshgraphicsprogramming: well obviously get the "new" index buffer!

AnastaZIuk: because it would always launch 

```C++
else if (indices && indexType != asset::EIT_UNKNOWN)
    {
        header += "element face ";
        header += std::to_string(faceCount) + '\n';
        idxT = indexType;
        const std::string idxTypeStr = idxT == asset::EIT_32BIT ? "uint32" : "uint16";
        header += "property list uchar " + idxTypeStr + " vertex_indices\n";
    }
 ```

AnastaZIuk: disscused on discord, it's all right

PR #63, opened 2021-02-08
FFT Extensions and Bloom

devshgraphicsprogramming:
## TODO list:
- [ ] Provide a nice kernel in an `.exr` file to use
- [ ] Constant smem usage and last FFT iterations in local memory


examples_tests/49.ComputeFFT/convolve.comp
```diff
@@ -0,0 +1,44 @@
...
+#version 430 core
+
+#include <nbl/builtin/glsl/math/complex.glsl>
+
+layout(local_size_x=256, local_size_y=1, local_size_z=1) in;
+ 
+struct nbl_glsl_ext_FFT_output_t
+{
+	vec2 complex_value;
+};
```
devshgraphicsprogramming, line 14: we'd usually do a "typedef" for this

in GLSL land it would be 
```glsl
#define nbl_glsl_ext_FFT_output_t vec2
```

devshgraphicsprogramming: ah and one more thing... this code is not part of the engine, its part of your example... so just how in C++ you wouldn't start putting declarations from example source code into the `nbl::` namespaces, so should you not add `nbl_glsl_` prefixes to declarations of types and functions (unless you're "overriding", i.e. providing a definition for a function already declared in the builtin GLSL "engine space")

include/nbl/builtin/glsl/math/complex.glsl
```diff
@@ -0,0 +1,35 @@
...
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+#ifndef _NBL_MATH_COMPLEX_INCLUDED_
+#define _NBL_MATH_COMPLEX_INCLUDED_
+
+#include <nbl/builtin/glsl/math/constants.glsl>
+#include <nbl/builtin/glsl/math/functions.glsl>
+
+vec2 nbl_glsl_eITheta(in float _theta)
```
devshgraphicsprogramming, line 11: I'd call it `nbl_glsl_expI` or `nbl_glsl_expImaginary`

include/nbl/builtin/glsl/math/complex.glsl
```diff
@@ -0,0 +1,35 @@
...
+    float i = rhs.x * lhs.y + rhs.y * lhs.x;
+    return vec2(r, i);
+}
+
+vec2 nbl_glsl_complex_add(in vec2 rhs, in vec2 lhs)
+{
+    return rhs + lhs;
+}
+
+vec2 nbl_glsl_complex_conjugate(in vec2 complex) {
```
devshgraphicsprogramming, line 31: nice functions... although instead of a `vec2` type I'd define a 
```glsl
#define nbl_glsl_complex vec2
#define nbl_glsl_cvec2 mat2
#define nbl_glsl_cvec3 mat3x2
#define nbl_glsl_cvec4 mat4x2
```

include/nbl/builtin/glsl/math/complex.glsl
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef _NBL_MATH_COMPLEX_INCLUDED_
+#define _NBL_MATH_COMPLEX_INCLUDED_
+
+#include <nbl/builtin/glsl/math/constants.glsl>
+#include <nbl/builtin/glsl/math/functions.glsl>
+
+vec2 nbl_glsl_eITheta(in float _theta)
+{
+    // Use sincos from math/functions.glsl?
```
devshgraphicsprogramming, line 13: probably not because `nbl_glsl_sincos` is has a comment "valid only for `theta` in [-PI,PI]"
(not sure if thats a problem anymore, but lets play it safe)

examples_tests/49.ComputeFFT/main.cpp
```diff
@@ -0,0 +1,629 @@
...
+	auto fftDescriptorSet_Src_FFT_Y = driver->createGPUDescriptorSet(core::smart_refctd_ptr<const IGPUDescriptorSetLayout>(fftPipelineLayout_SSBOInput->getDescriptorSetLayout(0u)));
+	FFTClass::updateDescriptorSet(driver, fftDescriptorSet_Src_FFT_Y.get(), fftOutputBuffer_0, fftOutputBuffer_1);
+
+	// Convolution
+	auto convolveDescriptorSet = driver->createGPUDescriptorSet(core::smart_refctd_ptr<const IGPUDescriptorSetLayout>(convolvePipelineLayout->getDescriptorSetLayout(0u)));
+	updateDescriptorSet_Convolution(driver, convolveDescriptorSet.get(), fftOutputBuffer_1, fftOutputBuffer_KernelNormalized, fftOutputBuffer_0);
+
+	// IFFT X
+	auto fftDescriptorSet_IFFT_X = driver->createGPUDescriptorSet(core::smart_refctd_ptr<const IGPUDescriptorSetLayout>(fftPipelineLayout_SSBOInput->getDescriptorSetLayout(0u)));
+	FFTClass::updateDescriptorSet(driver, fftDescriptorSet_IFFT_X.get(), fftOutputBuffer_0, fftOutputBuffer_1);
```
devshgraphicsprogramming, line 479: you can join these 3 dispatches into a single one

examples_tests/49.ComputeFFT/convolve.comp
```diff
@@ -0,0 +1,44 @@
...
+	nbl_glsl_ext_FFT_output_t out_data[];
+};
+
+layout(push_constant) uniform PushConstants
+{
+    layout (offset = 0) uvec3 padded_dimension;
+    layout (offset = 16) uint numSrcChannels;
+} pc;
+
+void main()
```
devshgraphicsprogramming, line 37: its wasteful to launch a kernel just to perform the complex multiplication of spectra

devshgraphicsprogramming: you could fold forward Y FFT together with complex multiplication and then Y iFFT

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+
+#ifndef _NBL_GLSL_EXT_FFT_PUSH_CONSTANTS_DEFINED_
+#define _NBL_GLSL_EXT_FFT_PUSH_CONSTANTS_DEFINED_
+layout(push_constant) uniform PushConstants
+{
+    layout (offset = 0) uvec3 dimension;
+    layout (offset = 16) uvec3 padded_dimension;
+	layout (offset = 32) uint direction;
+    layout (offset = 36) uint is_inverse;
+    layout (offset = 40) uint padding_type; // clamp_to_edge or fill_with_zero
```
devshgraphicsprogramming, line 55: could you go a bit easy on the memory and pack all these 3 `uint`s into a single `uint` via using all the bits?

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+#define _NBL_GLSL_EXT_FFT_DIRECTION_X_ 0
+#define _NBL_GLSL_EXT_FFT_DIRECTION_Y_ 1
+#define _NBL_GLSL_EXT_FFT_DIRECTION_Z_ 2
+
+#define _NBL_GLSL_EXT_FFT_CLAMP_TO_EDGE_ 0
+#define _NBL_GLSL_EXT_FFT_FILL_WITH_ZERO_ 1
+
+#ifndef _NBL_GLSL_EXT_FFT_PUSH_CONSTANTS_DEFINED_
+#define _NBL_GLSL_EXT_FFT_PUSH_CONSTANTS_DEFINED_
+layout(push_constant) uniform PushConstants
```
devshgraphicsprogramming, line 49: generally speaking I'd advise to create a struct
```glsl
struct nbl_glsl_ext_FFT_Parameters_t
{
    uvec3 dimension;
    uint flags;
    uvec3 padded_dimension;
    uint padding;
};
```

and requiring a `nbl_glsl_ext_FFT_Parameters_t nbl_glsl_ext_FFT_getParameters()` from the user just as we require definitions of get and set of data

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+
+#ifndef _NBL_GLSL_EXT_FFT_GET_DATA_DEFINED_
+#error "You need to define `nbl_glsl_ext_FFT_getData` and mark `_NBL_GLSL_EXT_FFT_GET_DATA_DEFINED_`!"
+#endif
+#ifndef _NBL_GLSL_EXT_FFT_SET_DATA_DEFINED_
+#error "You need to define `nbl_glsl_ext_FFT_setData` and mark `_NBL_GLSL_EXT_FFT_SET_DATA_DEFINED_`!"
+#endif
+
+// Count Leading Zeroes (naive?)
+uint clz(in uint x)
```
devshgraphicsprogramming, line 77: it could be just `31-findMSB(x)` because `findMSB` is a GLSL builtin

also sounds like a useful function to have in `builtin/math/functions.glsl`

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+       x >>= 1; 
+       count--; 
+    } 
+    reverse_num <<= count;
+    return reverse_num; 
+}
+
+uint calculate_twiddle_power(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+{
+    return (threadId & ((N / (1u << (logTwoN - iteration))) * 2 - 1)) * ((1u << (logTwoN - iteration)) / 2);;
```
devshgraphicsprogramming, line 108: please dont use integer divisions in shaders... unless your life is being threatened

Never use integer divisions if you know that the denominator is a power of two

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+	if (x == 0) { return 32; }
+    if (x <= 0x0000ffff) { n += 16; x <<= 16; }
+    if (x <= 0x00ffffff) { n +=  8; x <<= 8; }
+    if (x <= 0x0fffffff) { n +=  4; x <<= 4; }
+    if (x <= 0x3fffffff) { n +=  2; x <<= 2; }
+    if (x <= 0x7fffffff) { n++; };
+    return n;
+}
+
+uint reverseBits(in uint x)
```
devshgraphicsprogramming, line 89: already present in GLSL
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/bitfieldReverse.xhtml

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+       reverse_num <<= 1;        
+       reverse_num |= x & 1; 
+       x >>= 1; 
+       count--; 
+    } 
+    reverse_num <<= count;
+    return reverse_num; 
+}
+
+uint calculate_twiddle_power(in uint threadId, in uint iteration, in uint logTwoN, in uint N)
```
devshgraphicsprogramming, line 106: every single function you ever define in a GLSL builtin needs the correct prefix to the name (in this case `nbl_glsl_ext_FFT_`)... otherwise chaos ensues

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+    reverse_num <<= count;
+    return reverse_num; 
+}
+
+uint calculate_twiddle_power(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+{
+    return (threadId & ((N / (1u << (logTwoN - iteration))) * 2 - 1)) * ((1u << (logTwoN - iteration)) / 2);;
+}
+
+vec2 twiddle(in uint threadId, in uint iteration, in uint logTwoN, in uint N)
```
devshgraphicsprogramming, line 111: prefix

include/nbl/builtin/glsl/ext/FFT/fft.glsl
devshgraphicsprogramming: also wouldn't your life be easier if this function just took `log2N` without `N` at all?

include/nbl/builtin/glsl/ext/FFT/fft.glsl
devshgraphicsprogramming: I think this is rewritable to (under one assertion)
```glsl
const uint shiftSuffix = logTwoN -1u - iteration; // can we assert that iteration<logTwoN always??
const uint suffixMask = (2u << iteration) - 1u;
return (threadId & suffixMask) << shiftSuffix;
```

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+    return (threadId & ((N / (1u << (logTwoN - iteration))) * 2 - 1)) * ((1u << (logTwoN - iteration)) / 2);;
+}
+
+vec2 twiddle(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+{
+    uint k = calculate_twiddle_power(threadId, iteration, logTwoN, N);
+    return nbl_glsl_eITheta(-1 * 2 * nbl_glsl_PI * k / N);
+}
+
+vec2 twiddle_inv(in uint threadId, in uint iteration, in uint logTwoN, in uint N)
```
devshgraphicsprogramming, line 117: prefix

include/nbl/builtin/glsl/ext/FFT/fft.glsl
devshgraphicsprogramming: and do you need the `N`

include/nbl/builtin/glsl/ext/FFT/fft.glsl
devshgraphicsprogramming: and do you need the `N`

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+
+uint calculate_twiddle_power(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+{
+    return (threadId & ((N / (1u << (logTwoN - iteration))) * 2 - 1)) * ((1u << (logTwoN - iteration)) / 2);;
+}
+
+vec2 twiddle(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+{
+    uint k = calculate_twiddle_power(threadId, iteration, logTwoN, N);
+    return nbl_glsl_eITheta(-1 * 2 * nbl_glsl_PI * k / N);
```
devshgraphicsprogramming, line 114: instead of integer division... do a `(k>>logTwoN)`

devshgraphicsprogramming: ah ok... I see its not an integer division...

because of my confusion I'd advise you to explicitly cast floats to ints and vice-versa

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+}
+
+vec2 twiddle_inv(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+{
+    float k = calculate_twiddle_power(threadId, iteration, logTwoN, N);
+    return nbl_glsl_eITheta(2 * nbl_glsl_PI * k / N);
```
devshgraphicsprogramming, line 120: keep your code "DRY", a.k.a. "Dont Repeat Yourself"

it would be better if you defined this function as being the conjugate of `twiddle`

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+#error "_NBL_GLSL_EXT_FFT_MAX_DIM_SIZE_ should be defined."
+#endif
+
+#ifndef _NBL_GLSL_EXT_FFT_MAX_ITEMS_PER_THREAD
+#error "_NBL_GLSL_EXT_FFT_MAX_ITEMS_PER_THREAD should be defined."
+#endif
+
+
+// TODO: Investigate why +1 solves all glitches
+#define _NBL_GLSL_EXT_FFT_SHARED_SIZE_NEEDED_ _NBL_GLSL_EXT_FFT_MAX_DIM_SIZE_ + 1
```
devshgraphicsprogramming, line 24: ah yeah we still have the issue of smem usage left to address... right?

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+    return nbl_glsl_eITheta(-1 * 2 * nbl_glsl_PI * k / N);
+}
+
+vec2 twiddle_inv(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+{
+    float k = calculate_twiddle_power(threadId, iteration, logTwoN, N);
+    return nbl_glsl_eITheta(2 * nbl_glsl_PI * k / N);
+}
+
+uint getChannel()
```
devshgraphicsprogramming, line 123: I have a simpler solution `gl_WorkGroupID[pc.direction]`

devshgraphicsprogramming: also prefix

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+    } else if (pc.direction == _NBL_GLSL_EXT_FFT_DIRECTION_Y_) {
+        return gl_WorkGroupID.y;
+    } else if (pc.direction == _NBL_GLSL_EXT_FFT_DIRECTION_Z_) {
+        return gl_WorkGroupID.z;
+    } else {
+        return 0;
+    }
+}
+
+uvec3 getCoordinates(in uint tidx)
```
devshgraphicsprogramming, line 136: prefix and 
I have a simpler solution
```glsl
uvec3 tmp = gl_WorkGroupID;
tmp[pc.direction] = tidx;
return tmp;
```

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+    } else if (pc.direction == _NBL_GLSL_EXT_FFT_DIRECTION_Y_) {
+        return uvec3(gl_WorkGroupID.x, tidx, gl_WorkGroupID.z);
+    } else if (pc.direction == _NBL_GLSL_EXT_FFT_DIRECTION_Z_) {
+        return uvec3(gl_WorkGroupID.x, gl_WorkGroupID.y, tidx);
+    } else {
+        return uvec3(0,0,0);
+    }
+}
+
+uvec3 getBitReversedCoordinates(in uvec3 coords, in uint leadingZeroes)
```
devshgraphicsprogramming, line 149: prefix and get rid of the branches

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+        return uvec3(coords.x, bitReversedIndex, coords.z);
+    } else if (pc.direction == _NBL_GLSL_EXT_FFT_DIRECTION_Z_) {
+        uint bitReversedIndex = reverseBits(coords.z) >> leadingZeroes;
+        return uvec3(coords.x, coords.y, bitReversedIndex);
+    } else {
+        return uvec3(0,0,0);
+    }
+}
+
+uint getDimLength(uvec3 dimension)
```
devshgraphicsprogramming, line 165: prefix and get rid of the branches

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+            return vec2(0, 0);
+        }
+        
+        actual_coord.x = coordinate.x - min_x;
+        actual_coord.y = coordinate.y - min_y;
+        actual_coord.z = coordinate.z - min_z;
+
+    }
+    
+    return nbl_glsl_ext_FFT_getData(actual_coord, channel);
```
devshgraphicsprogramming, line 232: the user's `getData` should handle the padding and wrapping... you should just call it with out of range coodinates and expect it to live

Basically all the above should be in userspace

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+            actual_coord.y = coordinate.y - min_y;
+        }
+        
+        if (coordinate.z < min_z) {
+            actual_coord.z = 0;
+        } else if (coordinate.z > max_z) {
+            actual_coord.z = pc.dimension.z - 1;
+        } else {
+            actual_coord.z = coordinate.z - min_z;
+        }
```
devshgraphicsprogramming, line 216: the GLSL `max(coordinate,max_coord)` function should take care of all this ;)

btw your "less than minimum" branches will never be taken, because you're comparing unsigned coordinates for being less than 0

devshgraphicsprogramming: for "fill with zero" all you need is
```glsl
const uvec3 clamped_coord = min(coordinate,last_coord);
if (padding_type==FILL_WITH_ZERO && any(clamped_coord!=coordinate))
    return nbl_glsl_complexf(0.f,0.f);
return readFromSomethingAtCoord(clamped_coord,channel);
```

src/nbl/ext/FFT/FFT.cpp
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	return {bnd, bnd+sizeof(bnd)/sizeof(IGPUDescriptorSetLayout::SBinding)};
+}
+
+core::smart_refctd_ptr<video::IGPUSpecializedShader> FFT::createShader(video::IVideoDriver* driver, DataType inputType, uint32_t maxPaddedDimensionSize)
+{
+	assert(core::isPoT(maxPaddedDimensionSize));
+
+	const char* sourceFmt =
```
devshgraphicsprogramming, line 105: we avoid having GLSL in C++ source as strings like the plague

please move these to `nbl/builtin/glsl/ext/FFT/something.glsl` befrore I review

src/nbl/ext/FFT/FFT.cpp
```diff
@@ -0,0 +1,368 @@
...
+void FFT::defaultBarrier()
+{
+	COpenGLExtensionHandler::pGlMemoryBarrier(GL_SHADER_STORAGE_BARRIER_BIT);
+}
+
+// Kernel Normalization
+
+core::smart_refctd_ptr<video::IGPUSpecializedShader> FFT::createKernelNormalizationShader(video::IVideoDriver* driver)
+{
+	const char* sourceFmt =
```
devshgraphicsprogramming, line 252: we avoid having GLSL in C++ source as strings like the plague

please move these to `nbl/builtin/glsl/ext/FFT/something.glsl` befrore I review

src/nbl/ext/FFT/FFT.cpp
devshgraphicsprogramming: since its supposed to be a compute shader then it could have the `.comp` extension

src/nbl/ext/FFT/FFT.cpp
devshgraphicsprogramming: since its supposed to be a compute shader then it could have the `.comp` extension

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,328 @@
...
+    }
+    
+    return nbl_glsl_ext_FFT_getData(actual_coord, channel);
+}
+
+void nbl_glsl_ext_FFT()
+{
+    // Virtual Threads Calculation
+    uint dataLength = getDimLength(pc.padded_dimension);
+    uint num_virtual_threads = uint(ceil(float(dataLength) / float(_NBL_GLSL_EXT_FFT_BLOCK_SIZE_X_DEFINED_)));
```
devshgraphicsprogramming, line 239: how to divide integers rounding up
```glsl
uint num_virtual_threads = (dataLength-1u)/(_NBL_GLSL_EXT_FFT_BLOCK_SIZE_X_DEFINED_)+1u;
```

examples_tests/49.ComputeFFT/main.cpp
```diff
@@ -0,0 +1,629 @@
...
+		driver->bindComputePipeline(fftPipeline_SSBOInput.get());
+		driver->bindDescriptorSets(EPBP_COMPUTE, fftPipelineLayout_SSBOInput.get(), 0u, 1u, &fftDescriptorSet_IFFT_X.get(), nullptr);
+		FFTClass::pushConstants(driver, fftPipelineLayout_SSBOInput.get(), paddedDim, paddedDim, FFTClass::Direction::X, true);
+		FFTClass::dispatchHelper(driver, fftDispatchInfo_Horizontal);
+		
+		// Convolved IFFT Y
+		driver->bindComputePipeline(fftPipeline_SSBOInput.get());
+		driver->bindDescriptorSets(EPBP_COMPUTE, fftPipelineLayout_SSBOInput.get(), 0u, 1u, &fftDescriptorSet_IFFT_Y.get(), nullptr);
+		FFTClass::pushConstants(driver, fftPipelineLayout_SSBOInput.get(), paddedDim, paddedDim, FFTClass::Direction::Y, true);
+		FFTClass::dispatchHelper(driver, fftDispatchInfo_Vertical);
```
devshgraphicsprogramming, line 534: if you changed order of operation from 

```
FFT X
FFT Y
Convolve
iFFT X
iFFT Y
```
to
```
FFT X
FFT Y
Convolve
iFFT Y
iFFT X
```
then you could merge the middle 3 dispatches together

examples_tests/49.ComputeFFT/main.cpp
```diff
@@ -0,0 +1,629 @@
...
+
+		// Remove Padding and Copy to GPU Image
+		driver->bindComputePipeline(removePaddingPipeline.get());
+		driver->bindDescriptorSets(EPBP_COMPUTE, removePaddingPipelineLayout.get(), 0u, 1u, &removePaddingDescriptorSet.get(), nullptr);
+		driver->pushConstants(removePaddingPipelineLayout.get(), nbl::video::IGPUSpecializedShader::ESS_COMPUTE, 0u, sizeof(uint32_t) * 3, &paddedDim); // pc.numChannels
+		driver->pushConstants(removePaddingPipelineLayout.get(), nbl::video::IGPUSpecializedShader::ESS_COMPUTE, sizeof(uint32_t) * 4, sizeof(uint32_t) * 3, &kerDim); // numSrcChannels
+		driver->pushConstants(removePaddingPipelineLayout.get(), nbl::video::IGPUSpecializedShader::ESS_COMPUTE, sizeof(uint32_t) * 8, sizeof(uint32_t), &srcNumChannels); // numSrcChannels
+		dispatchHelper_RemovePadding(driver, removePaddingDispatchInfo);
```
devshgraphicsprogramming, line 542: by properly overriding `setData` in the last iFFT you dont need a special compute shader

examples_tests/49.ComputeFFT/remove_padding.comp
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2018-2020 - DevSH Graphics Programming Sp. z O.O.
```
devshgraphicsprogramming, line 1: as per, https://github.com/Devsh-Graphics-Programming/Nabla/pull/63/files#r572210425

I don't think you need this shader

include/nbl/builtin/glsl/ext/FFT/fft.glsl
Erfan-Ahmadi: Yep

include/nbl/ext/FFT/FFT.h
```diff
@@ -0,0 +1,240 @@
...
+			bool isInverse, 
+			PaddingType paddingType = PaddingType::CLAMP_TO_EDGE)
+		{
+			uint32_t is_inverse_u = isInverse;
+			driver->pushConstants(pipelineLayout, nbl::video::IGPUSpecializedShader::ESS_COMPUTE, 0u, sizeof(uint32_t) * 3, &inputDimension);
+			driver->pushConstants(pipelineLayout, nbl::video::IGPUSpecializedShader::ESS_COMPUTE, sizeof(uint32_t) * 4, sizeof(uint32_t) * 3, &paddedInputDimension);
+			driver->pushConstants(pipelineLayout, nbl::video::IGPUSpecializedShader::ESS_COMPUTE, sizeof(uint32_t) * 8, sizeof(uint32_t), &direction);
+			driver->pushConstants(pipelineLayout, nbl::video::IGPUSpecializedShader::ESS_COMPUTE, sizeof(uint32_t) * 9, sizeof(uint32_t), &is_inverse_u);
+			driver->pushConstants(pipelineLayout, nbl::video::IGPUSpecializedShader::ESS_COMPUTE, sizeof(uint32_t) * 10, sizeof(uint32_t), &paddingType);
+		}
```
devshgraphicsprogramming, line 210: If you defined a `nbl_glsl_ext_FFT_Parameters_t` in the GLSL in its own little header (lets say `nbl/builtin/glsl/ext/FFT/parameters.glsl`), you could include it in this C++ header just so

```glsl
namespace impl
{

using uint = uint32_t;
struct uvec3
{
    uint x,y,z;
};
#include "nbl/builtin/glsl/ext/FFT/parameters.glsl"

}
struct Parameters_t : impl::nbl_glsl_ext_FFT_Parameters_t
{
    // add some methods for C++ only, but no members
};
```

Then you could use this struct throughout the codebase and use only one `pushConstants` call and not calculate your offsets and sizes manually

include/nbl/ext/FFT/FFT.h
```diff
@@ -0,0 +1,240 @@
...
+		};
+
+		// returns dispatch size and fills the uniform data
+		static inline DispatchInfo_t buildParameters(
+			asset::VkExtent3D const & paddedInputDimensions,
+			Direction direction,
+			uint32_t num_channels)
+		{
+			assert(num_channels > 0);
+			assert(core::isPoT(paddedInputDimensions.width) && core::isPoT(paddedInputDimensions.height) && core::isPoT(paddedInputDimensions.depth));
```
devshgraphicsprogramming, line 57: move the dimension padding function to `FFT` and pad the dimension for the user!

include/nbl/ext/FFT/FFT.h
```diff
@@ -0,0 +1,240 @@
...
+			auto pcRange = getDefaultPushConstantRanges();
+			auto bindings = getDefaultBindings(driver, inputType);
+			return driver->createGPUPipelineLayout(
+				pcRange.begin(),pcRange.end(),
+				driver->createGPUDescriptorSetLayout(bindings.begin(),bindings.end()),nullptr,nullptr,nullptr
+			);
+		}
+		
+		//
+		static inline size_t getOutputBufferSize(asset::VkExtent3D const & paddedInputDimensions, uint32_t numChannels)
```
devshgraphicsprogramming, line 103: again, just pad the dimensions for the user.

include/nbl/ext/FFT/FFT.h
```diff
@@ -0,0 +1,240 @@
...
+{
+namespace ext
+{
+namespace FFT
+{
+
+class FFT : public core::TotalInterface
+{
+	public:
+
```
devshgraphicsprogramming, line 22: I'd probably add an option whether to use full `vec2` for storage of the intermediate data in the SSBO, or a packed half float (`pack2x16Half` and similar glsl builtins)

src/nbl/ext/FFT/FFT.cpp
```diff
@@ -0,0 +1,368 @@
...
+			sizeof(uint32_t) * 9,
+			sizeof(uint32_t)
+		},
+		{
+			ISpecializedShader::ESS_COMPUTE,
+			sizeof(uint32_t) * 10,
+			sizeof(uint32_t)
+		},
+	};
+	return {ranges, ranges+5};
```
devshgraphicsprogramming, line 45: this could all be just one range given by `sizeof(Parameters_t)`

src/nbl/ext/FFT/FFT.cpp
```diff
@@ -0,0 +1,368 @@
...
+	
+	if (!sampler)
+	{
+		IGPUSampler::SParams params =
+		{
+			{
+				ISampler::ETC_CLAMP_TO_EDGE,
+				ISampler::ETC_CLAMP_TO_EDGE,
+				ISampler::ETC_CLAMP_TO_EDGE,
+				ISampler::ETBC_FLOAT_OPAQUE_BLACK,
```
devshgraphicsprogramming, line 86: you could set `ETBC_FLOAT_TRANSPARENT_BLACK ` as the border colour and do conditional `fillWithZero ? ETC_CLAMP_TO_BORDER:ETC_CLAMP_TO_EDGE`

src/nbl/ext/FFT/FFT.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+vec2 nbl_glsl_ext_FFT_getData(in uvec3 coordinate, in uint channel)
+{
+	vec2 retValue = vec2(0, 0);
+#if USE_SSBO_FOR_INPUT > 0
+	uvec3 dimension = pc.dimension;
+	uint index = channel * (dimension.x * dimension.y * dimension.z) + coordinate.z * (dimension.x * dimension.y) + coordinate.y * (dimension.x) + coordinate.x;
+	retValue = inData[index].complex_value;
+#else
+	vec4 texelValue= texelFetch(inputImage, ivec2(coordinate.xy), 0);
```
devshgraphicsprogramming, line 198: does `texelFetch` respect sampler modes (when the unnormalized coords are outside the texture range)?

Please find the answer as to what happens, in the GLSL spec and paste here.

src/nbl/ext/FFT/FFT.cpp
```diff
@@ -0,0 +1,368 @@
...
+void main()
+{
+	nbl_glsl_ext_FFT();
+}
+
+)===";
+
+	const size_t extraSize = 32 + 32 + 32 + 32;
+
+	const uint32_t maxItemsPerThread = core::ceil(float(maxPaddedDimensionSize) / DEFAULT_WORK_GROUP_X_DIM);
```
devshgraphicsprogramming, line 220: dont use floats to achieve roundup.... also if we optimized the smem usage then we'd probably not need to give the max number of items per thread at all

src/nbl/ext/FFT/FFT.cpp
```diff
@@ -0,0 +1,368 @@
...
+}
+
+// Kernel Normalization
+
+core::smart_refctd_ptr<video::IGPUSpecializedShader> FFT::createKernelNormalizationShader(video::IVideoDriver* driver)
+{
+	const char* sourceFmt =
+R"===(#version 430 core
+
+layout(local_size_x=%u, local_size_y=1, local_size_z=1) in;
```
devshgraphicsprogramming, line 255: no need to vary local size, always keep it 256

src/nbl/ext/FFT/FFT.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+layout(set=0, binding=0) restrict readonly buffer InBuffer
+{
+	nbl_glsl_ext_FFT_output_t in_data[];
+};
+
+layout(set=0, binding=1) restrict buffer OutBuffer
+{
+	nbl_glsl_ext_FFT_output_t out_data[];
+};
```
devshgraphicsprogramming, line 270: a small note... you could do this in-place (normalize a buffer without having a copy) iby using the driver's GPU to GPU copy ability to copy the 8 bytes of `in_data[0]` to a temporary buffer used as an UBO in the shader and issue a barrier between the small copy and dispatch

Then you'd only have one descriptor in the set and no double copies of the buffer

devshgraphicsprogramming:
Saved reviewing 
`examples_tests/49.ComputeFFT/main.cpp`
and 
`nbl_glsl_ext_FFT`
for later

src/nbl/ext/FFT/FFT.cpp
Erfan-Ahmadi: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/texelFetch.xhtml

It isn't mentioned in the OpenGL Reference explicitly. 
but the input is "sampler" which I believe Vulkan calls "CombinedImageSampler".
I may be wrong here but then it implicitly should honor the properties of the sampler "combined" with the image.

include/nbl/builtin/glsl/ext/FFT/fft.glsl
Erfan-Ahmadi: double-checked, it's correct :)

include/nbl/ext/FFT/FFT.h
Erfan-Ahmadi: For example the pushConstant(helper) for FFT has 2 distinct parameters named "inputDimension" and "paddedInputDimension"

I cannot simply get the paddedDimension from inputDimension  because kernelSize might differ because paddedDimension is not just nextPOT it's also considering the kernel's dimensions.

Just like in the shader I believe padding needs to be done by the example/user.

What do you think? maybe there is a better way that I'm missing.

Erfan-Ahmadi: @devshgraphicsprogramming

include/nbl/builtin/glsl/ext/FFT/fft.glsl
devshgraphicsprogramming: ok we should cook something up for that for the next review.

src/nbl/ext/FFT/FFT.cpp
devshgraphicsprogramming: Let m find that for you

> The integer coordinates (i; j; k) passed to texelFetch are used to pointsample
the texture image. The level of detail accessed is computed by adding the
specified level-of-detail parameter lod to the base level of the texture, levelbase.
Texel fetch proceeds similarly to the steps described for texture access in section
11.1.3.5, with the exception that none of the operations controlled by sampler
object state are performed, including:
 level of detail clamping;
 texture wrap mode application;
 filtering (however, a mipmapped minification filter is required to access any
level of detail other than the base level);
 depth comparison.
The steps that are performed are:
 validation of texel coordinates as described below, including the computed
level-of-detail, (i; j; k), the specified level for array textures, and texture
completeness;
 sRGB conversion of fetched values as described in section 8.24;
 conversion to base color Cb;
 component swizzling.
The results of texelFetch builtins are undefined if any of the following
conditions hold:
 the computed level of detail is less than the texture’s base level (levelbase) or
greater than the maximum defined level, q (see section 8.14.3)
 the computed level of detail is not the texture’s base level and the texture’s
minification filter is NEAREST or LINEAR
 the layer specified for array textures is negative, or greater than or equal to
the number of layers in the array texture
the texel coordinates (i; j; k) refer to a texel outside the defined extents of
the computed level of detail, where any of
i < 0 i  ws
j < 0 j  hs
k < 0 k  ds
and ws, hs, and ds refer to the width, height, and depth of the image, as
defined in section 8.5.3.

devshgraphicsprogramming: You cannot use `texelFetch` when you expect wrapping.

Use `textureLod` everywhere instead

examples_tests/49.ComputeFFT/convolve.comp
```diff
@@ -8,24 +8,21 @@
 
 layout(local_size_x=256, local_size_y=1, local_size_z=1) in;
  
-struct nbl_glsl_ext_FFT_output_t
-{
-	vec2 complex_value;
-};
+ #define complex_value vec2
```
devshgraphicsprogramming, line 11: I see you use `complex_value` in multiple shaders, so its a good idea to have a `nbl_glsl_cvec2` and other definitions in `include/nbl/builtin/glsl/math/complex.glsl `

devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/63#discussion_r572151748

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -66,258 +66,161 @@ vec2 nbl_glsl_ext_FFT_getData(in uvec3 coordinate, in uint channel);
...
 #ifndef _NBL_GLSL_EXT_FFT_SET_DATA_DEFINED_
 #error "You need to define `nbl_glsl_ext_FFT_setData` and mark `_NBL_GLSL_EXT_FFT_SET_DATA_DEFINED_`!"
 #endif
+#ifndef _NBL_GLSL_EXT_FFT_GET_PADDED_DATA_DEFINED_
+#error "You need to define `nbl_glsl_ext_FFT_getPaddedData` and mark `_NBL_GLSL_EXT_FFT_GET_PADDED_DATA_DEFINED_`!"
+#endif
 
 // Count Leading Zeroes (naive?)
-uint clz(in uint x) 
+uint nbl_glsl_ext_FFT_clz(in uint x)
```
devshgraphicsprogramming, line 85: move to `builtin/math/functions.glsl` and remove `_ext_FFT` from the prefix

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -66,258 +66,161 @@ vec2 nbl_glsl_ext_FFT_getData(in uvec3 coordinate, in uint channel);
...
-    if (x <= 0x00ffffff) { n +=  8; x <<= 8; }
-    if (x <= 0x0fffffff) { n +=  4; x <<= 4; }
-    if (x <= 0x3fffffff) { n +=  2; x <<= 2; }
-    if (x <= 0x7fffffff) { n++; };
-    return n;
+    return 31u - findMSB(x);
 }
 
-uint reverseBits(in uint x)
+uint nbl_glsl_ext_FFT_reverseBits(in uint x)
```
devshgraphicsprogramming, line 90: just use `bitfieldReverse` in your code, why make a pointless wrapper function?

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -66,258 +66,161 @@ vec2 nbl_glsl_ext_FFT_getData(in uvec3 coordinate, in uint channel);
...
+    return (threadId & suffixMask) << shiftSuffix;
 }
 
-vec2 twiddle(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+nbl_glsl_complex nbl_glsl_ext_FFT_twiddle(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
 {
-    uint k = calculate_twiddle_power(threadId, iteration, logTwoN, N);
-    return nbl_glsl_eITheta(-1 * 2 * nbl_glsl_PI * k / N);
+    uint k = nbl_glsl_ext_FFT_calculateTwiddlePower(threadId, iteration, logTwoN);
+    return nbl_glsl_expImaginary(-1 * 2 * nbl_glsl_PI * k / N);
```
devshgraphicsprogramming, line 105: make it clear what are types of the literal constants

also you're got `k/N` there which looks like integer division, its misleading because if you had integer division there, your code wouldn't work at all

devshgraphicsprogramming: see my previous comment about not having implicit casts

Also you don't need `N` in the function arguments

include/nbl/builtin/glsl/ext/FFT/normalization.comp
```diff
@@ -0,0 +1,27 @@
+#version 430 core
+
+#ifndef _NBL_GLSL_EXT_FFT_NORMALIZATION_INCLUDED_
+#define _NBL_GLSL_EXT_FFT_NORMALIZATION_INCLUDED_
+
+layout(local_size_x=256, local_size_y=1, local_size_z=1) in;
+ 
+ #define complex_value vec2
```
devshgraphicsprogramming, line 8: use the definition of `nbl_glsl_cvec2` from the complex header

examples_tests/49.ComputeFFT/remove_padding.comp
devshgraphicsprogramming: this file should be gone

include/nbl/builtin/glsl/ext/FFT/default_compute_fft.comp
```diff
@@ -0,0 +1,127 @@
...
+#define _NBL_GLSL_EXT_FFT_BLOCK_SIZE_Y_DEFINED_ 1
+#endif
+
+#ifndef _NBL_GLSL_EXT_FFT_BLOCK_SIZE_Z_DEFINED_
+#define _NBL_GLSL_EXT_FFT_BLOCK_SIZE_Z_DEFINED_ 1
+#endif
+
+#define _NBL_GLSL_WORKGROUP_SIZE_ (_NBL_GLSL_EXT_FFT_BLOCK_SIZE_X_DEFINED_*_NBL_GLSL_EXT_FFT_BLOCK_SIZE_Y_DEFINED_*_NBL_GLSL_EXT_FFT_BLOCK_SIZE_Z_DEFINED_)
+
+layout(local_size_x=_NBL_GLSL_EXT_FFT_BLOCK_SIZE_X_DEFINED_, local_size_y=_NBL_GLSL_EXT_FFT_BLOCK_SIZE_Y_DEFINED_, local_size_z=_NBL_GLSL_EXT_FFT_BLOCK_SIZE_Z_DEFINED_) in;
```
devshgraphicsprogramming, line 24: you should always run a `WG_SIZEx1x1` workgroup to perform an FFT there's no point to non-1D shapes

include/nbl/builtin/glsl/ext/FFT/default_compute_fft.comp
```diff
@@ -0,0 +1,127 @@
...
+
+ 
+#define _NBL_GLSL_EXT_FFT_GET_DATA_DEFINED_
+#define _NBL_GLSL_EXT_FFT_SET_DATA_DEFINED_
+#define _NBL_GLSL_EXT_FFT_GET_PADDED_DATA_DEFINED_
```
devshgraphicsprogramming, line 29: where are the definitions?

include/nbl/builtin/glsl/ext/FFT/default_compute_fft.comp
```diff
@@ -0,0 +1,127 @@
...
+#endif
+
+// Get/Set Data Function
+
+nbl_glsl_complex nbl_glsl_ext_FFT_getData(in uvec3 coordinate, in uint channel)
+{
+	vec2 retValue = vec2(0, 0);
+#if USE_SSBO_FOR_INPUT > 0
+	uvec3 dimension = pc.dimension;
+	uint index = channel * (dimension.x * dimension.y * dimension.z) + coordinate.z * (dimension.x * dimension.y) + coordinate.y * (dimension.x) + coordinate.x;
```
devshgraphicsprogramming, line 92: ok here you better repeat or clamp the `coordinate` according to the paratemeters, cause this aint a sampler

devshgraphicsprogramming: ah ok I see the clamping happens in `nbl_glsl_ext_FFT_getPaddedData` but you could skip it for textures, and just use `textureLod` + the correct sampler and border color
https://github.com/Devsh-Graphics-Programming/Nabla/pull/63#discussion_r572230516

include/nbl/ext/FFT/FFT.h
devshgraphicsprogramming: > I cannot simply get the paddedDimension from inputDimension because kernelSize might differ because paddedDimension is not just nextPOT it's also considering the kernel's dimensions.

Provide a `static` function for computing the padded input dimension from input dimension, and make this function take a parameter `uint32_t workgroupSize` (assume workgroup is 1D) and use the static.

> Addding parameters like "sourceImageSize", "kernelImageSize", "minSize" in FFTClass would be out of scope of an extension that has the job only of computing FFT.

did I tell you to add these? are you refering to the idea of `nbl_glsl_ext_FFT_Parameters_t` ?

src/nbl/ext/FFT/FFT.cpp
devshgraphicsprogramming: Could save a lot of memory in lets say the blooming of 8k HDR images (512mb per padded frame) because the kernel would need to be padded to next PoT size of 8k too

examples_tests/49.ComputeFFT/remove_padding.comp
Erfan-Ahmadi: Yea I know. I haven't started on that yet
some shaders need to get merged

include/nbl/builtin/glsl/ext/FFT/default_compute_fft.comp
Erfan-Ahmadi: default_compute_fft.comp, can you update the PR or give me access somehow? 
There is more commits not showing up here

examples_tests/49.ComputeFFT/remove_padding.comp
Erfan-Ahmadi: also please give me access to update the PR with the latest commits @devshgraphicsprogramming

examples_tests/49.ComputeFFT/fft_convolve_ifft.comp
```diff
@@ -0,0 +1,96 @@
...
+	
+    barrier();
+    memoryBarrierShared();
+
+	convolve();	// outData+kerData->inData
+	
+    barrier();
+    memoryBarrierShared();
+
+	nbl_glsl_ext_FFT(!nbl_glsl_ext_FFT_getIsInverse()); // inData->outData
```
devshgraphicsprogramming, line 95: tip, use local variables, declare them in global scope so `getData` and `setData` fetch from the thread-local variables

handle the pre-loading of the local variables and saving in the `main()` function

This is what your local variable could look like.
```glsl
float dataStorage[NBL_GLSL_EXT_FFT_AXIS_LENGTH/_NBL_GLSL_WORKGROUP_SIZE_][NBL_GLSL_EXT_FFT_CHANNEL_COUNT];
```

P.S. the result of `NBL_GLSL_EXT_FFT_AXIS_LENGTH/_NBL_GLSL_WORKGROUP_SIZE_` would need to be rounded up to a power of two.

devshgraphicsprogramming: hmm actually for this kernel you could have separate channels in separate threads/workgroups because the input/ouput is an SSBO

devshgraphicsprogramming: but I'd make `nbl_glsl_ext_FFT` take a `uint channelIndex` parameter

examples_tests/49.ComputeFFT/fft_convolve_ifft.comp
```diff
@@ -0,0 +1,96 @@
...
+{
+	nbl_glsl_complex kerData[];
+};
+
+// Output Descriptor
+
+layout(set=0, binding=2) restrict buffer OutputBuffer
+{
+	nbl_glsl_complex outData[];
+};
```
devshgraphicsprogramming, line 31: if you have the same invocation responsible for each element and use `barrier` to sync the entire axis, then you can just use a single SSBO as both input and output

examples_tests/49.ComputeFFT/main.cpp
```diff
@@ -519,18 +519,18 @@ int main()
...
 		FFTClass::pushConstants(driver, fftPipelineLayout_SSBOInput.get(), paddedDim, paddedDim, FFTClass::Direction::Y, true);
 		FFTClass::dispatchHelper(driver, fftDispatchInfo_Vertical);
 
+		// Convolved IFFT X
+		driver->bindComputePipeline(fftPipeline_SSBOInput.get());
+		driver->bindDescriptorSets(EPBP_COMPUTE, fftPipelineLayout_SSBOInput.get(), 0u, 1u, &fftDescriptorSet_IFFT_X.get(), nullptr);
+		FFTClass::pushConstants(driver, fftPipelineLayout_SSBOInput.get(), paddedDim, paddedDim, FFTClass::Direction::X, true);
+		FFTClass::dispatchHelper(driver, fftDispatchInfo_Horizontal);
+		
 		// Remove Padding and Copy to GPU Image
```
devshgraphicsprogramming, line 534: https://github.com/Devsh-Graphics-Programming/Nabla/pull/63#discussion_r572210425

include/nbl/builtin/glsl/ext/FFT/default_compute_fft.comp
```diff
@@ -78,41 +56,54 @@ struct nbl_glsl_ext_FFT_output_t
...
-	vec2 retValue = vec2(0, 0);
+	nbl_glsl_complex retValue = nbl_glsl_complex(0, 0);
 #if USE_SSBO_FOR_INPUT > 0
-	uvec3 dimension = pc.dimension;
+	uvec3 dimension = nbl_glsl_ext_FFT_getDimensions();
 	uint index = channel * (dimension.x * dimension.y * dimension.z) + coordinate.z * (dimension.x * dimension.y) + coordinate.y * (dimension.x) + coordinate.x;
-	retValue = inData[index].complex_value;
+	retValue = inData[index];
 #else
 	vec4 texelValue= texelFetch(inputImage, ivec2(coordinate.xy), 0);
```
devshgraphicsprogramming, line 84: https://github.com/Devsh-Graphics-Programming/Nabla/pull/63#discussion_r575705439

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -81,71 +79,86 @@ vec2 nbl_glsl_ext_FFT_getPaddedData(in uvec3 coordinate, in uint channel);
...
+    return (params.dimension.w >> 16) & 0x000000ff;
+}
+bool nbl_glsl_ext_FFT_getIsInverse() {
+    nbl_glsl_ext_FFT_Parameters_t params = nbl_glsl_ext_FFT_getParameters();
+    return bool((params.dimension.w >> 8) & 0x000000ff);
+}
+uint nbl_glsl_ext_FFT_getPaddingType() {
+    nbl_glsl_ext_FFT_Parameters_t params = nbl_glsl_ext_FFT_getParameters();
+    return (params.dimension.w) & 0x000000ff;
 }
```
devshgraphicsprogramming: these I'd more to the parameters include and make them "methods" of the struct (just mangle the name)
`nbl_glsl_ext_FFT_Parameters_t_getDirection(nbl_glsl_ext_FFT_Parameters_t params)`

Its kinda bad when something has a custom bit encoding of various flags and properties and the "how they're packed" is not next to definition of the datatype/variable that holds them

include/nbl/builtin/glsl/ext/FFT/parameters.glsl
```diff
@@ -0,0 +1,14 @@
...
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+#ifndef _NBL_GLSL_EXT_FFT_PARAMETERS_INCLUDED_
+#define _NBL_GLSL_EXT_FFT_PARAMETERS_INCLUDED_
+
+struct nbl_glsl_ext_FFT_Parameters_t
+{
+    uvec4   dimension; // settings packed into the w component
+    uvec3   padded_dimension;
```
devshgraphicsprogramming, line 11: btw even under `std140` rules if you define it as the following
```glsl
uvec3 dimension;
uint flags;
uvec3 padded_dimension;
```
the data layout won't change

include/nbl/ext/FFT/FFT.h
```diff
@@ -9,24 +9,37 @@
...
 namespace ext
 {
 namespace FFT
 {
 
+typedef uint32_t uint;
+struct uvec3 {
+	uint x,y,z;
+};
+struct uvec4 {
```
devshgraphicsprogramming, line 24: add `alignas(16)` after `struct` in both cases

include/nbl/ext/FFT/FFT.h
```diff
@@ -9,24 +9,37 @@
...
+	uint x,y,z,w;
+};
+#include "nbl/builtin/glsl/ext/FFT/parameters.glsl";
+
 class FFT : public core::TotalInterface
 {
 	public:
+		struct Parameters_t : nbl_glsl_ext_FFT_Parameters_t {
```
devshgraphicsprogramming, line 32: because of alignment to 16 bytes, I'd add a `uint padding` member to the GLSL struct.

include/nbl/builtin/glsl/ext/FFT/default_compute_fft.comp
Erfan-Ahmadi: I actually tried textureLod, it worked with the "clamp_to_border" sampler as expected with dark edges.
I used the "prev_texture_coordinates/textureSize" as the coordinates to load from the image
but "clamp_to_edge" sampler just gave me a fully black coloured result. 
I didn't investigate further. but I'll test that again after everything else is set up.

examples_tests/49.ComputeFFT/remove_padding.comp
devshgraphicsprogramming: > weirdly I don't see my recent commits here. but It should appear

you didn't push probably?

No magical settings have changed.

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -81,71 +79,86 @@ vec2 nbl_glsl_ext_FFT_getPaddedData(in uvec3 coordinate, in uint channel);
...
     const uint suffixMask = (2u << iteration) - 1u;
     return (threadId & suffixMask) << shiftSuffix;
 }
 
-nbl_glsl_complex nbl_glsl_ext_FFT_twiddle(in uint threadId, in uint iteration, in uint logTwoN, in uint N) 
+nbl_glsl_complex nbl_glsl_ext_FFT_twiddle(in uint threadId, in uint iteration, in uint logTwoN) 
 {
     uint k = nbl_glsl_ext_FFT_calculateTwiddlePower(threadId, iteration, logTwoN);
-    return nbl_glsl_expImaginary(-1 * 2 * nbl_glsl_PI * k / N);
+    return nbl_glsl_expImaginary(-1.0f * 2.0f * nbl_glsl_PI * float(k) / (1 << logTwoN));
```
devshgraphicsprogramming, line 92: you still have an implicit int to float conversion here, because the left hand side to your `/` operator is uint

include/nbl/builtin/glsl/ext/FFT/parameters.glsl
Erfan-Ahmadi: I tested that first and didn't work, I completely expected it to work. 
When I used renderdoc the data was just weirdly wrong

devshgraphicsprogramming: might have been caused by https://github.com/Devsh-Graphics-Programming/Nabla/pull/63#discussion_r576138293

or an engine bug @Crisspl ?

Crisspl: this struct is used in some buffer or push constants or? anyway i dont see a chance for engine bug here

devshgraphicsprogramming:
Also check if the last half of the FFT output array is always 0, because AFAIK performing an DFFT of a real sequence should yield a complex sequence of half the size

include/nbl/builtin/glsl/ext/FFT/fft.glsl
Erfan-Ahmadi: Do you mean the right-hand side?

devshgraphicsprogramming: yeah sorry, got distracted.

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,244 @@
...
+    nbl_glsl_complex shuffled_values[_NBL_GLSL_EXT_FFT_MAX_ITEMS_PER_THREAD];
+
+    // Load Initial Values into Local Mem (bit reversed indices)
+    for(uint t = 0u; t < num_virtual_threads; t++)
+    {
+        uint tid = thread_offset + t * _NBL_GLSL_EXT_FFT_BLOCK_SIZE_X_DEFINED_;
+        uvec3 coords = nbl_glsl_ext_FFT_getCoordinates(tid);
+        uvec3 bitReversedCoords = nbl_glsl_ext_FFT_getBitReversedCoordinates(coords, leadingZeroes);
+
+        current_values[t] = nbl_glsl_ext_FFT_getPaddedData(bitReversedCoords, channel);
```
devshgraphicsprogramming, line 174: when faced a choice between a scattered access of faster memory vs. scattered access to slower memory always choose the former.

So read `getPaddedData` as much "in order" as you can, and write `current_values` out of order

devshgraphicsprogramming: I'm talking about you reading the padded data with `bitReversedCoords` which will kill any caching/coalescing.

if would be best to read it with `tid` and then perform a `nbl_glsl_workgroupShuffle(readValue[j],nbl_glsl_ext_FFT_getBitReversedCoordinate(axisCoordinate,leadingZeroes))` for `j in 0,1`

examples_tests/49.ComputeFFT/fft_convolve_ifft.comp
```diff
@@ -13,23 +13,16 @@ layout(local_size_x=_NBL_GLSL_WORKGROUP_SIZE_, local_size_y=1, local_size_z=1) i
 
 // Input Descriptor
 
-layout(set=0, binding=0) readonly restrict buffer InputBuffer
+layout(set=0, binding=0) buffer InputOutputBuffer
```
devshgraphicsprogramming, line 16: you can keep the `restrict` keyword, it means "no aliasing" that's all

include/nbl/builtin/glsl/ext/FFT/fft.glsl
Erfan-Ahmadi: Awesome point, will do that!
thanks for letting me know

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,244 @@
...
+#include <nbl/builtin/glsl/workgroup/shared_arithmetic.glsl>
+#include <nbl/builtin/glsl/math/functions.glsl>
+#include <nbl/builtin/glsl/ext/FFT/parameters.glsl>
+
+
+#ifndef _NBL_GLSL_EXT_FFT_MAX_DIM_SIZE_
+#error "_NBL_GLSL_EXT_FFT_MAX_DIM_SIZE_ should be defined."
+#endif
+
+#ifndef _NBL_GLSL_EXT_FFT_MAX_ITEMS_PER_THREAD
```
devshgraphicsprogramming, line 20: btw is `_NBL_GLSL_EXT_FFT_MAX_ITEMS_PER_THREAD` required to be a power of two?

examples_tests/49.ComputeFFT/fft_convolve_ifft.comp
Erfan-Ahmadi: oh I didn't notice that

include/nbl/builtin/glsl/ext/FFT/default_compute_fft.comp
devshgraphicsprogramming: btw when you convert integer coordinates you need to do this
`(vec3(integerCoords)+vec3(0.5))/vec3(textureSize)`

This is because pixels are sampled at centers

P.S. Make sure you dont divide by zero for the z coordinate

include/nbl/builtin/glsl/ext/FFT/parameters.glsl
Erfan-Ahmadi: @devshgraphicsprogramming  @Crisspl 
I remember I tried many things including your suggestion to add padding to the end. 

I was confused a lot because looking at my renderdoc capture I saw many correct values for the push constants for the first dispatches but for the last dispatches the uint in the middle had wrong values. 
maybe because I used the same pipelinelayout(?) I have no idea.

I just tested this again and It didn't work even with the padding at the end

Crisspl: hmmm.. maybe engine bug then, would need to inspect further. But it would need some working code presenting  this behaviour

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,247 @@
...
+    {
+        uint tid = thread_offset + t * _NBL_GLSL_EXT_FFT_WORKGROUP_SIZE_;
+        uvec3 coords = nbl_glsl_ext_FFT_getCoordinates(tid);
+        uvec3 bitReversedCoords = nbl_glsl_ext_FFT_getBitReversedCoordinates(coords, leadingZeroes);
+
+        current_values[t] = nbl_glsl_ext_FFT_getPaddedData(bitReversedCoords, channel);
+    }
+
+    // For loop for each stage of the FFT (each virtual thread computes 1 buttefly wing)
+	for(uint i = 0u; i < logTwo; ++i)
```
devshgraphicsprogramming, line 181: my idea for reducing shared memory usage.

Make a `nbl_glsl_ext_workgroup_FFT` function which is specialized in performing an FFT for a `2*_NBL_GLSL_WORKGROUP_SIZE_` sized input (you can use `4*_NBL_GLSL_WORKGROUP_SIZE_ ` shared memory to not have to use so many barriers for the data exchanges)

Then have this function combine `0x1u<<max(findMSB(dataLength-1)-_NBL_GLSL_WORKGROUP_SIZE_LOG_2_,0)` of these "double workgroup sized" FFTs together

Note that in this case your shuffle mask is `1u<j` where `j>_NBL_GLSL_WORKGROUP_SIZE_LOG_2_` so you end up requesting a shuffled value from a shared memory index assigned to the same "physical thread"

Meaning that instead of going through shared memory and worrying about barriers, you can simply do `shuffled_values[t] = current_values[t^(mask>>(_NBL_GLSL_WORKGROUP_SIZE_LOG_2_+1u))];` without any barriers

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,247 @@
...
+             : nbl_glsl_ext_FFT_twiddleInverse(tid, i, logTwo);
+
+            nbl_glsl_complex this_value = current_values[t];
+
+            if(0u < uint(tid & mask)) {
+                current_values[t] = shuffled_value + nbl_glsl_complex_mul(twiddle, this_value); 
+            } else {
+                current_values[t] = this_value + nbl_glsl_complex_mul(twiddle, shuffled_value); 
+            }
+        }
```
devshgraphicsprogramming, line 232: only have half the "virtual threads" perform work and do both parts of the butterly each, if you look at the FFT diagram/algorithm the butterflies happen in pairs and the twiddle factor is nearly identical for the pair (so you can compute it only once)

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,247 @@
...
+        // Data Exchange for virtual threads :
+        // X and Y are seperate to use less shared memory for complex numbers
+        // Get Shuffled Values X for virtual threads
+        for(uint t = 0u; t < num_virtual_threads; t++)
+        {
+            uint tid = thread_offset + t * _NBL_GLSL_EXT_FFT_WORKGROUP_SIZE_;
+            _NBL_GLSL_SCRATCH_SHARED_DEFINED_[tid] = floatBitsToUint(current_values[t].x);
+        }
+        barrier();
+        memoryBarrierShared();
```
devshgraphicsprogramming, line 194: `barrier` implies `memoryBarrierShared`

include/nbl/builtin/glsl/ext/FFT/fft.glsl
Erfan-Ahmadi: seems like it,
the max dim is power of two
so _NBL_GLSL_EXT_FFT_MAX_ITEMS_PER_THREAD is power of two if WORKGROUP_SIZE is power of two
which it is

include/nbl/builtin/glsl/ext/FFT/fft.glsl
Erfan-Ahmadi: totally agreed.
It was originally this way because you wanted me to use ShuffleXOR (in the first version) which would've had no use when a thread computes a full butterfly

include/nbl/builtin/glsl/ext/FFT/default_compute_fft.comp
Erfan-Ahmadi: ohhhh nice!

Erfan-Ahmadi: ok that's done now

examples_tests/49.ComputeFFT/remove_padding.comp
Erfan-Ahmadi: No, It was either me being stupid or the GitHub website having a bad interface

include/nbl/ext/FFT/FFT.h
Erfan-Ahmadi: @devshgraphicsprogramming 
There are 2 choices for me.

Either get the other images size as an input and do 
`padToNextPOT(max(InputDimensions, otherImageSize));`
or
The user has to give me the maximum of the 2 dimensions (kernel and srcImage) to Pad.

Since these 2 choices seemed misleading to me, I initially decided to just get the `paddedDimensions` as the input.

Anyways. just let me know what to do and I'll do it. 

I can also move the `padDimensionsToNextPOT` function from main.cpp to FFT Class.

include/nbl/ext/FFT/FFT.h
devshgraphicsprogramming: for both `uvec3` and `uvec4`

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,376 @@
...
+    uint thread_offset = gl_LocalInvocationIndex;
+
+	// Pass 0: Bit Reversal
+	uint leadingZeroes = nbl_glsl_clz(dataLength) + 1u;
+	uint logTwo = 32u - leadingZeroes;
+	
+    const uint numChannels = nbl_glsl_ext_FFT_Parameters_t_getNumChannels();
+
+    nbl_glsl_complex even_values[_NBL_GLSL_EXT_FFT_MAX_ITEMS_PER_THREAD][_NBL_GLSL_EXT_FFT_MAX_CHANNELS]; // should be half the prev version
+    nbl_glsl_complex odd_values[_NBL_GLSL_EXT_FFT_MAX_ITEMS_PER_THREAD][_NBL_GLSL_EXT_FFT_MAX_CHANNELS];
```
devshgraphicsprogramming, line 171: no need to store all channels at once

include/nbl/builtin/glsl/ext/FFT/fft.glsl
```diff
@@ -0,0 +1,376 @@
...
+                even_values[t][ch].y = uintBitsToFloat(_NBL_GLSL_SCRATCH_SHARED_DEFINED_[even_rev_bits]);
+                odd_values[t][ch].y = uintBitsToFloat(_NBL_GLSL_SCRATCH_SHARED_DEFINED_[odd_rev_bits]);
+            }
+
+            barrier();
+        }
+    }
+
+    // For loop for each stage of the FFT (each virtual thread computes 1 buttefly)
+	for(uint i = 0u; i < logTwo; ++i)
```
devshgraphicsprogramming, line 261: what I imagined it
```glsl
for (uint t=0u; t<num_virtual_threads; t++)
{
    nbl_glsl_ext_workgroup_FFT(even_values[t],odd_values[t]); // remember both parameters are `inout`
    barrier(); // because nbl_glsl_ext_workgroup_FFT reuses the same shared memory
}

for(uint i = findMSB(_NBL_GLSL_EXT_FFT_WORKGROUP_SIZE_-1u); i < logTwo; ++i)
{
    // do it without barriers and using shared memory
}
```

Issue #60, opened 2021-02-06
Rework `io` namespace into `system` and `ui`


devshgraphicsprogramming:
## Solution proposal

Use C++17 filesystem for all the path and folder stuff.

## Additional context

An old pseudo-code idea snippet: https://github.com/buildaworldnet/IrrlichtBAW/issues/7#issuecomment-679280914

PR #64, opened 2021-02-10
Asset restore, fixed ex18, CDerivativeMapCreator

Crisspl:
## TODO list:
In next commit im going to fix VT to work with non-NV GPUs by making use of our spirv-cross contribution


include/nbl/asset/IAssetManager.h
```diff
@@ -574,6 +574,7 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
         //! @returns 0xdeadbeefu on failure or 0-based index on success.
         uint32_t addAssetLoader(core::smart_refctd_ptr<IAssetLoader>&& _loader)
         {
+            _loader->initialize();
```
devshgraphicsprogramming, line 577: GET RID OF THIS

include/nbl/asset/interchange/SAssetBundle.h
```diff
@@ -61,6 +61,10 @@ class SAssetBundle
 		{
 			return core::SRange<const core::smart_refctd_ptr<IAsset>>(m_contents->begin(),m_contents->end());
 		}
+		inline core::SRange<core::smart_refctd_ptr<IAsset>> getContents()
+		{
+			return core::SRange<core::smart_refctd_ptr<IAsset>>(m_contents->begin(), m_contents->end());
+		}
```
devshgraphicsprogramming, line 67: now you've just allowed anyone to change the contents of the bundle!

src/nbl/asset/IAssetManager.cpp
```diff
@@ -185,9 +185,6 @@ void IAssetManager::addLoadersAndWriters()
 #ifdef _NBL_COMPILE_WITH_GLI_WRITER_
 	addAssetWriter(core::make_smart_refctd_ptr<asset::CGLIWriter>());
 #endif
-
-    for (auto& loader : m_loaders.vector)
-        loader->initialize();
```
devshgraphicsprogramming: restore

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -1139,12 +937,29 @@ SContext::tex_ass_type CMitsubaLoader::cacheTexture(SContext& ctx, uint32_t hier
...
+							auto bundle = interm_getRestoredAssetInHierarchy(2u, m_assetMgr, tex->bitmap.filename.svalue, ctx.inner.params, hierarchyLevel, ctx.override_);
+							if (bundle.getContents().empty() || image->isADummyObjectForCache())
+							{
+								// if for some reason restore failed, force recreating whole view
+								auto removeBundle = asset::SAssetBundle(nullptr, { view });
+								m_assetMgr->removeAssetFromCache(removeBundle);
+								view = nullptr;
+							}
+						}
+					}
```
devshgraphicsprogramming, line 955: why isnt this whole thing in some alternative to `IAssetLoaderOverride::findCachedAsset` that does restores from dummy?

src/nbl/video/COpenGLRenderpassIndependentPipeline.h
```diff
@@ -106,30 +106,11 @@ class COpenGLRenderpassIndependentPipeline final : public IGPURenderpassIndepend
...
-			            break;
-		        }
-		        assert(false);
-		        return *(stageUpdateStamps+IGPURenderpassIndependentPipeline::SHADER_STAGE_COUNT);
+                return const_cast<std::atomic_uint32_t&>(
+                    const_cast<const PushConstantsState*>(this)->getStamp(_stage)
+                );
 	        }
-            inline const uint32_t& getStamp(IGPUSpecializedShader::E_SHADER_STAGE _stage) const
+            inline const std::atomic_uint32_t& getStamp(IGPUSpecializedShader::E_SHADER_STAGE _stage) const
```
devshgraphicsprogramming, line 113: is the timestamp supposed to change value on us `volatile` style while we attempt to use it to compare with stuff?

include/nbl/asset/interchange/SAssetBundle.h
```diff
@@ -109,6 +109,10 @@ class SAssetBundle
 			m_contents->operator[](offset) = std::move(_asset);
 			assert(allSameTypeAndNotNull());
 		}
+		inline core::SRange<core::smart_refctd_ptr<IAsset>> getMutableContents()
```
devshgraphicsprogramming, line 112: why would we even have this?

devshgraphicsprogramming: ah ok I see private scope

devshgraphicsprogramming: `getMutableContentPointers` more appropriate, we can always mutate content assets anyway from the `getContents() const`

Crisspl: oh, right, via `smart_refctd_ptr::get()` yea, then this is not needed. 
Actually this is a little inconsistent since `operator->() const` returns `const T*` and `get() const` returns `T*`

devshgraphicsprogramming: yeah its a bit weird about the `smart_refctd_ptr::operator->() const` returning `const T*`

should probably be fixed

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -301,6 +337,50 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
+		return interm_getAssetInHierarchy(_mgr, _filename, getFindParams(_params), _hierarchyLevel, _override);
+	}
+	SAssetBundle interm_getAssetInHierarchy_find(IAssetManager* _mgr, io::IReadFile* _file, const std::string& _supposedFilename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel)
+	{
+		return interm_getAssetInHierarchy(_mgr, _file, _supposedFilename, getFindParams(_params), _hierarchyLevel);
+	}
+	SAssetBundle interm_getAssetInHierarchy_find(IAssetManager* _mgr, const std::string& _filename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel)
+	{
+		return interm_getAssetInHierarchy(_mgr, _filename, getFindParams(_params), _hierarchyLevel);
+	}
```
devshgraphicsprogramming, line 362: could we do this via "perfect forwarding" or a similar template so I'm not seeing all this duplicate code

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -301,6 +337,50 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
+		return interm_getAssetInHierarchy(_mgr, _filename, getReloadParams(_params, _restoreLevels, _hierarchyLevel), _hierarchyLevel);
+	}
+	SAssetBundle interm_getAssetInHierarchy_reload(uint32_t _restoreLevels, IAssetManager* _mgr, io::IReadFile* _file, const std::string& _supposedFilename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel)
+	{
+		return interm_getAssetInHierarchy(_mgr, _file, _supposedFilename, getReloadParams(_params, _restoreLevels, _hierarchyLevel), _hierarchyLevel);
+	}
+	SAssetBundle interm_getAssetInHierarchy_reload(uint32_t _restoreLevels, IAssetManager* _mgr, const std::string& _filename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel)
+	{
+		return interm_getAssetInHierarchy(_mgr, _filename, getReloadParams(_params, _restoreLevels, _hierarchyLevel), _hierarchyLevel);
+	}
```
devshgraphicsprogramming, line 383: could we do this via "perfect forwarding" or a similar template so I'm not seeing all this duplicate code

Crisspl: i tried and not really, im ofc in a tricky way but wont be readable. Because of the need to modify load params

devshgraphicsprogramming: Could something like this work?

```cpp
template<typename T>
T forwardButModify(T&& t, extra_parameter_t extra)
{
   if constexpr (std::is_same_v<T,TypeWeWantToModify>)
      return modifiedType(t); // should probably do separate versions for when the type is a value, const ref or rvalue ref
   return std::forward<T>(t);
}


template<typename... Args>
auto f_with_extra(extra_parameter_t extra, Args&&... args)
{
   return f(forwardButModify<Args>(args)...);
}
```

Crisspl: yea, i've thought abut something very similar but decided those explicit overloads will be much more legible
if its ok for you, i can do something like this though

Crisspl: but no, not this, forward is a cast to reference type, so this is not taht easy

devshgraphicsprogramming: hmm fine, leave as is.. ist just painful to read.

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -289,6 +289,42 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
+		if (_restoreLevels == 0u)
+			return interm_getAssetInHierarchy(std::forward<Args>(args)...);
+
+		auto any_dummy = [](const SAssetBundle& b) {
+			for (const auto& a : b.getContents())
+				if (a->isADummyObjectForCache())
+					return true;
+		};
+
+		auto bundle = interm_getAssetInHierarchy_find(std::forward<Args>(args)...);
```
devshgraphicsprogramming, line 306: this is wrong, because if you wanted to get a restored asset but are not sure if it was actually loaded before, then:
- this will use the loader to load any unloaded asset, but will not add it to the cache
- return the uncached bundle if none of the objects are dummy 

So it wont be a drop-in replacement for getAssetInHierarchy

The proper way would be to:
- do a "classical" load according to the flags, so cached, uncached or fully duplicate == no overriding except from the usual by the asset loader overload
- use the loader overload to choose your asset as usual
- if the chosen asset is a dummy or their dependant at any point up to the restore level (will never happen if you used duplicate flags for everything, so no need to check explicitly for that) 
   - do a duplicate load up to restore level, after the restore level use the flags specified
   [while doing the duplicate load, you need to allow for the asset loader override's to make about files and stuff if they're not found, as usual, but let it know you're doing it for a restore from dummy = because right now it cant tell the difference between the first primary load and the special reload]
   - for every asset in the duplicate load bundle and the cached load bundle (O(n^2) if one wishes so) = this part I'd implement in the IAssetLoaderOVerride
      - use the asset loader override to choose what should be restored and restore it
  - return the asset the IAssetLoaderOVerride originally chose (which should now not be a dummy)

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -289,6 +289,42 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
+		const uint32_t count = bundle.getMutableContents().size();
+		auto* dummies = bundle.getMutableContents().begin();
+		auto* reloaded = reloadedBundle.getMutableContents().begin();
+		for (uint32_t i = 0u; i < count; ++i)
+			if (dummies[i]->isADummyObjectForCache() && !dummies[i]->canBeRestoredFrom(reloaded[i].get()))
+				return {}; // return empty bundle
+
+		for (uint32_t i = 0u; i < count; ++i)
+			if (dummies[i]->isADummyObjectForCache())
+				dummies[i]->restoreFromDummy(reloaded[i].get(), _restoreLevels);
```
devshgraphicsprogramming, line 323: this part should be inside the IAssetLoaderOverride and it should be the one choosing what in the first bundle gets restored from what in the second bundle

include/nbl/asset/interchange/IAssetLoader.h
devshgraphicsprogramming, line 362: these functions shouldn't exist.

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -301,6 +337,50 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
+	{
+		return interm_getAssetInHierarchy(_mgr, _file, _supposedFilename, getFindParams(_params), _hierarchyLevel);
+	}
+	SAssetBundle interm_getAssetInHierarchy_find(IAssetManager* _mgr, const std::string& _filename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel)
+	{
+		return interm_getAssetInHierarchy(_mgr, _filename, getFindParams(_params), _hierarchyLevel);
+	}
+	static IAssetLoader::SAssetLoadParams getReloadParams(IAssetLoader::SAssetLoadParams _params, uint32_t _restoreLevels, uint32_t _hierLevel)
+	{
+		_params.cacheFlags = ECF_DUPLICATE_UNTIL_LEVEL(_restoreLevels + _hierLevel);
```
devshgraphicsprogramming, line 365: you should only overwrite the flags between `[_hierLevel,_hierLevel+_restoreLevels]`

include/nbl/asset/interchange/IAssetLoader.h
devshgraphicsprogramming: actually in that set-up we could replace most `SAssetBundle interm_getAssetInHierarchy` calls with `IAsset* getRestoredAssetInHierarchy`

devshgraphicsprogramming:
would be probably good to add a `restoreLevels` , either as an argument to `IAssetManager::getAsset()` or as a member to `IAssetLoader::SAssetLoadParams`

Crisspl:
> would be probably good to add a `restoreLevels` , either as an argument to `IAssetManager::getAsset()` or as a member to `IAssetLoader::SAssetLoadParams`

why? thats redundant then

devshgraphicsprogramming:
> 
> 
> > would be probably good to add a `restoreLevels` , either as an argument to `IAssetManager::getAsset()` or as a member to `IAssetLoader::SAssetLoadParams`
> 
> why? thats redundant then

how would it be redundant?

The loaders have a way to restore a part of an IAsset DAG which has been loaded from a file... the user doesn't.

include/nbl/asset/interchange/IAssetLoader.h
Crisspl: > this will use the loader to load any unloaded asset, but will not add it to the cache

not true, my flags modifications make sure that we always look into cache, but caching or not caching is still user's decision. Then if assets were just loaded, they're instantly returned (w/o reload and restore pass) and they're alerady in the cache if flags say they're wanted to be cached

> return the uncached bundle if none of the objects are dummy

Obvs not true as well

Maybe `interm_getAssetInHierarchy_find` name is misleading?

include/nbl/asset/interchange/IAssetLoader.h
Crisspl: hm, right, i mean flags for levels `[0, _hierLevel)` doesnt matter since nobody will ever look at them, but i shouldnt zero-out levels `[_hierLevel+_restoreLevels+1, INF)`

Crisspl: hm, actually i think i should because if original flags said to not look into cache for levels `>_hierLevel+_restoreLevels`, then those levels will be pointlessly loaded again. So flags for those levels should be 0 (look into cache and put into cache if loaded; or at least just look into cache -- caching/not caching flags doesnt really matter while reloading since we can be sure they were alerady loaded at least once so they'll always be found)

devshgraphicsprogramming: > 
> 
> hm, right, i mean flags for levels `[0, _hierLevel)` doesnt matter since nobody will ever look at them, but i shouldnt zero-out levels `[_hierLevel+_restoreLevels+1, INF)`

but for the sake of formalism you shouldn't overwrite the range `[0, _hierLevel)`

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
devshgraphicsprogramming: I understand now.

devshgraphicsprogramming: its still very fucked though

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -289,6 +289,42 @@ class IAssetLoader : public virtual core::IReferenceCounted
 	SAssetBundle interm_getAssetInHierarchy(IAssetManager* _mgr, const std::string& _filename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel, IAssetLoader::IAssetLoaderOverride* _override);
 	SAssetBundle interm_getAssetInHierarchy(IAssetManager* _mgr, io::IReadFile* _file, const std::string& _supposedFilename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel);
 	SAssetBundle interm_getAssetInHierarchy(IAssetManager* _mgr, const std::string& _filename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel);
+	//! ECF_DUPLICATE_* flags are ignored (since there's no point in **restoring** if we're not even looking into the cache) -- only ECF_DONT_CACHE_* are honored
+	//! `Args` ought to be same parameter types as interm_getAssetInHierarchy(...) overloads offer
+	template <typename ...Args>
+	SAssetBundle interm_getRestoredAssetInHierarchy(uint32_t _restoreLevels, Args&&... args)
+	{
+		if (_restoreLevels == 0u)
```
devshgraphicsprogramming, line 297: this should probably be
```cpp
const uint64_t duplicateBitmask =ECF_DUPLICATE_REFERENCES-ECF_DONT_CACHE_REFERENCES;
const uint64_t duplicateBitmaskForRestoreLevels = duplicateBitmask&((0x1u<<(restoreLevels*2u))-1u);
const bool duplicateAll = (cacheFlags>>(hierarchyLevel*2u)&duplicateBitmaskForRestoreLevels)==duplicateBitmaskForRestoreLevels;
if (duplicateAll)
```

include/nbl/asset/interchange/IAssetLoader.h
devshgraphicsprogramming:
```cpp
chosenAsset = override->handleRestore(std::move(chosenAsset), initialBundle, reloadedBundle)
if (chosenAsset->isAnyDependencyDummy(restoreLevel-1))
   return nullptr;
return chosenAsset;
```

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -301,6 +337,60 @@ class IAssetLoader : public virtual core::IReferenceCounted
...
+		return interm_getAssetInHierarchy(_mgr, _filename, getFindParams(_params, _restoreLevels, _hierarchyLevel), _hierarchyLevel, _override);
+	}
+	SAssetBundle interm_getAssetInHierarchy_find(uint32_t _restoreLevels, IAssetManager* _mgr, io::IReadFile* _file, const std::string& _supposedFilename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel)
+	{
+		return interm_getAssetInHierarchy(_mgr, _file, _supposedFilename, getFindParams(_params, _restoreLevels, _hierarchyLevel), _hierarchyLevel);
+	}
+	SAssetBundle interm_getAssetInHierarchy_find(uint32_t _restoreLevels, IAssetManager* _mgr, const std::string& _filename, const IAssetLoader::SAssetLoadParams& _params, uint32_t _hierarchyLevel)
+	{
+		return interm_getAssetInHierarchy(_mgr, _filename, getFindParams(_params, _restoreLevels, _hierarchyLevel), _hierarchyLevel);
+	}
```
devshgraphicsprogramming, line 365: as per today's voice call, these functions are not needed.

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -109,6 +109,9 @@ class IAssetLoader : public virtual core::IReferenceCounted
         const char* relativeDir;
         E_LOADER_PARAMETER_FLAGS loaderFlags;				//!< Flags having an impact on extraordinary tasks during loading process
 		IMeshManipulator* meshManipulatorOverride = nullptr;    //!< pointer used for specifying custom mesh manipulator to use, if nullptr - default mesh manipulator will be used
+		uint32_t restoreLevels = 0u;
+		// user should not ever write to this flag
+		bool reload = false;
```
devshgraphicsprogramming, line 114: make it private or something?

devshgraphicsprogramming: also its a bool (undefined size)

include/nbl/asset/interchange/IAssetLoader.h
```diff
@@ -261,6 +264,8 @@ class IAssetLoader : public virtual core::IReferenceCounted
 		//! After a successful load of an asset or sub-asset
 		//TODO change name
 		virtual void insertAssetIntoCache(SAssetBundle& asset, const std::string& supposedKey, const SAssetLoadContext& ctx, const uint32_t hierarchyLevel);
+
+		virtual bool handleRestore(core::smart_refctd_ptr<IAsset>&& _chosenAsset, SAssetBundle& _bundle, SAssetBundle& _reloadedBundle, uint32_t _restoreLevels);
```
devshgraphicsprogramming, line 268: what about the method that handles entire bundles being restored?

src/nbl/asset/interchange/CGraphicsPipelineLoaderMTL.cpp
```diff
@@ -519,17 +519,19 @@ CGraphicsPipelineLoaderMTL::image_views_set_t CGraphicsPipelineLoaderMTL::loadIm
...
             SAssetBundle bundle;
             if (i != CMTLMetadata::CRenderpassIndependentPipeline::EMP_BUMP)
-                bundle = interm_getAssetInHierarchy(m_assetMgr, relDir+_mtl.maps[i], lp, _ctx.topHierarchyLevel+ICPURenderpassIndependentPipeline::IMAGE_HIERARCHYLEVELS_BELOW, _ctx.loaderOverride);
+                bundle = interm_getAssetInHierarchy(m_assetMgr, relDir+_mtl.maps[i], lp, hierarchyLevel, _ctx.loaderOverride);
             else
             {
                 // we need bumpmap restored to create derivative map from it
                 const uint32_t restoreLevels = 3u; // 2 in case of image (image, texel buffer) and 3 in case of image view (view, image, texel buffer)
-                bundle = interm_getRestoredAssetInHierarchy(restoreLevels, m_assetMgr, relDir+_mtl.maps[i], lp, _ctx.topHierarchyLevel+ICPURenderpassIndependentPipeline::IMAGE_HIERARCHYLEVELS_BELOW, _ctx.loaderOverride);
+                lp.restoreLevels = hierarchyLevel + restoreLevels; // TODO what if `lp.restoreLevels` already had some non-zero value? ehhh this is complex, and this logic must be in some common function
```
devshgraphicsprogramming, line 533: its easy... take a `max(originalRestoreLevels,hierarchyLevel+wantedREstoreLEvelsFromHere)`

include/nbl/asset/IAssetManager.h
```diff
@@ -203,28 +203,37 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
+                flags_t mask = ~static_cast<flags_t>(0);
+                mask = core::bitfieldInsert<flags_t>(mask, IAssetLoader::ECF_DONT_CACHE_REFERENCES, 2u*_hierarchyLevel, 2u*restoreLevels);
+                params.cacheFlags = static_cast<IAssetLoader::E_CACHING_FLAGS>(params.cacheFlags & mask);
+            }
 
             IAssetLoader::SAssetLoadContext ctx{params, _file};
 
             std::string filename = _file ? _file->getFileName().c_str() : _supposedFilename;
-            io::IReadFile* file = _override->getLoadFile(_file, filename, ctx, _hierarchyLevel);
+            io::IReadFile* file = _override->getLoadFile(_file, filename, ctx, _hierarchyLevel); // WARNING: mem-leak possibility: _override should return smart_ptr<IReadFile> (TODO, inspect this)
```
devshgraphicsprogramming, line 224: yes 100% possibility, overall the design of the `io` namespace sucks ass, to be addressed in #60

include/nbl/asset/IAssetManager.h
```diff
@@ -203,28 +203,37 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
             {
                 params.meshManipulatorOverride = m_meshManipulator.get();
             }
+            if (restoreLevels)
+            {
+                using flags_t = std::underlying_type_t<IAssetLoader::E_CACHING_FLAGS>;
+                flags_t mask = ~static_cast<flags_t>(0);
+                mask = core::bitfieldInsert<flags_t>(mask, IAssetLoader::ECF_DONT_CACHE_REFERENCES, 2u*_hierarchyLevel, 2u*restoreLevels);
+                params.cacheFlags = static_cast<IAssetLoader::E_CACHING_FLAGS>(params.cacheFlags & mask);
+            }
```
devshgraphicsprogramming, line 219: you're only supposed to do this during the second call to `getAssetInHierarchy` meant for a reload (not the first call, because then you wont cache anything new you might load), remember?

DUPLICATE too, not DONT CACHE references

devshgraphicsprogramming: oh wait wtf, you're masking out the DONT_CACHE bit on [hierarchyLevel,hierarchyLevel+restoreLevels) ?

devshgraphicsprogramming: then for these levels
CACHE -> CACHE
DONT_CACHE -> CACHE ???
DUPLICATE -> CACHE ???

maan this bit arithmetic is confusing AF, I'd like to know its intended purpose

Crisspl: masking is not *out*
ignoring DUPLICATE, leaving DONT_CACHE as is

Crisspl: > then for these levels
> CACHE -> CACHE
> DONT_CACHE -> CACHE ???
> DUPLICATE -> CACHE ???
> 
> maan this bit arithmetic is confusing AF, I'd like to know its intended purpose

we've ben talking about this
CACHE -> CACHE
DONT_CACHE -> DONT_CACHE
DUPLICATE -> DONT_CACHE

include/nbl/asset/IAssetManager.h
```diff
@@ -235,30 +244,61 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
+                    reloadFlags = core::bitfieldInsert<flags_t>(reloadFlags, IAssetLoader::ECF_DUPLICATE_REFERENCES, _hierarchyLevel*2u, restoreLevels*2u);
+                    reloadParams.cacheFlags = static_cast<IAssetLoader::E_CACHING_FLAGS>(reloadFlags);
+
+                    reloadParams.restoreLevels = 0u; // make sure it wont turn into infinite recursion
+                    reloadParams.reload = true; // TODO (consider): alternative to this flag: another method in override just to let user choose asset for restore
+                }
+                IAssetLoader::SAssetLoadContext ctx(params, file);
+                auto asset = _override->chooseDefaultAsset(bundle, ctx);
+                if (!asset->isAnyDependencyDummy(restoreLevels - 1u))
+                    return bundle;
```
devshgraphicsprogramming, line 292: well in this case you need to check entire bundle (restore from the asset manager), no place for override choosing asset to check and restore

include/nbl/asset/IAssetManager.h
```diff
@@ -235,30 +244,61 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
+
+                auto reloadBundle = getAssetInHierarchy(_file, _supposedFilename, reloadParams, _hierarchyLevel, _override);
+
+                bool restoreSucceeded = _override->handleRestore(std::move(asset), bundle, reloadBundle, restoreLevels);
+                if (!restoreSucceeded) // hm? return empty bundle if restore was requested, but did not succeeded? or just return the bundle? (TODO consider)
```
devshgraphicsprogramming, line 297: the other `handleRestore` (whole bundle variant), also dont return a bool, check that none of the assets are dummy in the range that interests you.

Return empty bundle, OR return a bundle which only has the assets which were sucessfully restored?

devshgraphicsprogramming: (hmm, then maybe in that case the override could choose the assets for restore and construct a new bundle via a non-overridable method which actually checks if restore was successful before adding them to the new bundle)

However if we returned a custom bundle, we'd have a problem because we kinda expect that the `SAssetBundle` we get back from the `getAsset` is the same one that is inside the cache multimap...


So its a bad idea...


Return empty bundle then.

src/nbl/asset/interchange/IAssetLoader.cpp
```diff
@@ -35,6 +35,24 @@ void IAssetLoader::IAssetLoaderOverride::insertAssetIntoCache(SAssetBundle& asse
         m_manager->insertAssetIntoCache(asset, ASSET_MUTABILITY_ON_CACHE_INSERT);
 }
 
+bool IAssetLoader::IAssetLoaderOverride::handleRestore(core::smart_refctd_ptr<IAsset>&& _chosenAsset, SAssetBundle& _bundle, SAssetBundle& _reloadedBundle, uint32_t _restoreLevels)
+{
+    if (_bundle.getContents().size() != _reloadedBundle.getContents().size())
+        return false;
+
+    auto dummies = _bundle.getMutableContents();
```
devshgraphicsprogramming, line 43: why the call to `getMutableContents` its not like you'll be changing the pointers values stores in the bundle (the contents of the dynamic array)

src/nbl/asset/interchange/IAssetLoader.cpp
```diff
@@ -35,6 +35,24 @@ void IAssetLoader::IAssetLoaderOverride::insertAssetIntoCache(SAssetBundle& asse
         m_manager->insertAssetIntoCache(asset, ASSET_MUTABILITY_ON_CACHE_INSERT);
 }
 
+bool IAssetLoader::IAssetLoaderOverride::handleRestore(core::smart_refctd_ptr<IAsset>&& _chosenAsset, SAssetBundle& _bundle, SAssetBundle& _reloadedBundle, uint32_t _restoreLevels)
```
devshgraphicsprogramming, line 38: dont trust the bool, because you'll need to check yourself, so dont make a function that returns a `bool`

also return the `_chosenAsset`

src/nbl/asset/interchange/IAssetLoader.cpp
```diff
@@ -35,6 +35,24 @@ void IAssetLoader::IAssetLoaderOverride::insertAssetIntoCache(SAssetBundle& asse
...
+    auto found_it = std::find(dummies.begin(), dummies.end(), _chosenAsset);
+    if (found_it == dummies.end())
+        return false;
+    const uint32_t ix = found_it - dummies.begin();
+
+    auto reloaded = _reloadedBundle.getMutableContents();
+    if (dummies.begin()[ix]->isADummyObjectForCache() && !dummies.begin()[ix]->canBeRestoredFrom(reloaded.begin()[ix].get()))
+        return false;
+
+    return dummies.begin()[ix]->restoreFromDummy(reloaded.begin()[ix].get(), _restoreLevels);
```
devshgraphicsprogramming, line 53: little note, `dummies.begin()[ix]` should literally be the same thing as `_chosenAsset` so just use that.

include/nbl/asset/interchange/IAssetLoader.h
devshgraphicsprogramming: still waiting on making it impossible for the user to write to `reload`

src/nbl/asset/interchange/IAssetLoader.cpp
```diff
@@ -35,22 +35,36 @@ void IAssetLoader::IAssetLoaderOverride::insertAssetIntoCache(SAssetBundle& asse
         m_manager->insertAssetIntoCache(asset, ASSET_MUTABILITY_ON_CACHE_INSERT);
 }
 
-bool IAssetLoader::IAssetLoaderOverride::handleRestore(core::smart_refctd_ptr<IAsset>&& _chosenAsset, SAssetBundle& _bundle, SAssetBundle& _reloadedBundle, uint32_t _restoreLevels)
+void IAssetLoader::IAssetLoaderOverride::handleRestore(core::smart_refctd_ptr<IAsset>&& _chosenAsset, SAssetBundle& _bundle, SAssetBundle& _reloadedBundle, uint32_t _restoreLevels)
```
devshgraphicsprogramming, line 38: you've rvalue moved the chosen asset into the function, the function should be able to return a different chosen asset (smart pointer) 

but the default implementation should obvs just return `_chosenAsset` on success, and `nullptr` on failure

src/nbl/asset/interchange/IAssetLoader.cpp
```diff
@@ -35,22 +35,36 @@ void IAssetLoader::IAssetLoaderOverride::insertAssetIntoCache(SAssetBundle& asse
...
+    auto dummies = _bundle.getContents();
     auto found_it = std::find(dummies.begin(), dummies.end(), _chosenAsset);
     if (found_it == dummies.end())
-        return false;
+        return;
     const uint32_t ix = found_it - dummies.begin();
 
-    auto reloaded = _reloadedBundle.getMutableContents();
+    auto reloaded = _reloadedBundle.getContents();
     if (dummies.begin()[ix]->isADummyObjectForCache() && !dummies.begin()[ix]->canBeRestoredFrom(reloaded.begin()[ix].get()))
```
devshgraphicsprogramming, line 50: little note, in this case dummies.begin()[ix] should literally be the same thing as _chosenAsset so just use that.

src/nbl/asset/interchange/IAssetLoader.cpp
devshgraphicsprogramming: on a second thought, does anyone else except for `getAssetInHierarchy_impl` call this function?

include/nbl/asset/IAssetManager.h
```diff
@@ -342,19 +394,37 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
+        //TODO change name
+        SAssetBundle getAssetWholeBundleRestore(const std::string& _filename, const IAssetLoader::SAssetLoadParams& _params)
+        {
+            return getAssetWholeBundleRestore(_filename, _params, &m_defaultLoaderOverride);
+        }
+        //TODO change name
+        SAssetBundle getAssetWholeBundleRestore(io::IReadFile* _file, const std::string& _supposedFilename, const IAssetLoader::SAssetLoadParams& _params)
+        {
+            return getAssetWholeBundleRestore(_file, _supposedFilename, _params, &m_defaultLoaderOverride);
+        }
```
devshgraphicsprogramming, line 426: for the public API user code.. this should be default behaviour of `getAsset`, no new functions.

devshgraphicsprogramming: The philosophies...

User : Play it safe, Restore whole bundles, When inserting into cache insert with IMMUTABLE by default
Loader : Play it fast, Restore on a need-to-know basis, When inserting into cache insert with MUTABLE by default

include/nbl/asset/IAssetManager.h
```diff
@@ -286,30 +294,37 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
-                auto reloadBundle = getAssetInHierarchy(_file, _supposedFilename, reloadParams, _hierarchyLevel, _override);
+                auto reloadBundle = getAssetInHierarchy_impl<RestoreWholeBundle>(_file, _supposedFilename, reloadParams, _hierarchyLevel, _override);
 
-                bool restoreSucceeded = _override->handleRestore(std::move(asset), bundle, reloadBundle, restoreLevels);
-                if (!restoreSucceeded) // hm? return empty bundle if restore was requested, but did not succeeded? or just return the bundle? (TODO consider)
-                    return {};
+                if constexpr (RestoreWholeBundle)
+                {
+                    IAssetLoader::SAssetLoadContext ctx(params, file);
+                    auto asset = _override->chooseDefaultAsset(bundle, ctx);
```
devshgraphicsprogramming, line 303: the check that you do `whole_bundle_not_dummy` is overly conservative for this... you better check if the chosen asset is not dummy up to `restoreLevels` and skip the restore.

include/nbl/asset/IAssetManager.h
```diff
@@ -286,30 +294,37 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
+
+                    // user responsible for checking if assets he wanted to be restored are in fact restored
+                    _override->handleRestore(std::move(asset), bundle, reloadBundle, restoreLevels);
+                }
+                else
+                {
+                    // user responsible for checking if assets he wanted to be restored are in fact restored
```
devshgraphicsprogramming, line 310: if you shift the responsibility for checking if the restore was successful onto the user, we'll just end up with asstons of duplicate code in the loaders for checking that

devshgraphicsprogramming: also this is why it would be useful to have a
```cpp
core::smart_refctd_ptr<IAsset> IAssetLoader::IAssetLoaderOverride::handleRestore(core::smart_refctd_ptr<IAsset>&& _chosenAsset, SAssetBundle& _bundle, SAssetBundle& _reloadedBundle, uint32_t _restoreLevels)
```
so that you may do
```cpp
if (asset->isAnyDependencyDummy(restoreLevels))
{
   auto newChosenAsset = _override->handleRestore(std::move(asset), bundle, reloadBundle, restoreLevels);
   if (newChosenAsset->isAnyDependencyDummy(restoreLevels))
      return {};
}
```

include/nbl/asset/IAssetManager.h
```diff
@@ -286,30 +294,37 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
-                if (!asset->isAnyDependencyDummy(restoreLevels - 1u))
-                    return bundle;
 
-                auto reloadBundle = getAssetInHierarchy(_file, _supposedFilename, reloadParams, _hierarchyLevel, _override);
+                auto reloadBundle = getAssetInHierarchy_impl<RestoreWholeBundle>(_file, _supposedFilename, reloadParams, _hierarchyLevel, _override);
 
-                bool restoreSucceeded = _override->handleRestore(std::move(asset), bundle, reloadBundle, restoreLevels);
-                if (!restoreSucceeded) // hm? return empty bundle if restore was requested, but did not succeeded? or just return the bundle? (TODO consider)
-                    return {};
+                if constexpr (RestoreWholeBundle)
```
devshgraphicsprogramming, line 300: I think you've got your if-statement the wrong way round here

include/nbl/asset/IAssetManager.h
devshgraphicsprogramming: I still dont get why you feel the need to override DUPLICATE into DONT_CACHE of any level in the restore level range

If user/loader/override wanted a DUPLICATE, the better for US! Then at least we can load the relevant level assets in duplicate on first attempt, then they won't report being dummies (because freshly loaded duplicates never are) and we'll do the loading once instead of twice.

I see no downside to leaving the flags as they are on the preliminary load!

devshgraphicsprogramming: Also by overriding DUPLICATE into DONT_CACHE you're violating what the user/loader/override wants, because they want a fresh asset, not a pointer to some old crap in the cache that you're now hellbent to restore.

include/nbl/asset/IAssetManager.h
Crisspl: oh, right, thanks XD

src/nbl/asset/interchange/IAssetLoader.cpp
Crisspl: > on a second thought, does anyone else except for getAssetInHierarchy_impl call this function?

nah

include/nbl/asset/IAssetManager.h
Crisspl: > When inserting into cache insert with IMMUTABLE by default

i thought we agreed on CPU_PERSISTENT

include/nbl/asset/IAssetManager.h
Crisspl: i cant if `handleRestore` can choose again https://github.com/Devsh-Graphics-Programming/Nabla/pull/64#discussion_r575505953

include/nbl/asset/IAssetManager.h
Crisspl: ok, you've convinced me

include/nbl/asset/IAssetManager.h
devshgraphicsprogramming: yeah sorry I mis-wrote, yes user is CPU_PERSISTENT.

include/nbl/asset/IAssetManager.h
devshgraphicsprogramming: did I say to check the same asset as chosen before?
You check twice, once before restore and once after

PR #102, opened 2021-04-26
Visibility buffer

devshgraphicsprogramming:
## Description
Implement the Visibility Buffer rendering technique.

No bindless, everything done painfully and manually so it will even run on OpenGL ES 3.1 in the future!

## Testing
There's an extra example for this!

## TODO list:
- [] Compute Barycentrics from analytical formulas (no AMD/NV intrinsics)
- [] Compute Barycentric derivatives for the above
- [] Provide utility functions for applying the chain rule to triangle attributes
- [] Test analytical `dUVdScreen`
- [] Rasterize a Visibility Buffer
- [] Hijack all the fragment shader code to the Compute Pass 
- [] Cull triangle batches (individual MDIs)


examples_tests/41.VisibilityBuffer/common.glsl
```diff
@@ -9,42 +9,65 @@
...
-} virtualAttribTable;
+    BatchInstanceData batchInstanceData[];
+};
+#if 0
+// VG
+layout(set = 1, binding = 1, std430) readonly buffer IndexBuffer // TODO: VG descriptor include
+{
+    uint indexBuffer[];
+};
+#endif
```
devshgraphicsprogramming, line 45: @Przemog1 how can I add this to the VG descriptors and header?

devshgraphicsprogramming: I not only need access to the vertex data but also index buffer a lot of the time

examples_tests/41.VisibilityBuffer/shadeVBuffer.comp
```diff
@@ -40,24 +13,22 @@ layout (set = 2, binding = 3, std430) restrict readonly buffer PrecomputedStuffS
     uint layer_to_sampler_ix[_NBL_VT_MAX_PAGE_TABLE_LAYERS];
 } precomputed;
 
-layout(set = 3, binding = 0) sampler2D vBuffer;
-layout(set = 3, binding = 1) image2D frameBuffer;
+layout(set = 3, binding = 0) uniform sampler2D vBuffer;
+layout(set = 3, binding = 1, r32ui) uniform uimage2D frameBuffer;
```
devshgraphicsprogramming, line 17: note to self, figure out what to use here

PR #117, opened 2021-06-08
Window events, keyboard/mouse input, clipboard management

sadiuk:
## TODO list:

- [x] Windows
- [ ] X11
- [ ] Android
- [ ] Wayland


include/nbl/system/IAsyncQueueDispatcher.h
```diff
@@ -143,38 +144,44 @@ class IAsyncQueueDispatcher : public IThreadHandler<CRTP, InternalStateType>, pu
...
     }
 
+    void background_work() {}
 private:
     template <typename... Args>
     void work(lock_t& lock, Args&&... optional_internal_state)
     {
         static_assert(sizeof...(optional_internal_state) <= 1u, "How did this happen");
 
+        static_cast<CRTP*>(this)->backgroung_work();
```
devshgraphicsprogramming, line 154: typo

include/nbl/ui/IClipboardManager.h
```diff
@@ -11,9 +12,11 @@ namespace ui
 class IClipboardManager : public core::IReferenceCounted
 {
 public:
-    virtual const char* getClipboard() = 0;
-    virtual bool setClipboard(const char* contents) = 0;
-
+    virtual std::string getClipboardText() = 0;
+    virtual bool setClipboardText(const std::string_view& data) = 0;
+    virtual core::smart_refctd_ptr<asset::ICPUImage> getClipboardImage() = 0;
+    virtual bool setClipboardImage(asset::ICPUImage* image, asset::ICPUImage::SImageCopy data) = 0;
```
devshgraphicsprogramming, line 18: make a struct similar to `SImageCopy`, nested in IClipboardManager which only has the fields from `SImageCopy` that you'd use

include/nbl/ui/IInputEventChannel.h
```diff
@@ -107,19 +135,30 @@ class IMouseEventChannel : public impl::IEventChannelBase<SMouseEvent>
     virtual ~IMouseEventChannel() = default;
 
 public:
+    IMouseEventChannel(size_t circular_buffer_capacity) : base_t(circular_buffer_capacity)
```
devshgraphicsprogramming, line 138: does `base_t` assert that the capacity is a power of two?

include/nbl/ui/IWindow.h
```diff
@@ -57,7 +59,7 @@ class IWindow : public core::IReferenceCounted
             w->m_flags |= ECF_HIDDEN;
             onWindowHidden_impl();
         }
-        void onWindowMoved(IWindow* w, uint32_t x, uint32_t y)
+        void onWindowMoved(IWindow* w, int x, int y)
```
devshgraphicsprogramming, line 62: use `int32_t`, unsized types are cancer

include/nbl/ui/IWindow.h
```diff
@@ -120,7 +122,7 @@ class IWindow : public core::IReferenceCounted
     protected:
         virtual void onWindowShown_impl() {}
         virtual void onWindowHidden_impl() {}
-        virtual void onWindowMoved_impl(uint32_t x, uint32_t y) {}
+        virtual void onWindowMoved_impl(int x, int y) {}
```
devshgraphicsprogramming, line 125: use `int32_t`, unsized types are cancer

devshgraphicsprogramming:
What's with the empty files you've committed?

include/nbl/ui/IInputEventChannel.h
sadiuk: Yeah, circular buffers  have those assertions

include/nbl/ui/CWindowManagerX11.h
```diff
@@ -18,18 +19,29 @@ class CWindowManagerX11 : public IWindowManager
...
-		void init() {}
 		void exit() {}
 		void work(lock_t& lock)
 		{
+			XEvent event;
+			XNextEvent(display, event);
+			Window nativeWindow = event.xany.window;
+
+			XPointer windowCharPtr = nullptr;
+			XFindContext(display, nativeWindow, &windowCharPtr);
```
devshgraphicsprogramming, line 33: can any of these X11 functions block for extended period of time?

src/nbl/ui/CWindowX11.cpp
```diff
@@ -227,6 +230,84 @@ CWindowX11::CWindowX11(core::smart_refctd_ptr<system::ISystem>&& sys, uint32_t _
 
     m_dpy = dpy;
     m_native = win;
+
+    XContext classId = XUniqueContext();
+    XSaveContext(m_dpy, m_native, classId, (XPointer)this);
```
devshgraphicsprogramming, line 235: use @Crisspl's dynamic `x11` function caller for any `X` functions (dont want to compile with that lib being linked at compile time)

Crisspl: yep, he knows, @sadiuk add more functions to the caller in case they're missing (not all x11 fucntions are being loaded)

src/nbl/ui/CWindowX11.cpp
```diff
@@ -227,6 +230,84 @@ CWindowX11::CWindowX11(core::smart_refctd_ptr<system::ISystem>&& sys, uint32_t _
...
+            {
+                eventCallback->onWindowResized(this, e.width, e.height);
+                x11.pXResizeWindow(m_dpy, m_native, e.width, e.height);
+
+            }
+            // Moved
+            if(e.x != m_x || e.y != m_y)
+            {
+                eventCallback->onWindowMoved(this, e.x, e.y);
+                x11.pXMoveWindow(m_dpy, m_native, e.x, e.y);
```
devshgraphicsprogramming, line 258: I think that `onWindowXYZ` should return a bool to let us know if we're allowed to proceed with the request

src/nbl/ui/CWindowWin32.cpp
```diff
@@ -1,286 +1,423 @@
 #include "nbl/ui/CWindowWin32.h"
+#include <hidusage.h>
```
Crisspl, line 2: hidusage.h should be included under `#ifdef _NBL_PLATFORM_WINDOWS_`

src/nbl/ui/CWindowX11.cpp
sadiuk: I what cases will it return false?

devshgraphicsprogramming: when you're not allowing the backend to move the window (no call to `x11.pXMoveWindow`)

src/nbl/ui/CWindowX11.cpp
```diff
@@ -301,15 +310,53 @@ void CWindowX11::processEvent(XEvent event)
...
         }
         case FocusOut:
         {
+            
             break;
         }
     }
 }
 
+void CWindowManagerX11::init()
```
Crisspl, line 324: i believe those are methods of thread handler, not CWindowManagerX11's

devshgraphicsprogramming: isn't X11 safe to use in threads? with `XinitThreads` ?

devshgraphicsprogramming: do you even need a thread handler for X11 ?

include/nbl/ui/CWindowManagerX11.h
```diff
@@ -9,37 +10,73 @@ class CWindowManagerX11 : public IWindowManager
...
+		ERT_DESTROY_WINDOW
+	};
+	template <E_REQUEST_TYPE ERT>
+	struct SRequestParamsBase
+	{
+		static inline constexpr E_REQUEST_TYPE type = ERT;
+	};
+	struct SRequestParams_CreateWindow : SRequestParamsBase<ERT_CREATE_WINDOW>
+	{
+		SRequestParams_CreateWindow(int32_t _x, int32_t _y, uint32_t _w, uint32_t _h, CWindowWin32::E_CREATE_FLAGS _flags, CWindowWin32::native_handle_t* wnd, const std::string_view& caption) :
```
devshgraphicsprogramming, line 25: what's with the `Win32` ?

src/nbl/ui/CWindowX11.cpp
```diff
@@ -28,6 +31,9 @@ NBL_SYSTEM_DECLARE_DYNAMIC_FUNCTION_CALLER_CLASS(X11, system::DefaultFuncPtrLoad
     ,XGetErrorDatabaseText
     ,XGetErrorText
     ,XGetGeometry
+    ,XFindContext
+    ,XUniqueContext
+    ,XSaveContext
 );
```
devshgraphicsprogramming, line 37: regarding the TODO, its the `ui` namespace (system is not for interfacing with the user)

src/nbl/ui/CWindowX11.cpp
Crisspl: for events yes, ofc. For window creation/destruction im not sure it's mandatory like in case of Windows, so i assumed yes.

XNextEvent is blocking until an event appears so in case of of window creation/destruction request, we need to generate a dummy event to wake it up

include/nbl/ui/CWindowManagerX11.h
```diff
@@ -9,37 +10,73 @@ class CWindowManagerX11 : public IWindowManager
...
+			x(_x), y(_y), width(_w), height(_h), flags(_flags), nativeWindow(wnd), windowCaption(caption)
+		{}
+		int32_t x, y;
+		uint32_t width, height;
+		CWindowX11::E_CREATE_FLAGS flags;
+		CWindowX11::native_handle_t* nativeWindow;
```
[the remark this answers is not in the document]
Crisspl: it's output param

src/nbl/ui/CWindowX11.cpp
Crisspl: i already moved all the window classes to ui

src/nbl/ui/CWindowX11.cpp
devshgraphicsprogramming: > XNextEvent is blocking until an event appears so in case of of window creation/destruction request, we need to generate a dummy event to wake it up

you only need a dedicated thread to poll events then, nothing else (and a dummy event to make the loop quit)

include/nbl/ui/CWindowManagerX11.h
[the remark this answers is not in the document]
devshgraphicsprogramming: ah ok

src/nbl/ui/CWindowX11.cpp
Crisspl: are you sure? on Windows the thread polling events and the thread creating windows must be the same thread. Im not sure it's not the case here

include/nbl/ui/CWindowManagerX11.h
```diff
@@ -79,7 +156,17 @@ class CWindowManagerX11 : public IWindowManager
...
 	private:
 		Display* display;
+		X11 x11("X11");
+		Xinput xinput("Xinput");
+#ifdef _NBL_LINUX_X11_RANDR_
+    	Xrandr xrandr("Xrandr");
+#endif
+#ifdef _NBL_LINUX_X11_VIDMODE_
+    	Xxf86vm xxf86vm("Xxf86vm");
+#endif
```
devshgraphicsprogramming, line 166: dont even bother with the ifdef, remove the defines from the new code (and any place in the engine really)

src/nbl/ui/CWindowX11.cpp
```diff
@@ -321,31 +130,178 @@ void CWindowX11::processEvent(XEvent event)
...
 }
 
-void CWindowManagerX11::process_request(SRequest& req)
+void CWindowManagerX11::CThreadHandler::process_request(SRequest& req)
 {
     switch (req.type)
     {
     case ERT_CREATE_WINDOW:
     {
+        // XInitThreads() call not needed unless windows are created concurrently, spoof EGL synchronizes per-display access itself
```
devshgraphicsprogramming, line 145: just call `XInitThreads` somewhere at the beginning (preferably right after loading the X11 .so) thre's no harm from it

include/nbl/ui/CWindowManagerX11.h
```diff
@@ -4,11 +4,64 @@
...
+    ,XGetErrorDatabaseText
+    ,XGetErrorText
+    ,XGetGeometry
+    ,XFindContext
+    ,XUniqueContext
+    ,XSaveContext
+);
+// TODO add more
+NBL_SYSTEM_DECLARE_DYNAMIC_FUNCTION_CALLER_CLASS(Xinput, system::DefaultFuncPtrLoader
+	,XListInputDevices)
```
Crisspl, line 31: semicolon needed

src/nbl/ui/CWindowWin32.cpp
Crisspl: but CWindowWin32.h must be included above it

Crisspl:
What happended to EGL submodule in this commit ? https://github.com/Devsh-Graphics-Programming/Nabla/pull/117/commits/998e96a63546e4ef16db7055f5239121d214cb0b

include/CConcurrentObjectCache.h
```diff
@@ -129,9 +129,9 @@ namespace impl
 
         inline bool outputAll(size_t& _inOutStorageSize, MutablePairType* _out) const
         {
-            auto lk = lock_read();
-            const bool r = BaseCache::outputAll(_inOutStorageSize, _out);
-            return r;
+            //auto lk = lock_read();
+            //const bool r = BaseCache::outputAll(_inOutStorageSize, _out);
+            return true;
```
Crisspl, line 134: ???

include/IFileSystem.h
```diff
@@ -186,12 +186,12 @@ class IFileSystem : public virtual core::IReferenceCounted
 		virtual bool moveFileArchive(uint32_t sourceIndex, int32_t relative) =0;
 
 		//! Get the archive at a given index.
-		virtual IFileArchive* getFileArchive(uint32_t index) =0;
+		virtual system::IFileArchive* getFileArchive(uint32_t index) =0;
```
Crisspl, line 189: get rid of this overload

include/IFileSystem.h
```diff
@@ -186,12 +186,12 @@ class IFileSystem : public virtual core::IReferenceCounted
 		virtual bool moveFileArchive(uint32_t sourceIndex, int32_t relative) =0;
```
Crisspl, line 186: get rid of this

include/IFileSystem.h
Crisspl: ah nvm, i forgot those are old api

include/nbl/asset/IAssetManager.h
```diff
@@ -129,7 +127,7 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
     public:
         //! Constructor
         explicit IAssetManager(core::smart_refctd_ptr<io::IFileSystem>&& _fs) :
```
Crisspl, line 129: i guess you have to change ctor param as well

include/CConcurrentObjectCache.h
sadiuk: There is an issue that some methods of that class are declared `const` but call `lock_read()`/`lock_write()` which isn't `const`

include/nbl/asset/IAssetManager.h
```diff
@@ -322,7 +320,7 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
 
             std::string filePath = _filePath;
             _override->getLoadFilename(filePath, ctx, _hierarchyLevel);
-            io::IReadFile* file = m_fileSystem->createAndOpenFile(filePath.c_str());
+            system::IFile* file = m_system->createAndOpenFile(filePath.c_str());
```
Crisspl, line 323: I guess you just did find&replace here, but note that theres no `createAndOpenFile` in ISystem + you wont have to do `.c_str()` any more

include/CConcurrentObjectCache.h
sadiuk: So for now i just removed `const` qualifiers/removed the `const` versions of the methods

include/nbl/system/ISystem.h
```diff
@@ -172,21 +172,48 @@ class ISystem final : public core::IReferenceCounted
...
+        return future.get();
+    }
+    core::smart_refctd_ptr<IFile> createAndWriteFile(const std::filesystem::path& filename)
+    {
+        future_t<core::smart_refctd_ptr<IFile>> future;
+        if (!createFile(future, filename, IFile::ECF_WRITE))
+            return nullptr;
+
+        return future.get();
+    }
```
Crisspl, line 215: nope, nope, get rid of this, just `createFile` is perfectly sufficient

include/nbl/system/IFile.h
```diff
@@ -37,6 +37,10 @@ class IFile : public core::IReferenceCounted
 	virtual void* getMappedPointer() = 0;
 	virtual const void* getMappedPointer() const = 0;
 
+	virtual void seek(uint32_t pos) = 0;
+	virtual void read(char* outData, size_t count) const;
+	virtual size_t getPos() const = 0;
```
Crisspl, line 42: nah, we want to get rid of this kind of API. User must use ISystem API (`readFile`, `writeFile`) to access file contents

include/nbl/asset/IAssetManager.h
sadiuk: Yeah i actually added `createAndOpenFile` to `ISystem`. but yeah, that `.c_str()` is unnecessary

include/nbl/system/ISystem.h
sadiuk: Ah ok got it

include/nbl/system/IFile.h
Crisspl: You can have pure virtual methods for file content access here (i'd say it's encouraged implementation approach since we want to have memory files as well), but they should not be in public IFile API + avoid `seek()`-like approaches

include/nbl/ui/CWindowManagerWin32.h
```diff
@@ -1,5 +1,6 @@
 #ifndef C_WINDOWMANAGER_WIN32
 #define C_WINDOWMANAGER_WIN32
+#ifdef _NBL_PLATFORM_WINDOWS_
```
Crisspl, line 3: IWindowManager.h `#include` must be above this (or anything else that includes `compile_config.h` unconditionally

include/nbl/ui/CWindowManagerX11.h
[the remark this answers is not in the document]
Crisspl: actually, let's call `out_nativeWindow` or something, it may be misleading

include/nbl/ui/CWindowX11.h
```diff
@@ -37,11 +38,18 @@ class CWindowX11 final : public IWindowX11
...
 
+	void processEvent(XEvent event);
 private:
     CWindowX11(core::smart_refctd_ptr<system::ISystem>&& sys, uint32_t _w, uint32_t _h, E_CREATE_FLAGS _flags);
 
     Display* m_dpy;
     native_handle_t m_native;
+	CWindowManagerX11* m_manager;
+	// Not sure yet if i need these or i can do without them 
+	bool isMaximized = false, isMinimized = false;
```
Crisspl, line 49: `IWindow::m_flags` should be always updated to tell you whether the window is minimized, maximized or neither

Crisspl: so those booleans are not needed

include/nbl/ui/IWindowManager.h
```diff
@@ -1,11 +1,14 @@
 #ifndef I_WINDOWMANAGER
 #define I_WINDOWMANAGER
-
+#include <nbl/core/IReferenceCounted.h>
+#include "IWindow.h"
 namespace nbl::ui
 {
-	class IWindowManager
+	class IWindowManager : core::IReferenceCounted
```
Crisspl, line 7: public inheritance

include/CConcurrentObjectCache.h
Crisspl: ah, so instead make sure locking doesnt require mutability (use `mutable` keyword or something)

include/nbl/asset/IAssetManager.h
Crisspl: > Yeah i actually added createAndOpenFile to ISystem

yea, saw that later, get rid of it

src/nbl/ui/CWindowX11.cpp
devshgraphicsprogramming: this is X11 we're talking about

include/CConcurrentObjectCache.h
devshgraphicsprogramming: `mutable` on the lock or?

include/nbl/ui/IWindowManager.h
devshgraphicsprogramming: public virtual inheritance

include/CConcurrentObjectCache.h
Crisspl: yea

include/nbl/ui/IWindowManager.h
Crisspl: why virtual

devshgraphicsprogramming: because almost everything inherits from `IReferenceCounted` so you eventually always run into diamonds involving `IReferenceCounted`, plus you really dont want two copies of IReferenceCounted in your hierarchy (think about how having two ref counters would fuck everything up in terms of autodestruction)

Crisspl: yea, i just tend to avoid virtual inheritance unless theres an actual need

devshgraphicsprogramming: yeah IReferenceCounted is the only place you want to be using virtual inheritance

include/nbl/asset/IAssetManager.h
```diff
@@ -320,12 +320,12 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
...
             _override->getLoadFilename(filePath, ctx, _hierarchyLevel);
-            system::IFile* file = m_system->createAndOpenFile(filePath.c_str());
 
-            SAssetBundle asset = getAssetInHierarchy_impl<RestoreWholeBundle>(file, _filePath, _params, _hierarchyLevel, _override);
-
-            if (file)
-                file->drop();
+            system::ISystem::future_t<core::smart_refctd_ptr<system::IFile>> future;
+            bool validInput = m_system->createFile(future, _filePath, system::IFile::ECF_READ);
+            assert(validInput);
```
devshgraphicsprogramming, line 326: what's the boolean and the assertion?

include/nbl/system/ICancellableAsyncQueueDispatcher.h
```diff
@@ -165,7 +165,7 @@ class ICancellableAsyncQueueDispatcher : public IAsyncQueueDispatcher<CRTP, Requ
     {
         auto& req_base = static_cast<request_base_t&>(req);
         auto* future = static_cast<future_t<T>*>(req.future);
-        future->notify(std::forward(args)...);
+        future->notify(std::forward<Args...>(args)...);
```
devshgraphicsprogramming, line 168: when you do perfect forwarding you put the `...` in only one place, at the end

devshgraphicsprogramming: see other usages in the engine

include/nbl/system/ISystem.h
```diff
@@ -197,23 +198,10 @@ class ISystem final : public core::IReferenceCounted
...
-        future_t<core::smart_refctd_ptr<IFile>> future;
-        if (!createFile(future, filename, IFile::ECF_WRITE))
-            return nullptr;
-
-        return future.get();
-    }
 
+    virtual void seek(IFile* file, uint32_t pos) = 0;
+    virtual void read(IFile* file, char* outData, size_t count) const;
+    virtual size_t getPos(IFile* file) const = 0;
```
devshgraphicsprogramming, line 204: no, no, no 

bad API (multithreaded operation on the same file impossible)

We've already established with @Crisspl (see other PR) that files should not be keeping positions as intermediate state, the read and write calls should take absolute offsets into the file.

include/nbl/ui/CWindowX11.h
```diff
@@ -1,9 +1,9 @@
 #ifndef __C_WINDOW_X11_H_INCLUDED__
 #define __C_WINDOW_X11_H_INCLUDED__
 
+#ifdef _NBL_PLATFORM_LINUX_x1
```
devshgraphicsprogramming, line 4: typo?

include/nbl/asset/IAssetManager.h
```diff
@@ -219,9 +219,9 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
 
             IAssetLoader::SAssetLoadContext ctx{params, _file};
 
-            std::string filename = _file ? _file->getFileName().c_str() : _supposedFilename;
+            std::string filename = _file ? _file->getFileName().string() : _supposedFilename;
             system::IFile* file = _override->getLoadFile(_file, filename, ctx, _hierarchyLevel); // WARNING: mem-leak possibility: _override should return smart_ptr<IReadFile> (TODO, inspect this)
-            filename = file ? file->getFileName().c_str() : _supposedFilename;
+            filename = file ? file->getFileName().string() : _supposedFilename;
```
devshgraphicsprogramming: fil the memleak issue, also use `std::filesystem::path` for the `filename` throughout consistently

src/nbl/asset/utils/IGLSLEmbeddedIncludeLoader.h
```diff
@@ -52,12 +52,19 @@ class IGLSLEmbeddedIncludeLoader : public IBuiltinIncludeLoader
...
 			auto path = "nbl/builtin/" + _name;
-			auto data = fs->loadBuiltinData(path);
-			if (!data)
-				return "";
-			auto begin = reinterpret_cast<const char*>(data->getPointer());
-			auto end = begin+data->getSize();
-			return std::string(begin,end);
+			system::ISystem::future_t<core::smart_refctd_ptr<system::IFile>> future;
+			bool validInput = s->createFile(future, path, system::IFile::ECF_READ);
+			assert(validInput);
```
devshgraphicsprogramming, line 57: what's the boolean? why the assert?

src/nbl/asset/utils/IGLSLEmbeddedIncludeLoader.h
```diff
@@ -52,12 +52,19 @@ class IGLSLEmbeddedIncludeLoader : public IBuiltinIncludeLoader
...
+			core::smart_refctd_ptr<system::IFile> file = future.get();
+
+			size_t fileSize = file->getSize();
+			std::string content(fileSize, '/0');
+			system::ISystem::future_t<uint32_t> read_future;
+			validInput = s->readFile(read_future, file.get(), content.data(), 0, fileSize);
+			assert(validInput);
```
devshgraphicsprogramming, line 64: again same question

include/nbl/system/ISystem.h
sadiuk: Yeah i moved them there cause criss said we probably should have those functions somewhere in the future. Not using them though.

include/nbl/system/ICancellableAsyncQueueDispatcher.h
sadiuk: ah yeah

include/nbl/asset/IAssetManager.h
sadiuk: `createFile` returns false in case the filename is too long, so i assumed an assertion would be enough

include/nbl/system/ISystem.h
Crisspl: wait what, no, you must have misunderstood something. We dont want any of this in public API, we want to have file access API that allows access like memory (`read(offset,size)`, `write(offset,size)`)

include/nbl/asset/IAssetManager.h
devshgraphicsprogramming: how long is too long?

no assertion is not acceptable (it crashes the program, but only in debug mode), you need to fail gracefully.

include/nbl/system/ISystem.h
devshgraphicsprogramming: basically like OpenGL/Vulkan copy/update buffer sub data but for a file

include/nbl/system/ISystem.h
```diff
@@ -203,6 +203,56 @@ class ISystem final : public core::IReferenceCounted
     virtual void read(IFile* file, char* outData, size_t count) const;
     virtual size_t getPos(IFile* file) const = 0;
 
+
+    inline core::smart_refctd_ptr<asset::ICPUBuffer> loadBuiltinData(const std::string& builtinPath)
```
devshgraphicsprogramming, line 207: maybe dont declare this function as `inline` and hide it in a .cpp

include/nbl/system/ISystem.h
```diff
@@ -203,6 +203,56 @@ class ISystem final : public core::IReferenceCounted
...
+    {
+#ifdef _NBL_EMBED_BUILTIN_RESOURCES_
+        std::pair<const uint8_t*, size_t> found = nbl::builtin::get_resource<StringUniqueType>();
+        if (found.first && found.second)
+        {
+            auto returnValue = core::make_smart_refctd_ptr<asset::ICPUBuffer>(found.second);
+            memcpy(returnValue->getPointer(), found.first, returnValue->getSize());
+            return returnValue;
+        }
+        return nullptr;
```
devshgraphicsprogramming, line 250: this is duplicate code, could you maybe make a private common function to implement this for both the templated (static) and non templated (runtime) version?

src/nbl/asset/interchange/CImageLoaderJPG.cpp
```diff
@@ -147,7 +147,7 @@ namespace jpeg
 #endif // _NBL_COMPILE_WITH_LIBJPEG_
 
 //! returns true if the file maybe is able to be loaded by this class
-bool CImageLoaderJPG::isALoadableFileFormat(io::IReadFile* _file) const
+bool CImageLoaderJPG::isALoadableFileFormat(system::IFile* _file) const
```
devshgraphicsprogramming, line 150: in the asset `.cpp`s you can just do `using namespace nbl::system;` to save yourself some writin (but obvs not in headers)

devshgraphicsprogramming:
@Crisspl should `IFile` keep a refcounted smart pointer to the `ISystem` that created them?

That would simplify a lot of stuff:
- making sure the creator/manager doesn't die before the last item it manages
- loaders and asset manager keeping a pointer to `ISystem` but there's no guarantee that its the same one that made the file you're currently passing in

devshgraphicsprogramming:
@Crisspl should `IFile` keep a refcounted smart pointer to the `ISystem` that created them?

That would simplify a lot of stuff:
- making sure the creator/manager doesn't die before the last item it manages
- loaders and asset manager keeping a pointer to `ISystem` but there's no guarantee that its the same one that made the file you're currently passing in

include/nbl/system/CMemoryFile.h
```diff
@@ -0,0 +1,54 @@
...
+		}
+		memcpy(buffer, m_buffer.data(), sizeToRead);
+		return sizeToRead;
+	}
+
+	virtual int32_t write(const void* buffer, uint32_t offset, uint32_t sizeToWrite) override
+	{
+		if (offset + sizeToWrite > m_buffer.size())
+		{
+			m_buffer.resize(sizeToWrite);
```
Crisspl, line 37: should be `resize(offset+sizeToWrite)`

include/nbl/system/IFile.h
```diff
@@ -43,6 +43,9 @@ class IFile : public core::IReferenceCounted
 		return (m_flags & ECF_COHERENT) == ECF_COHERENT;
 	}
 
+protected:
+	virtual int32_t read(void* buffer, uint32_t offset, uint32_t sizeToRead) = 0;
+	virtual int32_t write(const void* buffer, uint32_t offset, uint32_t sizeToWrite) = 0;
```
Crisspl, line 48: size_t instead of uint32_t

include/nbl/system/ISystem.h
```diff
@@ -35,6 +37,7 @@ class ISystem final : public core::IReferenceCounted
     enum E_REQUEST_TYPE
     {
         ERT_CREATE_FILE,
+        ERT_CREATE_BUFFERED_FILE,
```
Crisspl, line 40: you dont need a request to thread to create memory file

src/nbl/asset/interchange/CGLILoader.cpp
```diff
@@ -251,20 +251,20 @@ namespace nbl
 		bool CGLILoader::isALoadableFileFormat(system::IFile* _file) const
 		{
 			const auto fileName = std::string(_file->getFileName().string());
-			const auto beginningOfFile = _file->getPos();
 
 			constexpr auto ddsMagic = 0x20534444;
 			constexpr std::array<uint8_t, 12> ktxMagic = { 0xAB, 0x4B, 0x54, 0x58, 0x20, 0x32, 0x30, 0xBB, 0x0D, 0x0A, 0x1A, 0x0A };
 			constexpr std::array<uint8_t, 16> kmgMagic = { 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55 };
 
 			system::ISystem::future_t<uint32_t> future1;
```
Crisspl, line 259: you dont need `future1`, unused

src/nbl/asset/interchange/CGLILoader.h
```diff
@@ -20,11 +20,12 @@ namespace asset
 //! Texture loader capable of loading in .ktx, .dds and .kmg file extensions
 class CGLILoader final : public asset::IAssetLoader
 {
+	core::smart_refctd_ptr<system::ISystem> m_system;
```
Crisspl, line 23: you could move it to IAssetLoader, all loaders will use ISystem anyway

src/nbl/asset/interchange/CGraphicsPipelineLoaderMTL.cpp
```diff
@@ -78,10 +78,12 @@ void CGraphicsPipelineLoaderMTL::initialize()
...
-    auto default_mtl_file = m_assetMgr->getFileSystem()->createMemoryReadFile(DUMMY_MTL_CONTENT, strlen(DUMMY_MTL_CONTENT), "default IrrlichtBAW material");
+    system::ISystem::future_t<core::smart_refctd_ptr<system::CMemoryFile>> future;
+    m_assetMgr->getSystem()->createBufferedFile(future, "default IrrlichtBAW material", system::IFile::ECF_READ);
+    auto default_mtl_file = future.get();
 
     SAssetLoadParams assetLoadParams;
-    auto bundle = loadAsset(default_mtl_file, assetLoadParams, &dfltOver);
+    auto bundle = loadAsset(default_mtl_file.get(), assetLoadParams, &dfltOver);
 
     default_mtl_file->drop();
```
Crisspl, line 88: dont call drop on smart pointer

src/nbl/asset/interchange/CGraphicsPipelineLoaderMTL.cpp
```diff
@@ -78,10 +78,12 @@ void CGraphicsPipelineLoaderMTL::initialize()
     }
 
     // default pipelines
-    auto default_mtl_file = m_assetMgr->getFileSystem()->createMemoryReadFile(DUMMY_MTL_CONTENT, strlen(DUMMY_MTL_CONTENT), "default IrrlichtBAW material");
+    system::ISystem::future_t<core::smart_refctd_ptr<system::CMemoryFile>> future;
+    m_assetMgr->getSystem()->createBufferedFile(future, "default IrrlichtBAW material", system::IFile::ECF_READ);
```
Crisspl, line 82: change the filename to "Nabla default MTL material" and store it in `constexpr const char*` (we dont want literal constants in the code)

include/nbl/system/CMemoryFile.h
```diff
@@ -0,0 +1,54 @@
...
+
+	}
+
+	virtual const std::filesystem::path& getFileName() const override
+	{
+		return m_name;
+	}
+
+	virtual void* getMappedPointer() override { return nullptr; }
+	virtual const void* getMappedPointer() const override { return nullptr; }
```
Crisspl, line 21: make it return m_buffer.data()
and force constructor adding MAPPABLE and COHERENT to flags

src/nbl/asset/interchange/CGraphicsPipelineLoaderMTL.cpp
devshgraphicsprogramming: yeah its a remnant of the old API @sadiuk, creation methods didnt use to return smart pointers, now that they do, you shouldn't need to call drop at all

include/nbl/system/CMemoryFile.h
devshgraphicsprogramming: can you guys call it a `CFileView` instead of memroy file?

src/nbl/asset/interchange/CGLILoader.h
devshgraphicsprogramming: @Crisspl I already brought it up, but I'm of the opinion that the file should keep a smart pointer to the `system` that created it (and it needs for performing async reads and writes)

Because if someone gets the idea to create multiple system objects and multiple files, this turns into a clusterfuck (wrong system reading a file that doesn't belong to it)

Also file should keep the system alive while the file is alive, so a backward pointer makes sense.

include/nbl/system/ISystem.h
```diff
@@ -50,6 +53,13 @@ class ISystem final : public core::IReferenceCounted
         char filename[MAX_FILENAME_LENGTH] {};
         IFile::E_CREATE_FLAGS flags;
     };
+    struct SRequestParams_CREATE_BUFFERED_FILE : SRequestParamsBase<ERT_CREATE_BUFFERED_FILE>
```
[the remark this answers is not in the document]
devshgraphicsprogramming: I second this

include/nbl/system/ISystem.h
devshgraphicsprogramming: I second this

include/nbl/system/IFile.h
devshgraphicsprogramming: yeah you dont want to fail on files > 4GB

src/nbl/asset/interchange/CGLILoader.h
Crisspl: yeah, but in ISystem.h there needs some declaration like `smart_refctd_ptr<IFile>` and in IFile then would need to be a declaration like `smart_refctd_ptr<ISystem>`, but you cannot do this with forward-declaring the classes since at that point it must be known that T in smart_refctd_ptr<T> is derivative of IReferenceCounted. So it's basically impossible to do easy way (well, unless we get rid of static_assert in smart_refctd_ptr). So we postponed it to not lose time on not that important things

include/nbl/system/CMemoryFile.h
Crisspl: kk

src/nbl/asset/interchange/CGLILoader.h
devshgraphicsprogramming: > So it's basically impossible to do easy way (well, unless we get rid of static_assert in smart_refctd_ptr).

Is it just the assert that's causing this? not the function definitions, etc?

Yeah this is kinda annoying, can we move the assert somewhere else?

Crisspl: Hm, right, whole class definition must be known to use it with smart_refctd_ptr, with or without static_assert. So eh, it's basically impossible to do

include/nbl/ui/CWindowManagerWin32.h
```diff
@@ -19,8 +19,18 @@ namespace nbl::ui
 	{
 	public:
 		CWindowManagerWin32() = default;
-		~CWindowManagerWin32() = default;
+		~CWindowManagerWin32() {};
 
+		IWindowWin32::native_handle_t createNativeWindow(int _x, int _y, uint32_t _w, uint32_t _h, IWindow::E_CREATE_FLAGS _flags, const std::string_view& caption)
```
Crisspl, line 24: dont make it public,
IWindowManager interface should creater and return `IWindow` and each derived class should implement it

include/nbl/ui/CWindowWin32.h
```diff
@@ -9,42 +9,44 @@
...
+	//
+	//	m_width = rect.right - rect.left;
+	//	m_height = rect.bottom - rect.top;
+	//	// TODO m_flags
+	//}
 
 	native_handle_t getNativeHandle() const override { return m_native; }
 
-	static core::smart_refctd_ptr<CWindowWin32> create(core::smart_refctd_ptr<system::ISystem>&& sys, int32_t _x, int32_t _y, uint32_t _w, uint32_t _h, E_CREATE_FLAGS _flags)
+	static core::smart_refctd_ptr<CWindowWin32> create(CWindowManagerWin32* winManager, core::smart_refctd_ptr<system::ISystem>&& sys, int32_t _x, int32_t _y, uint32_t _w, uint32_t _h, E_CREATE_FLAGS _flags, const std::string_view& caption)
```
devshgraphicsprogramming, line 33: why do you need both a pointer to the window manager and the ISystem!?

src/nbl/asset/interchange/CImageLoaderPNG.h
```diff
@@ -23,16 +23,16 @@ namespace asset
 class CImageLoaderPng : public asset::IAssetLoader
 {
     core::smart_refctd_ptr<system::ISystem> m_system;
+public:
     struct SUserData
```
devshgraphicsprogramming, line 27: we usually call these `SContext`

src/nbl/ui/CWindowWin32.cpp
```diff
@@ -9,16 +11,21 @@ namespace nbl {
 namespace ui
 {
 
-	CWindowWin32::CWindowWin32(core::smart_refctd_ptr<system::ISystem>&& sys, int _x, int _y, uint32_t _w, uint32_t _h, E_CREATE_FLAGS _flags)
+	CWindowWin32::CWindowWin32(CWindowManagerWin32* winManager, core::smart_refctd_ptr<system::ISystem>&& sys, int _x, int _y, uint32_t _w, uint32_t _h, E_CREATE_FLAGS _flags, const std::string_view& caption) : m_windowManager(winManager)
 	{
 		m_width = _w; m_height = _h;
-		windowThreadHandler.createWindow(_x, _y, _w, _h, _flags, m_native);
+		m_native = m_windowManager->createNativeWindow(_x, _y, _w, _h, _flags, caption);
```
devshgraphicsprogramming, line 17: this is really roundabout, I'd much rather prefer a constructor that takes a native window and the window manager

devshgraphicsprogramming: because we dont want to use exceptions, so the only way to have a "constructor which can fail" is to have a private "constructor that cannot fail" and either a friended factory class OR a static create method, which does all the setup and validation before calling the ctor.

src/nbl/asset/interchange/CGLILoader.h
devshgraphicsprogramming: ok so the solution here is to split `ISystem` and to inherit from `IFileOperator` and `IFileFactory` where:
- File Factory would hand out the `smart_refctd_ptr<IFile>` and the files need not know about the factory
- File Operator's protected methods would operate on "dumb pointers" to `IFile`s and its the only reason why an `IFile` wants to hold a smartpointer to an `IFileOperator` (to ensure operator is still alive when it wants to use the operator to implement its own methods)

Crisspl: do you think both of them would have a thread? Or they share one thread under the hood?
Anyway.. IMO it's easier and cleaner to just have IFile hold ISystem with raw pointer (include ISystem in .cpp file and use grab/drop directly). Maybe even have a base class for backward linking (an object holding its creator) like
```
// IBackwardLink.h
// have it templated for case when theres more than 1 backward links needed (possible?)
template <typename T> // T is some forward-declared class
class IBackwardLink
{
explicit IBackwardLink(T* c);
~IBackwardLink();
T* getCreator() { return creator; }
private:
T* creator;
};

//IBackwardLink.cpp
// have to explicitely instantiate (force compiling them) those ctor and dtor when needed
template <typename T>
IBackwardLink<T>::IBackwardLink(T* c) : creator(c) { c->grab(); }
template<typename T>
IBackwardLink<T>::~IBackwardLink() { creator->drop(); }
```

A little weird but DRY and dont have to split ISystem which might be problematic. IFileFactory would have to have pointer to IFileOperator to give it to files being created (if you want files to be able to be read and written to w/o direct usage of IFileOperator API as you said). AND often it's needed to be able to both create/open and operate on files, so they would passed in pair to things (IAssetManager, IGLSLCompiler, more?)
I'm team IBackwardLink

include/nbl/ui/CWindowWin32.h
Crisspl: why not?

src/nbl/ui/CWindowWin32.cpp
Crisspl: window manager should have `createWindow()` method which returns `smart_refctd_ptr<IWindow>`. Already commented about this, a while ago

Crisspl: `createWindow()` should validate and return nullptr in case of failed validation and if native handle creation failed

include/nbl/ui/CWindowWin32.h
Crisspl: ok, im not sure what ISystem is needed for here

include/nbl/ui/CWindowManagerWin32.h
devshgraphicsprogramming: important

include/nbl/ui/CWindowX11.h
devshgraphicsprogramming: deffo  a typo

include/nbl/asset/IAssetManager.h
devshgraphicsprogramming: again

include/nbl/ui/CWindowWin32.h
devshgraphicsprogramming: > why not?

because I assume the window manager will hold a reference to a IUI and that will hold a pointer to ISystem

src/nbl/ui/CWindowWin32.cpp
devshgraphicsprogramming: > `createWindow()` should validate and return nullptr in case of failed validation and if native handle creation failed

yep so thats stuff that happens before calling a constructor on the window class

include/nbl/core/containers/CCircularBuffer.h
```diff
@@ -163,12 +163,12 @@ class CCircularBufferBase : public Base
...
         auto xormask = static_cast<typename base_t::atomic_alive_flags_block_t::value_type>(1) << local_ix;
 
-        typename base_t::atomic_alive_flags_block_t* flags = base_t::getAliveFlagsStorage();
+        auto* flags = base_t::getAliveFlagsStorage();
         flags[block_n].fetch_xor(xormask);
     }
 
 private:
-    static inline counter_t wrapAround(counter_t x)
+    counter_t wrapAround(counter_t x)
```
devshgraphicsprogramming, line 171: why did you remove static inline?

include/nbl/ui/CWindowManagerWin32.h
```diff
@@ -20,7 +20,26 @@ namespace nbl::ui
...
+				creationParams.flags,
+				creationParams.windowCaption);
+			// TODO: params validation
+			if (false)
+			{
+				return nullptr;
+			}
+			return core::make_smart_refctd_ptr<CWindowWin32>(this, std::move(creationParams), handle);
+		}
+		virtual void destroyWindow(IWindow* wnd) override
```
devshgraphicsprogramming, line 38: unless you plan on inheriting from this class again and overriding these overrides, dont put `virtual` in front of the function (moreover you could make this the `final` override)

include/nbl/system/IAsyncQueueDispatcher.h
```diff
@@ -15,8 +15,11 @@ namespace impl
     class IAsyncQueueDispatcherBase
     {
     public:
+        IAsyncQueueDispatcherBase() {};
+        virtual ~IAsyncQueueDispatcherBase() {};
         struct request_base_t
         {
+            virtual ~request_base_t() {}
```
Crisspl, line 22: why would we need it?

include/nbl/system/IAsyncQueueDispatcher.h
```diff
@@ -15,8 +15,11 @@ namespace impl
     class IAsyncQueueDispatcherBase
     {
     public:
+        IAsyncQueueDispatcherBase() {};
+        virtual ~IAsyncQueueDispatcherBase() {};
```
Crisspl, line 19: it's not meant to be used with runtime polymorphism

include/nbl/system/IAsyncQueueDispatcher.h
```diff
@@ -73,7 +77,12 @@ class IAsyncQueueDispatcher : public IThreadHandler<CRTP, InternalStateType>, pu
         return x & Mask;
     }
 
+
 public:
+
+    IAsyncQueueDispatcher() {}
+    virtual ~IAsyncQueueDispatcher() {}
```
Crisspl, line 84: get rid of virtual, why do you need them actually?

include/nbl/system/IAsyncQueueDispatcher.h
Crisspl: why do you need them actually?

include/nbl/ui/CWindowManagerWin32.h
```diff
@@ -20,7 +20,26 @@ namespace nbl::ui
...
-
+		virtual core::smart_refctd_ptr<IWindow> createWindow(IWindow::SCreationParams&& creationParams) override
+		{
+			CWindowWin32::native_handle_t handle = createNativeWindow(creationParams.x,
+				creationParams.y,
+				creationParams.width,
+				creationParams.height,
+				creationParams.flags,
+				creationParams.windowCaption);
+			// TODO: params validation
```
Crisspl, line 31: also return nullptr in case of failed native handle creation

include/nbl/system/IAsyncQueueDispatcher.h
devshgraphicsprogramming: because stuff inherits from `IAsyncQueueDispatcher` and the correct destructors dont get called in the right order?

include/nbl/ui/CWindowManagerWin32.h
```diff
@@ -68,12 +87,15 @@ namespace nbl::ui
 				SRequestParams_CreateWindow createWindowParam;
 				SRequestParams_DestroyWindow destroyWindowParam;
 			};
+			SRequest() {}
+			~SRequest() {}
```
Crisspl, line 91: why?

include/nbl/system/IAsyncQueueDispatcher.h
devshgraphicsprogramming: ah yeah I see the comment about the class not being meant for runtime polymorphism

include/nbl/ui/CWindowManagerWin32.h
```diff
@@ -92,7 +114,9 @@ namespace nbl::ui
 			{
 				this->start();
 			}
-			~CThreadHandler() {};
+			~CThreadHandler() override
```
Crisspl, line 117: this is extremely weird, thread handlers shouldnt anything virtual,
They use 100% static polymorphism (CRTP stuff)

include/nbl/ui/CWindowWin32.h
```diff
@@ -16,32 +16,22 @@ class CWindowWin32 final : public IWindowWin32
...
-	//	m_width = rect.right - rect.left;
-	//	m_height = rect.bottom - rect.top;
-	//	// TODO m_flags
-	//}
+	CWindowWin32(CWindowManagerWin32* winManager, SCreationParams&& params, native_handle_t hwnd);
 
 	native_handle_t getNativeHandle() const override { return m_native; }
 
-	static core::smart_refctd_ptr<CWindowWin32> create(CWindowManagerWin32* winManager, core::smart_refctd_ptr<system::ISystem>&& sys, int32_t _x, int32_t _y, uint32_t _w, uint32_t _h, E_CREATE_FLAGS _flags, const std::string_view& caption)
+	static core::smart_refctd_ptr<CWindowWin32> create(CWindowManagerWin32* winManager, core::smart_refctd_ptr<system::ISystem>&& sys, SCreationParams&& params)
```
Crisspl, line 23: `CWindow*::create()` should exist any more, only windows manager should be able to create windows

src/nbl/video/IAPIConnection.cpp
```diff
@@ -41,7 +41,9 @@ IAPIConnection::IAPIConnection()
 #else
         "";
 #endif
-    m_system = core::make_smart_refctd_ptr<system::ISystem>(std::move(builtinResourceDirectoryPath));
+    assert(false);
+    //TODO: system implementation
+    m_system = core::make_smart_refctd_ptr<system::ISystem>(nullptr);
```
Crisspl, line 46: what happened here?

src/nbl/asset/interchange/CGLILoader.h
devshgraphicsprogramming: > do you think both of them would have a thread? Or they share one thread under the hood?
Well the most derived one (operator) would hold most of the stuff, the base (factory) is just an interface.

> Anyway.. IMO it's easier and cleaner to just have IFile hold ISystem with raw pointer (include ISystem in .cpp file and use grab/drop directly).
Would the usage of the backward link util class be via inheritance, or composition?
I'm liking this direction, i'd call it `IFactoryKeepalive` or `IFactoryLifetimeSupport` though.

In theory definitely better than forced splits because of how `smart_refctd_ptr` works.

But there's one huge caveat, this is essentially a `smart_refctd_ptr` but the constructor and destructor are not inlined, but defined in a different TU and explicitly instantiated for every class you might want to use them with.

include/nbl/system/ISystem.h
```diff
@@ -23,7 +23,7 @@ class ISystem final : public core::IReferenceCounted
 
     public:  
         virtual core::smart_refctd_ptr<IFile> createFile(ISystem* sys, const std::filesystem::path& filename, IFile::E_CREATE_FLAGS flags) = 0;
-        virtual core::smart_refctd_ptr<CMemoryFile> createBufferedFile(ISystem* sys, const std::filesystem::path& filename, IFile::E_CREATE_FLAGS flags) = 0;
+        virtual core::smart_refctd_ptr<CFileView> createBufferedFile(ISystem* sys, const std::filesystem::path& filename, IFile::E_CREATE_FLAGS flags) = 0;
```
devshgraphicsprogramming, line 26: dont even need a factory for this, CFileView can create itself no problem

include/nbl/ui/CWindowManagerWin32.h
devshgraphicsprogramming: above NBL_PLATFORM_WINDOWS

include/nbl/system/ISystem.h
Crisspl: lets have this function in ISystem API and public ctor in CFileView as well (CFileView.h should be in `include/` ofc)

Crisspl: also you dont need it to be `virtual`

include/nbl/system/CFileWin32.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+void* nbl::system::CFileWin32::getMappedPointer()
+{
+	return nullptr;
+}
+
+const void* nbl::system::CFileWin32::getMappedPointer() const
+{
+	return nullptr;
+}
```
Crisspl, line 48: mark as TODO

include/nbl/system/CFileWin32.cpp
```diff
@@ -0,0 +1,70 @@
...
+#define HIDWORD(_qw)    ((DWORD)(((_qw) >> 32) & 0xffffffff))
+
+nbl::system::CFileWin32::CFileWin32(const std::string_view& _filename, std::underlying_type_t<E_CREATE_FLAGS> _flags) : base_t(_flags), m_filename{ _filename }
+{
+	DWORD access = m_flags | ECF_READ_WRITE ? GENERIC_READ | GENERIC_WRITE :
+		(m_flags | ECF_READ ? GENERIC_READ : (m_flags | ECF_WRITE ? GENERIC_WRITE : 0));
+	const bool canOpenWhenOpened = false;
+	SECURITY_ATTRIBUTES secAttribs{ sizeof(SECURITY_ATTRIBUTES), nullptr, FALSE };
+	m_native = CreateFile(m_filename.string().data(), access, canOpenWhenOpened, &secAttribs, OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
+	if (m_native != INVALID_HANDLE_VALUE) [[likely]] // let this idle here until c++20 :)
```
Crisspl, line 14: what happens on c++20?

include/nbl/system/CFileWin32.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+	return nullptr;
+}
+
+const void* nbl::system::CFileWin32::getMappedPointer() const
+{
+	return nullptr;
+}
+
+int32_t nbl::system::CFileWin32::read(void* buffer, size_t offset, size_t sizeToRead)
```
Crisspl, line 50: why does it return signed type?
it should be unsigned 64bit type (return 0 on error)

include/nbl/system/CFileWin32.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+	seek(offset);
+	DWORD numOfBytesRead;
+	ReadFile(m_native, buffer, sizeToRead, &numOfBytesRead, nullptr);
+	return numOfBytesRead;
+}
+
+int32_t nbl::system::CFileWin32::write(const void* buffer, size_t offset, size_t sizeToWrite)
```
Crisspl, line 58: same as read, unsigned and 64bit

include/nbl/system/CFileWin32.cpp
```diff
@@ -0,0 +1,70 @@
+#include "C:/dev/work/Nabla-vs/src/nbl/CMakeFiles/Nabla.dir/Debug/cmake_pch.hxx"
+#include "CFileWin32.h"
+
+#define LODWORD(_qw)    ((DWORD)(_qw))
+#define HIDWORD(_qw)    ((DWORD)(((_qw) >> 32) & 0xffffffff))
```
Crisspl, line 5: you sure winapi headers dont already define those?

include/nbl/system/CFileWin32.cpp
```diff
@@ -0,0 +1,70 @@
+#include "C:/dev/work/Nabla-vs/src/nbl/CMakeFiles/Nabla.dir/Debug/cmake_pch.hxx"
```
Crisspl, line 1: wtf?

include/nbl/system/CFileWin32.h
```diff
@@ -0,0 +1,36 @@
+#ifndef	CFILE_WIN32_H
```
Crisspl, line 1: include guards within engine source must look like `__NBL_<header name>_INCLUDED__`

include/nbl/system/CSystemWin32.h
```diff
@@ -0,0 +1,14 @@
+#ifndef CSYSTEM_WIN32
```
Crisspl, line 1: include guards within engine source must look like `__NBL_<header name>_INCLUDED__`

devshgraphicsprogramming: `_NBL_<NAMESPACE>_<ADDITIONAL_NAMESPACE>_<HEADER_NAME>_INCLUDED_`

include/nbl/system/CFileWin32.h
devshgraphicsprogramming: no exactly, but close

include/nbl/system/CSystemWin32.h
Crisspl: thats new but yeah good idea

include/nbl/system/CSystemWin32.h
```diff
@@ -1,14 +1,40 @@
...
-
+#include "CFileWin32.h"
 namespace nbl::system
 {
-	class CSystemWin32 : public ISystem
-	{
-
-	};
+	
+class CSystemCallerWin32 : public ISystem::ISystemCaller
```
Crisspl, line 8: make it `final`

include/nbl/system/IFile.h
```diff
@@ -8,9 +8,12 @@
 namespace nbl {
 namespace system
 {
-
 class IFile : public core::IReferenceCounted
 {
+	// Don't like this approach, gonna stack up many forward declarations, @criss any suggestions?
+	// I also had an idea to make interfaces public in the implementations of IFile
+	// and static_cast files before calling, but that would invalidate any other IFile descendants calls
+	friend class CSystemCallerWin32;
```
Crisspl, line 16: i dont get it, whats the problem?
`IFile` must not know about any platform-specific stuff, get rid of this `friend`

Crisspl: just `static_cast` in implementation classes, why would you need this `friend`?

include/nbl/system/CFileWin32.h
```diff
@@ -1,5 +1,5 @@
-#ifndef	CFILE_WIN32_H
-#define	CFILE_WIN32_H
+#ifndef	_NBL_SYSTEM_CFILEWIN32_H_INCLUDED_
+#define	_NBL_SYSTEM_CFILEWIN32_H_INCLUDED_
```
Crisspl, line 2: double `_` on both sides

devshgraphicsprogramming: I believe thats reserved in the C preprocessor

include/nbl/system/ISystem.h
```diff
@@ -16,14 +16,14 @@ namespace system
 class ISystem final : public core::IReferenceCounted
 {
 public:
+    static core::smart_refctd_ptr<ISystem> create();
```
devshgraphicsprogramming, line 19: nope, you dont want the creator in the interface, the user will still need to know what implementation to construct.

src/nbl/asset/interchange/CGraphicsPipelineLoaderMTL.cpp
```diff
@@ -78,10 +78,8 @@ void CGraphicsPipelineLoaderMTL::initialize()
     }
 
     // default pipelines
-    system::ISystem::future_t<core::smart_refctd_ptr<system::CFileView>> future;
     constexpr std::string_view filename = "Nabla default MTL material";
-    m_assetMgr->getSystem()->createFileView(DUMMY_MTL_CONTENT, sizeof(DUMMY_MTL_CONTENT), system::IFile::ECF_READ, filename);
-    auto default_mtl_file = future.get();
+    auto default_mtl_file = m_assetMgr->getSystem()->createFileView(DUMMY_MTL_CONTENT, sizeof(DUMMY_MTL_CONTENT), system::IFile::ECF_READ, filename);
```
devshgraphicsprogramming, line 82: this is ugly, make file view be able to create itself (why do you even need the system to create a file view!?_

src/nbl/system/ISystem.cpp
```diff
@@ -0,0 +1,13 @@
...
+namespace nbl::system
+{
+core::smart_refctd_ptr<ISystem> nbl::system::ISystem::create()
+{
+	core::smart_refctd_ptr<ISystem::ISystemCaller> caller = nullptr;
+#ifdef _NBL_PLATFORM_WINDOWS_
+	caller = core::make_smart_refctd_ptr<CSystemCallerWin32>();
+#endif
+	return core::make_smart_refctd_ptr<ISystem>(std::move(caller));
+}
```
devshgraphicsprogramming, line 12: oh god no, get rid of this

devshgraphicsprogramming: or more accurately, move it to CommonAPI.h for the examples

src/nbl/video/IAPIConnection.cpp
```diff
@@ -41,9 +41,13 @@ IAPIConnection::IAPIConnection()
 #else
         "";
 #endif
-    assert(false);
-    //TODO: system implementation
-    m_system = core::make_smart_refctd_ptr<system::ISystem>(nullptr);
+    core::smart_refctd_ptr<system::ISystem::ISystemCaller> caller;
+#ifdef _NBL_PLATFORM_WINDOWS_
+    caller = core::make_smart_refctd_ptr<system::CSystemCallerWin32>();
```
devshgraphicsprogramming, line 46: reasoning? why is a system caller created here!?

include/nbl/ui/IInputEventChannel.h
```diff
@@ -97,21 +97,28 @@ namespace impl
 
 struct SMouseEvent
 {
+    SMouseEvent()
+    {
+        using namespace std::chrono;
+        timeStamp = duration_cast<microseconds>(system_clock::now().time_since_epoch());
```
devshgraphicsprogramming, line 103: oh fuck no, don't do that, either get the timestamp directly from the OS, or record the time **as soon as you return from polling for the event**

include/nbl/ui/IInputEventChannel.h
```diff
@@ -152,13 +160,19 @@ class IMouseEventChannel : public impl::IEventChannelBase<SMouseEvent>
 
 struct SKeyboardEvent
 {
+    SKeyboardEvent() {
+        using namespace std::chrono;
+        timeStamp = duration_cast<microseconds>(system_clock::now().time_since_epoch());
+    }
```
devshgraphicsprogramming, line 166: this constructor and timeStamp member variable repeats a lot, have you thought about making a common `SEvent` and having the other structs inherit from it?

devshgraphicsprogramming, line 166: this constructor and timeStamp member variable repeats a lot, have you thought about making a common `SEvent` and having the other structs inherit from it?

include/nbl/system/CFileWin32.h
Crisspl: well, we do this all over the engine XD

include/nbl/ui/CWindowWin32.h
```diff
@@ -28,37 +28,57 @@ class CWindowWin32 final : public IWindowWin32
...
-	std::map<HANDLE, core::smart_refctd_ptr<IKeyboardEventChannel>> m_keyboardEventChannel;
-
+	std::map<HANDLE, core::smart_refctd_ptr<IMouseEventChannel>> m_mouseEventChannels;
+	std::map<HANDLE, core::smart_refctd_ptr<IKeyboardEventChannel>> m_keyboardEventChannels;
+	
+	/* 
+	*  Storing this data is required for the device removal to work properly
+	*  When you get a message about the device removal, its type isn't accessible anymore.
+	*/
+	std::map<HANDLE, uint32_t> m_deviceTypes;
```
Crisspl, line 38: NABLA CONTAINERS!!! nbl::core::

include/nbl/system/CFileWin32.h
devshgraphicsprogramming: point is, we probably shouldn't XD

include/IEventReceiver.h
```diff
@@ -6,7 +6,7 @@
 #ifndef __NBL_I_EVENT_RECEIVER_H_INCLUDED__
 #define __NBL_I_EVENT_RECEIVER_H_INCLUDED__
```
devshgraphicsprogramming, line 7: noke this file

include/nbl/asset/IAssetManager.h
```diff
@@ -86,6 +86,7 @@ class IAssetManager : public core::IReferenceCounted, public core::QuitSignallin
         static void refCtdDispose(T* _asset) { _asset->drop(); }
 
         core::smart_refctd_ptr<system::ISystem> m_system;
+        system::logger_opt_smart_ptr m_logger;
```
devshgraphicsprogramming, line 89: why are you holding an logger in the asset manager?

include/nbl/asset/IImage.h
```diff
@@ -235,7 +236,7 @@ class IImage : public IDescriptor
 			E_SHARING_MODE								sharingMode = ESM_EXCLUSIVE;
 			core::smart_refctd_dynamic_array<uint32_t>	queueFamilyIndices = nullptr;
 			E_IMAGE_LAYOUT								initialLayout = EIL_UNDEFINED;
-
+			system::logger_opt_ptr						logger = nullptr;
```
devshgraphicsprogramming, line 239: why IImage holds onto a logger?

include/nbl/system/ILogger.h
```diff
@@ -0,0 +1,284 @@
...
+		void logP(const std::string_view& fmtString, ...)
+		{
+			if (ELL_PERFORMANCE & m_logLevelMask)
+			{
+				va_list args;
+				va_start(args, ELL_PERFORMANCE);
+				log_impl(fmtString, ELL_PERFORMANCE, args);
+				va_end(args);
+			}
+		}
```
devshgraphicsprogramming, line 99: why!?

AnastaZIuk:
@devshgraphicsprogramming @sadiuk please change base to https://github.com/Devsh-Graphics-Programming/Nabla/tree/newAPIExamplesPort and I will resolve conflicts unless Dan wants to

devshgraphicsprogramming:
Address the comments and implement X11 later on.

sadiuk:
> why does MTL need to know about the system?

Unfortunately that's necessary bc MTL loader creates a `CFileView` which needs an `ISystem` as a ctor param

include/nbl/ui/CWindowManagerX11.h
devshgraphicsprogramming: @sadiuk ?

src/nbl/ui/CWindowX11.cpp
devshgraphicsprogramming: TODO @sadiuk (might want to coordinate with @achalpandeyy to make an example that allows swapchain resizing)

include/nbl/ui/CWindowManagerX11.h
devshgraphicsprogramming: TODO: @sadiuk

src/nbl/ui/CWindowX11.cpp
devshgraphicsprogramming: I dont think you need a threadhandler for X11

include/nbl/ui/CWindowManagerWin32.h
devshgraphicsprogramming: @sadiuk TODO

include/nbl/ui/CWindowX11.h
devshgraphicsprogramming: @sadiuk TODO

include/nbl/asset/IAssetManager.h
devshgraphicsprogramming: @sadiuk urgent TODO

include/nbl/asset/IAssetManager.h
devshgraphicsprogramming: @sadiuk done?

src/nbl/asset/utils/IGLSLEmbeddedIncludeLoader.h
devshgraphicsprogramming: @sadiuk what did you change ?

include/nbl/system/IFile.h
devshgraphicsprogramming: @sadiuk  changed?

include/nbl/system/CMemoryFile.h
devshgraphicsprogramming: @sadiuk changed?

include/nbl/ui/CWindowManagerWin32.h
devshgraphicsprogramming: fixed it myself

include/nbl/core/containers/CCircularBuffer.h
devshgraphicsprogramming: ok, runtime sized variant, I see.

include/nbl/ui/CWindowManagerWin32.h
devshgraphicsprogramming: because of union I guess?

include/nbl/ui/CWindowManagerWin32.h
devshgraphicsprogramming: @sadiuk TODO

include/nbl/ui/IInputEventChannel.h
devshgraphicsprogramming: TODO: @sadiuk

include/nbl/ui/IInputEventChannel.h
devshgraphicsprogramming: also this is not when the time should be polled!

devshgraphicsprogramming: also this is not when the time should be polled!

include/nbl/asset/IAssetManager.h
devshgraphicsprogramming: @sadiuk fixed?

devshgraphicsprogramming:
> 
> 
> > why does MTL need to know about the system?
> 
> Unfortunately that's necessary bc MTL loader creates a `CFileView` which needs an `ISystem` as a ctor param

when? in the ctor?

Crisspl:
> > why does MTL need to know about the system?
> 
> Unfortunately that's necessary bc MTL loader creates a `CFileView` which needs an `ISystem` as a ctor param

huh? why CFileView needs ISystem for anything?

sadiuk:
> > > why does MTL need to know about the system?
> > 
> > 
> > Unfortunately that's necessary bc MTL loader creates a `CFileView` which needs an `ISystem` as a ctor param
> 
> huh? why CFileView needs ISystem for anything?

it's just right now `IFile`'s `read` and `write` methods are not virtual. they have a predefined algorithm that everything should go through ISystem first, which calls `read_impl` and `write_impl`. But yeah, it obvs makes sense to make it virtual.

devshgraphicsprogramming:
> 
> 
> > > > why does MTL need to know about the system?
> > > 
> > > 
> > > Unfortunately that's necessary bc MTL loader creates a `CFileView` which needs an `ISystem` as a ctor param
> > 
> > 
> > huh? why CFileView needs ISystem for anything?
> 
> it's just right now `IFile`'s `read` and `write` methods are not virtual. they have a predefined algorithm that everything should go through ISystem first, which calls `read_impl` and `write_impl`. But yeah, it obvs makes sense to make it virtual.

yes please make virtual.

include/nbl/ui/CWindowManagerWin32.h
devshgraphicsprogramming: still a tODO?

src/nbl/ui/CWindowX11.cpp
devshgraphicsprogramming: its safe to call `XInitThreads` multiple times but not concurrently, so mutex it and put it in the `CX11WindowManager` then you're fine

devshgraphicsprogramming:
I'll make them virtual

Issue #59, opened 2021-02-06
Environment Map Importance Sampling


devshgraphicsprogramming:
## Description

Need to be able to importance sample an envmap.

## Description of the related problem

Most discussed in https://github.com/buildaworldnet/IrrlichtBAW/issues/494

## Solution proposal

No idea if we should do column choice followed by row choice or some sort of hierarchical descent (quadtree / BSP style).

devshgraphicsprogramming:
Getting finished off by @Erfan-Ahmadi

PR #122, opened 2021-06-19
Importance sampling env maps

achalpandeyy:
## Description

For #59 
Known issues:
* slightly low exposure when importance sampling is enabled as compared to current results with uniform sampling
* ~~slightly incorrect code for uniform sampling wrt MIS~~


examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
```diff
@@ -0,0 +1,669 @@
...
+    normal = Sphere_getNormal(sphere,intersection);
+    return sphere.bsdfLightIDs;
+}
+
+float nbl_glsl_light_deferred_pdf(in Light light, in Ray_t ray)
+{
+#ifdef IMPORTANCE_SAMPLING
+    const vec3 direction = ray._immutable.direction;
+    const vec2 uv = SampleSphericalMap(direction);
+    return (getLuma(textureLod(envMap, uv, 0.0).rgb))/(2.0*nbl_glsl_PI*nbl_glsl_PI);
```
devshgraphicsprogramming, line 292: normalization factor depends on the integral of the luma of the envmap

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
```diff
@@ -0,0 +1,669 @@
...
+
+    vec3 L = vec3(sin(theta)*cos(phiPdf.x), cos(theta), sin(theta)*sin(phiPdf.x));
+    pdf = phiPdf.y;
+
+#else
+    const float phi = acos(2*xi[0]-1);
+    const float theta = 2*nbl_glsl_PI*xi[1];
+    vec3 L = vec3(sin(theta)*cos(phi), cos(theta), sin(phi)*sin(theta));
+    pdf = nbl_glsl_RECIPROCAL_PI*0.25;
+#endif
```
devshgraphicsprogramming, line 320: remove this `#else` the whole function shouldn't exist  if no IS (because only Importance sampled NEE calls it)

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
```diff
@@ -0,0 +1,669 @@
...
+    return Light_getRadiance(light, normalize(ray._immutable.direction));
+}
+
+vec3 nbl_glsl_light_generate_and_pdf(out float pdf, out float newRayMaxT, in vec3 origin, in nbl_glsl_AnisotropicViewSurfaceInteraction interaction, in bool isBSDF, in vec3 xi, in uint objectID)
+{   
+#ifdef IMPORTANCE_SAMPLING
+    const vec2 phiPdf = texture(phiPdfLUT, xi.xy).xy;
+    const float theta = texture(thetaLUT, xi.y).x;
+
+    vec3 L = vec3(sin(theta)*cos(phiPdf.x), cos(theta), sin(theta)*sin(phiPdf.x));
```
devshgraphicsprogramming, line 312: there's a `nbl_glsl_sincos` function of computing sin/cos pairs (you need to make sure the input argument ranges are correct though)

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
+// Copyright (C) 2018-2020 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+#define _NBL_STATIC_LIB_
+#include <nabla.h>
+#include "../../../include/nbl/asset/filters/CSummedAreaTableImageFilter.h"
```
devshgraphicsprogramming, line 7: this should reachable via `nbl/asset/asset.h`, if not please fix

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+#include "nbl/ext/FullScreenTriangle/FullScreenTriangle.h"
+#include "nbl/ext/ScreenShot/ScreenShot.h"
+#include "glm/glm.hpp"
```
devshgraphicsprogramming, line 10: please dont use glm

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+		descriptorInfos[3].desc = phiPdfLUTImageView;
+		{
+			ISampler::SParams samplerParams = { ISampler::ETC_CLAMP_TO_EDGE, ISampler::ETC_CLAMP_TO_EDGE, ISampler::ETC_CLAMP_TO_EDGE, ISampler::ETBC_FLOAT_OPAQUE_BLACK, ISampler::ETF_NEAREST, ISampler::ETF_NEAREST, ISampler::ESMM_NEAREST, 0u, false, ECO_ALWAYS };
+			descriptorInfos[3].image.sampler = driver->createGPUSampler(samplerParams);
+			descriptorInfos[3].image.imageLayout = EIL_SHADER_READ_ONLY_OPTIMAL;
+		}
+
+		descriptorInfos[4].desc = thetaLUTImageView;
+		{
+			ISampler::SParams samplerParams = { ISampler::ETC_CLAMP_TO_EDGE, ISampler::ETC_CLAMP_TO_EDGE, ISampler::ETC_CLAMP_TO_EDGE, ISampler::ETBC_FLOAT_OPAQUE_BLACK, ISampler::ETF_NEAREST, ISampler::ETF_NEAREST, ISampler::ESMM_NEAREST, 0u, false, ECO_ALWAYS };
```
devshgraphicsprogramming, line 585: maybe enable LINEAR filtering for this one

devshgraphicsprogramming: nah lets just do quadtree mapping in the future

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+			envmapPixel += channelCount;
+		}
+	}
+
+	return outBuffer;
+}
+
+// Returns the offset into the passed array the element at which is <= the passed element (`x`)
+// returns offset = -1 if passed element is < the element at index 0
+static int32_t bisectionSearch(const double* arr, const uint32_t arrCount, const double x, double* xFound)
```
devshgraphicsprogramming, line 55: move into `nbl/core/algorithm/bisection_search.h`, rename to `bisection_search` and have it have a templated signature similar to `upper_bound`

```cpp
template<typename Iterator, typename Value>
Value bisection_search(Iterator begin, Iterator end, Value val, Iterator& one_less_upper_bound)
template<typename Iterator, typename Value>
Value bisection_search(Iterator begin, Iterator end, Value val)
{
    Iterator dummy;
    return bisection_search(begin,end,val,dummy);
}
```
dont convert the upper_bound_minus_1 to an offset relative to `begin` and normalize `xFound` to the 0,1 range

then we can reuse it.

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+	int32_t offset = std::upper_bound(arr, arr + arrCount, x) - arr - 1u;
+	double dx = 0.0;
+	if (offset == -1)
+		dx = x / arr[offset + 1];
+	else
+		dx = (x - arr[offset]) / (arr[offset + 1] - arr[offset]);
+
+	// This assumes array values to be in the range [0,1) which is fine for our purposes because we use them as texture coordinates
+	if (xFound)
+		*xFound = (offset + 1 + dx) / arrCount;
```
devshgraphicsprogramming, line 66: offset+1 is the location of the value that is greater than `x` , and then you're adding `dx` to it... this seems wrong

shouldn't it be `(offset+dx)` ?

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+using namespace video;
+
+using SATFilter = CSummedAreaTableImageFilter<false>;
+
+static core::smart_refctd_ptr<ICPUBuffer> computeLuminancePdf(smart_refctd_ptr<ICPUImage> envmap)
+{
+	const core::vector2d<uint32_t> envmapExtent = { envmap->getCreationParameters().extent.width, envmap->getCreationParameters().extent.height };
+	const uint32_t channelCount = getFormatChannelCount(envmap->getCreationParameters().format);
+	
+	const core::vector2d<uint32_t> pdfDomainExtent = { envmapExtent.X, envmapExtent.Y };
```
devshgraphicsprogramming, line 25: `core::vector<>` is legacy (trying to get rid of it), just use a plain `std::array<uint32_t>`

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
```diff
@@ -0,0 +1,669 @@
...
+            bool validPath = nee_sample.NdotL>0.0;
+            // but if we allowed non-watertight transmitters (single water surface), it would make sense just to apply this line by itself
+            nbl_glsl_AnisotropicMicrofacetCache _cache;
+            validPath = validPath && nbl_glsl_calcAnisotropicMicrofacetCache(_cache, interaction, nee_sample, monochromeEta);
+            if (validPath)
+            {
+                float bsdfPdf;
+                neeContrib *= nbl_glsl_bsdf_cos_remainder_and_pdf(bsdfPdf,nee_sample,interaction,bsdf,monochromeEta,_cache)*throughput;
+                const float oc = bsdfPdf*rcpChoiceProb;
+                neeContrib /= 1.0/oc+oc/(lightPdf*lightPdf); // MIS weight
```
devshgraphicsprogramming, line 522: replace these two lines with
```glsl
float trueLightPdf = lightPdf*neeProbability;
neeContrib *= neeProbability/(bsdfPdf+trueLightPdf*trueLightPdf/bsdfPdf);
```

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
```diff
@@ -0,0 +1,669 @@
...
+    }
+
+    vec3 throughput = ray._payload.throughput;
+
+    // add emissive and finish MIS
+    const uint lightID = bitfieldExtract(bsdfLightIDs,16,16);
+    if (lightID != INVALID_ID_16BIT) // has emissive
+    {
+        float lightPdf;
+        ray._payload.accumulation += nbl_glsl_light_deferred_eval_and_prob(lightPdf, lights[lightID], ray) * throughput/(1.0 + lightPdf*lightPdf*ray._payload.otherTechniqueHeuristic);
```
devshgraphicsprogramming, line 473: just to make sure you want to `#ifdef IMPORTANCE_SAMPLING` the division by `1.0+lightPdf*lightPdf*ray._payload.otherTechniqueHeuristic`

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
```diff
@@ -0,0 +1,669 @@
...
+        //rand
+        mat2x3 epsilon = rand3d(depth,_sample,scramble_state);
+
+        // thresholds
+        const float bsdfPdfThreshold = 0.0001;
+        const float lumaContributionThreshold = getLuma(nbl_glsl_eotf_sRGB(vec3(1.0)/255.0)); // OETF smallest perceptible value
+        const vec3 throughputCIE_Y = transpose(nbl_glsl_sRGBtoXYZ)[1]*throughput;
+        const float monochromeEta = dot(throughputCIE_Y,BSDFNode_getEta(bsdf)[0])/(throughputCIE_Y.r+throughputCIE_Y.g+throughputCIE_Y.b);
+
+        // do NEE
```
devshgraphicsprogramming, line 501: wrap this whole codeblock in `#ifdef IMPORTANCE_SAMPLING`

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
```diff
@@ -0,0 +1,669 @@
...
+            bsdfSampleL = bsdf_sample.L;
+        }
+        
+        // additional threshold
+        const float lumaThroughputThreshold = lumaContributionThreshold;
+        if (bsdfPdf>bsdfPdfThreshold && getLuma(throughput)>lumaThroughputThreshold)
+        {
+            ray._payload.throughput = throughput;
+            ray._payload.otherTechniqueHeuristic = neeProbability/bsdfPdf; // numerically stable, don't touch
+            ray._payload.otherTechniqueHeuristic *= ray._payload.otherTechniqueHeuristic;
```
devshgraphicsprogramming, line 545: wrap these two in `#ifdef IMPORTANCE_SAMPLING`

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+		for (uint32_t x = 0; x < pdfDomainExtent.X; ++x)
+		{
+			double result = 0.0;
+			for (uint32_t ch = 0; ch < channelCount; ++ch)
+				result += luminanceScales[ch] * envmapPixel[ch];
+
+			*outPixel++ = result * sinTheta;
+			envmapPixel += channelCount;
+		}
+	}
```
devshgraphicsprogramming, line 48: the buffer's pointer being the first pixel, wont always work like that, there are offsets, strides, formats (you can't just cast to float), etc.

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+#include "../common/QToQuitEventReceiver.h"
+
+using namespace nbl;
+using namespace asset;
+using namespace core;
+using namespace video;
+
+using SATFilter = CSummedAreaTableImageFilter<false>;
+
+static core::smart_refctd_ptr<ICPUBuffer> computeLuminancePdf(smart_refctd_ptr<ICPUImage> envmap)
```
devshgraphicsprogramming, line 20: I was expecting this to be implemented as a new IImageFilter (you could copy most of the impl from  a CopyAndConvert filter, which doesnt do any scaling, just decodes pixels from one image, saves to another, all you need to do is just add a modification step)

devshgraphicsprogramming: actually upon thought I dont think you need and intermediate Luma image at all (could work directly from multi-channel SAT)

`Sum[dot(coeffs,rgb*sin(theta))] == dot(coeffs,Sum[rgb*sin(theta)])`

Also upon further reflection, a slight modification to `CSwizzleableAndDitherableFilterBase` (refactor into a `TranscodeFilterBase`) to allow a generalpurpose `encodePrelude` and `decodePostlude` lambdas to be called in the first thing in `onEncode` and last thing in `onDecode` (and have the `CSwizzleableAndDitherableFilterBase` refactor its `onEncode` and `onDecode` into preludes and postludes that apply swizzles and dithers) would enable the implementation `sin(theta)` weighting (and luma computation if you wish) as part of an image filter.

Then you could do
```cpp
using EnvmapLuminancePdfComputationFilter = TranscodeFilterBase<decltype(lumaDotLambda),decltype(sinThetaLamda)>; // cvt to luma on decode, weigh by sin on encode
```

devshgraphicsprogramming: Or, even less effort, you could add the custom per-pixel lambda to the SAT filter instead!

change signature to 
```cpp
template<bool ExclusiveMode = false, class CustomPerPixelLambda=CSummedAreaTableImageFilterBase::IdentityPerPixelLambda/*NOOP*/>
class CSummedAreaTableImageFilter 
```
and add a `CustomPerPixelLambda*` to the `CState` if it inherits from `CSummedAreaTableImageFilterBase::StatefulPerPixelLambda`

then right after the any calls to `asset::decodePixelsRuntime` you can do 
```cpp
if constexpr (std::is_base_of_v<CSummedAreaTableImageFilterBase::StatefulPerPixelLambda,CustomPerPixelLambda>)
    state.perPixelLambda(decodeBuffer,blockX,blockY);
else
    CustomPerPixelLambda()(decodeBuffer,blockX,blockY);
```

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+			luminanceImageRegions->begin()->bufferOffset = 0ull;
+			luminanceImageRegions->begin()->bufferRowLength = luminanceImageParams.extent.width;
+			luminanceImageRegions->begin()->bufferImageHeight = 0u;
+			luminanceImageRegions->begin()->imageSubresource = {};
+			luminanceImageRegions->begin()->imageSubresource.layerCount = 1u;
+			luminanceImageRegions->begin()->imageOffset = { 0, 0, 0 };
+			luminanceImageRegions->begin()->imageExtent = { luminanceImageParams.extent.width, luminanceImageParams.extent.height, 1 };
+
+			core::smart_refctd_ptr<ICPUImage> luminanceImage = ICPUImage::create(std::move(luminanceImageParams));
+			luminanceImage->setBufferAndRegions(core::smart_refctd_ptr(luminancePdfBuffer), luminanceImageRegions);
```
devshgraphicsprogramming, line 235: if you follow my advice of adding a `CustomPerPixelLambda` to the SAT filter, you'll never need the luma image

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+			conditionalIntegralsPixel = (double*)conditionalIntegrals->getPointer();
+
+			// now normalize
+			for (uint32_t y = 0; y < pdfDomainExtent.Y; ++y)
+			{
+				for (uint32_t x = 0; x < pdfDomainExtent.X; ++x)
+				{
+					conditionalCdfPixel[y * pdfDomainExtent.X + x] /= conditionalIntegralsPixel[y];
+				}
+			}
```
devshgraphicsprogramming, line 290: could be folded into SAT filter (it already has a `Normalize` boolean)
`/= SAT_VALUES[SCRATCH_ADDRESS(sumX ? MaxX:x,sumY ? y:maxY,sumZ ? z:maxZ)]`

would also "fix" the filter normalization option after your axis-summing choice

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
```diff
@@ -0,0 +1,668 @@
...
+				const int32_t offset = bisectionSearch((double*)conditionalCdfImage->getBuffer()->getPointer() + rowToSample * pdfDomainExtent.X, pdfDomainExtent.X, xi.X, &xiRemapped.X);
+				const double conditionalPdf = ((double*)luminancePdfBuffer->getPointer())[rowToSample * pdfDomainExtent.X + offset + 1] / ((double*)conditionalIntegrals->getPointer())[rowToSample];
+
+				const double phi = xiRemapped.X * 2.0 * core::PI<double>();
+				const double pdf = (core::sin(theta) == 0.0) ? 0.0 : (marginalPdf * conditionalPdf) / (2.0 * core::PI<double>() * core::PI<double>() * core::sin(theta));
+
+				*phiPdfLUTPixel++ = (float)phi;
+				*phiPdfLUTPixel++ = (float)pdf;
+			}
+		}
```
devshgraphicsprogramming, line 408: turn this into a simple, new image filter called `Inverse2DConditionalCDFFilter`, inherit from `CMatchedSizeInOutImageFilterCommon`, add extra image parameters in the state for the `inMarginalCDF` and `outMarginalInvCDF` (use the already exisitng `CMatchedSizeInOutImageFilterCommon` in and out images as the conditional 2D CDF and the inverse 2D CDF)

Kinda important cause I want to reuse a lot of the code here for example 22

devshgraphicsprogramming:
looks good, could do with a bit of polish

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
achalpandeyy: done

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
achalpandeyy: done

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
achalpandeyy: discarded

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
```diff
@@ -0,0 +1,669 @@
...
+    vec3 L = vec3(sin(theta)*cos(phi), cos(theta), sin(phi)*sin(theta));
+    pdf = nbl_glsl_RECIPROCAL_PI*0.25;
+#endif
+
+    newRayMaxT = FLT_MAX;
+
+    return L;
+}
+
+nbl_glsl_LightSample nbl_glsl_light_generate_and_remainder_and_pdf(out vec3 remainder, out float pdf, out float newRayMaxT, in vec3 origin, in nbl_glsl_AnisotropicViewSurfaceInteraction interaction, in bool isBSDF, in vec3 xi, in uint depth)
```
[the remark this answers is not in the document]
achalpandeyy: done

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
achalpandeyy: done

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
achalpandeyy: done

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
achalpandeyy: done

examples_tests/0.ImportanceSamplingEnvMaps/fullscreen.frag
achalpandeyy: done

devshgraphicsprogramming:
Ok @achalpandeyy do the filter stuff at some point, watch out for when I'll finish the `raytracing` branch, then merge it into yourself and then we can merge ex 0 and ex42 together

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
devshgraphicsprogramming: I've made a similar utility for the property pool manager, we should unify

include/nbl/asset/filters/CSummedAreaTableImageFilter.h
```diff
@@ -30,6 +30,7 @@ class CSummedAreaTableImageFilterBase
 				uint8_t*	scratchMemory = nullptr;										//!< memory covering all regions used for temporary filling within computation of sum values
 				size_t	scratchMemoryByteSize = {};											//!< required byte size for entire scratch memory
 				bool normalizeImageByTotalSATValues = false;								//!< after sum performation division will be performed for the entire image by the max sum values in (maxX, 0, z) depending on input image - needed for UNORM and SNORM
+				uint8_t axesToSum = 0u;														//!< which axes you want to sum; X: bit0, Y: bit1, Z: bit2
```
devshgraphicsprogramming, line 33: should be an 8 value enum instead of a random `uint8_t`

examples_tests/0.ImportanceSamplingEnvMaps/main.cpp
Erfan-Ahmadi: @devshgraphicsprogramming It was misleading me at first, but notice that @achalpandeyy subtracts -1 from upper_bound returned offset so here actually the offset is actually refering the value before that which is less than `x`, I think he's just making up for that `-1` a few lines back.

devshgraphicsprogramming: thanks

PR #123, opened 2021-06-19
Build System, Android

AnastaZIuk:


src/nbl/CMakeLists.txt
```diff
@@ -625,7 +607,26 @@ propagate_changed_variables_to_parent_scope()
...
+	${NBL_ROOT_PATH}/source/Nabla/CPakReader.h
+	${NBL_ROOT_PATH}/source/Nabla/CTarReader.h
+	${NBL_ROOT_PATH}/source/Nabla/CZipReader.h
+	${NBL_ROOT_PATH}/source/Nabla/nbl_CLogger.h
+#	${NBL_ROOT_PATH}/source/Nabla/COSOperator.h
+	${NBL_ROOT_PATH}/source/Nabla/nbl_os.h
+)
+
 nbl_install_headers("${NABLA_HEADERS_TO_INSTALL}" "${NBL_ROOT_PATH}/include")
+nbl_install_headers("${NABLA_HEADERS_TO_INSTALL_SOURCE_REFACTOR}" "${NBL_ROOT_PATH}/source/Nabla")
```
devshgraphicsprogramming, line 629: this is evil, but necessary anymore?

src/nbl/CMakeLists.txt
```diff
@@ -651,20 +655,12 @@ if (_NBL_COMPILE_WITH_OPEN_EXR_)
 	nbl_install_lib(Imath)
 endif()
 nbl_install_lib(png_static)
-if(WIN32)
+
+if(NOT (ANDROID AND "${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows")) # OpenSSL can't be built for Android platform using windows OS host
```
devshgraphicsprogramming, line 659: make an issue out of it on GH

CMakeLists.txt
```diff
@@ -295,23 +296,19 @@ macro(nbl_adjust_flags)
...
-			
-			list(APPEND CMAKE_C_FLAGS_DEBUG -g -glldb)
-			set(CMAKE_C_FLAGS_DEBUG ${CMAKE_C_FLAGS_DEBUG} CACHE STRING "" FORCE)
-			
-			list(APPEND CMAKE_EXE_LINKER_FLAGS_DEBUG -g -glldb)
-			set(CMAKE_EXE_LINKER_FLAGS_DEBUG ${CMAKE_EXE_LINKER_FLAGS_DEBUG} CACHE STRING "" FORCE)
-			
-			list(APPEND CMAKE_MODULE_LINKER_FLAGS_DEBUG -g -glldb)
-			set(CMAKE_MODULE_LINKER_FLAGS_DEBUG ${CMAKE_MODULE_LINKER_FLAGS_DEBUG} CACHE STRING "" FORCE)
+			function(NBL_INSERT_ANDROID_FLAGS _NBL_CMAKE_FLAGS_VARIABLE_NAME_)
```
devshgraphicsprogramming, line 299: call it "insert debug flags" ?

src/nbl/CMakeLists.txt
```diff
@@ -406,31 +403,12 @@ macro(nbl_target_include_parallel_hashmap _trgt)
...
-	else()
-		target_link_libraries(${_trgt} INTERFACE
-				"$<TARGET_PROPERTY:openssl_build,BINARY_DIR>/build/lib/libcrypto.a"
-				"$<TARGET_PROPERTY:openssl_build,BINARY_DIR>/build/lib/libssl.a"
-			)
-	endif()
 
-	if(NOT (ANDROID AND "${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows"))
-		target_include_directories(${_trgt} PUBLIC "$<TARGET_PROPERTY:openssl_build,BINARY_DIR>/build/include")
+		target_include_directories(${_trgt} PUBLIC "$<TARGET_PROPERTY:OpenSSL,BINARY_DIR>/build/include")
```
devshgraphicsprogramming, line 411: what does this path evaluate to?

src/nbl/CMakeLists.txt
```diff
@@ -582,19 +559,25 @@ if (UNIX AND NOT ANDROID AND NOT APPLE)
...
+	PRIVATE	_NBL_STATIC_LIB_ EGL_STATIC SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS
+	PUBLIC UNW_LOCAL_ONLY
 )
+
+if(WIN32)
+	target_compile_definitions(Nabla PRIVATE VK_USE_PLATFORM_WIN32_KHR)
+endif()
+
 if (ANDROID)
 	target_compile_definitions(Nabla PUBLIC $<$<BOOL:${OpenGL_EGL_FOUND}>:WL_EGL_PLATFORM>)
```
devshgraphicsprogramming, line 576: isnt `WL_EGL_PLATFORM` wayland?

btw nothing should depend on `*_FOUND` in the OpenGL/EGL/GLX/Vulkan space, this is why we dynamically link stuff so even if the dev machine doesn't have something we're able to compile for everything

CMakeLists.txt
AnastaZIuk: okay

src/nbl/CMakeLists.txt
AnastaZIuk: it is the same as it was, but OpenSSL is a IMPORTED library target - not custom target. The thing is that custom targets aren't allowed to be exported and because we have to reference it's properties by generator expression there would always be an error while exporting targets, hence the change that is a must.

src/nbl/CMakeLists.txt
AnastaZIuk: ask @Crisspl please, but the `VK_USE_PLATFORM_WIN32_KHR` change is due to generating definitions to the header file.

src/nbl/CMakeLists.txt
AnastaZIuk: TODO!

src/nbl/CMakeLists.txt
AnastaZIuk: it is, as soon as anybody may reference those headers (like even @Crisspl for instance, but I think users mostly), they need to be installed at include directory. We will remove those files when they finally get refactored.

src/nbl/CMakeLists.txt
AnastaZIuk: uh small mistake - INTERFACE, not IMPORTED

PR #127, opened 2021-06-29
Android build fixes

AnastaZIuk:


android-sample/main.cpp
```diff
@@ -80,7 +80,7 @@ static int engine_init_display(struct nabla* engine) {
 
     engine->window = core::make_smart_refctd_ptr<ui::CWindowAndroid>(engine->app->window);
 
-    engine->api = video::IAPIConnection::create(video::EAT_OPENGL_ES, 0, "android-sample", /*&dbgcb*/nullptr);
+    engine->api = video::IAPIConnection::create(video::EAT_OPENGL_ES, 0, "android-sample", /*&dbgcb*/{});
```
Crisspl, line 83: this will crash most likely since callback func ptr is null. 
I think we could change IAPIConnection::create() to have last arg defaulted to `{}` (nullptr callback and nullptr user data) as u have now and do not forward ptr to the struct to physical device if callback is nullptr (just pass nullptr instead)

devshgraphicsprogramming: @Crisspl what about just creating a `system::ILogger` interface, the debug callback needs to log stuff from the gfx API

but I'm sure a lot of loaders, etc. also want to log (and we could finally get rid of the `os::Printer`)

The one could log to `stdout`, the default implementation could be provided to log to a file, if the user needs anything else let them inherit and make their own.

P.S. This would also have the added advantage of stopping the Nabla windows examples being console applications by default.

PR #129, opened 2021-07-01
removed ex22 optix dependency in cmake

Erfan-Ahmadi:
## Description
Example 22 does not use Optix but has build dependency on it

## Testing 
Config/Generate button on Cmake and the NblExtOPTIX was gone.


devshgraphicsprogramming:
tell me why it didnt work?

I only appended the libs and includes if NBL_BUILD_WITH_OPTIX was enabled, didnt append anything otherwise

Erfan-Ahmadi:
> tell me why it didnt work?
> 
> I only appended the libs and includes if NBL_BUILD_WITH_OPTIX was enabled, didnt append anything otherwise

Because I enabled it and but the folder not found, meanwhile config and generate was successfull but the path to optix was invalid or by defaul: "include/"

to put it shortly "NBL_BUILD_WITH_OPTIX " was true but the path was invalid

devshgraphicsprogramming:
> 
> 
> > tell me why it didnt work?
> > I only appended the libs and includes if NBL_BUILD_WITH_OPTIX was enabled, didnt append anything otherwise
> 
> Because I enabled it but the folder not found, meanwhile config and generate was successfull but the path to optix was invalid or by defaul: "include/"
> 
> to put it shortly "NBL_BUILD_WITH_OPTIX " was true but the path was invalid

so why wasnt this fixed by you disabling `NBL_BUILD_WITH_OPTIX` on your machine?

Erfan-Ahmadi:
@devshgraphicsprogramming If I disabled it It would've been fixed, I thought it was neccessary for it to be on so disabling it didn't occur to me.
So basically everything was and is fine, but since ex22 doesn't use optix why consider building it at all?

devshgraphicsprogramming:
> 
> 
> @devshgraphicsprogramming If I disabled it It would've been fixed, I thought it was neccessary for it to be on, and disabling it didn't occur to me.
> So basically everything was and is fine, but since ex22 doesn't use optix why consider building it at all?

because its "not using optix" temporarily (cause I need to figure out how to define/declare albedo and normals for complex materials)

Erfan-Ahmadi:
> > @devshgraphicsprogramming If I disabled it It would've been fixed, I thought it was neccessary for it to be on, and disabling it didn't occur to me.
> > So basically everything was and is fine, but since ex22 doesn't use optix why consider building it at all?
> 
> because its "not using optix" temporarily (cause I need to figure out how to define/declare albedo and normals for complex materials)

Then this PR is useless.

PR #140, opened 2021-07-10
[WIP] Polyphase Image Filtering (kernel weight caching)

achalpandeyy:
## Description

* Example 0 (and `enableLUTUsage` state variable) is just for testing and will be removed before merge


include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -187,13 +187,15 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Normalize,Clamp,Sw
...
-				KernelY					kernelY;
-				KernelZ					kernelZ;
+				uint32_t							inMipLevel = 0u;
+				uint32_t							outMipLevel = 0u;
+				ICPUImage*							inImage = nullptr;
+				ICPUImage*							outImage = nullptr;
+				KernelX								kernelX;
+				KernelY								kernelY;
+				KernelZ								kernelZ;
+				core::smart_refctd_ptr<ICPUBuffer>	phaseSupportLUT = nullptr;
```
devshgraphicsprogramming, line 197: store the LUT in some part of the state's `scratch` (its for storing everything, really)

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -372,6 +375,12 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Normalize,Clamp,Sw
 				return core::vectorSIMDi32(kernelX.getWindowMinCoord(halfTexelOffset).x-1,kernelY.getWindowMinCoord(halfTexelOffset).y-1,kernelZ.getWindowMinCoord(halfTexelOffset).z-1,0);
 			}();
 			const auto windowMinCoordBase = inOffsetBaseLayer+startCoord;
+
+			const core::vectorSIMDu32 phaseCount = getPhaseCount(inExtentLayerCount, outExtentLayerCount);
+
+			const size_t maxPhaseSupportLUTSize = core::max(phaseCount[0], core::max(phaseCount[1], phaseCount[2])) * core::max(kernelX.getWindowSize()[0], core::max(kernelY.getWindowSize()[1], kernelZ.getWindowSize()[2])) * sizeof(double);
+			state->phaseSupportLUT = core::make_smart_refctd_ptr<ICPUBuffer>(maxPhaseSupportLUTSize);
```
devshgraphicsprogramming, line 382: compute and prepare in a `CState` method (so you can reuse for multiple `execute` calls)

this also requires the LUT to be `sum(phaseCount[i]*kernel[i].getWindowSize()[i])` instead of `max_element(phaseCount[i]*kernel[i].getWindowSize()[i])`

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -465,7 +530,7 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Normalize,Clamp,Sw
 							lineBuffer = intermediateStorage[axis-1]+core::dot(static_cast<const core::vectorSIMDi32&>(intermediateStrides[axis-1]),localTexCoord)[0];
 						else
 						{
-							const auto windowEnd = inExtent.width+window_last.x;
+							const auto windowEnd = inExtent.width+window_last.x+1;
```
devshgraphicsprogramming, line 533: you'll overwrite your own linebuffer

achalpandeyy: i now realize its basically same as putting a `<=` in the `for` below (apart from other consequences of `window_last` ofcourse)

examples_tests/0.Polyphase/main.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+	blitFilterState.axisWraps[0] = ISampler::ETC_CLAMP_TO_EDGE;
+	blitFilterState.axisWraps[1] = ISampler::ETC_CLAMP_TO_EDGE;
+	blitFilterState.axisWraps[2] = ISampler::ETC_CLAMP_TO_EDGE;
+	blitFilterState.borderColor = ISampler::E_TEXTURE_BORDER_COLOR::ETBC_FLOAT_OPAQUE_WHITE;
+
+	blitFilterState.enableLUTUsage = useLUT;
+
+	blitFilterState.scratchMemoryByteSize = BlitFilter::getRequiredScratchByteSize(&blitFilterState);
+	blitFilterState.scratchMemory = reinterpret_cast<uint8_t*>(_NBL_ALIGNED_MALLOC(blitFilterState.scratchMemoryByteSize, 32));
```
devshgraphicsprogramming, line 73: so here you'd want the LUT to be counted into the scratch size and taken from an offset

And yes the LUT population function would need to be split off from `execute` so the LUT could be reused for blitting images with identical phases (usually this means identical source to destination extent rations), obviously not by the filter, the user would have invent such a caching scheme.

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -409,7 +418,63 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Normalize,Clamp,Sw
...
+						const IImageFilterKernel::UserData* unused2) -> void
+					{
+						for (auto h = 0; h < MaxChannels; h++)
+							windowSample[h] = value_type(1);
+					};
+
+					// There's no real reason that this should have `MaxChannels` number of elements except for the fact that
+					// `CFloatingPointSeparableImageFilterKernelBase<CRTP>::sample_functor_t<PreFilter,PostFilter>::operator()` expects that.
+					// So this, for now, just replicates the same value in all 4 channels.
+					value_type kernelWeight[MaxChannels];
```
devshgraphicsprogramming, line 444: there are special kernel wrappers which make them behave differently for different channels, the derivative map creator uses a blit filter with a separate-per-channel-kernel

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -409,7 +418,63 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Normalize,Clamp,Sw
...
+
+					// a dummy load functor
+					// does nothing but fills up the `windowSample` with 1s (identity) so we can preserve the value of kernel
+					// weights when eventually `windowSample` gets multiplied by them later in
+					// `CFloatingPointSeparableImageFilterKernelBase<CRTP>::sample_functor_t<PreFilter,PostFilter>::operator()`
+					// this exists only because `evaluateImpl` expects a pre filtering step.
+					// 
+					// Here I suggest to make something like a CImageFilterKernel<CRTP,value_type>::computeWeights which goes through
+					// the same pipeline of injecting other pre filtering stages (like `wrap` for the `CScaledImageFilterKernel`, for example)
+					// and which doesn't take this functor as argument.
```
devshgraphicsprogramming, line 432: there's an easier API/function for just evaluating a kernel
https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/asset/filters/kernels/CommonImageFilterKernels.h#L83

did you not see it before or did it not fullfill some criteria?

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: in the end `evaluateImpl` just ends up being calls to `weight`

https://github.com/Devsh-Graphics-Programming/Nabla/blob/124355408c8a96428205f9decb2c7eb6c75a4073/include/nbl/asset/filters/kernels/kernels.h#L87

P.S. We should probably make kernel tags (like pre-C++20 iterator tags) for separable, symmetric, etc.

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -502,16 +567,20 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Normalize,Clamp,Sw
 								}
 							}
 						}
+
+						double* phaseSupportLUTPixel = (double*)state->phaseSupportLUT->getPointer();
```
devshgraphicsprogramming, line 571: here you could compute an offset into the scratch

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: wanted to use the existing infrastructure of injecting other prefiltering stages
for example, in `CScaledImageFilterKernel` you have this https://github.com/achalpandeyy/Nabla/blob/master/include/nbl/asset/filters/kernels/CScaledImageFilterKernel.h#L93
where the input kernel evaluation point gets rescaled (potentially) multiple times to get a value in the defined domain of the kernel (for example [-1/2,1/2] for the Box kernel)

if I would have just called `weight` I would have to implement something like the above myself

achalpandeyy: and thats just `CScaledImageFilterKernel` other kernels might inject their own pre and post filtering stages

devshgraphicsprogramming: hmm ok I guess you're right, that you need this weirdness for it to work with scaling,

but it whoul all be moved to some static method to precompute the LUT

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -520,19 +589,32 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Normalize,Clamp,Sw
...
 							{
-								value_type windowSample[MaxChannels];
+								if (state->enableLUTUsage)
+								{
+									value_type windowSample[MaxChannels];
+									for (auto ch = 0; ch < MaxChannels; ch++)
+										windowSample[ch] = phaseSupportLUTPixel[phaseIndex*windowSize+h] * lineBuffer[(windowCoord[axis] - windowMinCoord[axis]) * MaxChannels + ch];
+
+									for (auto ch = 0; ch < MaxChannels; ch++)
+										value[ch] += windowSample[ch];
```
devshgraphicsprogramming, line 608: accumulate straight into `value[ch]`

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -520,19 +589,32 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Normalize,Clamp,Sw
...
 							core::vectorSIMDf tmp;
 							tmp[axis] = float(i)+0.5f;
 							core::vectorSIMDi32 windowCoord;
 							windowCoord[axis] = kernel.getWindowMinCoord(tmp*fScale,tmp)[axis];
 							auto relativePos = tmp[axis]-float(windowCoord[axis]);
+							const uint32_t phaseIndex = i % phaseCount[axis];
 							for (auto h=0; h<windowSize; h++)
 							{
-								value_type windowSample[MaxChannels];
+								if (state->enableLUTUsage)
```
devshgraphicsprogramming, line 601: from now on, you always want to use a LUT

devshgraphicsprogramming:
after I make a release (after #152 is merged), get this all polished  up and switch all instances of `BlitImageFilter` in the engine (especially the ones used in Virtual Texturing) to use polyphase

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: 1. Is there something to do here or it can be resolved?
2. Should I remove the comment?

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: I think this is already fixed in [this](https://github.com/Devsh-Graphics-Programming/Nabla/commit/d91717e6fe5b9a655bf851bcdca424dafe6e5ff0) commit. 

Resolved?

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved.

examples_tests/0.Polyphase/main.cpp
achalpandeyy: If I understand this correctly, I think it is resolved now.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved.

achalpandeyy:
Okay @devshgraphicsprogramming give me in the green signal before you merge and I'll remove the `enableLUTUsage` flag and make it always use the LUT.

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -189,21 +267,43 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
+				ICPUImage*							inImage = nullptr;
+				ICPUImage*							outImage = nullptr;
+				KernelX								kernelX;
+				KernelY								kernelY;
+				KernelZ								kernelZ;
+				bool								enableLUTUsage = true; // temporary state for testing, will remove before merge
 		};
 		using state_type = CState;
 		
+		// If you're changing this method don't forget to change the one below (getPhaseSupportLUTByteOffset)
```
devshgraphicsprogramming, line 281: DRY your code, then you won't need this comment

```cpp
static inline uint32_t getRequiredScratchByteSize(const state_type* state)
{
    uint32_t retval = getPhaseSupportLUTByteOffset(state);
    // need to add the memory for phase support LUT
			const core::vectorSIMDu32 phaseCount = getPhaseCount(state->inExtentLayerCount, state->outExtentLayerCount, state->inImage->getCreationParameters().type);
			retval += ((phaseCount[0]*scaledKernelX.getWindowSize().x) + (phaseCount[1]*scaledKernelY.getWindowSize().y) + (phaseCount[2]*scaledKernelZ.getWindowSize().z)) * sizeof(value_type);

			return retval;
}
```

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -511,16 +616,20 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
 								}
 							}
 						}
+
+						value_type* phaseSupportLUTPixel = reinterpret_cast<value_type*>(state->scratchMemory + getPhaseSupportLUTByteOffset(state) + axisOffsets[axis]);
```
devshgraphicsprogramming, line 620: this is a hot code path, please hoist the `getPhaseSupportLUTByteOffset(state)+axisOffsets[axis]`

(maybe pre-add to `axisOffsets`)

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -529,19 +638,28 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
 									value[h] += windowSample[h];
 							};
-							// do the filtering 
+							// do the filtering
 							core::vectorSIMDf tmp;
 							tmp[axis] = float(i)+0.5f;
 							core::vectorSIMDi32 windowCoord(0);
 							windowCoord[axis] = kernel.getWindowMinCoord(tmp*fScale,tmp)[axis];
 							auto relativePos = tmp[axis]-float(windowCoord[axis]);
+							const uint32_t phaseIndex = i % phaseCount[axis];
```
devshgraphicsprogramming, line 647: modulo is expensive (several hundred cycles), can you hoist it to above the `for` on line 623
```c
uint32_t phaseIndex = 0;
```
and then at the end of the loop do
```c
if (++phaseIndex==phaseCount[axis])
   phaseIndex = 0;
```

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: you can also get rid of `load` and `evaluate` lambdas declared above.

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -529,19 +638,28 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
 							for (auto h=0; h<windowSize; h++)
 							{
-								value_type windowSample[MaxChannels];
+								if (state->enableLUTUsage)
+								{
+									for (auto ch = 0; ch < MaxChannels; ch++)
+										value[ch] += phaseSupportLUTPixel[phaseIndex * windowSize + h] * lineBuffer[(windowCoord[axis] - windowMinCoord[axis]) * MaxChannels + ch];
```
devshgraphicsprogramming, line 653: 1) ok I can see that as you move positively in the LUT, you also move positively in the window coord.

However convolution moves in reverse in the second function (as you can see with `relativePos -= 1.f`), so is your LUT inverted then, and did you test and validate the blit still works with a non-symmetric Kernel such as one used for derivative maps?

devshgraphicsprogramming: 2. I'd turn `phaseSupportLUTPixel[phaseIndex * windowSize + h] * lineBuffer[(windowCoord[axis] - windowMinCoord[axis]) * MaxChannels + ch]` into a lambda

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: If you do https://github.com/Devsh-Graphics-Programming/Nabla/pull/140/files#r863621098 then you can hoist the `h=0`  case outside and swap the `+=` for a `=` and start the loop at `h=1` and move `windowCoord[axis]++` to be the first line in the loop.

This will allow you to get rid of the `std::fill` at line 627

include/nbl/asset/filters/CMipMapGenerationImageFilter.h
```diff
@@ -96,6 +96,7 @@ class CMipMapGenerationImageFilter : public CImageFilter<CMipMapGenerationImageF
 			for (auto inMipLevel=state->startMipLevel; inMipLevel!=state->endMipLevel; inMipLevel++)
 			{
 				auto blit = buildBlitState(state, inMipLevel);
+				pseudo_base_t::state_type::computePhaseSupportLUT(&blit);
```
devshgraphicsprogramming, line 99: the call to this should be in `buildBlitState` already

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -614,6 +743,14 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
 			// obviously we have multiple channels and each channel has a certain type for arithmetic
 			return texelCount*MaxChannels*sizeof(value_type);
 		}
+
+		static inline core::vectorSIMDu32 getPhaseCount(const core::vectorSIMDu32& in, const core::vectorSIMDu32& out, const asset::IImage::E_TYPE inImageType)
+		{
+			core::vectorSIMDu32 result(0u);
```
devshgraphicsprogramming, line 749: initialize to 1 maybe?

devshgraphicsprogramming: then an assert that none of the phases is 0 is not troublesome

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -588,22 +706,33 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
 		}
+		static inline core::vectorSIMDu32 getPhaseSupportLUTAxisOffsets(const core::vectorSIMDu32& phaseCount,
+			const CScaledImageFilterKernel<KernelX>& kernelX,
+			const CScaledImageFilterKernel<KernelY>& kernelY,
+			const CScaledImageFilterKernel<KernelZ>& kernelZ)
+		{
+			core::vectorSIMDu32 result;
+			result.x = 0u;
+			result.y = (phaseCount[0] * kernelX.getWindowSize().x) * sizeof(value_type);
+			result.z = ((phaseCount[0] * kernelX.getWindowSize().x) + (phaseCount[1] * kernelY.getWindowSize().y)) * sizeof(value_type);
```
devshgraphicsprogramming, line 724: can we keep DRY and write it as
```cpp
core::vectorSIMDu32 result(0u);
result.y = phaseCount.x * kernelX.getWindowSize().x;
result.z = result.y + phaseCount.y * kernelY.getWindowSize().y;
return result*sizeof(value_type);
```

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -588,22 +706,33 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
-			core::vectorSIMDi32 last(kernelX.getWindowSize().x,0,0,0);
+			core::vectorSIMDi32 last(kernelX.getWindowSize().x-1,0,0,0);
 			if (inImageType>=IImage::ET_2D)
-				last.y = kernelY.getWindowSize().x;
+				last.y = kernelY.getWindowSize().y-1;
 			if (inImageType>=IImage::ET_3D)
-				last.z = kernelZ.getWindowSize().x;
+				last.z = kernelZ.getWindowSize().z-1;
 			return last;
 		}
```
devshgraphicsprogramming, line 715: rename the function to `getRealWindowSize()`, and rename any variables called `window_end` to `real_window_size`

yes its supposed to be WITHOUT the `-1`

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: also use `real_window_size` as input instead of the 3 kernels? (less opportunity for fuckups)

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -588,22 +706,33 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
+			const CScaledImageFilterKernel<KernelY>& scaledKernelY,
+			const CScaledImageFilterKernel<KernelZ>& scaledKernelZ)
 		{
 			const auto inType = state->inImage->getCreationParameters().type;
-			const auto kernelX = state->contructScaledKernel(state->kernelX);
-			const auto kernelY = state->contructScaledKernel(state->kernelY);
-			const auto kernelZ = state->contructScaledKernel(state->kernelZ);
 
-			const auto window_end = getWindowEnd(state->inImage->getCreationParameters().type,kernelX,kernelY,kernelZ);
+			const auto window_end = getWindowEnd(state->inImage->getCreationParameters().type,scaledKernelX,scaledKernelY,scaledKernelZ);
```
devshgraphicsprogramming, line 735: what was wrong with constructing the scaled kernels locally here, function took less arguments before... was cleaner

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -588,22 +706,33 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
 			// TODO: account for the size needed for coverage adjustment
```
devshgraphicsprogramming, line 736: can you fix this TODO or is that fixed in #343 ?

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -588,22 +706,33 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
 			// the first pass will be along X, so new temporary image will have the width of the output extent, but the height and depth will need to be padded
 			// but the last pass will be along Z and the new temporary image will have the exact dimensions of `outExtent` which is why there is a `core::max`
```
devshgraphicsprogramming, line 738: probably in the interest of dryness we should make a function called `computeIntermediateExtents`

because there's some duplication here in working out the intermediate extents and in the `execute()` call

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: yeah, you were right about the -1 being wrong.

a few renamings are necessary:
https://github.com/Devsh-Graphics-Programming/Nabla/pull/140#discussion_r863651029


Also I didn't get to comment on this line (because it never changed), but this:
```cpp
							const auto windowEnd = inExtent.width+window_end.x;
							decode_offset = alloc_decode_scratch();
							lineBuffer = intermediateStorage[1]+decode_offset*MaxChannels*windowEnd;
							for (auto& i=localTexCoord.x; i<windowEnd; i++)
```
should be renamed for clarity into this
```cpp
							const auto inputEnd = inExtent.width+real_window_size.x;
							decode_offset = alloc_decode_scratch();
							lineBuffer = intermediateStorage[1]+decode_offset*MaxChannels*inputEnd;
							for (auto& i=localTexCoord.x; i<inputEnd; i++)
```

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -152,6 +152,84 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
+							windowCoord[axis] = kernel.getWindowMinCoord(tmp * fScale, tmp)[axis];
+
+							float relativePos = tmp[axis] - float(windowCoord[axis]); // relative position of the last pixel in window from current (ith) output pixel having a unique phase sequence of kernel evaluation points
+
+							for (int32_t j = 0; j < windowSize; ++j)
+							{
+								core::vectorSIMDf tmp(relativePos, 0.f, 0.f);
+								kernel.evaluateImpl(dummyLoad, dummyEvaluate, kernelWeight, tmp, windowCoord, &scale); // `windowCoord` is unused
+								phaseSupportLUTPixel[i * windowSize + j] = kernelWeight[0];
+								relativePos -= 1.f;
```
devshgraphicsprogramming, line 223: ok I see that forward iterated LUT entries store the kernel entries in reverse, you need to document (comment that) so that anyone reading the `execute()` function doesn't go "thIs cOdE Is wrOng It ImpLEmEnts cOrrELAtIOn nOt cOnvOLUtIOn!"

devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/140#discussion_r863620797

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -152,6 +152,84 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
+						const auto windowSize = kernel.getWindowSize()[axis];
+
+						IImageFilterKernel::ScaleFactorUserData scale(1.f / fScale[axis]);
+						const IImageFilterKernel::ScaleFactorUserData* otherScale = IImageFilterKernel::ScaleFactorUserData::cast(state->kernelX.getUserData());
+
+						if (otherScale)
+						{
+							for (auto k = 0; k < MaxChannels; k++)
+								scale.factor[k] *= otherScale->factor[k];
+						}
```
devshgraphicsprogramming, line 205: erm, remind me what this is about?

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: what I meant is, the kernel weights **are not monochrome** so your LUT shouldn't be either.

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -152,6 +152,84 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
+						for (auto h = 0; h < MaxChannels; h++)
+							windowSample[h] = value_type(1);
+					};
+
+					// There's no real reason that this should have `MaxChannels` number of elements except for the fact that
+					// `CFloatingPointSeparableImageFilterKernelBase<CRTP>::sample_functor_t<PreFilter,PostFilter>::operator()` expects that.
+					// So this, for now, just replicates the same value in all 4 channels.
+					value_type kernelWeight[MaxChannels];
+					// actually used to put values in the LUT
+					auto dummyEvaluate = [&kernelWeight](const value_type* windowSample, const core::vectorSIMDf& unused0, const core::vectorSIMDi32& unused1, const IImageFilterKernel::UserData* unused2) -> void
```
devshgraphicsprogramming, line 185: instead of calling them `unused` you can just not give them names
https://stackoverflow.com/questions/2319663/why-does-c-code-missing-a-formal-argument-name-in-a-function-definition-compil

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -152,6 +152,84 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
+							tmp[axis] = float(i) + 0.5f;
+
+							core::vectorSIMDi32 windowCoord;
+							windowCoord[axis] = kernel.getWindowMinCoord(tmp * fScale, tmp)[axis];
+
+							float relativePos = tmp[axis] - float(windowCoord[axis]); // relative position of the last pixel in window from current (ith) output pixel having a unique phase sequence of kernel evaluation points
+
+							for (int32_t j = 0; j < windowSize; ++j)
+							{
+								core::vectorSIMDf tmp(relativePos, 0.f, 0.f);
```
devshgraphicsprogramming, line 220: initialize tmp to 0 and set `tmp[axis] = relativePos;`

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -152,6 +152,84 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
+								kernel.evaluateImpl(dummyLoad, dummyEvaluate, kernelWeight, tmp, windowCoord, &scale); // `windowCoord` is unused
+								phaseSupportLUTPixel[i * windowSize + j] = kernelWeight[0];
```
devshgraphicsprogramming, line 222: kernel weights are not monochrome, LUT shouldn't be either!

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -152,6 +152,84 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
...
+								scale.factor[k] *= otherScale->factor[k];
+						}
+
+						value_type* phaseSupportLUTPixel = reinterpret_cast<value_type*>(state->scratchMemory + getPhaseSupportLUTByteOffset(state) + axisOffsets[axis]);
+						for (uint32_t i = 0u; i < phaseCount[axis]; ++i)
+						{
+							core::vectorSIMDf tmp;
+							tmp[axis] = float(i) + 0.5f;
+
+							core::vectorSIMDi32 windowCoord;
```
devshgraphicsprogramming, line 213: just use an `int32_t`

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Nothing wrong, just repeated computation 2x in `execute` 1x in `getRequiredScratchByteSize`.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Oh yeah.. it was weird.

Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: On the contrary, a default value of 1 would be troublesome. It would make [this size calculation](https://github.com/Devsh-Graphics-Programming/Nabla/blob/fd9be09ea1b0770f439554a1e2fd4035a7049e42/include/nbl/asset/filters/CBlitImageFilter.h#L294) incorrect. If we put a check there to see if `phaseCount == 1`, then to know if its truly a monophase case or just a default value for a non-existent axis, we would need more checks.

achalpandeyy: Related: https://github.com/Devsh-Graphics-Programming/Nabla/pull/140#discussion_r863613193

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Removed and Resolved in #343 .

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Removed the comment and the LUT not monochrome in #343.
Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343. Except the `assert` because of reasons mentioned [here](https://github.com/Devsh-Graphics-Programming/Nabla/pull/140#discussion_r865541796).

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CMipMapGenerationImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: I don't get it. Coverage adjustment already works.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: It will break the program!

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: I would still need to pass a `core::vectorSIMDi32` in the loop below which the LOC:
```
kernel.evaluateImpl(dummyLoad, dummyEvaluate, kernelWeight, tmp, windowCoord, &scale);
```

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: ah see, if you would have initialized the `phaseCount` to 1 for any axes that won't be filtered then no assert needed.

You start with a phaseIndex = 0, increment to 1, end up equaling 1, go back to 0

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: ok

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: maybe don't change the def of `getPhaseCount` but `min(,1)` the `phaseCount you compute here.

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: so why is this TODO comment still here?

devshgraphicsprogramming: it clearly about something....

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: I see you've done it in #343, so it works anyway?

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: yeah, you should create the `core::vectorSIMDi32` on demand there, within the loop, compiler might optimize a few registers away.

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: ah ok, i goes forward, relative pos goes backward

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: i think its because coverage adjustment uses some scratch too, but the amount it will use for the temp data is not taken into account in this `max`

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: I don't see how it could be using any temporary memory.
Most of the operations are done in place. The places where it accesses intermediate/scratch storages are [here](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L337), [here](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L343) (though we might want to replace the 4 here with `MaxChannels`), and [here](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L356), and they all appear to be valid, in-bound indices.
Apart from this there are just two `atomic_int32_t` needed to keep the state for coverage adjustment.

I'm failing to understand the purpose of `alloc_decode_scratch` [here](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L422), but since it is common for both coverage adjustment and non-coverage adjustment cases, I doubt here is the issue.

Probably try asking the person who wrote this TODO?

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved in #343.

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: > (though we might want to replace the 4 here with MaxChannels),

please do.

> I'm failing to understand the purpose of alloc_decode_scratch [here](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L422), but since it is common for both coverage adjustment and non-coverage adjustment cases, I doubt here is the issue.

It's because the decode line is a captured-by-reference variable outside the lambdas which could run in parallel "for_each", now if we have multiple threads (or even SIMD lanes with `par_unseq`) hammering the same decode scratch space we'll get data races.

This is what the `(state->inExtent.width + real_window_size[0]) * std::thread::hardware_concurrency() * VectorizationBoundSTL` bound is for.

You can add that in a comment I guess.

> don't see how it could be using any temporary memory.
Most of the operations are done in place. The places where it accesses intermediate/scratch storages are [here](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L337), [here](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L343) (though we might want to replace the 4 here with MaxChannels), and [here](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L356), and they all appear to be valid, in-bound indices.

Ok I went back into the code and figured out what its about.

Normally when you blit you'll decode your image into a 1-double-per-channel array of size:
`[output.W]x[inputH+kernelH_support]x[input.D]`
then as you move through the dimensions you get
`[output.W]x[output.H]x[input.D+kernelD_support]`
and then
`output.W x output.H x output.D`

The TODO is referring to the fact that coverage adjustment uses the scratch of the "next" axis to run `nth_element` 
https://github.com/Devsh-Graphics-Programming/Nabla/blob/a3c958b2659ddcfb90e3f6c22531a567f680251a/include/nbl/asset/filters/CBlitImageFilter.h#L337-356
its really an obnoxious use of memory.

The whole idea with `nth_element` was a dumb hack, it should be rewritten to mirror what you do on the GPU, which is a small histogram mapped to fixed precision floats (rescaled integers), prefix sum, binary search for "nth" that way.


So we should clean up my old sphagetti code to something like this:
```cpp
                enum E_SCRATCH_USAGE
                {
                	ESU_PHASE_SUPPORT_LUT=0,
                	ESU_DECODE_WRITE=1,
                	ESU_BLIT_X_AXIS_READ=ESU_DECODE_WRITE,
                	ESU_BLIT_X_AXIS_WRITE=2,
                	ESU_BLIT_Y_AXIS_READ=ESU_BLIT_X_AXIS_WRITE,
                	ESU_BLIT_Y_AXIS_WRITE=3,
                	ESU_BLIT_Z_AXIS_READ=ESU_BLIT_Y_AXIS_WRITE,
                	ESU_BLIT_Z_AXIS_WRITE=4,
                	ESU_ALPHA_VALUE_HISTOGRAM=5,
                	ESU_COUNT=6 // when you call `getScratchOffset(state,ESU_COUNT)` it would give the total scratch size needed
                };
		static inline uint32_t getScratchOffset(const state_type* state, const E_SCRATCH_USAGE usage)
		{
			const auto inType = state->inImage->getCreationParameters().type;
			const auto kernelX = state->contructScaledKernel(state->kernelX);
			const auto kernelY = state->contructScaledKernel(state->kernelY);
			const auto kernelZ = state->contructScaledKernel(state->kernelZ);
			
			// need to add the memory for phase support LUT
			const core::vectorSIMDu32 phaseCount = getPhaseCount(state->inExtentLayerCount, state->outExtentLayerCount,inType);
			const size_t phaseLUTElementCount = (phaseCount[0]*scaledKernelX.getWindowSize().x) + (phaseCount[1]*scaledKernelY.getWindowSize().y) + (phaseCount[2]*scaledKernelZ.getWindowSize().z);

			const auto real_window_size = getRealWindowSize(inType,kernelX,kernelY,kernelZ);
			core::vectorSIMDi32 intermediateExtent[3];
			getIntermediateExtents(intermediateExtent, state, real_window_size);
			assert(intermediateExtent[0].x == intermediateExtent[2].x);
			
			const size_t MaxParallelism = std::thread::hardware_concurrency() * VectorizationBoundSTL;
			
			// decode
			uint32_t pingBufferElementCount = (state->inExtent.width + real_window_size[0]) * MaxParallelism;
			// x-axis filter output
			uint32_t pongBufferElementCount = intermediateExtent[0].x*intermediateExtent[0].y*intermediateExtent[0].z;

			const uint32_t yWriteElementCount = intermediateExtent[1].x*intermediateExtent[1].y*intermediateExtent[1].z;
			if (inType>=2D && yWriteElementCount>pingBufferElementCount) // will there will be a y-pass
				pingBufferElementCount = yWriteElementCount;

			const uint32_t zWriteElementCount = intermediateExtent[2].x*intermediateExtent[2].y*intermediateExtent[2].z;
			if (inType>=3D && zWriteElementCount>pongBufferElementCount) // will there will be a z-pass
				pongBufferElementCount = zWriteElementCount;
			
			uint32_t retval = phaseLUTElementCount; // LUT is first and immutable
			switch (usage)
			{
				case ESU_PHASE_SUPPORT_LUT:
					retval = 0u; // LUT is first and immutable
					break;
				// ping
				case ESU_DECODE_WRITE:
				case ESU_BLIT_Y_AXIS_WRITE:
					// just don't add anything
					break;
				// pong (need to skip same memory used both for decode and Y-axis)
				case ESU_BLIT_X_AXIS_WRITE:
				case ESU_BLIT_Z_AXIS_WRITE:
					retval += pingBufferElementCount;
					break;
				case ESU_ALPHA_VALUE_HISTOGRAM:
					retval += pingBufferElementCount+pongBufferElementCount;
					[[fallthrough]];
				default: // ESU_COUNT
					// if we use coverage adjustment we needed some extra memory for histogram
					if (state->alphaSemantic==CState::EAS_REFERENCE_OR_COVERAGE)
						retval += kAlphaHistogramSize*MaxParallelism; // duplicate histogram, avoid atomics
					break;
			}
			return retval * sizeof(value_type) * MaxChannels;
		}

```
effectively putting the LUT first, DECODE/Y_WRITE next, then X_WRITE/Z_WRITE afterwards and the alpha value histogram last.

Then get rid of all other `getXXXOffset` functions and rephrase the `getXXXXSize` in terms of this one.

achalpandeyy: Wondering why we don't use the `core::LinearAddressAllocator` here? 🤔

achalpandeyy: This snippet has a bug because `ESU_COUNT` will directly go to the `default:` case and won't report the total scratch size correctly.
I'll restructure this a bit.

achalpandeyy: I think its best to not rely on fallthroughs (except for the obvious ones --overlapping memory for different passes) and write it more explicitly.

achalpandeyy: > This snippet has a bug because ESU_COUNT will directly go to the default: case and won't report the total scratch size correctly.
I'll restructure this a bit.

Another bug is, since the LUT data type can vary `{16, 32, 64}`, we would be overestimating the scratch size most of the times.

achalpandeyy: I think [this](https://github.com/Devsh-Graphics-Programming/Nabla/blob/92228e74ddb54145e5865713b84780146a42f2d1/include/nbl/asset/filters/CBlitImageFilter.h#L51) can now go?
After implementing the new scratch usage regime, I don't see this being used at all.

achalpandeyy: You have this in the above snippet:

```C++
if (state->alphaSemantic==CState::EAS_REFERENCE_OR_COVERAGE)
    retval += kAlphaHistogramSize*MaxParallelism; // duplicate histogram, avoid atomics
```

[I currently did it with atomics (`std::atomic_ref`)](https://github.com/Devsh-Graphics-Programming/Nabla/blob/01f0c5202fceb96d8603bbb563e150b0aace0c9d/include/nbl/asset/filters/CBlitImageFilter.h#L412), which we don't want.

I was wondering how do you want me to approach this, since you cannot control the number of threads in `std::execution` policy, and also (AFAIK) you can't get the number of threads invoked by it.
One way I could think of is:
1. Fork `std::thread::hardware_concurrency` number of `std::thread`s.
2. Make them operate on their own copy of histogram.
3. `std::thread::join`.
4. Make the main thread merge all the histogram into one, which will be used for further processing.

If we do decide to go down this route, then I don't understand why we have the multiplication with `MaxParallelism` above, because it also has a factor of `VectorizationBoundSTL` and AFAICT we're not using any SIMD anywhere (that could've been a possibility with execution policy, I guess)?

devshgraphicsprogramming: > You have this in the above snippet:
> 
> ```c++
> if (state->alphaSemantic==CState::EAS_REFERENCE_OR_COVERAGE)
>     retval += kAlphaHistogramSize*MaxParallelism; // duplicate histogram, avoid atomics
> ```
> 
> [I currently did it with atomics (`std::atomic_ref`)](https://github.com/Devsh-Graphics-Programming/Nabla/blob/01f0c5202fceb96d8603bbb563e150b0aace0c9d/include/nbl/asset/filters/CBlitImageFilter.h#L412), which we don't want.
> 
> I was wondering how do you want me to approach this, since you cannot control the number of threads in `std::execution` policy, and also (AFAIK) you can't get the number of threads invoked by it. One way I could think of is:
> 
>     1. Fork `std::thread::hardware_concurrency` number of `std::thread`s.
> 
>     2. Make them operate on their own copy of histogram.
> 
>     3. `std::thread::join`.
> 
>     4. Make the main thread merge all the histogram into one, which will be used for further processing.
> 
> 
> If we do decide to go down this route, then I don't understand why we have the multiplication with `MaxParallelism` above, because it also has a factor of `VectorizationBoundSTL` and AFAICT we're not using any SIMD anywhere (that could've been a possibility with execution policy, I guess)?

It works for decode scratch memory, I've outlined this before
https://github.com/Devsh-Graphics-Programming/Nabla/blob/gpu_blit_filter/include/nbl/asset/filters/CBlitImageFilter.h#L506

We already "allocate" scratch per-thread & per-simd lane (yes the execution policy can make that possible, supposedly).

`MaxParallelism` is a good bound, and there's an [assert if it fails](https://github.com/Devsh-Graphics-Programming/Nabla/blob/gpu_blit_filter/include/nbl/asset/filters/CBlitImageFilter.h#L521) so we'll know if it does.

devshgraphicsprogramming: > I think [this](https://github.com/Devsh-Graphics-Programming/Nabla/blob/92228e74ddb54145e5865713b84780146a42f2d1/include/nbl/asset/filters/CBlitImageFilter.h#L51) can now go? After implementing the new scratch usage regime, I don't see this being used at all.

yes it can.

devshgraphicsprogramming: > > This snippet has a bug because ESU_COUNT will directly go to the default: case and won't report the total scratch size correctly.
> > I'll restructure this a bit.
> 
> Another bug is, since the LUT data type can vary `{16, 32, 64}`, we would be overestimating the scratch size most of the times.

Good catches, as for the second one, add a parameter for the LUT data type and make it compute accordingly.

achalpandeyy: Resolved.

PR #158, opened 2021-07-24
Non-blocking wait for fences

devshgraphicsprogramming:
@Crisspl review my fix for non-greedy fences, if all good then merge

<!-
By creating this pull request into Nabla, you agree to release all your past (even from previous commits) and present contributions in the Nabla repository under the Apache 2.0 license. If you're not the sole contributor, ensure that all contributors have signed the CLA agreeing to this.
-->

### DIFF
--- a/src/nbl/video/IOpenGL_LogicalDevice.h
+++ b/src/nbl/video/IOpenGL_LogicalDevice.h
@@ -211,9 +211,10 @@ namespace impl
         {
             static inline constexpr E_REQUEST_TYPE type = ERT_WAIT_FOR_FENCES;
             using retval_t = IGPUFence::E_STATUS;
+            using clock_t = std::chrono::steady_clock;
             core::SRange<IGPUFence*const,IGPUFence*const*,IGPUFence*const*> fences = { nullptr, nullptr };
+            clock_t::time_point timeoutPoint;
             bool waitForAll;
-            uint64_t timeout;
         };
         struct SRequestGetFenceStatus
         {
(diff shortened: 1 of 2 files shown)


src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -792,17 +788,17 @@ class IOpenGL_LogicalDevice : public ILogicalDevice, protected impl::IOpenGL_Log
...
+            for (bool notFirstRun=false; true; notFirstRun=true)
             {
                 for (uint32_t i=0u; i<_count; )
                 {
-                    const auto now = clock_t::now();
+                    const auto now = SRequestWaitForFences::clock_t::now();
                     if (timeout)
                     {
-                        if(now>=end)
+                        if(notFirstRun && now>=timeoutPoint)
```
Crisspl, line 798: maybe first `now` should be taken from outside (as param, queried from clock before submitting the request)? with very low timeouts it's possible it'll always be timeout without any GL call

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -792,17 +788,17 @@ class IOpenGL_LogicalDevice : public ILogicalDevice, protected impl::IOpenGL_Log
                 _waitAll = true;
                 timeout = 0xdeadbeefBADC0FFEull;
```
Crisspl, line 789: hmm, so in case _waitAll==false and _count!=1u, timeout is always 0 at first run? i dont get this

src/nbl/video/IOpenGL_LogicalDevice.h
devshgraphicsprogramming: `notFirstRun` condition should enforce at least one call to `glClientSyncWait`

src/nbl/video/IOpenGL_LogicalDevice.h
Crisspl: or is it on purpose?

src/nbl/video/IOpenGL_LogicalDevice.h
devshgraphicsprogramming: `notFirstRun` condition should enforce at least one call to `glClientSyncWait`

src/nbl/video/IOpenGL_LogicalDevice.h
devshgraphicsprogramming: if you're not waiting for all fences, it means that if any fence succeeds the function should return a sucess (according to the vulkan spec)

so even if the user passes in a large timeout, you cannot start waiting on the set of _count>=2 fences with large timeouts, first you need to poll them all (because a success from one should cause the execution to resume immediately)

Crisspl: neat, ok, makes perfect sense

PR #170, opened 2021-08-04
Danylo system

Erfan-Ahmadi:
## Description
Merging Dan's work and fixes into input_channel_demo


include/nbl/system/ISystem.h
```diff
@@ -102,7 +102,7 @@ class ISystem final : public core::IReferenceCounted
         void request_impl(SRequestType& req, FutureType& future, RequestParams&& params)
         {
             req.type = params.type;
-            req.params = std::move(params);
+            req.params = params;
```
devshgraphicsprogramming, line 105: @sadiuk why did you get rid of the `std::move` ?

sadiuk: oops, wasn't intending to. I had some issues with that `=` and forgot to return when resolved.

PR #173, opened 2021-08-06
How do you like them synchronisation apples

devshgraphicsprogramming:

### DIFF
--- a/include/nbl/asset/utils/CDirQuantCacheBase.h
+++ b/include/nbl/asset/utils/CDirQuantCacheBase.h
@@ -476,7 +476,10 @@ class CDirQuantCacheBase : public impl::CDirQuantCacheBase
 		template<E_FORMAT CacheFormat>
 		static inline size_t getSerializedCacheSizeInBytes_impl(size_t capacity)
 		{
-			return 1u+sizeof(size_t)*2u+phmap::container_internal::Group::kWidth+(sizeof(typename cache_type_t<CacheFormat>::slot_type)+1u)*capacity;
+			exit(0x45);
+			return 0x45;
+			// @Przemog TODO: I updated phmap and now kWdith went somewhere else
+			//return 1u+sizeof(size_t)*2u+phmap::container_internal::Group::kWidth+(sizeof(typename cache_type_t<CacheFormat>::slot_type)+1u)*capacity;
 		}
 		template<E_FORMAT CacheFormat>
 		static inline bool validateSerializedCache(const SBufferRange<const ICPUBuffer>& buffer)
--- a/include/nbl/asset/utils/IIncludeHandler.h
+++ b/include/nbl/asset/utils/IIncludeHandler.h
@@ -16,7 +16,7 @@ namespace asset
 class IIncludeHandler : public core::IReferenceCounted
 {
 	public:
-		_NBL_STATIC_INLINE_CONSTEXPR char* BUILTIN_PREFIX = "nbl/builtin/";
+		static constexpr const char* BUILTIN_PREFIX = "nbl/builtin/";
 		static bool isBuiltinPath(const system::path& _p)
 		{
 			const size_t prefix_len = strlen(BUILTIN_PREFIX);
--- a/include/nbl/system/IAsyncQueueDispatcher.h
+++ b/include/nbl/system/IAsyncQueueDispatcher.h
@@ -15,59 +15,79 @@ namespace impl
     public:
         IAsyncQueueDispatcherBase() = default;
         ~IAsyncQueueDispatcherBase() = default;
+        // dont want to play around with relaxed memory ordering yet
         struct request_base_t
         {
+                enum E_STATE : uint32_t
+                {
+                    ES_INITIAL=0,
+                    ES_RECORDING=1,
+                    ES_PENDING=2,
+                    ES_EXECUTING=3,
+                    ES_READY=4
+                };
+                request_base_t() : state(ES_INITIAL)
+                {
+                    assert(std::atomic_uint32_t::is_lock_free());
+                }
                 ~request_base_t() = default;
 
                 // lock when overwriting the request
-                std::unique_lock<std::mutex> reset()
+                void reset()
                 {
-                    std::unique_lock<std::mutex> lk(mtx);
-                    ready.store(false);
-                    return lk;
+                    transition(ES_INITIAL,ES_RECORDING);
                 }
                 // unlock request after we've written everything into it
-                void finalize(std::unique_lock<std::mutex>&& lk)
+                void finalize()
                 {
-                    lk.unlock();
-                    ready_for_work.store(true);
-                    #if __cplusplus >= 202002L
-                        ready_for_work.notify_one(); // does an atomic need to be notified under a lock?
-                    #endif
+                    const auto prev = state.exchange(ES_PENDING);
+                    assert(prev==ES_RECORDING);
+                    state.notify_one();
                 }
                 // do NOT allow canceling of request while they are processed
-                std::unique_lock<std::mutex> wait_for_work()
+                void wait_for_work()
                 {
-                        while (!ready_for_work.load())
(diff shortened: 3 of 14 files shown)


include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -476,7 +476,10 @@ class CDirQuantCacheBase : public impl::CDirQuantCacheBase
 		template<E_FORMAT CacheFormat>
 		static inline size_t getSerializedCacheSizeInBytes_impl(size_t capacity)
 		{
-			return 1u+sizeof(size_t)*2u+phmap::container_internal::Group::kWidth+(sizeof(typename cache_type_t<CacheFormat>::slot_type)+1u)*capacity;
+			exit(0x45);
+			return 0x45;
+			// @Przemog TODO: I updated phmap and now kWdith went somewhere else
```
devshgraphicsprogramming, line 481: @Przemog1 TODO

include/nbl/asset/utils/IIncludeHandler.h
```diff
@@ -16,7 +16,7 @@ namespace asset
 class IIncludeHandler : public core::IReferenceCounted
 {
 	public:
-		_NBL_STATIC_INLINE_CONSTEXPR char* BUILTIN_PREFIX = "nbl/builtin/";
+		static constexpr const char* BUILTIN_PREFIX = "nbl/builtin/";
```
Crisspl, line 19: what about inline?

include/nbl/system/IAsyncQueueDispatcher.h
```diff
@@ -15,63 +15,79 @@ namespace impl
...
+                {
+                    ES_INITIAL=0,
+                    ES_RECORDING=1,
+                    ES_PENDING=2,
+                    ES_EXECUTING=3,
+                    ES_READY=4
+                };
+                request_base_t() : state(ES_INITIAL)
+                {
+                    assert(std::atomic_uint32_t::is_lock_free());
```
Crisspl, line 31: are the platforms where this could not be true? out of curiosity

include/nbl/system/IAsyncQueueDispatcher.h
```diff
@@ -15,63 +15,79 @@ namespace impl
...
         ~IAsyncQueueDispatcherBase() = default;
+        // dont want to play around with relaxed memory ordering yet
         struct request_base_t
         {
+                enum E_STATE : uint32_t
+                {
+                    ES_INITIAL=0,
+                    ES_RECORDING=1,
+                    ES_PENDING=2,
+                    ES_EXECUTING=3,
```
Crisspl, line 26: whats the difference between PENDING and EXECUTING state?

include/nbl/system/IAsyncQueueDispatcher.h
```diff
@@ -15,63 +15,79 @@ namespace impl
...
-                    #if __cplusplus >= 202002L
-                        ready_for_work.notify_one(); // does an atomic need to be notified under a lock?
-                    #endif
+                    const auto prev = state.exchange(ES_PENDING);
+                    assert(prev==ES_RECORDING);
+                    state.notify_one();
                 }
                 // do NOT allow canceling of request while they are processed
-                std::unique_lock<std::mutex> wait_for_work()
+                void wait_for_work()
```
Crisspl, line 48: isnt this name misleading? i mean it waits for PENDING and then changes to EXECUTING, i'd expect just waiting for some state

include/nbl/asset/utils/IIncludeHandler.h
devshgraphicsprogramming: you said yourself we dont need inline anymore?

include/nbl/system/IAsyncQueueDispatcher.h
Crisspl: ah, ok, recorded, but not yet being executed

include/nbl/system/IAsyncQueueDispatcher.h
devshgraphicsprogramming: dunno, RISC-V, some embedded CPUs?

include/nbl/system/IAsyncQueueDispatcher.h
devshgraphicsprogramming: PENDING is the state between the end of writing the request struct into the circular buffer and the start of execution of the request (between the two user defined CRTP static polymorphic methods)

EXECUTING is the state during the call to the user defined CRTP static polymorphic request execution routine

devshgraphicsprogramming: Originally I just had pending state for both cases, but the Cancellable requests required that I make a distinction between pending (which can be cancelled) and executing (which must not)

include/nbl/system/IAsyncQueueDispatcher.h
devshgraphicsprogramming: I guess, but right now the focus isnt on how well semantically named my methods are... rather its whether my solution will work at all.

include/nbl/asset/utils/IIncludeHandler.h
Crisspl: what, no. I said we dont need macro any more (macro was to not put `inline` on certain compilers)

devshgraphicsprogramming: yea and in this case MSVC never got `inline`

include/nbl/system/ICancellableAsyncQueueDispatcher.h
```diff
@@ -97,108 +106,124 @@ class ICancellableAsyncQueueDispatcherBase
...
-        void associate_request(RequestType* req)
-        {
-            request = req;
-        }
+                void cancel()
+                {
+                    bool valid = valid_flag.exchange(false);
+                    if (valid)
+                        future_storage_t<T>::getStorage()->~T();
+                    impl::ICancellableAsyncQueueDispatcherBase::future_base_t::cancel();
```
Crisspl, line 154: i believe this function should call `set_cancel` on request?

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -444,10 +444,12 @@ class IOpenGL_LogicalDevice : public ILogicalDevice, protected impl::IOpenGL_Log
         template <typename RequestParams>
         void waitForRequestCompletion(SRequest& req)
         {
-            auto lk = req.wait_for_result();
+            req.wait_ready();
```
Crisspl, line 447: did you make sure no logical device functions use request api directly? (and not waitForRequestCompletion) From what i see user is always required to call wait_ready() and then discard_storage(). 
btw: nabla's interfaces IAsyncQueueDispatcher and Cancellable one dont provide `waitForRequestCompletion` so it must be written by user (using request api directly), i hope it's not an issue

include/nbl/asset/utils/IIncludeHandler.h
Crisspl: yep, it didnt support `inline`, now it's no problem (since we enabled cpp17)

src/nbl/video/IOpenGL_LogicalDevice.h
devshgraphicsprogramming: yes.

include/nbl/system/ICancellableAsyncQueueDispatcher.h
devshgraphicsprogramming: `ICancellableAsyncQueueDispatcherBase::future_base_t::cancel` does exactly that

but I looked at the code and it seems like it should cancel before it tries to delete the storage instead of what its doing now and was doing before (not my bug)

Crisspl: hm, yeah, probbaly would be better to first cancel and then delete the storage

PR #195, opened 2021-08-24
Raytracing device extensions

Erfan-Ahmadi:
## Description
Added Extensions and internal helper functions to add extensions

## Testing 
It was tested with multiple extensions and features


include/nbl/video/IPhysicalDevice.h
```diff
@@ -136,13 +142,33 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+    {
+        assert(false); // Invalid code path
+        return false;
+    }
 
+    virtual bool isRayQuerySupported() const
+    {
+        assert(false); // Invalid code path
+        return false;
+    }
```
devshgraphicsprogramming, line 162: I would have thought these would correllate to the booleans?

P.S. Dont assert here, the user is only asking for availability not actually using the extensions

Erfan-Ahmadi: Understood, I just followed what `isSwapchainSupported` did
I guess I should remove the assert from that function as well?

devshgraphicsprogramming: yes please

include/nbl/video/IPhysicalDevice.h
```diff
@@ -113,7 +113,13 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
     virtual IDebugCallback* getDebugCallback() = 0;
 
     virtual bool isSwapchainSupported() const = 0;
+        
+    virtual bool isRayTracingPipelineSupported() const = 0;
 
+    virtual bool isAccelerationStructuresSupported() const = 0;
+
+    virtual bool isRayQuerySupported() const = 0;
```
devshgraphicsprogramming, line 121: I'm a bit concerned about redundancy here, lots and lots of code to find out what's already present in 
```cpp
    const SLimits& getLimits() const { return m_limits; }
    const SFeatures& getFeatures() const { return m_features; }
```
anyway

Erfan-Ahmadi: Do you want me to remove these functions?  @devshgraphicsprogramming

devshgraphicsprogramming: yes, would make most sense.

devshgraphicsprogramming:
@Erfan-Ahmadi have @Crisspl review instead of me

Erfan-Ahmadi:
Seems like It's already being used and integrated in vulkan_backend, deferring review for that branch, closing here and deleting this useless branch

PR #203, opened 2021-08-31
Bullet and some other examples corrected

AnastaZIuk:


AnastaZIuk:
@devshgraphicsprogramming remember that if u were to launch ex 06 it expects you to put sponza.zip's contents upon executable (we don't have archive loader yet)

PR #209, opened 2021-09-03
Bitflag

Przemog1:


include/nbl/asset/IImage.h
```diff
@@ -232,7 +233,7 @@ class IImage : public IDescriptor
 			E_SAMPLE_COUNT_FLAGS						samples;
 			// stuff below is irrelevant in OpenGL backend
 			E_TILING									tiling = static_cast<E_TILING>(0);
-			std::underlying_type_t<E_USAGE_FLAGS>		usage = 0u;
+			core::bitflag<E_USAGE_FLAGS>				usage = static_cast<E_USAGE_FLAGS>(0);
```
Przemog1, line 236: There is no `E_USAGE_FLAGS` enum that is equal to 0, so I'm not sure if it is correct

devshgraphicsprogramming: good

Przemog1:
merged into `erfan_common_api_vk` branch

PR #219, opened 2021-09-13
CommonAPI to work with Vulkan Backend

Erfan-Ahmadi:
- [x] Merge @Przemog1's Bitflag Fixes
- [x] Device/Queue Selection in CommonAPI
- [x] SurfaceFormat Selection in CommonAPI

PR #160, opened 2021-07-26
Vulkan Backend

devshgraphicsprogramming:

PR #226, opened 2021-09-21
Upstream Everything to Scene Manager

devshgraphicsprogramming:


examples_tests/02.ComputeShader/main.cpp
```diff
@@ -418,56 +313,115 @@ int main()
...
-	// vkGetBufferMemoryRequirements)
-	video::IDriverMemoryBacked::SDriverMemoryRequirements memoryRequirements = {};
-	memoryRequirements.vulkanReqs.size = sizeof(UniformBufferObject);
-	memoryRequirements.vulkanBufferUsageFlags = VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT;
-	memoryRequirements.sharingMode = asset::ESM_EXCLUSIVE;
-	memoryRequirements.queueFamilyIndices = nullptr;
-	for (uint32_t i = 0u; i < swapchainImageCount; ++i)
-		ubos[i] = device->createGPUBuffer(memoryRequirements);
+	// Uncomment once the KTX loader works
+#if 0
```
devshgraphicsprogramming, line 317: doesn't it?

include/nbl/asset/IBuffer.h
```diff
@@ -18,6 +18,27 @@ class IBuffer : public core::IBuffer, public IDescriptor
...
+            EUF_INDIRECT_BUFFER_BIT = 0x00000100
+        };
+
+        struct SCreationParams
+        {
+            core::bitflag<E_USAGE_FLAGS> usage;
+            E_SHARING_MODE sharingMode;
+            uint32_t queueFamilyIndexCount;
+            const uint32_t* queueFamilyIndices;
+        };
```
devshgraphicsprogramming, line 40: @Crisspl wtf did you merge?

include/nbl/asset/ICPUImage.h
```diff
@@ -177,6 +177,13 @@ class ICPUImage final : public IImage, public IAsset
 			return true;
 		}
 
+		inline bool setImageUsageFlags(core::bitflag<E_USAGE_FLAGS> usage)
```
devshgraphicsprogramming, line 180: @Erfan-Ahmadi I need an `add` to `|` onto existing flags!

include/nbl/asset/IImage.h
```diff
@@ -232,9 +238,10 @@ class IImage : public IDescriptor
 			E_SAMPLE_COUNT_FLAGS						samples;
 			// stuff below is irrelevant in OpenGL backend
 			E_TILING									tiling = static_cast<E_TILING>(0);
```
devshgraphicsprogramming, line 240: @Erfan-Ahmadi @achalpandeyy suggestions for a default value?

include/nbl/asset/IImage.h
```diff
@@ -232,9 +238,10 @@ class IImage : public IDescriptor
...
-			std::underlying_type_t<E_USAGE_FLAGS>		usage = 0u;
+			core::bitflag<E_USAGE_FLAGS>				usage = static_cast<E_USAGE_FLAGS>(0);
 			E_SHARING_MODE								sharingMode = ESM_EXCLUSIVE;
-			core::smart_refctd_dynamic_array<uint32_t>	queueFamilyIndices = nullptr;
+			uint32_t									queueFamilyIndexCount;
+			const uint32_t*								queueFamilyIndices;
```
devshgraphicsprogramming, line 244: @achalpandeyy & @Crisspl what is this stuff doing here!?

include/nbl/ext/ScreenShot/ScreenShot.h
```diff
@@ -41,9 +41,11 @@ inline core::smart_refctd_ptr<asset::ICPUImageView> createScreenShot(video::ILog
 		auto extent = gpuImage->getMipSize();
 		video::IGPUImage::SBufferCopy pRegions[1u] = { {0u,extent.x,extent.y,{static_cast<asset::IImage::E_ASPECT_FLAGS>(0u),0,0u,1u},{0u,0u,0u},{extent.x,extent.y,extent.z}} };
 
+		video::IGPUBuffer::SCreationParams unused = {};
+
 		auto deviceLocalGPUMemoryReqs = logicalDevice->getDownStreamingMemoryReqs();
 		deviceLocalGPUMemoryReqs.vulkanReqs.size = extent.x*extent.y*extent.z*asset::getTexelOrBlockBytesize(fetchedGpuImageParams.format);
-		gpuTexelBuffer = logicalDevice->createGPUBufferOnDedMem(deviceLocalGPUMemoryReqs, true);
+		gpuTexelBuffer = logicalDevice->createGPUBufferOnDedMem(unused, deviceLocalGPUMemoryReqs, true);
```
devshgraphicsprogramming, line 48: huh, @Crisspl did you just break screenshots?

include/nbl/video/decl/IBackendObject.h
```diff
@@ -25,13 +25,17 @@ class IBackendObject
...
+
+        const std::string& getObjectDebugName() const { return m_debugName; }
 
     protected:
         virtual ~IBackendObject();
 
     private:
         const core::smart_refctd_ptr<const ILogicalDevice> m_originDevice;
+
+        mutable std::string m_debugName;
```
devshgraphicsprogramming, line 38: why do we even need to remember the debug name?

include/nbl/video/surface/ISurfaceGL.h
```diff
@@ -77,12 +85,30 @@ class CSurfaceGLImpl final : public Base<Window>
...
+        RECT wr;
+        GetWindowRect(m_handle, &wr);
+        return wr.right - wr.left;
+    }
+    uint32_t getHeight() const override 
+    { 
+        RECT wr;
+        GetWindowRect(m_handle, &wr);
+        return wr.top - wr.bottom;
+    }
```
devshgraphicsprogramming, line 110: who did this? @achalpandeyy @Erfan-Ahmadi ?

what's wrong with the window size the `IWindowWin32` would report?

include/nbl/video/utilities/CDrawIndirectAllocator.h
```diff
@@ -30,15 +30,25 @@ class CDrawIndirectAllocator final : public IDrawIndirectAllocator
             const auto& limits = params.device->getPhysicalDevice()->getLimits();
             
             ExplicitBufferCreationParameters explicit_params;
+
+            video::IGPUBuffer::SCreationParams creationParams;
+            creationParams.usage = asset::IBuffer::EUF_STORAGE_BUFFER_BIT;
```
devshgraphicsprogramming, line 35: wrong, needs INDIRECT usage first and foremost

devshgraphicsprogramming: and then STORAGE_BUFFER_BIT

include/nbl/video/utilities/CPropertyPool.h
```diff
@@ -43,13 +43,20 @@ class CPropertyPool final : public IPropertyPool
         static inline core::smart_refctd_ptr<this_t> create(ILogicalDevice* device, const uint32_t capacity, const bool contiguous = false)
         {
             asset::SBufferRange<video::IGPUBuffer> blocks[PropertyCount];
-            for (auto i=0u; i<PropertyCount; i++)
+            video::IGPUBuffer::SCreationParams params;
+            params.queueFamilyIndexCount = 0u;
+            params.queueFamilyIndices = nullptr;
+            params.sharingMode = asset::ESM_CONCURRENT;
+            //params.usage = // should be user-provided probably
```
devshgraphicsprogramming, line 50: STORAGE_BUFFER first and foremost, you can take additional usages from the user and `|` them in

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -447,8 +448,10 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUBuffer** const _begin
         reqs.vulkanReqs.size = addrAllctr.get_allocated_size();
         if (reqs.vulkanReqs.size==0u)
             return;
+
+        IGPUBuffer::SCreationParams unused = {};
```
devshgraphicsprogramming, line 452: WTF!? @Crisspl explain yourself

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -760,8 +763,11 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
     const auto assetCount = std::distance(_begin, _end);
     auto res = core::make_refctd_dynamic_array<created_gpu_object_array<asset::ICPUImage> >(assetCount);
 
+    // This should be the other way round because if a queue supports either compute or graphics
+    // but not the other way round
```
devshgraphicsprogramming, line 767: user must always provide a queue with transfer capability when converting buffers or images

if an explicit compute queue was not provided (even the same one as transfer, i.e. the queue is `nullptr`), because lets say we wont be computing any mip maps, we still need to know the final queue's family index for an ownership transfer.

Setting `computeFamIx==transferFamIx` spares us a lot of checking at the end.

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -777,7 +783,10 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
 
         // TODO: @criss why isn't this buffer cached and why are we not going through recursive asset creation and getting ICPUBuffer equivalents? 
         //(we can always discard/not cache the GPU Buffers created only for image data upload)
-        auto gpubuf = _params.device->createDeviceLocalGPUBufferOnDedMem(cpuimg->getBuffer()->getSize());
+        IGPUBuffer::SCreationParams params = {};
+        params.usage = core::bitflag(video::IGPUBuffer::EUF_TRANSFER_SRC_BIT) | video::IGPUBuffer::EUF_TRANSFER_DST_BIT;
+        const size_t size = cpuimg->getBuffer()->getSize();
+        auto gpubuf = _params.device->createDeviceLocalGPUBufferOnDedMem(params, size);
```
devshgraphicsprogramming, line 789: @achalpandeyy this needs to get fixed to not use an intermediate buffer for the upload (basically use staging directly as per our conversation earlier).

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -855,76 +864,117 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
             barrier.barrier.dstAccessMask = asset::EAF_TRANSFER_WRITE_BIT;
-            cmdbuf_transfer->pipelineBarrier(asset::EPSF_TRANSFER_BIT, asset::EPSF_TRANSFER_BIT, 0, 0u, nullptr, 1u, &barrier, 0u, nullptr);
-            cmdbuf_transfer->copyBufferToImage(buf.get(), img, asset::EIL_UNDEFINED, cpuimg->getRegions().size(), cpuimg->getRegions().begin());
+
+            IGPUCommandBuffer::SImageMemoryBarrier toTransferDst = {};
+            toTransferDst.barrier.srcAccessMask = static_cast<asset::E_ACCESS_FLAGS>(0u);
+            toTransferDst.barrier.dstAccessMask = asset::EAF_TRANSFER_WRITE_BIT;
+            toTransferDst.oldLayout = asset::EIL_UNDEFINED;
+            toTransferDst.newLayout = asset::EIL_TRANSFER_DST_OPTIMAL;
+            toTransferDst.srcQueueFamilyIndex = transferFamIx;
```
devshgraphicsprogramming, line 873: who did the image belong to before? (you just created it really)

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -855,76 +864,117 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
+            toTransferDst.dstQueueFamilyIndex = transferFamIx;
+            toTransferDst.image = core::smart_refctd_ptr<video::IGPUImage>(img);
+            toTransferDst.subresourceRange.aspectMask = asset::IImage::EAF_COLOR_BIT; // this probably shoudn't be hardcoded
```
devshgraphicsprogramming, line 876: all supported aspects?

devshgraphicsprogramming: or you could iterate through the regions beforehand and gather an apect <mask,baselevel,levelCount,baseArray,arrayCount> "AABB" for your layout transition

devshgraphicsprogramming: you only need to transition the aspects for which you have a `SRegion` that will be transferred into it.

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -855,76 +864,117 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
+            toTransferDst.subresourceRange.baseMipLevel = 0u;
+            toTransferDst.subresourceRange.levelCount = img->getCreationParameters().mipLevels;
+            toTransferDst.subresourceRange.baseArrayLayer = 0u;
+            toTransferDst.subresourceRange.layerCount = cpuimg->getCreationParameters().arrayLayers;
+
+            cmdbuf_transfer->pipelineBarrier(asset::EPSF_TRANSFER_BIT, asset::EPSF_TRANSFER_BIT, asset::EDF_NONE, 0u, nullptr, 1u, &barrier, 1u, &toTransferDst);
```
devshgraphicsprogramming, line 882: source stage seems wrong, you just created the image, so I guess it would be TOP_OF_PIPE

devshgraphicsprogramming: ah ok, you're barriering against the transfer stage that wrote your IGPUBuffer (that you should remove anyway)

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -855,76 +864,117 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
-        for (auto& region : cpuimg->getRegions())
-            lowestPresentMip = (std::max)(lowestPresentMip, region.imageSubresource.mipLevel);
-        // generate temporary image view to make sure we don't screw up any explicit mip levels
-        IGPUImageView::SCreationParams tmpViewParams;
-        tmpViewParams.subresourceRange.levelCount = img->getCreationParameters().mipLevels + 1u - lowestPresentMip;
-        // if not all mip levels have been manually specified
-        if (tmpViewParams.subresourceRange.levelCount > 1u)
+        video::IGPUCommandBuffer::SImageMemoryBarrier barrier = {};
+        barrier.srcQueueFamilyIndex = ~0u;
+        barrier.dstQueueFamilyIndex = ~0u;
```
devshgraphicsprogramming, line 899: transfer to compute family, you already have them

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -855,76 +864,117 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
+        barrier.image = core::smart_refctd_ptr<video::IGPUImage>(gpuimg);
+        // TODO this is probably wrong (especially in case of depth/stencil formats), but i think i can leave it like this since we'll never have any depth/stencil images loaded (right?)
+        barrier.subresourceRange.aspectMask = cpuimg->getRegions().begin()->imageSubresource.aspectMask; 
+        barrier.subresourceRange.levelCount = 1u;
```
devshgraphicsprogramming, line 903: all mips you'll generate (see `tmpViewParams.subresourceRange.levelCount = img->getCreationParameters().mipLevels + 1u - lowestPresentMip`)

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -855,76 +864,117 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
-        IGPUImageView::SCreationParams tmpViewParams;
-        tmpViewParams.subresourceRange.levelCount = img->getCreationParameters().mipLevels + 1u - lowestPresentMip;
-        // if not all mip levels have been manually specified
-        if (tmpViewParams.subresourceRange.levelCount > 1u)
+        video::IGPUCommandBuffer::SImageMemoryBarrier barrier = {};
+        barrier.srcQueueFamilyIndex = ~0u;
+        barrier.dstQueueFamilyIndex = ~0u;
+        barrier.image = core::smart_refctd_ptr<video::IGPUImage>(gpuimg);
+        // TODO this is probably wrong (especially in case of depth/stencil formats), but i think i can leave it like this since we'll never have any depth/stencil images loaded (right?)
+        barrier.subresourceRange.aspectMask = cpuimg->getRegions().begin()->imageSubresource.aspectMask;
```
devshgraphicsprogramming, line 902: can you even write with a compute shader to a depth or stencil image view directly?

devshgraphicsprogramming: but yeah we'll probably not support the automatic mipping of depth or stencil, so this can always be the COLOR aspect.

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -855,76 +864,117 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
-            auto tmpView = _params.device->createGPUImageView(std::move(tmpViewParams));
-
-            // deprecated OpenGL path (do with compute shader in the future)
-            cmdbuf_compute->regenerateMipmaps(tmpView.get());
+            const uint32_t srcLoD = i - 1u;
+
+            barrier.barrier.srcAccessMask = asset::EAF_TRANSFER_WRITE_BIT;
+            barrier.barrier.dstAccessMask = asset::EAF_TRANSFER_READ_BIT;
+            barrier.oldLayout = asset::EIL_TRANSFER_DST_OPTIMAL;
+            barrier.newLayout = asset::EIL_TRANSFER_SRC_OPTIMAL;
```
devshgraphicsprogramming, line 930: a blit requires a transfer SRC layout!?

Really!? @achalpandeyy @Erfan-Ahmadi ?

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
devshgraphicsprogramming: ah ok, I see, the source mipmap needs to be in one layout and dest in another.

include/nbl/video/utilities/IUtilities.h
```diff
@@ -159,6 +226,7 @@ class IUtilities : public core::IReferenceCounted
             submit.pWaitSemaphores = semaphoresToWaitBeforeExecution;
             submit.pWaitDstStageMask = stagesToWaitForPerSemaphore;
             queue->submit(1u,&submit,fence);
+            m_device->waitForFences(1u, &fence, false, 9999999999ull);
```
devshgraphicsprogramming, line 229: remove this entire method (fold it into the one below)

include/nbl/video/utilities/IUtilities.h
```diff
@@ -104,6 +138,7 @@ class IUtilities : public core::IReferenceCounted
...
```
devshgraphicsprogramming, line 141: fix the above to do filling directly from the staging buffer without creating an intermediate, then you'll get the opportunity to add the fence and the commandbuffer to the staging buffer's zombie list during a `multi_free`

include/nbl/video/utilities/renderdoc_app.h
```diff
@@ -0,0 +1,688 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ ******************************************************************************/
+
```
devshgraphicsprogramming, line 24: WTF @Crisspl its already in `3rdparty` !

src/nbl/asset/interchange/CImageLoaderOpenEXR.cpp
```diff
@@ -376,7 +376,7 @@ namespace nbl
 						auto texelBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(image->getImageDataSizeInBytes());
 						auto regions = core::make_refctd_dynamic_array<core::smart_refctd_dynamic_array<ICPUImage::SBufferCopy>>(1u);
 						ICPUImage::SBufferCopy& region = regions->front();
-						//region.imageSubresource.aspectMask = ...; // waits for Vulkan
+						region.imageSubresource.aspectMask = IImage::E_ASPECT_FLAGS::EAF_COLOR_BIT; // TODO(Erfan) ? Deduec or always EAF_COLOR_BIT?
```
devshgraphicsprogramming, line 379: yes always, OpenEXR doesnt load other formats, you can remove the comment now @Erfan-Ahmadi

src/nbl/video/COpenGLESFunctionTable.h
```diff
@@ -624,6 +628,46 @@ class COpenGLESFunctionTable final : public IOpenGL_FunctionTable
...
+			case GL_SHADER:				identifier = GL_SHADER_OBJECT_EXT; break;
+			case GL_PROGRAM:			identifier = GL_PROGRAM_OBJECT_EXT; break;
+			case GL_VERTEX_ARRAY:		identifier = GL_VERTEX_ARRAY_OBJECT_EXT; break;
+			case GL_QUERY:				identifier = GL_QUERY_OBJECT_EXT; break;
+			case GL_PROGRAM_PIPELINE:	identifier = GL_PROGRAM_PIPELINE_OBJECT_EXT; break;
+			// other types does not need to be translated into extension-specific enums
+			}
+			glesDebug.pglGetObjectLabelEXT(identifier, name, bufsize, length, label);
+		}
+	}
```
devshgraphicsprogramming, line 670: what do you even need this for?

src/nbl/video/IOpenGL_FunctionTable.h
```diff
@@ -1080,10 +1091,10 @@ namespace nbl {
 			{
 				// Save the previous bound vertex array
 				GLint restoreVertexArray;
-				_NBL_GL_CALL(m_logger, glGeneral.pglGetIntegerv(GL_VERTEX_ARRAY_BINDING, &restoreVertexArray));
+				//_NBL_GL_CALL(m_logger, glGeneral.pglGetIntegerv(GL_VERTEX_ARRAY_BINDING, &restoreVertexArray));
 				_NBL_GL_CALL(m_logger, glVertex.pglBindVertexArray(vaobj));
 				_NBL_GL_CALL(m_logger, glVertex.pglEnableVertexAttribArray(index));
-				_NBL_GL_CALL(m_logger, glVertex.pglBindVertexArray(restoreVertexArray));
+				//_NBL_GL_CALL(m_logger, glVertex.pglBindVertexArray(restoreVertexArray));
```
devshgraphicsprogramming, line 1097: @Crisspl WTF!?

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -261,6 +263,16 @@ namespace impl
...
         };
+        struct SRequestSetDebugName
+        {
+            static inline constexpr E_REQUEST_TYPE type = ERT_SET_DEBUG_NAME;
+            using retval_t = void;
+
+            GLenum id;
+            GLuint object;
+            GLsizei len;
+            std::string label;
```
devshgraphicsprogramming, line 274: OpenGL only supports max 256 character long debug names

Crisspl: 256 is min max afaik

include/nbl/video/IDriverMemoryAllocation.h
```diff
@@ -141,8 +143,12 @@ class IDriverMemoryAllocation : public virtual core::IReferenceCounted
             currentMappingAccess = access;
         }
 
-        IDriverMemoryAllocation() : mappedPtr(nullptr), mappedRange(0,0), currentMappingAccess(EMCAF_NO_MAPPING_ACCESS) {}
+        IDriverMemoryAllocation(const ILogicalDevice* originDevice)
+            : m_originDevice(originDevice), mappedPtr(nullptr), mappedRange(0,0),
+            currentMappingAccess(EMCAF_NO_MAPPING_ACCESS)
+        {}
 
+        const ILogicalDevice* m_originDevice = nullptr;
```
devshgraphicsprogramming, line 151: @achalpandeyy remind me, did we decide AGAINST `IDriverMemoryAllocation` keeping a smartrefctd pointer back to the `ILogicalDevice`?

AFAIK we only decided to not keep smartpointers to `IDriverMemoryAllocation`

src/nbl/video/IOpenGL_FunctionTable.h
Crisspl: i guess i forgot to bring back, 
anyway this rebinding previous state is basically pointless, we dont let anyone call GL on our internal threads

src/nbl/video/alloc/SimpleGPUBufferAllocator.cpp
```diff
@@ -16,5 +16,15 @@ SimpleGPUBufferAllocator::value_type SimpleGPUBufferAllocator::allocate(size_t b
...
+    // Todo(achal): Need to get these values from the user somehow
+    // This doesn't matter right now because updateBufferRangeViaStagingBuffer 
+    // only makes use of upstreaming buffer which would be TRANSFER_SRC
+    IGPUBuffer::SCreationParams creationParams = {};
+    creationParams.usage = IGPUBuffer::EUF_TRANSFER_SRC_BIT;
+    creationParams.sharingMode = asset::ESM_EXCLUSIVE;
+    creationParams.queueFamilyIndexCount = 0u;
+    creationParams.queueFamilyIndices = nullptr;
+
+    return mDriver->createGPUBufferOnDedMem(creationParams, reqs, canUpdateViaCmdBuff);
```
devshgraphicsprogramming, line 29: @achalpandeyy could we fold `canUpdateViaCmdBuff` into `IGPUBuffer::ScreationParams` ?

include/nbl/asset/IImage.h
devshgraphicsprogramming: that doesn't impact our ability to transfer in and out of them?

include/nbl/video/declarations.h
```diff
@@ -23,6 +23,8 @@
 // platform and API specific stuff
 #include "nbl/video/COpenGL_Connection.h"
 #include "nbl/video/surface/ISurfaceGL.h"
+//for some reason CVulkanConnection is in src/ ???
+//#include "nbl/video/CVulkanConnection.h"
```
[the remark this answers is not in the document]
devshgraphicsprogramming: `CVulkanConnection.h` should be in `include/nbl/video` not `src`

src/nbl/video/IOpenGL_LogicalDevice.h
devshgraphicsprogramming: still play it safe, also having `std::string` floating around in GPU API objects is awful (malloc every !*("$@ where)

include/nbl/asset/IImage.h
Erfan-Ahmadi: I'm not sure didn't find anything in the spec, there is only 2 options for a default and OPTIMAL seems like a better choice to me.

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
Erfan-Ahmadi: 1.
> VUID-vkCmdBlitImage-srcImageLayout-01398
srcImageLayout must be VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR, VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL or VK_IMAGE_LAYOUT_GENERAL
> dstImage must have been created with VK_IMAGE_USAGE_TRANSFER_DST_BIT usage flag
2. 
> VK_PIPELINE_STAGE_TRANSFER_BIT specifies the following commands:
All copy commands, including vkCmdCopyQueryPoolResults
vkCmdBlitImage2KHR and vkCmdBlitImage
vkCmdResolveImage2KHR and vkCmdResolveImage
All clear commands, with the exception of vkCmdClearAttachments
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPipelineStageFlagBits.html

src/nbl/video/SOpenGLContextLocalCache.cpp
```diff
@@ -282,6 +284,14 @@ void SOpenGLContextLocalCache::flushStateGraphics(IOpenGL_FunctionTable* gl, uin
...
 
             currentState.pipeline.graphics.pipeline = nextState.pipeline.graphics.pipeline;
+
+            if (currentState.pipeline.graphics.pipeline->getObjectDebugName().size())
+            {
+                const GLuint glname = currentState.pipeline.graphics.usedPipeline;
+                const auto& dbgnm = currentState.pipeline.graphics.pipeline->getObjectDebugName();
+
+                gl->extGlObjectLabel(GL_PROGRAM_PIPELINE, glname, dbgnm.size(), dbgnm.c_str());
+            }
```
devshgraphicsprogramming, line 294: why is this here? inside of a flush state!?

src/nbl/video/SOpenGLContextLocalCache.cpp
```diff
@@ -604,6 +614,13 @@ void SOpenGLContextLocalCache::flushStateGraphics(IOpenGL_FunctionTable* gl, uin
             }
 
             GLuint vao = currentState.vertexInputParams.vaoval.GLname;
+
+            if (currentState.pipeline.graphics.pipeline->getObjectDebugName().size())
+            {
+                const auto& dbgnm = currentState.pipeline.graphics.pipeline->getObjectDebugName();
+                gl->extGlObjectLabel(GL_VERTEX_ARRAY, vao, dbgnm.size(), dbgnm.c_str());
+            }
+
```
devshgraphicsprogramming, line 623: why is this here? inside of a flush state!?

src/nbl/video/COpenGL_Queue.h
```diff
@@ -245,6 +257,12 @@ class COpenGL_Queue final : public IGPUQueue
                     _state.ctxlocal.removePipelineEntry(&gl, hash);
                 }
                 break;
+                case ERT_BEGIN_CAPTURE:
+                    m_rdoc_api->StartFrameCapture(NULL, NULL);
+                break;
+                case ERT_END_CAPTURE:
+                    m_rdoc_api->EndFrameCapture(NULL, NULL);
+                break;
```
devshgraphicsprogramming, line 265: @Crisspl this wont work
> RENDERDOC_DevicePointer and RENDERDOC_WindowHandle are described above in SetActiveWindow(). device and wndHandle can either or both be set to NULL to wildcard match against active device/window combinations. This wildcard matching can be used if the handle is difficult to obtain where frame captures are triggered.

> For example if device is NULL but wndHandle is set, RenderDoc will begin a capture on the first API it finds that is active on that window.

> If the wildcard match has multiple possible candidates, it is not defined which will be chosen. Wildcard matching should only be used when e.g. it is known that only one API is active on a window, or there is only one window active for a given API.

> If no window has been created and all rendering is off-screen, NULL can be specified for the window handle and the device object can be passed to select that API. If both are set to NULL, RenderDoc will simply choose one at random so is only recommended for the case where only one is present.


You should at least pass in `RENDERDOC_DevicePointer` which is
> For OpenGL it must be the HGLRC, GLXContext, or EGLContext context object.
> For OpenGLES it must be the EGLContext context object.

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -76,10 +75,18 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
 
+        VkPresentModeKHR vkPresentMode;
+        if((params.presentMode & ISurface::E_PRESENT_MODE::EPM_IMMEDIATE) == ISurface::E_PRESENT_MODE::EPM_IMMEDIATE)
+            vkPresentMode = VK_PRESENT_MODE_IMMEDIATE_KHR;
+        else if((params.presentMode & ISurface::E_PRESENT_MODE::EPM_MAILBOX) == ISurface::E_PRESENT_MODE::EPM_MAILBOX)
+            vkPresentMode = VK_PRESENT_MODE_MAILBOX_KHR;
+        else if((params.presentMode & ISurface::E_PRESENT_MODE::EPM_FIFO) == ISurface::E_PRESENT_MODE::EPM_FIFO)
+            vkPresentMode = VK_PRESENT_MODE_FIFO_KHR;
+        else if((params.presentMode & ISurface::E_PRESENT_MODE::EPM_FIFO_RELAXED) == ISurface::E_PRESENT_MODE::EPM_FIFO_RELAXED)
+            vkPresentMode = VK_PRESENT_MODE_FIFO_RELAXED_KHR;
```
devshgraphicsprogramming, line 88: should have been a switch with a `assert(false)` in the `default:` case

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -304,92 +324,249 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
-        // Todo(achal): Hoist creation out of constructor
-        return nullptr; // return core::make_smart_refctd_ptr<CVulkanRenderpass>(this, params);
+        // auto* vk = m_vkdev->getFunctionTable();
+
+        VkRenderPassCreateInfo createInfo = { VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO };
+        createInfo.pNext = nullptr;
+        createInfo.flags = static_cast<VkRenderPassCreateFlags>(0u); // No flags are supported
+        createInfo.attachmentCount = params.attachmentCount;
+
+        core::vector<VkAttachmentDescription> attachments(createInfo.attachmentCount); // TODO reduce number of allocations/get rid of vectors
```
devshgraphicsprogramming, line 334: this usage of a vector is actually fine, its out of the hot path and its sized up front.

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -304,92 +324,249 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
+            auto& vkatt = attachments[i];
+            vkatt.flags = static_cast<VkAttachmentDescriptionFlags>(0u); // No flags are supported
+            vkatt.format = getVkFormatFromFormat(att.format);
+            vkatt.samples = static_cast<VkSampleCountFlagBits>(att.samples);
+            vkatt.loadOp = static_cast<VkAttachmentLoadOp>(att.loadOp);
+            vkatt.storeOp = static_cast<VkAttachmentStoreOp>(att.storeOp);
+
+            // Todo(achal): Do we want these??
+            vkatt.stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+            vkatt.stencilStoreOp = VK_ATTACHMENT_STORE_OP_DONT_CARE;
```
devshgraphicsprogramming, line 347: YES

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -304,92 +324,249 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
+        VkMappedMemoryRange vk_memoryRanges[MAX_MEMORY_RANGE_COUNT];
+
+        const uint32_t memoryRangeCount = static_cast<uint32_t>(ranges.size());
+        assert(memoryRangeCount <= MAX_MEMORY_RANGE_COUNT);
+
+        getVkMappedMemoryRanges(vk_memoryRanges, ranges.begin(), ranges.end());
+
+        if (vkInvalidateMappedMemoryRanges(m_vkdev, memoryRangeCount, vk_memoryRanges) != VK_SUCCESS)
+            printf("invalidateMappedMemoryRanges failed!\n");
     }
```
devshgraphicsprogramming, line 485: @achalpandeyy do it, change them into boolean returns

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -304,92 +324,249 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
+        constexpr uint32_t MAX_MEMORY_RANGE_COUNT = 408u;
+        VkMappedMemoryRange vk_memoryRanges[MAX_MEMORY_RANGE_COUNT];
+
+        const uint32_t memoryRangeCount = static_cast<uint32_t>(ranges.size());
+        assert(memoryRangeCount <= MAX_MEMORY_RANGE_COUNT);
+
+        getVkMappedMemoryRanges(vk_memoryRanges, ranges.begin(), ranges.end());
+        
+        if (vkFlushMappedMemoryRanges(m_vkdev, memoryRangeCount, vk_memoryRanges) != VK_SUCCESS)
+            printf("flushMappedMemoryRanges failed\n");
```
devshgraphicsprogramming, line 469: either dont print or use the logger provided to the device.

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -304,92 +324,249 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
-            if ((pBindInfos[i].buffer->getAPIType() != EAT_VULKAN) /*|| (pBindInfos[i].memory->getAPIType() != EAT_VULKAN)*/)
+            const auto& bindInfo = pBindInfos[i];
+            
+            if ((bindInfo.buffer->getAPIType() != EAT_VULKAN) || (bindInfo.memory->getAPIType() != EAT_VULKAN))
                 continue;
 
-            VkBuffer vk_buffer = static_cast<const CVulkanBuffer*>(pBindInfos[i].buffer)->getInternalObject();
+            CVulkanBuffer* vulkanBuffer = static_cast<CVulkanBuffer*>(bindInfo.buffer);
+            vulkanBuffer->setMemoryAndOffset(
+                core::smart_refctd_ptr<IDriverMemoryAllocation>(bindInfo.memory), bindInfo.offset);
```
devshgraphicsprogramming, line 499: remember that its also an error to bind memory more than once to a buffer (unless the buffer is sparse, but then its a sparse bind request which is something completely different)

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -453,39 +652,60 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
                 continue;
-            VkImage vk_image = static_cast<const CVulkanImage*>(pBindInfos[i].image)->getInternalObject();
 
-            // if (pBindInfos[i].memory->getAPIType() != EAT_VULKAN)
-            //     continue;
-            VkDeviceMemory vk_deviceMemory = static_cast<const CVulkanMemoryAllocation*>(pBindInfos[i].memory)->getInternalObject();
+            CVulkanImage* vulkanImage = static_cast<CVulkanImage*>(bindInfo.image);
+            vulkanImage->setMemoryAndOffset(
+                core::smart_refctd_ptr<IDriverMemoryAllocation>(bindInfo.memory),
+                bindInfo.offset);
```
devshgraphicsprogramming, line 666: remember that its also an error to bind memory more than once to an image (unless the image is sparse, but then its a sparse bind request which is something completely different)

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -616,97 +836,56 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
+        vk_createInfo.mipLodBias = _params.LodBias;
+        vk_createInfo.anisotropyEnable = _params.AnisotropicFilter; // Todo(achal): Verify
+        vk_createInfo.maxAnisotropy = std::log2(_params.AnisotropicFilter); // Todo(achal): Verify
+        vk_createInfo.compareEnable = _params.CompareEnable;
+        vk_createInfo.compareOp = static_cast<VkCompareOp>(_params.CompareFunc);
+        vk_createInfo.minLod = _params.MinLod;
+        vk_createInfo.maxLod = _params.MaxLod;
+        assert(_params.BorderColor < asset::ISampler::ETBC_COUNT);
+        vk_createInfo.borderColor = static_cast<VkBorderColor>(_params.BorderColor);
+        vk_createInfo.unnormalizedCoordinates = VK_FALSE; // Todo(achal): Verify
```
devshgraphicsprogramming, line 878: good

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -794,8 +973,49 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
+        // assert(!(m_params.flags & ECF_IMAGELESS_BIT));
+
+        constexpr uint32_t MemSize = 1u << 12;
+        constexpr uint32_t MaxAttachments = MemSize / sizeof(VkImageView);
+
+        VkImageView vk_attachments[MaxAttachments];
+        uint32_t attachmentCount = 0u;
+        for (uint32_t i = 0u; i < params.attachmentCount; ++i)
+        {
+            if (params.attachments[i]->getAPIType() == EAT_VULKAN)
```
devshgraphicsprogramming, line 986: you need to hardcore assert this, or bail out (retur nullptr) on the first one that doesnt fullfill this

otherwise the indices to the attachments in the subpass dependencies, etc. will all be messed up

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -0,0 +1,149 @@
...
+    const IDriverMemoryBacked::SDriverMemoryRequirements& additionalReqs)
+{
+    if (getUpStreamingMemoryReqs().memoryHeapLocation != additionalReqs.memoryHeapLocation)
+        return nullptr;
+
+    IDriverMemoryBacked::SDriverMemoryRequirements memoryReqs = getUpStreamingMemoryReqs();
+    memoryReqs.vulkanReqs.alignment = core::max(memoryReqs.vulkanReqs.alignment, additionalReqs.vulkanReqs.alignment);
+    memoryReqs.vulkanReqs.size = core::max(memoryReqs.vulkanReqs.size, additionalReqs.vulkanReqs.size);
+    memoryReqs.vulkanReqs.memoryTypeBits &= additionalReqs.vulkanReqs.memoryTypeBits;
+    memoryReqs.mappingCapability = additionalReqs.mappingCapability;
```
devshgraphicsprogramming, line 51: Upstreaming shall only have WRITE mapping capability

CACHED and COHERENT could be enabled if there exists a heap in `memoryTypeBits` which supports that.

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -0,0 +1,149 @@
...
+    const IDriverMemoryBacked::SDriverMemoryRequirements& additionalReqs)
+{
+    if (getDownStreamingMemoryReqs().memoryHeapLocation != additionalReqs.memoryHeapLocation)
+        return nullptr;
+
+    IDriverMemoryBacked::SDriverMemoryRequirements memoryReqs = getDownStreamingMemoryReqs();
+    memoryReqs.vulkanReqs.alignment = core::max(memoryReqs.vulkanReqs.alignment, additionalReqs.vulkanReqs.alignment);
+    memoryReqs.vulkanReqs.size = core::max(memoryReqs.vulkanReqs.size, additionalReqs.vulkanReqs.size);
+    memoryReqs.vulkanReqs.memoryTypeBits &= additionalReqs.vulkanReqs.memoryTypeBits;
+    memoryReqs.mappingCapability = additionalReqs.mappingCapability;
```
devshgraphicsprogramming, line 69: READ only mapping capability

COHERENT if possible (`memoryTypeBits` exist that support it)

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
-		colorAttRef.layout = asset::EIL_UNDEFINED;
+		colorAttRef.layout = asset::EIL_COLOR_ATTACHMENT_OPTIMAL;
 
 		video::IGPURenderpass::SCreationParams::SSubpassDescription::SAttachmentRef depthStencilAttRef;
 		depthStencilAttRef.attachment = 1u;
-		depthStencilAttRef.layout = asset::EIL_UNDEFINED;
+		depthStencilAttRef.layout = asset::EIL_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
 
 		video::IGPURenderpass::SCreationParams::SSubpassDescription sp;
+		sp.pipelineBindPoint = asset::EPBP_GRAPHICS;
```
devshgraphicsprogramming, line 933: @Crisspl why does `subpassdescription` even have a pipeline bind point?
CC: @achalpandeyy is this something to do with mesh shaders?

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
+		}
+
+		video::ISwapchain::SCreationParams sc_params = {};
 		sc_params.width = width;
 		sc_params.height = height;
 		sc_params.arrayLayers = 1u;
 		sc_params.minImageCount = imageCount;
-		sc_params.presentMode = presentMode;
+		sc_params.presentMode = requestedPresentMode;
+		sc_params.imageUsage = static_cast<asset::IImage::E_USAGE_FLAGS>(asset::IImage::EUF_COLOR_ATTACHMENT_BIT | asset::IImage::EUF_STORAGE_BIT | asset::IImage::EUF_TRANSFER_DST_BIT);;
```
devshgraphicsprogramming, line 896: why transfer DST?

For screenshots it should be SRC

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
-	static nbl::core::smart_refctd_ptr<nbl::video::ISwapchain> createSwapchain(uint32_t width,
+	static nbl::core::smart_refctd_ptr<nbl::video::ISwapchain> createSwapchain(
+		nbl::video::E_API_TYPE api_type,
+		uint32_t width,
 		uint32_t height,
 		uint32_t imageCount,
 		const nbl::core::smart_refctd_ptr<nbl::video::ILogicalDevice>& device,
 		const nbl::core::smart_refctd_ptr<nbl::video::ISurface>& surface,
-		nbl::video::ISurface::E_PRESENT_MODE presentMode)
+		nbl::video::ISurface::E_PRESENT_MODE requestedPresentMode,
```
devshgraphicsprogramming, line 829: make FIFO_RELAXED the default argument for that one

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
+		nbl::video::ISurface::SFormat requestedSurfaceFormat,
```
devshgraphicsprogramming, line 830: SRGB 8bit default

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
+			}
+
+			bool presentModeSupported = (gpuInfo.availablePresentModes & requestedPresentMode) != 0;
+			assert(presentModeSupported);
+			if(!presentModeSupported) // fallback 
+			{
+				requestedPresentMode = nbl::video::ISurface::E_PRESENT_MODE::EPM_FIFO;
+				assert(false && "Fallback: requested 'present mode' is not supported");
+			}
+		}
```
devshgraphicsprogramming, line 884: @achalpandeyy what happened to making OpenGL report its surface formats and present modes?

achalpandeyy: It reports present modes correctly. Need to do surface formats.

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
+		nbl::video::ISurface::SFormat surfaceFormat;
+
+		if(api_type == video::EAT_VULKAN)
+		{
+			uint32_t found_format_and_colorspace = ~0u;
+			uint32_t found_format = ~0u;
+			for(uint32_t i = 0; i < gpuInfo.availableSurfaceFormats.size(); ++i)
+			{
+				const auto & supportedFormat = gpuInfo.availableSurfaceFormats[i];
+				if(requestedSurfaceFormat.format == supportedFormat.format)
```
devshgraphicsprogramming, line 850: format with matching eotf and colorspace, but with wider bitdepth is an acceptable substitute

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
-		q_params.priorities = &priority;
-		dev_params.queueCreateInfos = &q_params;
+		dev_params.queueParamsCount = actualQueueCount;
+		dev_params.queueParams = qcp;
 		result.logicalDevice = gpu->createLogicalDevice(dev_params);
 
 		result.utilities = core::make_smart_refctd_ptr<video::IUtilities>(core::smart_refctd_ptr(result.logicalDevice));
 
-		auto queue = result.logicalDevice->getQueue(familyIndex, 0);
+		result.mainQueue = result.logicalDevice->getQueue(mainQueueFamilyIndex, 0);
```
devshgraphicsprogramming, line 765: what do you need a `mainQueue` for?

examples_tests/common/CommonAPI.h
```diff
@@ -360,53 +567,84 @@ class CommonAPI
...
+		{
+			result.apiConnection = video::COpenGLConnection::create(core::smart_refctd_ptr(result.system), 0, app_name.data(), video::COpenGLDebugCallback(core::smart_refctd_ptr(result.logger)));
+		}
+		else if(api_type == EAT_OPENGL_ES)
+		{
+			result.apiConnection = video::COpenGLESConnection::create(core::smart_refctd_ptr(result.system), 0, app_name.data(), video::COpenGLDebugCallback(core::smart_refctd_ptr(result.logger)));
+		}
+		else
+		{
+			_NBL_TODO();
```
devshgraphicsprogramming, line 584: `api_type = rand()%3u` and go again

examples_tests/common/CommonAPI.h
```diff
@@ -293,6 +293,211 @@ class CommonAPI
...
+		if(ret == ~0u)
+		{
+			assert(false);
+			ret = 0;
+		}
+
+		return ret;
+	}
 
 	template<uint32_t sc_image_count> //! input with window creation
```
devshgraphicsprogramming, line 502: no idea why swapchain image count is a template argument

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
+			actualQueueCount++;
+		}
+		if(gpuInfo.queueFamilyProps.transfer.index != gpuInfo.queueFamilyProps.compute.index && gpuInfo.queueFamilyProps.transfer.index != gpuInfo.queueFamilyProps.graphics.index)
+		{
+			qcp[actualQueueCount].flags = static_cast<video::IGPUQueue::E_CREATE_FLAGS>(0);
+			qcp[actualQueueCount].familyIndex = gpuInfo.queueFamilyProps.transfer.index;
+			qcp[actualQueueCount].count = 1u;
+			qcp[actualQueueCount].priorities = &queuePriority;
+			actualQueueCount++;
+		}
```
devshgraphicsprogramming, line 746: I think I'll always want 2 separate queues for upload and download even if they're from the same family

examples_tests/common/CommonAPI.h
```diff
@@ -293,6 +293,211 @@ class CommonAPI
...
+				if(extractedInfo.queueFamilyProps.present.index != extractedInfo.queueFamilyProps.graphics.index && extractedInfo.queueFamilyProps.graphics.supportsPresent) 
+				{
+					extractedInfo.queueFamilyProps.present = extractedInfo.queueFamilyProps.graphics;
+				}
+
+				// If a unique Compute is not found but Graphics supports Compute, then use Graphics for Compute
+				if(extractedInfo.queueFamilyProps.compute.index == QueueFamilyProps::InvalidIndex && extractedInfo.queueFamilyProps.graphics.supportsCompute)
+				{
+					extractedInfo.queueFamilyProps.compute = extractedInfo.queueFamilyProps.graphics;
+				}
```
devshgraphicsprogramming, line 420: its good to have graphics and compute be separate queues but probably not separate families

your function should know if you require graphics (headless compute) or not then pick a family that definitely has compute, but conditionally graphics

the fallback should be to have two separate families for compute and graphics (dunno if spec even allows for not having a family that doesnt support both compute and graphics)

examples_tests/common/CommonAPI.h
```diff
@@ -293,6 +293,211 @@ class CommonAPI
...
+				if ((extractedInfo.queueFamilyProps.compute.index != QueueFamilyProps::InvalidIndex) &&
+					(extractedInfo.queueFamilyProps.transfer.index != QueueFamilyProps::InvalidIndex))
+					isGPUSuitable = true;
+			}
+
+			if(extractedInfo.isSwapChainSupported == false)
+				isGPUSuitable = false;
+
+			if(extractedInfo.hasSurfaceCapabilities == false)
+				isGPUSuitable = false;
```
devshgraphicsprogramming, line 483: for headless compute its fine (no graphics and no swapchain).

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
devshgraphicsprogramming: lets transfer the entire image to `VK_IMAGE_LAYOUT_GENERAL` once at the beginning after the region upload (instead of SHADER_READ_ONLY), less PITA and less code. Then SHADER_READ_ONLY at the end of mipmapping.

Furthermore it will already be as its needed to be for the compute shader mipmapping.

src/nbl/video/alloc/SimpleGPUBufferAllocator.cpp
```diff
@@ -16,5 +16,15 @@ SimpleGPUBufferAllocator::value_type SimpleGPUBufferAllocator::allocate(size_t b
...
-    return mDriver->createGPUBufferOnDedMem(reqs,canUpdateViaCmdBuff);
+
+    // Todo(achal): Need to get these values from the user somehow
+    // This doesn't matter right now because updateBufferRangeViaStagingBuffer 
+    // only makes use of upstreaming buffer which would be TRANSFER_SRC
+    IGPUBuffer::SCreationParams creationParams = {};
+    creationParams.usage = IGPUBuffer::EUF_TRANSFER_SRC_BIT;
+    creationParams.sharingMode = asset::ESM_EXCLUSIVE;
+    creationParams.queueFamilyIndexCount = 0u;
+    creationParams.queueFamilyIndices = nullptr;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: @achalpandeyy

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
+				assert(false && "Fallback: requested 'format' and 'colorspace' is not supported.");
+				surfaceFormat = gpuInfo.availableSurfaceFormats[0];
+			}
+
+			bool presentModeSupported = (gpuInfo.availablePresentModes & requestedPresentMode) != 0;
+			assert(presentModeSupported);
+			if(!presentModeSupported) // fallback 
+			{
+				requestedPresentMode = nbl::video::ISurface::E_PRESENT_MODE::EPM_FIFO;
+				assert(false && "Fallback: requested 'present mode' is not supported");
```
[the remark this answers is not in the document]
Erfan-Ahmadi: Noted.

examples_tests/common/CommonAPI.h
```diff
@@ -443,128 +681,256 @@ class CommonAPI
...
+			requestedFormat.colorSpace.eotf = asset::EOTF_sRGB;
+			requestedFormat.colorSpace.primary = asset::ECP_SRGB;
+		}
+		else
+		{
+			// Temporary to make previous examples work
+			requestedFormat.format = asset::EF_R8G8B8A8_SRGB;
+			requestedFormat.colorSpace.eotf = asset::EOTF_sRGB;
+			requestedFormat.colorSpace.primary = asset::ECP_SRGB;
+		}
```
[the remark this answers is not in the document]
Erfan-Ahmadi: I wrote this only because these were the reported from physicalDevice and surface. and they were different in the APIs

examples_tests/common/CommonAPI.h
Erfan-Ahmadi: This is the queue that was used to create initOutput.commandPool and other stuff CommonAPI uses the queue for.

examples_tests/common/CommonAPI.h
Erfan-Ahmadi: It was here from the start and I don't know why it's the way it is.
@AnastaZIuk any ideas?

examples_tests/common/CommonAPI.h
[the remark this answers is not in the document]
devshgraphicsprogramming: Vulkan reports `UNORM and `sRGB` EOTF.... a bit strange
https://vulkan.gpuinfo.org/displayreport.php?id=12537#surfaceformats

should usually report SRGB format and SRGB EOTF as well
https://vulkan.gpuinfo.org/listsurfaceformats.php

examples_tests/common/CommonAPI.h
Erfan-Ahmadi: Now that I think about it why should CommonAPI even create commandPool?

include/nbl/video/IGPUQueue.h
```diff
@@ -46,8 +46,8 @@ class IGPUQueue : public core::Interface, public core::Unmovable
         }
 
         // for renderdoc and friends
-        virtual bool startCapture() {}// = 0;
-        virtual bool endCapture() {}// = 0;
+        virtual bool startCapture() { return false; }// = 0;
+        virtual bool endCapture() { return false; }// = 0;
```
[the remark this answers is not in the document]
Crisspl: i left them to return false for now by default becasue i didnt want to impl it for vulkan

examples_tests/common/CommonAPI.h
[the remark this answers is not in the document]
Erfan-Ahmadi: It does sorry,
I changed Vulkan's to SRGB in my own branch a few days ago and I see that the change is not applied here

include/nbl/video/decl/IBackendObject.h
Crisspl: 256 is minimum it supports
but well ok

> why do we even need to remember the debug name?

because most GL resources are created in deferred fashion (pretty much everything that cant be shared between contexts, please read the code)

include/nbl/video/surface/ISurfaceGL.h
Crisspl: me
for surfaces created from native window handles
Needed for baw

include/nbl/video/utilities/renderdoc_app.h
Crisspl: hm? i've found it in /src and just moved to /include

src/nbl/video/IOpenGL_LogicalDevice.h
Crisspl: yeah, thats right

src/nbl/video/IOpenGL_FunctionTable.h
Crisspl: ive commented and will push it today anyway, but i dont think so.

src/nbl/video/SOpenGLContextLocalCache.cpp
Crisspl: becasue thats where GL object is created

src/nbl/video/SOpenGLContextLocalCache.cpp
Crisspl: becasue thats where GL object is created

include/nbl/asset/IImage.h
achalpandeyy: This comment https://github.com/Devsh-Graphics-Programming/Nabla/pull/160/files/6634885ebe66f4e9ba1657631f0902e0773ff7ba#r691105872

examples_tests/common/CommonAPI.h
achalpandeyy: Probably @Erfan-Ahmadi needs to see this

Erfan-Ahmadi: Because of `vkCmdBlitImage`
and EUF_STORAGE_BIT doesn't seem to be well supported so I removed it from my branch, I will also add `TRANSFER_SRC`

examples_tests/common/CommonAPI.h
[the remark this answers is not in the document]
devshgraphicsprogramming: ok gimme the commit ID, maybe we can cherrypick.

examples_tests/common/CommonAPI.h
devshgraphicsprogramming: ok we'll address it with your upcoming changes where you'll be creating up to 4 commandPools for the 4 queues (and always 4 cmdbuffers)

then user can just get the cmdbuffer for the Graphics/Compute queue as necessary

include/nbl/asset/IBuffer.h
devshgraphicsprogramming: assign @achalpandeyy , this needs to move IGPUBuffer

in ICPUBuffer and ICPUImage you only need to track `core::bitflag<E_USAGE_FLAGS> m_usage;`

include/nbl/asset/IImage.h
devshgraphicsprogramming: ok change it.

include/nbl/asset/IImage.h
devshgraphicsprogramming: > 
> 
> i thought you guys established something and this is new way of specifying queue fam indices

yeah that's fine, I have no clue what this is doing in the base of ICPUImage

same thing as with ICPUBuffer 

assign @achalpandeyy , this needs to move IGPUImage

in ICPUBuffer and ICPUImage you only need to track core::bitflag<E_USAGE_FLAGS> m_usage;

include/nbl/ext/ScreenShot/ScreenShot.h
devshgraphicsprogramming: you're passing garbage (default) buffer creation parameters?

no use, nothing

include/nbl/video/IGPUQueue.h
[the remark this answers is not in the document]
devshgraphicsprogramming: ok this TODO passes to @Erfan-Ahmadi

include/nbl/video/decl/IBackendObject.h
devshgraphicsprogramming: > please read the code

yeah I read 1 hour later into the review and I was appalled by debug names being set on cached objects every time their state was flushed.

include/nbl/video/surface/ISurfaceGL.h
devshgraphicsprogramming: aah ok, can it break because of threading / out of sync stuff?

src/nbl/video/alloc/SimpleGPUBufferAllocator.cpp
devshgraphicsprogramming: ok lets do it

include/nbl/video/utilities/renderdoc_app.h
AnastaZIuk: it is in 3rdparty, but not sure if in the following PR

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
Crisspl: should be mem barrier on buffer, gonna fix this

examples_tests/common/CommonAPI.h
devshgraphicsprogramming: probably because someone was banking on returning a `std::array<>` of swapchain images, but its moot because as you've correctly notices the actual number of swapchain images returned might be less and determined at runtime

So need to have this statically seared into CommonAPI

include/nbl/video/utilities/renderdoc_app.h
Crisspl: ok yeah, found it, so it was doubled

src/nbl/video/SOpenGLContextLocalCache.cpp
Crisspl: decided to set names on GL objects that are shared between nabla objects

examples_tests/common/CommonAPI.h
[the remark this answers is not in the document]
Erfan-Ahmadi: https://github.com/Devsh-Graphics-Programming/Nabla/pull/219/commits/b994374c14f13dee20449f1e2994d6d2196d625f

include/nbl/asset/IImage.h
Crisspl: i think matt means that SCreationParams should be in IGPUBuffer, not in IBuffer (common base) since queue indices have nothing to do with asset namespace. Same goes for IImage

src/nbl/video/COpenGL_Queue.h
Crisspl: fixed

include/nbl/asset/IImage.h
Erfan-Ahmadi: Yea moving it to IGPUBuffer will work and I'll take care of the usage flags part.

devshgraphicsprogramming: ok but then also take care of the IImage problem

include/nbl/video/utilities/renderdoc_app.h
Crisspl: fixed

examples_tests/common/CommonAPI.h
Erfan-Ahmadi: I think Vulkan's mindset is more like this: if you want to render to the swapchain, you must either have that as COLOR_ATTACHMENT or write to another image (in compute or whatever) with supported usage flags and VkCmdBlit to the swapchain image.
Example 42. runs fine without it.

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -449,9 +449,14 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUBuffer** const _begin
         if (reqs.vulkanReqs.size==0u)
             return;
 
-        IGPUBuffer::SCreationParams unused = {};
+        IGPUBuffer::SCreationParams bufparams;
+        bufparams.sharingMode = _params.sharingMode;
+        //bufparams.usage = //this has to be sourced from somewhere..
```
devshgraphicsprogramming, line 454: ICPUBuffer shall track

Erfan-Ahmadi: Yep, already done. On #219

examples_tests/06.MeshLoaders/oit.glsl
```diff
@@ -0,0 +1,202 @@
+#ifndef _OIT_GLSL_INCLUDED_
```
devshgraphicsprogramming, line 1: can you make this into example 16 instead of overwriting an existing one?

include/nbl/video/surface/ISurfaceGL.h
devshgraphicsprogramming: whenever you call `GetWindowRect` you're calling it from whatever thread calls `ISurfaceGL::getWidth` etc. while all the window stuff gets done on the window manager thread?

include/nbl/video/surface/ISurfaceGL.h
```diff
@@ -65,10 +65,18 @@ class CSurfaceGLImpl final : public Base<Window>
...
         {
-            return false;
+            // Todo(achal): Fill it properly
+            capabilities.minImageCount = 2u;
+            capabilities.maxImageCount = 8u;
+            capabilities.currentExtent = { this->getWidth(), this->getHeight() };
+            capabilities.minImageExtent = { 1u, 1u };
+            capabilities.maxImageExtent = { this->getWidth(), this->getHeight() };
+            capabilities.maxImageArrayLayers = 1u;
+            capabilities.supportedUsageFlags = static_cast<asset::IImage::E_USAGE_FLAGS>(0u);
```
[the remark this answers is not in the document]
devshgraphicsprogramming: could pretty much have them all

examples_tests/common/CommonAPI.h
```diff
@@ -360,53 +567,84 @@ class CommonAPI
 		result.logger = core::make_smart_refctd_ptr<system::CColoredStdoutLoggerWin32>(); // we should let user choose it?
 		result.inputSystem = core::make_smart_refctd_ptr<InputSystem>(system::logger_opt_smart_ptr(result.logger));
 
-		assert(api_type == video::EAT_OPENGL); // TODO: more choice OR EVEN RANDOM CHOICE!
-
-		auto _apiConnection = video::COpenGLConnection::create(nbl::core::smart_refctd_ptr(result.system), 0, app_name.data(), video::COpenGLDebugCallback(core::smart_refctd_ptr(result.logger)));
-		result.apiConnection = _apiConnection;
+		if(api_type == EAT_VULKAN) 
+		{
+			result.apiConnection = video::CVulkanConnection::create(core::smart_refctd_ptr(result.system), 0, app_name.data(), true);
```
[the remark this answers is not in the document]
achalpandeyy: resolved

examples_tests/common/CommonAPI.h
achalpandeyy: Can I make it templated it on `MAX_SWAPCHAIN_IMAGE_COUNT` instead?
@devshgraphicsprogramming @Erfan-Ahmadi

devshgraphicsprogramming: no need for any templates really

examples_tests/02.ComputeShader/main.cpp
achalpandeyy: Doesn't look like it

src/nbl/video/CVulkanLogicalDevice.h
achalpandeyy: Can I remove the TODO?

examples_tests/common/CommonAPI.h
Erfan-Ahmadi: I will write code that tries to achieve this. But note that there can be a case where there is only 1 queue family and maxQueueCount = 1 which supports all (I even have the device, It's intel UHD something)
I'm on it.

examples_tests/common/CommonAPI.h
```diff
@@ -293,6 +293,211 @@ class CommonAPI
...
+				}
+				
+				// If a unique Transfer is not found then use either Graphics or Compute for Transfer (prefer compute)
+				if(extractedInfo.queueFamilyProps.transfer.index == QueueFamilyProps::InvalidIndex)
+				{
+					if(extractedInfo.queueFamilyProps.compute.supportsTransfer)
+						extractedInfo.queueFamilyProps.transfer = extractedInfo.queueFamilyProps.compute;
+					else if(extractedInfo.queueFamilyProps.graphics.supportsTransfer)
+						extractedInfo.queueFamilyProps.transfer = extractedInfo.queueFamilyProps.graphics;
+				}
```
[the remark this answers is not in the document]
Erfan-Ahmadi: > actually the way it should work, is you try to find a queue family that supports transfer ONLY (no compute or graphics)
I do that at first, before going into this function;
> if that doesnt exist, then find any queue family with max queue count >=3 that doesnt do graphics or compute but does transfer.
I know what you mean and want, I'll try to achive this but there can be a case where there is only 1 queue family and maxQueueCount = 1. (I'll handle this case as well)

I'm on it.

examples_tests/common/CommonAPI.h
Erfan-Ahmadi: > its good to have graphics and compute be separate queues but probably not separate families

I will have compute and graphics be different queues (if possible and maxQueues allow)
I'm not sure about them being different queue families or not and I'm still researching this, but I assumed them being different families would have benefits and have some more concurrency or whatever xD

vulkan spec:
> If an implementation exposes any queue family that supports graphics operations, at least one queue family of at least one physical device exposed by the implementation must support both graphics and compute operations.

devshgraphicsprogramming: > I'm not sure about them being different queue families or not and I'm still researching this, but I assumed them being different families would have benefits and have some more concurrency or whatever xD

I;ve never heard of having different families doing performance benefits.

devshgraphicsprogramming:
@achalpandeyy & @Erfan-Ahmadi let me know the next time @Crisspl can merge your work into `criss_baw`

PR #236, opened 2021-10-07
examples and media storage APKs, appropriate permissions for accessing resources

AnastaZIuk:


AnastaZIuk:
generates media APK properly, need to add write/read permissions and the question is how we want to access data from it querying from example's APK

PR #238, opened 2021-10-16
UTF8 and System utilities

AnastaZIuk:


include/nbl/core/string/stringutil.h
```diff
@@ -11,6 +11,7 @@
 #include <cctype>
 #include "stddef.h"
 #include "string.h"
+#include "utf8.h"
```
devshgraphicsprogramming, line 14: remove

devshgraphicsprogramming: after you remove, you can merge to your branch with the latest stuff (danylo and LoD system)

PR #262, opened 2021-11-11
Criss baw upstream

devshgraphicsprogramming:


src/nbl/video/COpenGLCommandBuffer.h
```diff
@@ -1043,18 +1043,45 @@ class COpenGLCommandBuffer final : public IGPUCommandBuffer
...
+                if (first == last)
+                {
+                    first = i;
+                    last = first + 1;
+                }
+                else
+                    ++last;
+
+                // Do a look ahead
+                if ((i + 1 > descriptorSetCount - 1) || !pDescriptorSets[i + 1])
```
devshgraphicsprogramming, line 1066: you need to check `descriptorSetCount!=0u` first above, and if it is, return `true` immediately

or assert that its not

or change the check to `((i+1)==descriptorSetCount)`

src/nbl/video/COpenGLCommandBuffer.h
```diff
@@ -1043,18 +1043,45 @@ class COpenGLCommandBuffer final : public IGPUCommandBuffer
...
+                {
+                    SCmd<impl::ECT_BIND_DESCRIPTOR_SETS> cmd;
+                    cmd.pipelineBindPoint = pipelineBindPoint;
+                    cmd.layout = core::smart_refctd_ptr<const pipeline_layout_t>(layout);
+                    cmd.firstSet = firstSet + first;
+                    cmd.dsCount = last - first;
+                    for (uint32_t b = 0u; b < cmd.dsCount; ++b)
+                        cmd.descriptorSets[b] = core::smart_refctd_ptr<const IGPUDescriptorSet>(pDescriptorSets[first + b]);
+                    cmd.dynamicOffsets = dynamicOffsets;
```
devshgraphicsprogramming, line 1075: if you have to split into two commands/requests, wont that mess up the `dynamicOffsets` assigned to a descriptor set?

Issue #305, opened 2022-01-26
NewAPI Fixes and Changes


Erfan-Ahmadi:
Here are the suggested changes to make to the NewAPI:

- [ ] Move CommandBuffer Allocation/Free from LogicalDevice to CommandPool

These changes affect most of the examples and will touch many places, they should come after most of the branches are synced together and trivial issues is fixed.

devshgraphicsprogramming:
> Here are the suggested changes to make to the NewAPI:
> 
> * [ ]  Move CommandBuffer Allocation/Free from LogicalDevice to CommandPool
> 
> These changes affect most of the examples and will touch many places, they should come after most of the branches are synced together and trivial issues is fixed.

seems that this one can be done by @achalpandeyy

PR #306, opened 2022-01-26
Erfan Fixes and New Buffer API

Erfan-Ahmadi:
Fixing PR issues [here](https://github.com/Devsh-Graphics-Programming/Nabla/pull/160) assigned to @Erfan-Ahmadi 

## Fixes

- [x] create 2 issues regarding CW/CCW and SwapChain Orientation
- [x] return false for IGPUCommandBuffer
- [x] IGPUCommandBuffer change param to r-value ref smart pointer
- [x] isCompatibleDevicewise, wasCreatedBy (templated casting function IBackendObject::cast)
  - [x] use it in every unprotected static_cast place
- [x] CSurfaceVulkan.h functions to `inline`
- [x] CSurfaceVulkan protected constructor
- [x] resetFences return bool
- [x] createGPUDescriptorSetLayout_impl -> convert some stack variables to vector
- [x] Vulkan backend should not be holding onto the spirv buffer after shader module creation, useless memory pressure. (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791847489)
- [x] UNDO -> CVulkanShader still refcounts the SPIRV module source (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791859399)
- [x] check that the OpenGL pipeline cache works, and doesn't collide on shaders that have the same shader stage
	- (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791707386)
- [x] investigate: SRGB swapchain with alpha make the window transparent in OpenGL?	
	- (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r728185231)
- [x] SPIRV version fix 
	- (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791899381)
- [x] CGeomCreator newUsageFlags should be added automagically when attached as Index/Vertex Buffer
	- (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791913657)
- [x] Expose MaxAnisotropyLog2 (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791796518)
- [x] assert alignment [1](https://github.com/Devsh-Graphics-Programming/Nabla/blob/erfan_fixes/src/nbl/video/COpenGLCommandBuffer.cpp#L147) [2](https://github.com/Devsh-Graphics-Programming/Nabla/blob/erfan_fixes/src/nbl/video/COpenGLCommandBuffer.cpp#L248)
- [x] QueryPool Reset for OpenGL [QueryPoolReset](https://github.com/Devsh-Graphics-Programming/Nabla/blob/erfan_fixes/src/nbl/video/COpenGLCommandBuffer.cpp#L910)
- [ ] Test and Fix QueryPools and different Queries for OpenGL/Vulkan (previously couldn't because of incomplete vulkan impl and examples)
  -  Should be consistent across the APIs for TIMESTAMP, OCCLUSION, ~~TRANSFORM_FEEDBACK_STREAM~~
- [x] Improve and Test texture download on GLES [link to code](https://github.com/Devsh-Graphics-Programming/Nabla/blob/erfan_fixes/src/nbl/video/COpenGLCommandBuffer.cpp#L288)
- [x] Remove Comment [line](https://github.com/Devsh-Graphics-Programming/Nabla/blob/92a06156ee45006f4836cf4e3a34f5f6299f2b28/src/nbl/video/COpenGLCommandBuffer.cpp#L829)
- [ ] Screenshot Changes
  - [ ] wait semaphore (https://github.com/Devsh-Graphics-Programming/Nabla/pull/219#discussion_r800543586)
  - [ ] add convert filter
  - [ ] add flip image option
  - [ ] dst access mask (https://github.com/Devsh-Graphics-Programming/Nabla/pull/219#discussion_r800546866)
## Features and Extension API

- [ ] LogicalDevice function to query Enabled Features/Extension 
- [ ] move `getExtraGLSLDefines()` to ILogicalDevice and Implement for Vulkan

## Delayed for later:

- [ ] getSizedOpenGLFormatFromOurFormat -> cache formats, and implement format feature functions for GL
 - will happen on the next PR when I'm working on format promotion
- [ ] freeCommandBuffers_impl (see [this](https://github.com/Devsh-Graphics-Programming/Nabla/issues/305))

## Changing the Buffer API:

- [ ] should only be called iff the vkBindBufferMemory call succeeds 
		- (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791849743)
		- (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791860130)
- [ ] NUKE `core::smart_refctd_ptr<IDriverMemoryAllocation> m_backedMemory;` 
 - The smartpointer to the IDriverMemoryAllocation needs to be pushed down into CVulkanImage and CVulkanBuffer
 - (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791769190)	
- [ ] Remove "GPU" from ILogicalDevice function names
- [ ] return m_cachedCreationParams.declaredSize in IGPUBuffer::getSize()
- [ ] E_DRIVER_VENDOR enum in `IPhysicalDevice` and implement for all backends
- [ ] Spoof OpenGL Heaps and MemoryTypes
```
Heap 0 : Device local, its actually possible to query the VRAM size using some weird AMD/NVidia/Intel undocumented GL extensions
Heap 1 : Not Device Local, set to 70% of total (not free or available) system RAM or RAM - 2GB, whichever is higher (can be queried from ISystem)
Heap 2 : 256MB Device Local
```
```
<PRop,HeapIx>

{VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT, 0} // required by spec

{VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT | VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT, 2}
{VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT | VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT, 2}

{ZERO_VK_MEMORY_PROPERTY_FLAG, 1}
{VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT, 1} // required by spec
{VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT, 1}
```

- [ ] Implement getXXXMemoryTypeMask and remove getXXXMemoryReqs
- [ ] Change the implementations of COpenGLBuffer and COpenGLImage to move the creation of the GL buffer/image to bindBufferMemory and bindImageMemory respectively + `isInitialized` boolean in IDriverMemoryBacked
- [ ] MemoryType Iterator
- [ ] AllocateInfo struct and allocateMemory function
- [ ] cleanup `createGPUBuffer` -> remove size param and rename to `createBuffer`
- [ ] SDriverMemoryRequirements Cleanup
- [ ] Remove all the createXXXOnDedMem methods 
- [ ] rename createGPUImage to createImage
- [ ]  replace `IDriverMemoryAllocation::getType()` with `uint8_t IDriverMemoryAllocation::getMemoryTypeIndex() const`
- [ ] NUKE E_MAPPING_CAPABILITY_FLAGS and fold values into `E_MEMORY_ALLOCATE_FLAGS`
- [ ] Querying VRAM size on OpenGL
### [VkImageViewUsageCreateInfo](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageViewUsageCreateInfoKHR.html)

Basically it means that you can override the `VkImageUsageFlags` you passed to VkImage when creating it when creating the ImageView

> When this structure is chained to VkImageViewCreateInfo the usage field overrides the implicit usage parameter inherited from image creation time and its value is used instead for the purposes of determining the valid usage conditions of VkImageViewCreateInfo.

```c
typedef struct VkImageViewUsageCreateInfo {
    [sType+pNext]
    VkImageUsageFlags    usage;
} VkImageViewUsageCreateInfo;
```

### [VkExternalFormatANDROID](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkExternalFormatANDROID.html)

There is a concept of `hardware buffers` for Android:
> Android hardware buffers may represent images using implementation-specific formats, layouts, color models, etc., which do not have Vulkan equivalents. Such external formats are commonly used by external image sources such as video decoders or cameras. Vulkan can import Android hardware buffers that have external formats, but since the image contents are in an undiscoverable and possibly proprietary representation, images with external formats must only be used as sampled images

External Format must retrieved from a "HardwareBuffer" in [VkAndroidHardwareBufferFormatPropertiesANDROID](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkAndroidHardwareBufferFormatPropertiesANDROID.html) via  `vkGetAndroidHardwareBufferPropertiesANDROID`

To learn more about AndroidHardwareBuffer I found these links
[developer.android](https://developer.android.com/reference/android/hardware/HardwareBuffer#:~:text=android.hardware.HardwareBuffer,buffers%20across%20different%20application%20processes.)
[ndk reference](https://developer.android.com/ndk/reference/group/a-hardware-buffer)

### [VkImageFormatListCreateInfo](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageFormatListCreateInfoKHR.html)

Specify that an image can: be used with a particular set of formats

If the pNext chain of VkImageCreateInfo includes a VkImageFormatListCreateInfo structure, then that structure contains a list of all formats that can be used when creating views of this image.
```c
typedef struct VkImageFormatListCreateInfo {
    [sType+pNext]
    uint32_t           viewFormatCount;
    const VkFormat*    pViewFormats;
} VkImageFormatListCreateInfo;
```

### [VkImageStencilUsageCreateInfo](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageStencilUsageCreateInfoEXT.html)

This extension(Promoted to Vulkan 1.2) allows specifying separate usage flags for the stencil aspect of images with a depth-stencil format at image creation time.

Goes in the pNext chain of VkImageCreateInfo 

```c
// Provided by VK_VERSION_1_2
typedef struct VkImageStencilUsageCreateInfo {
    [sType+pNext]
    VkImageUsageFlags    stencilUsage;
} VkImageStencilUsageCreateInfo;
```

### [VkImageSwapchainCreateInfoKHR](https://www.google.com/search?q=VkImageSwapchainCreateInfoKHR&sourceid=chrome&ie=UTF-8)
This is basically a swapChain handle that goes into the pNext chain of `VkImageCreateInfo`:

> If the pNext chain of VkImageCreateInfo includes a VkImageSwapchainCreateInfoKHR structure, then that structure includes a swapchain handle indicating that the image will be bound to memory from that swapchain.

```c
// Provided by VK_VERSION_1_1 with VK_KHR_swapchain, VK_KHR_device_group with VK_KHR_swapchain
typedef struct VkImageSwapchainCreateInfoKHR {
    [sType+pNext]
    VkSwapchainKHR     swapchain;
} VkImageSwapchainCreateInfoKHR;
```
> If swapchain is not VK_NULL_HANDLE, the fields of VkImageCreateInfo must match the implied image creation parameters of the swapchain

You may be asking which index of the swapchain images will the image be bound to?!

And I will refer you to [VkBindImageMemorySwapchainInfoKHR](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkBindImageMemorySwapchainInfoKHR.html) which goes into the pNextChain of the famous `VkBindImageMemoryInfo`.

> If the pNext chain of VkBindImageMemoryInfo includes a VkBindImageMemorySwapchainInfoKHR structure, then that structure includes a swapchain handle and image index indicating that the image will be bound to memory from that swapchain.
```c
// Provided by VK_VERSION_1_1 with VK_KHR_swapchain, VK_KHR_device_group with VK_KHR_swapchain
typedef struct VkBindImageMemorySwapchainInfoKHR {
    [sType+pNext]
    VkSwapchainKHR     swapchain;
    uint32_t           imageIndex;
} VkBindImageMemorySwapchainInfoKHR;
```

### [VkDedicatedAllocationImageCreateInfoNV](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkDedicatedAllocationImageCreateInfoNV.html)

Basically it's a boolean telling the driver if it **WILL** have a dedicated allocation bound to it.

> Using a dedicated allocation for color and depth/stencil attachments or other large images may improve performance on some devices.

It can be pNext of `VkImageCreateInfo` or `VkImageStencilUsageCreateInfo`

```c
// Provided by VK_NV_dedicated_allocation
typedef struct VkDedicatedAllocationImageCreateInfoNV {
    [sType+pNext]
    VkBool32           dedicatedAllocation;
} VkDedicatedAllocationImageCreateInfoNV;
```


src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -149,12 +149,13 @@ class CVulkanLogicalDevice final : public ILogicalDevice
         for (uint32_t i = 0u; i < _count; ++i)
         {
             if (_fences[i]->getAPIType() != EAT_VULKAN)
-                return;
+                return false;
```
devshgraphicsprogramming, line 152: this is a major error, you should also assert

devshgraphicsprogramming:
# VkImageViewUsageCreateInfo

Core in 1.1 good idea to use and add to frontend api (another member to ImageView creation parameters), however note that it **doesn't override the pparent image usage flags** it **narrows them down to a subset**.
So Nabla would need validation (for the sake of GL) that the imageview usages are indeed subsets.

devshgraphicsprogramming:
# VkExternalFormatANDROID

Above our current paygrade, won't do.

devshgraphicsprogramming:
# VkImageFormatListCreateInfo

Way too verbose and PITA. Plus I need to see the numbers on the performance claims before even considering such a ball&chain for out frontend's API.

devshgraphicsprogramming:
# VkDedicatedAllocationImageCreateInfoNV

Doesn't `VK_KHR_dedicated_allocation` supersede the NV extension?

devshgraphicsprogramming:
Did you fix up and make sure example 7 works as expected ? (basically that the secondary commandbuffer doesn't make the Validation Layers go ape£$"% because of missing inheritance info needed to continue a renderpass)

devshgraphicsprogramming:
Also there's no reason to clone the SPIR-V buffer (cloning with param 1 would copy the entire SPV buffer for no reason)
https://github.com/Devsh-Graphics-Programming/Nabla/blob/f1086b0257059556085cddceddc0809befab6aa9/src/nbl/video/CVulkanLogicalDevice.h#L583

include/nbl/core/execution.h
```diff
@@ -0,0 +1,64 @@
...
+#else
+#include <oneapi/dpl/algorithm>
+#include <oneapi/dpl/execution>
+#include <oneapi/dpl/iterator>
+#include "oneapi/dpl/pstl/execution_defs.h"
+#include "oneapi/dpl/pstl/glue_algorithm_defs.h"
+#include "oneapi/dpl/pstl/glue_algorithm_ranges_defs.h"
+#endif
+
+#define ALIAS_TEMPLATE_FUNCTION(highLevelF, lowLevelF) \
```
devshgraphicsprogramming, line 20: still missing the UNDEF for this

Erfan-Ahmadi:
Oh yea, I forgot about that xD
Will add

devshgraphicsprogramming:
> Also there's no reason to clone the SPIR-V buffer (cloning with param 1 would copy the entire SPV buffer for no reason)
> 
> https://github.com/Devsh-Graphics-Programming/Nabla/blob/f1086b0257059556085cddceddc0809befab6aa9/src/nbl/video/CVulkanLogicalDevice.h#L583
> 
> Old comment
> 
> > Also this clone is pointless if the real shader contents are GLSL, some shaders are a few megs of GLSL, useless memory pressure.

Ok I forgot we need to keep the SPIR-V around if we want to use the SPIR-V optimizer later.

https://github.com/Devsh-Graphics-Programming/Nabla/pull/219#discussion_r728618735

Erfan-Ahmadi:
I removed it, do I need to bring it back?

3rdparty/pstl/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang") # only non-multi-generators and clang supported
+	get_filename_component(CCBD_ABSOLUTE "${CMAKE_CURRENT_BINARY_DIR}" ABSOLUTE)
+
+	set(TBB_INSTALL_VARS ON CACHE BOOL "" FORCE)
+	set(TBB_TEST OFF CACHE BOOL "" FORCE)
+	set(TBB_STRICT ON CACHE BOOL "" FORCE)
+	set(PREVIOUS_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
+	set(BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
```
devshgraphicsprogramming, line 8: why are we building shared libs of pstl/TBB ?

devshgraphicsprogramming:
> 

bring back the clone of the SPIR-V and the VulkanShader keeping the SPIR-V around

Document in the source code with comments why its like this.

cmake/toolchains/android/build.cmake
```diff
@@ -0,0 +1,203 @@
...
+		message(STATUS "Using Windows as a host OS for Android cross-compiling!")
+	else()
+		message(FATAL_ERROR "Current host OS for Android cross-compiling isn't allowed!")
+	endif()
+	
+	# TODO - remove those bellow in future
+	set(_NBL_COMPILE_WITH_OPEN_EXR_ OFF CACHE BOOL "Off for Android purposes at the moment" FORCE) 		# TODO - check it out
+	set(_NBL_COMPILE_WITH_BAW_LOADER_ OFF CACHE BOOL "Off for Android purposes at the moment" FORCE)	# TODO - remove when BAW-v4 is done
+	set(_NBL_COMPILE_WITH_BAW_WRITER_ OFF CACHE BOOL "Off for Android purposes at the moment" FORCE)	# TODO - remove when BAW-v4 is done
+	set(_NBL_COMPILE_WITH_OBJ_LOADER_ ON CACHE BOOL "Off for Android purposes at the moment" FORCE)
```
devshgraphicsprogramming, line 95: can you remove the OBJ line? its ON anyway

PR #307, opened 2022-01-26
CI Test

AnastaZIuk:


devshgraphicsprogramming:
@AnastaZIuk can we nuke the branch and the PR?

AnastaZIuk:
I suggest we don't nuke it because soon I will be using the branch when creating actual proper CI job for android build and it needs solid testing and valid setup

PR #314, opened 2022-02-03
Oriented bounding box

Przemog1:


src/nbl/asset/utils/CMeshManipulator.cpp
```diff
@@ -1513,6 +1513,343 @@ core::smart_refctd_ptr<ICPUBuffer> IMeshManipulator::idxBufferFromTrianglesFanTo
 	return nullptr;
 }
 
+IMeshManipulator::OBB IMeshManipulator::calcOBB_DiTO26(ICPUMeshBuffer* mb)
```
devshgraphicsprogramming, line 1516: should really break it up into (in `IMeshManipulator)
```cpp
struct Slab
{
   float Min = FLT_MAX;
   float Max = -FLT_MAX;
};

template<uint8_t k>
struct kDOP
{
   // compute kDOP from meshbuffer
   kDOP(ICPUMeshBuffer*);

   Slab slabs[k];
};

template<uint8_t k>
inline OBB calcOBB(const kDOP<k>& kdop)
{
}
```

Not all `k`s need to be implemented, you can `static_assert` that `k` is 13.

AnastaZIuk:

devshgraphicsprogramming:
implemented and finished by @Pjbomb2

PR #332, opened 2022-02-24
File+System rework and Native Handle getters for SDK Native API Interop

devshgraphicsprogramming:


devshgraphicsprogramming:
Example 5 needs to be rid of Vulkan validation errors.

Examples 6,7 need testing after TAR/ZIP refactors.

Example 9 broken due to broken PNG writer.

Example 11 seems to deadlock.

PR #339, opened 2022-03-17
Point cloud compute rasterizer (Recruitment task)

deprilula28:
## Description
Will add an example with a rasterizer for point clouds done entirely on compute shaders.

## TODO list:
- The positions buffer currently reports as "No resource" in RenderDoc
- Instead of just the depth buffer, a 64-bit visibility buffer should be generated
- The visibility buffer should be shaded according to some information about the points in a later pass
- Everything needs to be tested


AnastaZIuk:

examples_tests/CMakeLists.txt
```diff
@@ -72,4 +72,5 @@ if (NOT NBL_BUILD_ANDROID)
 	add_subdirectory(58.MediaUnpackingOnAndroid EXCLUDE_FROM_ALL)
 endif()
 add_subdirectory(60.ClusteredRendering EXCLUDE_FROM_ALL)
+add_subdirectory(ComputePointCloud EXCLUDE_FROM_ALL)
```
devshgraphicsprogramming, line 75: give it a number, see the readme about which numbers are "free"

examples_tests/ComputePointCloud/rasterizer.comp
```diff
@@ -0,0 +1,63 @@
...
+layout (push_constant) uniform pushConstants
+{
+	uvec2 imgSize;
+	uint pointCount;
+	mat4 mvp;
+} u_pushConstants;
+
+layout (set = 0, binding = 0, r32ui) uniform coherent uimage2D outImage;
+layout (set = 0, binding = 1, std430) readonly buffer pointCloud {
+	vec3 position[];
```
devshgraphicsprogramming, line 20: `vec3` has a 16byte alignment, either:
- pad
- change format to PoT bitcount
- make a `struct{float x,y,z;};` and ue that instead of `vec3`

examples_tests/ComputePointCloud/rasterizer.comp
```diff
@@ -0,0 +1,63 @@
...
+} u_pointCloud;
+
+void main()
+{
+	if (gl_GlobalInvocationID.x >= u_pushConstants.pointCount) 
+	{
+		return;
+	}
```
devshgraphicsprogramming, line 28: need to use "virtual" (persistently resident) workgroups otherwise your rasterizer wont handle more than 256*64k = 16M points on AMD GPUs

devshgraphicsprogramming: you basically want to launch a fixed number of workgroups (see `CScanner` or any other built-in compute thing) which run on a loop and comupute a `virtualInvocationIndex` instead of using `gl_GlobalInvocationID.x`

examples_tests/ComputePointCloud/rasterizer.comp
```diff
@@ -0,0 +1,63 @@
...
+	vec3 pointPosition = u_pointCloud.position[gl_GlobalInvocationID.x];
+	vec4 transformed = u_pushConstants.mvp * vec4(pointPosition, 1.0);
+
+	// Behind the camera
+	if (transformed.z < 0.0)
```
devshgraphicsprogramming, line 33: you might want to bitwise XOR the most significant bit of `transformed.w` with the `transformed.z` (because a negative sign in W flips post divide NDC z)

examples_tests/ComputePointCloud/rasterizer.comp
```diff
@@ -0,0 +1,63 @@
...
+	vec3 ndc = transformed.xyz / vec3(transformed.w);
+	vec2 uv = (ndc.xy + 1.0) * 0.5;
+	const ivec2 size = ivec2(5, 5);
+	const vec2 halfTexelSize = (vec2(size) * 0.5) / vec2(u_pushConstants.imgSize);
+	// AABB for the point: uv - halfTexelSize to uv + halfTexelSize
+
+	ivec2 offset = ivec2((uv - halfTexelSize) * vec2(u_pushConstants.imgSize));
+	ivec2 curPixel = ivec2(0);
+	// Flattened iteration over the AABB
+	// TODO make the points circles
```
devshgraphicsprogramming, line 47: dont bother, leave them as points.

I'd appreciate the culling against the screen bounds to be hoisted up though.

examples_tests/ComputePointCloud/rasterizer.comp
```diff
@@ -0,0 +1,63 @@
...
+	
+	// If the Y has gone outside the AABB, it has no more pixels
+	while (curPixel.y > size.y) {
+		ivec2 coords = offset + curPixel;
+		if (all(greaterThanEqual(coords, ivec2(0))) && all(lessThan(coords, ivec2(u_pushConstants.imgSize))))
+		{
+			// Write out depth
+			imageAtomicMax(outImage, coords, floatBitsToUint(ndc.z));
```
devshgraphicsprogramming, line 55: when you have access to r64ui shove the depth in the MSB and pointID in LSB

when you don't, implement a spinlock (like the MLAB OIT sample's fallback path) OR a z-prepass (first render depth, then re-render with `imageAtomicCompSwap` to swap depth for pointID).

examples_tests/ComputePointCloud/main.cpp
```diff
@@ -0,0 +1,516 @@
...
+			}
+			m_visbuffer = logicalDevice->createDeviceLocalGPUImageOnDedMem(std::move(imageParams));
+
+			video::IGPUImageView::SCreationParams imgViewInfo;
+			{
+				imgViewInfo.flags = static_cast<video::IGPUImageView::E_CREATE_FLAGS>(0u);
+				imgViewInfo.image = m_visbuffer;
+				imgViewInfo.viewType = video::IGPUImageView::ET_2D;
+				imgViewInfo.format = asset::E_FORMAT::EF_R32_UINT;
+				imgViewInfo.subresourceRange.aspectMask = static_cast<asset::IImage::E_ASPECT_FLAGS>(0u);
```
devshgraphicsprogramming, line 283: you need a COLOR aspect, otherwise Vulkan validation error

examples_tests/ComputePointCloud/main.cpp
```diff
@@ -0,0 +1,516 @@
...
+			//for (uint32_t i = 0; i < posBuffer->getSize() / 12; i++) {
+			//	logger->log("Pos %f %f %f", system::ILogger::E_LOG_LEVEL::ELL_INFO, positions[i * 3], positions[i * 3 + 1], positions[i * 3 + 2]);
+			//}
+
+			m_pointCount = posBuffer->getSize() / 12;
+			auto posBuffer_cpu = posBuffer.get();
+			auto positionsVertexBuffer = cpu2gpu.getGPUObjectsFromAssets(&posBuffer_cpu, &posBuffer_cpu + 1, cpu2gpuParams)->front();
+			positionVbOffset = positionsVertexBuffer->getOffset();
+			positionVb = core::smart_refctd_ptr<video::IGPUBuffer>(positionsVertexBuffer->getBuffer());
+		}
```
devshgraphicsprogramming, line 332: MeshManipulator is more robust/resilient to de-interleaving different formats

devshgraphicsprogramming: Also you have no guarantee that vertex buffer offset is okay to use with SSBO binding alignment, so I'd always repack

examples_tests/ComputePointCloud/main.cpp
```diff
@@ -0,0 +1,516 @@
...
+				writeDescriptorSets[0].arrayElement = 0u;
+				writeDescriptorSets[0].count = 1u;
+				writeDescriptorSets[0].descriptorType = asset::EDT_STORAGE_IMAGE;
+				writeDescriptorSets[0].info = &descriptorInfos[0];
+			}
+
+			// Point cloud vertex buffer
+			{
+				descriptorInfos[1].buffer.offset = positionVbOffset;
+				descriptorInfos[1].buffer.size = positionVb->getSize();
```
devshgraphicsprogramming, line 358: erm, I think that has to be the `size` AFTER the offset, right now you're setting it to whole buffer size (disregarding the offset)

examples_tests/ComputePointCloud/main.cpp
```diff
@@ -0,0 +1,516 @@
...
+			}
+
+			logicalDevice->updateDescriptorSets(writeDescriptorCount, writeDescriptorSets, 0u, nullptr);
+		}
+
+		{
+			asset::SPushConstantRange pcRange = {};
+			pcRange.stageFlags = asset::IShader::ESS_COMPUTE;
+			pcRange.offset = 0u;
+			pcRange.size = (2 + 1 + 16) * sizeof(uint32_t);
```
devshgraphicsprogramming, line 376: I'd either make a shared "header" between c++ and GLSL, or get this value from shader introspector

examples_tests/ComputePointCloud/main.cpp
```diff
@@ -0,0 +1,516 @@
...
+	{
+		++resourceIx;
+		if (resourceIx >= FRAMES_IN_FLIGHT)
+			resourceIx = 0;
+
+		auto& commandBuffer = commandBuffers[resourceIx];
+		auto& fence = frameComplete[resourceIx];
+
+		if (fence)
+			while (logicalDevice->waitForFences(1u, &fence.get(), false, MAX_TIMEOUT) == video::IGPUFence::ES_TIMEOUT) {}
```
devshgraphicsprogramming, line 431: you can use `blockForFences` method now

examples_tests/ComputePointCloud/main.cpp
```diff
@@ -0,0 +1,516 @@
...
+		}
+		const double averageFrameTime = frameDataFilled ? (time_sum / (double)NBL_FRAMES_TO_AVERAGE) : (time_sum / frame_count);
+
+#ifdef NBL_MORE_LOGS
+		logger->log("renderDt = %f ------ averageFrameTime = %f", system::ILogger::ELL_INFO, renderDt, averageFrameTime);
+#endif // NBL_MORE_LOGS
+
+		auto averageFrameTimeDuration = std::chrono::duration<double, std::milli>(averageFrameTime);
+		auto nextPresentationTime = renderStart + averageFrameTimeDuration;
+		auto nextPresentationTimeStamp = std::chrono::duration_cast<std::chrono::microseconds>(nextPresentationTime.time_since_epoch());
```
devshgraphicsprogramming, line 458: @Erfan-Ahmadi do we have a utility for that now?

devshgraphicsprogramming:
> The positions buffer currently reports as "No resource" in RenderDoc

Set up a logger/callback for the OpenGL or Vulkan connection, put a breakpoint in there an intercept any errors.

I think the error might be to do with bad buffer binding.

examples_tests/ComputePointCloud/main.cpp
devshgraphicsprogramming: This might be why you're getting
> The positions buffer currently reports as "No resource" in RenderDoc

OpenGL throws silent errors unlike Vulkan, if you've bound an invalid range (too big size, or unaligned offset) OpenGL will just silently fail to bind.

deprilula28:
The no resource thing was fixed when I changed to use the IGPUBuffer as a descriptor instead of the IGPUBufferView; But does one of the examples have setting a callback?

devshgraphicsprogramming:
> The no resource thing was fixed when I changed to use the IGPUBuffer as a descriptor instead of the IGPUBufferView; But does one of the examples have setting a callback?

ah yeah BufferView is UTB/TBO so `samplerBuffer` not storage buffer ;)

examples_tests/59.ComputePointCloud/rasterizer.comp
```diff
@@ -2,19 +2,19 @@
...
 
-layout (local_size_x = 256, local_size_y = 1) in;
-
 layout (push_constant) uniform pushConstants
 {
 	layout(row_major) RasterizerPushConstants c;
 } u_pushConstants;
+
+#if defined(Z_PREPASS) || defined(USE_Z_PREPASS)
+layout (set = 0, binding = 0, r32ui) uniform coherent uimage2D outImage;
```
devshgraphicsprogramming, line 13: `outImage` needs not be `coherent` with zprepass

PR #341, opened 2022-04-03
WIP - k-DOP OBB

hiradyazdan:
## Description

This is a Work In Progress PR

### List of changes
+ created obb, lbt, kdop structures and compute funcs on separate file
+ created example for kdop obb with explicit cmake instructions
+ updated example tests cmake to include obb example
+ updated example tests readme
+ added quarter PI constant template to float utils
+ updated gitignore to exclude some project build/ide files
+ updated gitmodules to use https instead of ssh since most of them
  already use https (may need to revert since CI may need ssh)

## Testing
example is made for debug draw

## TODO list:
- the algorithm needs fixing in order to draw a correct OBB


AnastaZIuk:

include/nbl/asset/IMesh.h
```diff
@@ -22,6 +23,7 @@ namespace asset
 		protected:
 			//! The cached bounding box of this mesh
 			core::aabbox3d<float> m_cachedBoundingBox;
+      core::OBB m_orientedBoundingBox;
```
devshgraphicsprogramming, line 26: don't add an OBB to a mesh

include/aabbox3d.h
```diff
@@ -15,20 +24,33 @@ namespace core
 
 //! Axis aligned bounding box in 3d dimensional space.
 /** Has some useful methods used with occlusion culling or clipping.
+ * @TODO UVector was introduced to allow for usage other than of vector3d<float> (i.e. vectorSIMDf)
+ *   and defaulted to allow for gradual migration of vector3d-based usage/implementation of aabbox3df
+ *   to vectorSIMDf (i.e. vector3df_SIMD) everywhere in the codebase
 */
-template <class T>
+template<class T, class UVector>
```
devshgraphicsprogramming, line 31: these modifications are not needed, `aabbox3d` is a legacy class I intend to get rid of in the near future

include/nbl/asset/utils/IMeshManipulator.h
```diff
@@ -451,11 +457,32 @@ class IMeshManipulator : public virtual core::IReferenceCounted
...
-			meshbuffer->setBoundingBox(calculateBoundingBox(meshbuffer,meshbuffer->getJointAABBs()));
+      if(isOBBDisabled)
+      { meshbuffer->setBoundingBox(calculateBoundingBox(meshbuffer,meshbuffer->getJointAABBs())); }
+      else
+      { meshbuffer->setBoundingBox(calculateOrientedBBox(meshbuffer)); }
 		}
 
+    static inline core::OBB calculateOrientedBBox(const ICPUMeshBuffer* meshBuffer)
+    {
+      auto vtxCount = meshBuffer->getIndexCount();
```
devshgraphicsprogramming, line 470: this is wrong, Index Count != Vertex Count

include/nbl/asset/utils/IMeshManipulator.h
```diff
@@ -451,11 +457,32 @@ class IMeshManipulator : public virtual core::IReferenceCounted
...
+
+      std::vector<core::vectorSIMDf> vtxList;
+      vtxList.reserve(vtxCount);
+
+      for(auto i = 0u; i < vtxCount; i++)
+      { vtxList.push_back(meshBuffer->getPosition(i)); }
```
devshgraphicsprogramming, line 476: this will trigger an "out of bounds" assert for most indexed meshes, go and see how the `calculateBoundingBox` function iterates through a meshbuffer's vertices

include/nbl/asset/utils/IMeshManipulator.h
hiradyazdan: I am sure it gives me exact number of vertices on a mesh, checked out the math already!

include/nbl/ext/DebugDraw/CDraw3DLine.h
```diff
@@ -10,131 +10,189 @@
...
+            static_assert(
+              std::is_base_of<core::aabbox3dsf, TBox>::value ||
+              std::is_base_of<core::aabbox3df,  TBox>::value ||
+              std::is_base_of<core::aabbox3di,  TBox>::value,
+              "TBox should derive from aabbox3d<T, UVector>"
+            );
+
+            core::vectorSIMDf verts[10];
+
+            box.getEdges(verts);
```
devshgraphicsprogramming, line 127: this is horrible method, neither the AABB or the OBB should have it inside them, it inverts the ordering of logic (object should not know or care about metadata which is derived from it)

I know its present in the AABB, and this is likely @Przemog1's code, but its quite wrong

include/nbl/ext/DebugDraw/Draw3DLineShaders.h
```diff
@@ -12,25 +12,25 @@ namespace ext
...
+    gl_Position = pc.vp * vec4(vPos, 1.0);
     Color = vCol;
 }
 )===";
 
-static const char* Draw3DLineFragmentShader = R"===(
+static constexpr const auto Draw3DLineFragmentShader = R"===(
 #version 430 core
 
 layout(location = 0) in vec4 Color;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: although that would be more of a thing to point out to @Przemog1

include/nbl/asset/utils/IMeshManipulator.h
```diff
@@ -451,11 +457,32 @@ class IMeshManipulator : public virtual core::IReferenceCounted
...
+      else
+      { meshbuffer->setBoundingBox(calculateOrientedBBox(meshbuffer)); }
 		}
 
+    static inline core::OBB calculateOrientedBBox(const ICPUMeshBuffer* meshBuffer)
+    {
+      auto vtxCount = meshBuffer->getIndexCount();
+
+      std::vector<core::vectorSIMDf> vtxList;
+      vtxList.reserve(vtxCount);
```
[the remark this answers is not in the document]
hiradyazdan: unless i calculate the number vertices of a mesh at compile time before loading which doesn't conform with the procedural mesh creation. This is the only way and that is why i had to linearly allocate the memory to save some performance hit.

include/nbl/asset/utils/IMeshManipulator.h
hiradyazdan: no it doesn't

include/nbl/asset/utils/IMeshManipulator.h
hiradyazdan: the other method calcVertexCount is actually wrong which max out at 16 whatsoever!

include/nbl/asset/IMesh.h
hiradyazdan: only seemed logical to follow the pattern, as in the example test, like the aabb which was already calculated on loading asset, could get the obb as well. Though, I designed obb to be disabled if not needed anyway.

include/aabbox3d.h
hiradyazdan: `aabbox3d` is used everywhere in the codebase. If this is a legacy which is used everywhere else, then don't you think it should be gradually upgraded than just ignoring it? its methods are possibly wrong (e.g. getEdges) but then they can simply be rectified. It still plays an important role throughout the entire system's structure.

include/nbl/asset/utils/IMeshManipulator.h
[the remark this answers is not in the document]
hiradyazdan: regardless still needs to be dynamic.

include/nbl/asset/utils/IMeshManipulator.h
hiradyazdan: `upperBoundVertexID` method seems to be doing vertex count calculation. It simply can be replaced, though weird naming for the method. Eitherway, there's no reason to rule out other things like dynamic allocation of vertices as well as looping through the indices (either by vtx count or index count) to get the position vector claiming it triggers out of bounds assert as it doesn't in any case. The only problem remains the duplicate of vertices which simply can be avoided by correct calculation, but in any case, the final calculation is not affected by any of these factors as they will be overwritten in the iterations down the line as there's a constraint to which they iterated by, which is the projection count.

include/nbl/core/shapes/OBB.h
```diff
@@ -0,0 +1,62 @@
...
+//    }
+//
+//    inline core::OBB decompress() const
+//    {
+//      core::OBB obb;
+//      reinterpret_cast<rgb32f&>(obb.bMin) = rgb18e7s3_to_rgb32f(minVx18E7S3);
+//      reinterpret_cast<rgb32f&>(obb.bMax) = rgb18e7s3_to_rgb32f(maxVx18E7S3);
+//      return obb;
+//    }
+//  };
```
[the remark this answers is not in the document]
hiradyazdan: just prepared the landing site for future!

include/nbl/ext/DebugDraw/CDraw3DLine.h
hiradyazdan: I think, this method can be re-implemented in a way that for both aabb and obb, it outputs the side/edge vectors from the matrix instead of using center point. IMHO, it is not horrible, just needs a little more care :)

PR #342, opened 2022-04-03
BUGFIX for Builtin Resources Target Failing on Windows

hiradyazdan:
## Description

Builtin resources cannot be generated automatically with default windows API, due to an OS limit, as it cannot accept a command line instruction with parameters and length longer than a certain limit. Therefore, the instruction is invalid (currently the length of the cli instruction is more than 8191 characters, hence it fails).

One solution is to transform the command into a bash command which can take long arguments. However, it still needs to be manipulated manually as the required cmake paths point to a Windows location than a mounted location on WSL/bash. 

The best solution is to write all resource address into a file and make the build to read from the file to generate the builtin resources files which this PR represents.

### References

- https://docs.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation
- https://stackoverflow.com/questions/3205027/maximum-length-of-command-line-string
- https://stackoverflow.com/questions/19354870/bash-command-line-and-input-limit

### DIFF
--- a/src/nbl/builtin/CMakeLists.txt
+++ b/src/nbl/builtin/CMakeLists.txt
@@ -6,11 +6,18 @@ function(nbl_add_builtin_resources)
   list(APPEND builtin_dependencies_list  ${CMAKE_CURRENT_SOURCE_DIR}/builtin/builtinHeaderGen.py)
   list(APPEND builtin_dependencies_list  ${CMAKE_CURRENT_SOURCE_DIR}/builtin/builtinDataGen.py)
 
+  # Just in case if not found yet (3rd parties should've found it already though regardless of platform)
+  find_package(Python3 REQUIRED)
+
+  set(RESOURCES_FILE ${CMAKE_BINARY_DIR}/include/nbl/builtin/resources.txt)
+
+  string(REPLACE ";" "," RESOURCES_ARGS "${ARGV}")
   file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/include/nbl/builtin/)
+  file(WRITE ${RESOURCES_FILE} ${RESOURCES_ARGS})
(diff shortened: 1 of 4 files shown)


AnastaZIuk:

src/nbl/builtin/CMakeLists.txt
```diff
@@ -6,11 +6,18 @@ function(nbl_add_builtin_resources)
   list(APPEND builtin_dependencies_list  ${CMAKE_CURRENT_SOURCE_DIR}/builtin/builtinHeaderGen.py)
   list(APPEND builtin_dependencies_list  ${CMAKE_CURRENT_SOURCE_DIR}/builtin/builtinDataGen.py)
 
+  # Just in case if not found yet (3d parties should've found it already though regardless of platform)
+  find_package(PythonInterp 3 REQUIRED)
```
devshgraphicsprogramming, line 10: > Deprecated since version 3.12: Use [FindPython3](https://cmake.org/cmake/help/latest/module/FindPython3.html#module:FindPython3), [FindPython2](https://cmake.org/cmake/help/latest/module/FindPython2.html#module:FindPython2) or [FindPython](https://cmake.org/cmake/help/latest/module/FindPython.html#module:FindPython) instead.

from cmake docs

devshgraphicsprogramming:
I'll merge as soon as DCO is signed

src/nbl/builtin/CMakeLists.txt
hiradyazdan: just a little wake up call! on latest MSVC (VS 2022, works fine on VS 2019 though) on this weird build, this has to be `PythonInterp 3` to find the python3_EXECUTABLE otherwise it fails. So, you might want to reconsider.

PR #340, opened 2022-03-30
Spherical rectangle sampling

sunho:
## Description
This is an implementation of spherical rectangle sampling that can potentially improve convergence of scenes with rectangular light source. It's basically an implementation of the paper below with possibly additional optimizations.

https://www.arnoldrenderer.com/research/egsr2013_spherical_rectangle.pdf

## Testing 
It was compared with triangle reference implmentation and they match the same output by my eye balling.

## TODO list:
- [x] Optimize trigonometry calls
- [x] to be added

PR #344, opened 2022-04-06
Imgui

MohammadFakhreddin:
## Description

Example of Imgui ui by Nabla API. 

## Testing 

I created an example called 61.UI.  I checked some simple imgui functionality there

## TODO list:

1 - Entire implementation is inside the example_test/ui folder in a file called UI_System.h in my own style. It needs to be integrated into Nabla as a system in the future in Nabla code style.
2 - I haven't handled the keyboard events and keyboard mapping.
3- I haven't handled resize event


AnastaZIuk:

3rdparty/CMakeLists.txt
```diff
@@ -267,6 +267,22 @@ add_library(spirv_cross OBJECT
 target_compile_definitions(spirv_cross PUBLIC SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS)
 
 
+add_library(imgui
```
devshgraphicsprogramming, line 270: this needs a `NBL_BUILD_WITH` flag like we do for all the extensions like Radeon Rays

3rdparty/imgui/imgui.cpp
```diff
@@ -0,0 +1,11142 @@
...
+// - Glossary              https://github.com/ocornut/imgui/wiki/Glossary
+// - Wiki                  https://github.com/ocornut/imgui/wiki
+// - Issues & support      https://github.com/ocornut/imgui/issues
+
+// Developed by Omar Cornut and every direct or indirect contributors to the GitHub.
+// See LICENSE.txt for copyright and licensing details (standard MIT License).
+// This library is free but needs your support to sustain development and maintenance.
+// Businesses: you can support continued development via invoiced technical support, maintenance and sponsoring contracts. Please reach out to "contact AT dearimgui.org".
+// Individuals: you can support continued development via donations. See docs/README or web page.
+
```
devshgraphicsprogramming, line 24: is there no github repository you could have attached as a submodule?

devshgraphicsprogramming:
Why did you readd some submodules, I've deleted?

P.S. why are there so many submodule changes which are irrelevant?

examples_tests/common/CommonAPI.h
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
devshgraphicsprogramming, line 1: use proper header guards

include/nbl/core/decl/smart_refctd_ptr.h
```diff
@@ -98,6 +98,11 @@ class smart_refctd_ptr
 		inline I_REFERENCE_COUNTED& operator[](size_t idx) { return ptr[idx]; }
 		inline const I_REFERENCE_COUNTED& operator[](size_t idx) const { return ptr[idx]; }
 
+		[[nodiscard]]
+		bool isValid() const noexcept
+		{
+			return ptr != nullptr;
+		}
```
devshgraphicsprogramming, line 105: we already define an `opeerator(bool)`

So
```cpp
if (!ptr)
if (bool(ptr))
```
should work.

Please remove

examples_tests/61.UI/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+include(common RESULT_VARIABLE RES)
+if(NOT RES)
+	message(FATAL_ERROR "common.cmake not found. Should be in {repo_root}/cmake directory")
+endif()
+
+set(EXTRA_SOURCES "")
+list(APPEND EXTRA_SOURCES 
+	"UI_System.cpp" 
+	"UI_System.hpp"
+)
```
devshgraphicsprogramming, line 10: you want to turn this into an extension `nbl::ext::imgui`

devshgraphicsprogramming: see other extensions to see how they're supposed to look

examples_tests/61.UI/UI_System.cpp
```diff
@@ -0,0 +1,1102 @@
...
+		0x00000001,0x00000000,0x0003001b,0x00000014,0x00000013,0x00040020,0x00000015,0x00000000,
+		0x00000014,0x0004003b,0x00000015,0x00000016,0x00000000,0x0004002b,0x0000000e,0x00000018,
+		0x00000001,0x00040020,0x00000019,0x00000001,0x0000000a,0x00050036,0x00000002,0x00000004,
+		0x00000000,0x00000003,0x000200f8,0x00000005,0x00050041,0x00000010,0x00000011,0x0000000d,
+		0x0000000f,0x0004003d,0x00000007,0x00000012,0x00000011,0x0004003d,0x00000014,0x00000017,
+		0x00000016,0x00050041,0x00000019,0x0000001a,0x0000000d,0x00000018,0x0004003d,0x0000000a,
+		0x0000001b,0x0000001a,0x00050057,0x00000007,0x0000001c,0x00000017,0x0000001b,0x00050085,
+		0x00000007,0x0000001d,0x00000012,0x0000001c,0x0003003e,0x00000009,0x0000001d,0x000100fd,
+		0x00010038
+	};
```
devshgraphicsprogramming, line 123: dear lord, you do realize you can add these GLSL files to `nbl/builtin/glsl/ext/imgui/` and use the asset manager to get them?

examples_tests/61.UI/UI_System.hpp
```diff
@@ -0,0 +1,110 @@
+#pragma once
+
+#include <nabla.h>
+
+#include <glm/vec3.hpp>
+
+#include "../common/CommonAPI.h"
```
devshgraphicsprogramming, line 7: you cannot depend on `CommonAPI` in an extension

examples_tests/61.UI/UI_System.hpp
```diff
@@ -0,0 +1,110 @@
+#pragma once
+
+#include <nabla.h>
```
devshgraphicsprogramming, line 3: see how the other extensions include nabla, i think you should only include `video` namespace

examples_tests/61.UI/UI_System.cpp
```diff
@@ -0,0 +1,1102 @@
...
+
+		// TODO: Use a signal class instead.//Signal<> UIRecordSignal{};
+		struct Subscriber {
+			int id = -1;
+			std::function<void()> listener = nullptr;
+		};
+		std::vector<Subscriber> subscribers{};
+	};
+
+	static State* state = nullptr;
```
devshgraphicsprogramming, line 148: Globals are a big no-no at DevSH Graphics Programming, please refactor

devshgraphicsprogramming (commented):
Needs to be refactored into something that doesn't use globals

examples_tests/61.UI/UI_System.cpp
MohammadFakhreddin: This is a direct copy from imgui examples.  I'm going to move shaders into a separate file and compile them like other Nabla examples.

include/nbl/core/decl/smart_refctd_ptr.h
MohammadFakhreddin: Ok

examples_tests/61.UI/UI_System.hpp
MohammadFakhreddin: I need access to the keyboard and mouse inputs. I'm currently using the input system.  What should I do instead?

examples_tests/61.UI/UI_System.hpp
MohammadFakhreddin: Got it

3rdparty/imgui/imgui.cpp
MohammadFakhreddin: I'll try to make it a submodule instead

examples_tests/61.UI/UI_System.cpp
```diff
@@ -0,0 +1,1102 @@
...
+
+		// Creating gpu image then copying data into it which needs a transient command buffer
+		cpu2GpuParams.beginCommandBuffers();
+		IGPUObjectFromAssetConverter cpu2gpuConverter{};
+		auto const gpuImage = cpu2gpuConverter.getGPUObjectsFromAssets(
+			&cpuImage,
+			&cpuImage + 1,
+			cpu2GpuParams
+		);
+		cpu2GpuParams.waitForCreationToComplete(false);
```
devshgraphicsprogramming, line 372: its a decent enough approach, but you should proably create a IGPUImage straight away and then use the video::IUTilities class to "copy buffer to image"

I guess fix it last.

examples_tests/61.UI/UI_System.cpp
```diff
@@ -0,0 +1,1102 @@
...
+						io.MouseDown[buttonIndex] = false;
+					}
+				}
+			}
+		});
+
+		auto const position = state->window->getCursorControl()->getPosition();
+		auto& io = ImGui::GetIO();
+		io.MousePos.x = static_cast<float>(position.x) - state->window->getX();
+		io.MousePos.y = static_cast<float>(position.y) - state->window->getY();
```
devshgraphicsprogramming, line 603: I'd rather get the mouse position from an event, reason is that events distinguish between mice... cursor control doesn't (we should probably remove that method).

examples_tests/61.UI/UI_System.hpp
devshgraphicsprogramming: take a `const IMouseEventChannel::range_t& events` as an argument

So basically whoever is using your GUI system, will be in charge of "filtering" the events, and only giving you the ones you should be interested in.

examples_tests/61.UI/UI_System.cpp
```diff
@@ -0,0 +1,1102 @@
...
+	//}
+
+	//-------------------------------------------------------------------------------------------------
+
+	bool Render(IGPUCommandBuffer& commandBuffer, int frameIndex)
+	{
+		ImGuiIO& io = ImGui::GetIO();
+		assert(io.Fonts->IsBuilt() && "Font atlas not built! It is generally built by the renderer backend. Missing call to renderer _NewFrame() function? e.g. ImGui_ImplOpenGL3_NewFrame().");
+
+		UpdateMousePositionAndButtons();
```
devshgraphicsprogramming, line 637: dont couple update with render... have user call them seprately!

examples_tests/61.UI/UI_System.cpp
```diff
@@ -0,0 +1,1102 @@
...
+
+			auto & indexBuffer = state->indexBuffers[frameIndex];
+
+			if (indexBuffer.isValid() == false || indexBuffer->getSize() < indexSize)
+			{
+				indexBuffer = state->device->createCPUSideGPUVisibleGPUBufferOnDedMem(
+					indexCreationParams,
+					indexSize
+				);
+			}
```
devshgraphicsprogramming, line 686: I like the fact you're round robining buffers, and not re-creating every frame.

However we have an "Upstreaming" streaming buffer in `IUtilities`

Its basically a persistently mapped host visible buffer, which you can suballocate from.
The nice part is that you can "bind" and draw directly from it.
You just need to remember to flush.

examples_tests/61.UI/UI_System.cpp
```diff
@@ -0,0 +1,1102 @@
...
+							clip_rect.y = 0.0f;
+
+						{// Apply scissor/clipping rectangle
+							VkRect2D scissor{};
+							scissor.offset.x = static_cast<int32_t>(clip_rect.x);
+							scissor.offset.y = static_cast<int32_t>(clip_rect.y);
+							scissor.extent.width = static_cast<uint32_t>(clip_rect.z - clip_rect.x);
+							scissor.extent.height = static_cast<uint32_t>(clip_rect.w - clip_rect.y);
+							commandBuffer.setScissor(0, 1, &scissor);
+						}
```
devshgraphicsprogramming, line 802: an idea for optimization, use 4 programmable clip planes instead `gl_ClipDistance[0-4]` in the vertex shader of changing the scissor rect all the time.

Add a second vertex attribute with a per-instance divisor, which would be a vec4 containing the scissor rect in NDC.
(basically a linear buffer of `vec4` scissor rects scaled to be in [-1,1]^2 relative to th viewport)

Then have a third GPU buffer full of DrawElementsIndirect, always drawing one instance, but using `baseInstance` to identify which scissor rect you want to use.

Then draw everything with a single drawcall !

examples_tests/61.UI/UI_System.cpp
devshgraphicsprogramming: if you put them into the builtins (see #340 ) then their source is "baked" into the engine binary, and they'll always be found.

examples_tests/61.UI/UI_System.cpp
MohammadFakhreddin: I have an update method. I move them there.

examples_tests/61.UI/UI_System.cpp
```diff
@@ -0,0 +1,1102 @@
...
+		io.Fonts->GetTexDataAsRGBA32(&pixels, &width, &height);
+		assert(pixels != nullptr);
+		assert(width > 0);
+		assert(height > 0);
+		size_t const componentsCount = 4;
+		size_t const image_size = width * height * componentsCount * sizeof(uint8_t);
+
+		// Pixels buffer-----------------------
+		smart_refctd_ptr<ICPUBuffer> imageCpuBuffer = make_smart_refctd_ptr<ICPUBuffer>(image_size);
+		memcpy(imageCpuBuffer->getPointer(), pixels, imageCpuBuffer->getSize());   // TODO: Can we avoid this copy ?
```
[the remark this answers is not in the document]
MohammadFakhreddin: Thanks! I will.

examples_tests/common/CommonAPI.h
MohammadFakhreddin: Done

examples_tests/61.UI/CMakeLists.txt
MohammadFakhreddin: Done

examples_tests/61.UI/UI_System.cpp
MohammadFakhreddin: I rewrote it in a class .

examples_tests/61.UI/UI_System.cpp
devshgraphicsprogramming: @deprilula28 Open a GH issue to redo these in HLSL and use bindless while we're at it, its a bit limiting to have to rebind textures if some skip or display thing changes

examples_tests/61.UI/UI_System.cpp
devshgraphicsprogramming: @deprilula28 you might want to fix this in your fork

examples_tests/61.UI/UI_System.cpp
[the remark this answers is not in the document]
devshgraphicsprogramming: @deprilula28 should be easy to fix in your fork as well

PR #348, opened 2022-05-03
[WIP] Buffer API and Other Backend Fixes/Changes

Erfan-Ahmadi:
Fixing PR issues [here](https://github.com/Devsh-Graphics-Programming/Nabla/pull/160) assigned to @Erfan-Ahmadi 
and other TODOs including:
* Improving Buffer API
* Improving Feature/Extension Selection and Report API
* Format Promotion and Image Upload Utility Improvements
* Maintenance

## Examples
- [ ] Finalize Porting Example 0.ImportanceSampling
## Fixes
- [ ] Screenshot Changes
  - [ ] wait semaphore (https://github.com/Devsh-Graphics-Programming/Nabla/pull/219#discussion_r800543586)
  - [ ] add convert filter
  - [ ] add flip image option
  - [ ] dst access mask (https://github.com/Devsh-Graphics-Programming/Nabla/pull/219#discussion_r800546866)
- [ ] getSizedOpenGLFormatFromOurFormat -> cache formats, and implement format feature functions for GL
  - (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r801454717)
  - Swapchain colorformat + colospace + oetf reporting for GL backends
- [ ] freeCommandBuffers_impl (see [this](https://github.com/Devsh-Graphics-Programming/Nabla/issues/305))
- [ ] drawParameters extention and fix for Vulkan and GLES (https://discord.com/channels/593902898015109131/932586232591024128/934485710193459251)
- [ ] Improve Features Selection API and Report (link to issue)
- [ ] CGLTF DerivativeMap normalization factor and builtin shader changes (consider normalization + push_constants in a nice struct)
- [ ] Const-Correctness of IGPURenderpassIndependentPipeline constructor parameters
- [ ] Image Upload Utility
  - [ ] Convert to Format and Copy Filter instead of manual
  - [ ] Improve code by using an Iterator that gives you the next region to copy based on available memory
  - [ ] Utility for calculating image data size in Asset namespace + other functions that may be useful
- [ ] Format Promotion (link to issue)
- [ ] OpenGLES Extensions that allow more formats
  - https://discord.com/channels/593902898015109131/942802829918232576/942810567624757288
- [ ]  Assert in updateBufferRangeViaStagingBuffer that the commandbuffer was begun with EU_ONE_TIME_SUBMIT_BIT
## Features and Extension API

- [ ] LogicalDevice function to query Enabled Features/Extension 
- [ ] move `getExtraGLSLDefines()` to ILogicalDevice and Implement for Vulkan
- [ ] https://discord.com/channels/593902898015109131/936526907648925706/972634976422076476
## Changing the Buffer API:

- [x] should only be called iff the vkBindBufferMemory call succeeds 
		- (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791849743)
		- (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791860130)
- [x] NUKE `core::smart_refctd_ptr<IDriverMemoryAllocation> m_backedMemory;` 
 - The smartpointer to the IDriverMemoryAllocation needs to be pushed down into CVulkanImage and CVulkanBuffer
 - (https://github.com/Devsh-Graphics-Programming/Nabla/pull/160#discussion_r791769190)	
- [x] Reorganise the structs in IGPUBuffer memory layout wise
- [x] return m_cachedCreationParams.declaredSize in IGPUBuffer::getSize()
- [x] Query ComputeShader Shared Memory Size via  Physical Device
- [x] E_DRIVER_VENDOR enum in `IPhysicalDevice` and implement for all backends
- [x] Spoof OpenGL Heaps and MemoryTypes
- [x] Remove "GPU" from ILogicalDevice function names

```
Heap 0 : Device local, its actually possible to query the VRAM size using some weird AMD/NVidia/Intel undocumented GL extensions
Heap 1 : Not Device Local, set to 70% of total (not free or available) system RAM or RAM - 2GB, whichever is higher (can be queried from ISystem)
Heap 2 : 256MB Device Local
```
```
<PRop,HeapIx>

{VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT, 0} // required by spec

{VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT | VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT, 2}
{VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT | VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT, 2}

{ZERO_VK_MEMORY_PROPERTY_FLAG, 1}
{VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT, 1} // required by spec
{VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT, 1}
```

- [x] Implement getXXXMemoryTypeMask and remove getXXXMemoryReqs
- [x] COpenGL_MemoryAllocation
- [x] MemoryType Iterator

```c++
class MemoryTypeIterator
{
  public:
    inline uint32_t end() const {return 32u;}    

    inline uint32_t operator*() const
    {
      const uint32_t type = dereference();
      assert((memoryTypeMask>>type)&0x1u);
      return type;
    }
    inline uint32_t operator++()
    {
      advance();
      return operator*();
    }
  
  protected:
    MemoryTypeIterator(const uint32_t _memoryTypeMask) : memoryTypeMask(_memoryTypeMask) {}
    virtual uint32_t deference() const=0;    
    virtual void advance()=0;

    uint32_t memoryTypeMask;
};
```
- [x] AllocateInfo struct and allocateMemory function
```C++
struct AllocateInfo
{
  size_t size : 54 = 0ull;
  size_t flags : 5 = 0u; // IDriverMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS
  size_t memoryTypeIndex : 5 = 0u;
  IDriverMemoryBacked* dedication = nullptr;
  size_t opaqueCaptureAddress = 0xdeadbeefBadC0ffeull;
};
// if you make the info have a `dedication` the memory will be bound right away
virtual core::smart_refctd_ptr<IDriverMemoryAllocation> allocateMemory(const AllocateInfo& info) = 0;
```
- [x] NUKE E_MAPPING_CAPABILITY_FLAGS ~~and fold values into `E_MEMORY_ALLOCATE_FLAGS`~~
- [x] cleanup `createGPUBuffer` -> remove size param and rename to `createBuffer`
- [x] SDriverMemoryRequirements Cleanup
- [x] Remove all the createXXXOnDedMem methods 
- [x]  replace `IDriverMemoryAllocation::getType()` with `uint8_t IDriverMemoryAllocation::getMemoryTypeIndex() const`
- [x] Querying VRAM size on OpenGL

Example BufferAPI pseudo-code usage:
```c++
auto buffer = device->createBuffer(params);
{
Requirements reqs = buffer->getMemoryRequirements();
reqs.memoryTypeMask &= device->getStreamingMemoryTypeMask();
auto memory = device->allocate(reqs,EMAF_WRITE,buffer.get());
if (!memory)
  buffer = nullptr;
// no further `buffer->bind({0ull,std::move(memory)})` needed
}
```

Edits after API Changes:
- [ ] Fix Ex48 on Vulkan (use utilities downstreaming buffer to download data instead of multiple huge buffers)

## Acceleration Structures
- [ ] Compaction Example and in AssetConverter
- [ ] Use bitflags for accelerationStructure flags now


include/nbl/video/IGPUBuffer.h
```diff
@@ -27,18 +27,18 @@ class IGPUBuffer : public asset::IBuffer, public IDriverMemoryBacked, public IBa
     public:
 		struct SCachedCreationParams
 		{
+			size_t declaredSize = 0ull;
 			core::bitflag<E_USAGE_FLAGS> usage = EUF_NONE;
-			bool canUpdateSubRange = false; // whether `IGPUCommandBuffer::updateBuffer` can be used on this buffer
 			asset::E_SHARING_MODE sharingMode = asset::ESM_EXCLUSIVE;
-			size_t declaredSize = 0ull;
+			bool canUpdateSubRange = false; // whether `IGPUCommandBuffer::updateBuffer` can be used on this buffer
```
devshgraphicsprogramming, line 33: can you somehow make sure and static-assert `sizeof(SCachedCreationParams)==16` ?

include/nbl/video/IPhysicalDevice.h
```diff
@@ -26,75 +26,167 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
     public:
         //
         virtual E_API_TYPE getAPIType() const = 0;
-        
-        // TODO: fold into SLimits
+
+        enum E_TYPE : uint8_t {
+            ET_OTHER = 0,
```
devshgraphicsprogramming, line 31: maybe UNKNOWN?

include/nbl/video/IPhysicalDevice.h
```diff
@@ -26,75 +26,167 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            ET_INTEGRATED_GPU = 1,
+            ET_DISCRETE_GPU = 2,
+            ET_VIRTUAL_GPU = 3,
+            ET_CPU = 4,
+        };
+
+        enum E_DRIVER_ID : uint8_t
+        {
+            EDI_OTHER = 0,
```
devshgraphicsprogramming, line 40: UNKNOWN

include/nbl/video/IPhysicalDevice.h
```diff
@@ -26,75 +26,167 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
-            uint32_t UBOAlignment;
-            uint32_t SSBOAlignment;
-            uint32_t bufferViewAlignment;
-            float    maxSamplerAnisotropyLog2;
-            float    timestampPeriodInNanoSeconds; // timestampPeriod is the number of nanoseconds required for a timestamp query to be incremented by 1 (a float because vulkan reports), use core::rational in the future
-
+            //uint32_t              maxImageDimension1D;
+            //uint32_t              maxImageDimension2D;
+            //uint32_t              maxImageDimension3D;
+            //uint32_t              maxImageDimensionCube;
```
devshgraphicsprogramming, line 79: why are they commented out?

include/nbl/video/IPhysicalDevice.h
```diff
@@ -26,75 +26,167 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
-            uint32_t maxTextures;
-            uint32_t maxStorageImages;
-
-            uint64_t maxTextureSize;
-
+            //uint32_t              maxPushConstantsSize;
+            //uint32_t              maxMemoryAllocationCount;
+            //uint32_t              maxSamplerAllocationCount;
+            //VkDeviceSize          bufferImageGranularity;
+            //VkDeviceSize          sparseAddressSpaceSize;
```
devshgraphicsprogramming, line 88: we should uncomment

include/nbl/video/IPhysicalDevice.h
```diff
@@ -114,6 +206,13 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
             uint32_t           shaderGroupHandleAlignment;
             uint32_t           maxRayHitAttributeSize;
 
+            //--> Nabla:
+            uint32_t maxBufferSize;
+            uint64_t maxTextureSize; // TODO: Use maxImageDimensions1D/2D/3D/Cube instead for gl and get rid of this
```
devshgraphicsprogramming, line 211: ok

include/nbl/video/IPhysicalDevice.h
```diff
@@ -26,75 +26,167 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
         {
             uint32_t major : 5;
             uint32_t minor : 5;
             uint32_t patch : 22;
         };
-        const APIVersion& getAPIVersion() const { return m_apiVersion; }
-        
+
         //
         struct SLimits
```
devshgraphicsprogramming, line 74: doesn't look like you've exposed everything in 1.0, 1.1, 1.2, 1.3 and extensions

include/nbl/video/IPhysicalDevice.h
```diff
@@ -114,6 +206,13 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
             uint32_t           shaderGroupHandleAlignment;
             uint32_t           maxRayHitAttributeSize;
 
+            //--> Nabla:
+            uint32_t maxBufferSize;
```
devshgraphicsprogramming, line 210: actually you can replace this with 1.1's `maxMemoryAllocationSize`

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+        struct SProperties
+        {
+            //--> VkPhysicalDeviceProperties:
+            APIVersion  apiVersion;
+            // uint32_t driverVersion;
+            // uint32_t vendorID;
+            // uint32_t deviceID;
+            E_TYPE      deviceType;
+            // char     deviceName[VK_MAX_PHYSICAL_DEVICE_NAME_SIZE];
+            // uint8_t  pipelineCacheUUID[VK_UUID_SIZE];
```
devshgraphicsprogramming, line 238: expose, just fill `driverVersion,vendorID,deviceID` with bogus values on GL (all zeros?)

on OpenGL I guess `deviceName=getGetString(GL_RENDERER)` and `pipelineCacheUUID` could be a hash (we hasve some 256bit hash function in `core`) of that string

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            SLimits     limits;
+            // VkPhysicalDeviceSparseProperties    sparseProperties;
+
+            //--> VkPhysicalDeviceDriverProperties
+            E_DRIVER_ID driverID;
+            // char driverName[VK_MAX_DRIVER_NAME_SIZE];
+            // char driverInfo[VK_MAX_DRIVER_INFO_SIZE];
+            // VkConformanceVersion conformanceVersion;
```
devshgraphicsprogramming, line 246: Conformance in OpenGL:
INTEL = 4.4
NVIDIA = 4.4
AMD = 3.3

I guess `driverInfo` can be `GL_RENDERER` again, but `driverName` could just be the `E_DRIVER_ID` enum but as a string

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+
+            //--> VkPhysicalDeviceIDProperties
+            uint8_t deviceUUID[VK_UUID_SIZE];
+            // uint8_t driverUUID[VK_UUID_SIZE];
+            // uint8_t deviceLUID[VK_LUID_SIZE];
+            // uint32_t deviceNodeMask;
+            // VkBool32 deviceLUIDValid;
```
devshgraphicsprogramming, line 253: fill UUID with 0s on OpenGL, obviously LUID won't be valid

`deviceNodeMask`, make it look like its a single GPU

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+        };
+
+        const SProperties& getProperties() const { return m_properties; }
+        const SLimits& getLimits() const { return m_properties.limits; }
+        const APIVersion& getAPIVersion() const { return m_properties.apiVersion; }
```
devshgraphicsprogramming, line 258: the value is 32-bit, no need for a reference

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            //VkBool32    sparseResidencyImage3D;
+            //VkBool32    sparseResidency2Samples;
+            //VkBool32    sparseResidency4Samples;
+            //VkBool32    sparseResidency8Samples;
+            //VkBool32    sparseResidency16Samples;
+            //VkBool32    sparseResidencyAliased;
+            //VkBool32    variableMultisampleRate;
+            bool inheritedQueries = false;
+
+            //--> VkPhysicalDeviceSubgroupProperties: // TODO(Erfan): I think we should move these into SProperties::SLimits since it's part of properties and not features
```
devshgraphicsprogramming, line 319: ok, do it

include/nbl/video/IPhysicalDevice.h
```diff
@@ -148,49 +328,80 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
 
-            // Buffer Device Address
+            //--> VkPhysicalDeviceBufferDeviceAddressFeaturesKHR
             bool bufferDeviceAddress = false;
+            //VkBool32           bufferDeviceAddressCaptureReplay;
+            //VkBool32           bufferDeviceAddressMultiDevice;
+            
+            //--> Nabla:
+            bool dispatchBase = false;
+            bool drawIndirectCount = false;
```
devshgraphicsprogramming, line 362: pretty sure that's a feature in Vulkan

devshgraphicsprogramming: probably has a different name

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
 
         //
         struct SFeatures
         {
             bool robustBufferAccess = false;
+            //VkBool32    fullDrawIndexUint32;
             bool imageCubeArray = false;
+            //VkBool32    independentBlend;
+            bool geometryShader    = false;
+            //VkBool32    tessellationShader;
```
devshgraphicsprogramming, line 268: these we want to expose under GL definitely

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            //VkBool32    sampleRateShading;
+            //VkBool32    dualSrcBlend;
             bool logicOp = false;
+            bool multiDrawIndirect = false;
+            //VkBool32    drawIndirectFirstInstance;
+            //VkBool32    depthClamp;
+            //VkBool32    depthBiasClamp;
```
devshgraphicsprogramming, line 275: please expose

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            //VkBool32    fillModeNonSolid;
+            //VkBool32    depthBounds;
+            //VkBool32    wideLines;
+            //VkBool32    largePoints;
+            //VkBool32    alphaToOne;
```
devshgraphicsprogramming, line 280: need to expose

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            //VkBool32    textureCompressionASTC_LDR;
+            //VkBool32    textureCompressionBC;
+            //VkBool32    occlusionQueryPrecise;
+            //VkBool32    pipelineStatisticsQuery;
+            //VkBool32    vertexPipelineStoresAndAtomics;
+            //VkBool32    fragmentStoresAndAtomics;
+            //VkBool32    shaderTessellationAndGeometryPointSize;
+            //VkBool32    shaderImageGatherExtended;
+            //VkBool32    shaderStorageImageExtendedFormats;
+            //VkBool32    shaderStorageImageMultisample;
```
devshgraphicsprogramming, line 293: want to expose this

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            //VkBool32    shaderStorageImageReadWithoutFormat;
+            //VkBool32    shaderStorageImageWriteWithoutFormat;
+            //VkBool32    shaderUniformBufferArrayDynamicIndexing;
+            //VkBool32    shaderSampledImageArrayDynamicIndexing;
+            //VkBool32    shaderStorageBufferArrayDynamicIndexing;
+            //VkBool32    shaderStorageImageArrayDynamicIndexing;
+            //VkBool32    shaderClipDistance;
+            //VkBool32    shaderCullDistance;
```
devshgraphicsprogramming, line 301: want to expose

include/nbl/video/IPhysicalDevice.h
```diff
@@ -126,17 +225,98 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            //VkBool32    wideLines;
+            //VkBool32    largePoints;
+            //VkBool32    alphaToOne;
             bool multiViewport = false;
-            bool vertexAttributeDouble = false;
-            bool dispatchBase = false;
+            bool samplerAnisotropy = false;
+            //VkBool32    textureCompressionETC2;
+            //VkBool32    textureCompressionASTC_LDR;
+            //VkBool32    textureCompressionBC;
```
devshgraphicsprogramming, line 285: want to expose, because want to flag as enabled.

include/nbl/video/IPhysicalDevice.h
```diff
@@ -148,49 +328,80 @@ class IPhysicalDevice : public core::Interface, public core::Unmovable
...
+        {
+            EMPF_DEVICE_LOCAL_BIT = 0x00000001,
+            EMPF_HOST_VISIBLE_BIT = 0x00000002,
+            EMPF_HOST_COHERENT_BIT = 0x00000004,
+            EMPF_HOST_CACHED_BIT = 0x00000008,
+            EMPF_LAZILY_ALLOCATED_BIT = 0x00000010,
+            EMPF_PROTECTED_BIT = 0x00000020,
+            EMPF_DEVICE_COHERENT_BIT_AMD = 0x00000040,
+            EMPF_DEVICE_UNCACHED_BIT_AMD = 0x00000080,
+            EMPF_RDMA_CAPABLE_BIT_NV = 0x00000100,
```
devshgraphicsprogramming, line 377: maybe we should comment these out for now (haven't read the extension spec, unsure what extra validation we should do)

src/nbl/video/IOpenGL_PhysicalDeviceBase.h
```diff
@@ -237,15 +237,81 @@ class IOpenGL_PhysicalDeviceBase : public IOpenGLPhysicalDeviceBase
...
+		if(m_properties.driverID == E_DRIVER_ID::EDI_INTEL_OPEN_SOURCE_MESA || m_properties.driverID == E_DRIVER_ID::EDI_INTEL_PROPRIETARY_WINDOWS)
+			m_properties.deviceType = E_TYPE::ET_INTEGRATED_GPU;
+		else if(m_properties.driverID == E_DRIVER_ID::EDI_AMD_OPEN_SOURCE || m_properties.driverID == E_DRIVER_ID::EDI_AMD_PROPRIETARY)
+			m_properties.deviceType = E_TYPE::ET_DISCRETE_GPU;
+		else if(m_properties.driverID == E_DRIVER_ID::EDI_NVIDIA_PROPRIETARY)
+			m_properties.deviceType = E_TYPE::ET_DISCRETE_GPU;
+		else if(hasInString(renderer, "virgl", true))
+			m_properties.deviceType = E_TYPE::ET_VIRTUAL_GPU;
+		else
+			m_properties.deviceType = E_TYPE::ET_OTHER;
```
devshgraphicsprogramming, line 309: you can also use `#if defined(NBL_PLATFORM)` to help yourself differentiate between WINDOWS and others

include/nbl/video/IPhysicalDevice.h
Erfan-Ahmadi: Ok but I didn't comment anything out in actuality. I just reordered and wrote the ones we didn't use in Vulkan.
Everything I uncommented I will implement for opengl also.
I'll do these things at some point after BufferAPI when cleaning up features and stuff

include/nbl/video/IPhysicalDevice.h
Erfan-Ahmadi: Yea unknown for this makes sense

include/nbl/video/IPhysicalDevice.h
Erfan-Ahmadi: "Other" here has the same name as the Vulkan enum. It kinda makes sense. But for consistently I will make it UNKNOWN

include/nbl/video/IDriverMemoryAllocation.h
```diff
@@ -90,12 +90,13 @@ class IDriverMemoryAllocation : public virtual core::IReferenceCounted
...
 
         //! Memory allocate flags
-        enum E_MEMORY_ALLOCATE_FLAGS
+        enum E_MEMORY_ALLOCATE_FLAGS : uint32_t
         {
             EMAF_NONE = 0x00000000,
             EMAF_DEVICE_MASK_BIT = 0x00000001,
             EMAF_DEVICE_ADDRESS_BIT = 0x00000002,
             EMAF_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT = 0x00000004,
+            // Mapping Capability Flags
```
devshgraphicsprogramming, line 99: huh?

include/nbl/video/IDriverMemoryBacked.h
```diff
@@ -27,6 +27,17 @@ class IDriverMemoryBacked : public virtual core::IReferenceCounted
...
+        
+        struct SDriverMemoryRequirements2
+        {
+            size_t   size;
+            uint32_t memoryTypeBits;
+            uint32_t alignmentLog2 : 6;
+            uint32_t prefersDedicatedAllocation     : 1;
+            uint32_t requiresDedicatedAllocation    : 1;
+        };
+        static_assert(sizeof(SDriverMemoryRequirements2)==16);
```
devshgraphicsprogramming, line 39: why have a struct with a `2` prefix?

Arent we nuking the old one?

src/nbl/video/COpenGL_LogicalDevice.h
```diff
@@ -241,6 +241,11 @@ class COpenGL_LogicalDevice : public IOpenGL_LogicalDevice
         return output;
     }
 
+    SMemoryOffset allocate(const SAllocateInfo& info) override
+    {
+        return {}; //TODO: Erfan
```
devshgraphicsprogramming, line 246: its fine, return nullptr and invalid ;)

include/nbl/video/IDriverMemoryAllocation.h
Erfan-Ahmadi: Don't worry about this, I was going to move the mapping capability flags here, but saw that it was not trivial so I decided to do other tasks first

include/nbl/video/IDriverMemoryBacked.h
Erfan-Ahmadi: Yea please don't review right now :)))
I'm going to. I just didn't want the hassle to change every place to fix compile errors, after I get this working I'll remove the old one and rename

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -297,7 +297,79 @@ core::smart_refctd_ptr<IDriverMemoryAllocation> CVulkanLogicalDevice::allocateGP
...
+    VkMemoryDedicatedAllocateInfo vk_dedicatedInfo = {VK_STRUCTURE_TYPE_MEMORY_DEDICATED_ALLOCATE_INFO, nullptr};
+    VkMemoryAllocateFlagsInfo vk_allocateFlagsInfo = { VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_FLAGS_INFO, nullptr };
+    VkMemoryAllocateInfo vk_allocateInfo = { VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO, &vk_allocateFlagsInfo};
+
+    if (allocateFlags.hasValue(IDriverMemoryAllocation::EMAF_DEVICE_MASK_BIT))
+        vk_allocateFlagsInfo.flags |= VK_MEMORY_ALLOCATE_DEVICE_MASK_BIT;
+    else if(allocateFlags.hasValue(IDriverMemoryAllocation::EMAF_DEVICE_ADDRESS_BIT))
+        vk_allocateFlagsInfo.flags |= VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT;
+    else if (allocateFlags.hasValue(IDriverMemoryAllocation::EMAF_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT))
+        vk_allocateFlagsInfo.flags |= VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_CAPTURE_REPLAY_BIT;
```
devshgraphicsprogramming, line 313: so when do we use it and how do we use it?

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -297,7 +297,79 @@ core::smart_refctd_ptr<IDriverMemoryAllocation> CVulkanLogicalDevice::allocateGP
...
+    vk_allocateInfo.allocationSize = info.size;
+    vk_allocateInfo.memoryTypeIndex = info.memoryTypeIndex;
+
+    VkDeviceMemory vk_deviceMemory;
+    if(info.dedication)
+    {
+        // VK_KHR_dedicated_allocation is in core 1.1, no querying for support needed
+        static_assert(MinimumVulkanApiVersion >= VK_MAKE_API_VERSION(0, 1, 1, 0));
+        const auto& dedicationMemoryRequirements = info.dedication->getMemoryReqs();
+        if(dedicationMemoryRequirements.prefersDedicatedAllocation || dedicationMemoryRequirements.requiresDedicatedAllocation)
```
devshgraphicsprogramming, line 325: if user gave it a dedication then dedicate it!

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -297,7 +297,79 @@ core::smart_refctd_ptr<IDriverMemoryAllocation> CVulkanLogicalDevice::allocateGP
...
+            else if(info.dedication->getObjectType() == IDriverMemoryBacked::EOT_IMAGE)
+                vk_dedicatedInfo.image = static_cast<CVulkanImage*>(info.dedication)->getInternalObject();
+        }
+    }
+
+    auto vk_res = m_devf.vk.vkAllocateMemory(m_vkdev, &vk_allocateInfo, nullptr, &vk_deviceMemory);
+    if (vk_res == VK_SUCCESS)
+    {
+        ret.memory = core::make_smart_refctd_ptr<CVulkanMemoryAllocation>(this, info.size, false, vk_deviceMemory, allocateFlags);
+        ret.offset = 0ull; // currently we allocate on demand so offset will be 0 relative to the memory handle
```
devshgraphicsprogramming, line 339: update comment to "LogicalDevice doesn't suballocate, so offset always 0, if you want to suballocate, write/use an allocator"

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -297,7 +297,79 @@ core::smart_refctd_ptr<IDriverMemoryAllocation> CVulkanLogicalDevice::allocateGP
...
+                bindImageInfo.image = static_cast<IGPUImage*>(info.dedication);
+                bindImageInfo.memory = ret.memory.get();
+                bindImageInfo.offset = ret.offset;
+                dedicationSuccess = bindImageMemory(1u, &bindImageInfo);
+            }
+            if(!dedicationSuccess)
+            {
+                // automatically allocation goes out of scope and frees itself
+                ret = {nullptr, IDeviceMemoryAllocator::InvalidMemoryOffset};
+            }
```
devshgraphicsprogramming, line 365: turn it into a `switch` and assert/error/blowup in the default case

src/nbl/video/CVulkanLogicalDevice.cpp
Erfan-Ahmadi: Never, I just gote pasted it from previous allocate function.
I'll remove this.

include/nbl/video/IDeviceMemoryAllocator.h
```diff
@@ -3,6 +3,7 @@
 
 #include "IDriverMemoryAllocation.h"
 #include "IDriverMemoryBacked.h"
+#include "nbl/core/math/glslFunctions.h" // findLSB
```
devshgraphicsprogramming, line 6: just include `nbl/core/declarations.h`

devshgraphicsprogramming: or `definitions`

include/nbl/video/IDeviceMemoryAllocator.h
```diff
@@ -27,7 +28,93 @@ class IDeviceMemoryAllocator
...
+		}
+
+	protected:
+		virtual uint32_t dereference() const = 0;
+		virtual void advance() = 0;
+		virtual bool operator==(uint32_t rhs) const = 0;
+		virtual bool operator!=(uint32_t rhs) const = 0;
+		
+		uint32_t m_allocateFlags;
+		IDriverMemoryBacked::SDriverMemoryRequirements2 m_reqs;
```
devshgraphicsprogramming, line 67: swap the order (padding)

include/nbl/video/IDeviceMemoryAllocator.h
```diff
@@ -27,7 +28,93 @@ class IDeviceMemoryAllocator
...
+			ret.memoryTypeIndex = dereference();
+			ret.dedication = dedication;
+			return ret;
+		}
+
+	protected:
+		virtual uint32_t dereference() const = 0;
+		virtual void advance() = 0;
+		virtual bool operator==(uint32_t rhs) const = 0;
+		virtual bool operator!=(uint32_t rhs) const = 0;
```
devshgraphicsprogramming, line 64: what do you need those for?

Erfan-Ahmadi: easiest way I could find to != IMemoryTypeItr::end() which is a uint32_t
If ::end it's iterator type instead (like usual iterators) then it'll also need a operator +

src/nbl/video/COpenGL_LogicalDevice.h
```diff
@@ -135,7 +135,33 @@ class COpenGL_LogicalDevice : public IOpenGL_LogicalDevice
...
+        {
+            if (params.type == IGPUImage::ET_1D)
+                return nullptr;
+        }
 
+        core::smart_refctd_ptr<IGPUImage> retval;
+
+        SRequestImageCreate2 reqParams;
+
+        reqParams.mreqs.size = 0u; // TODO(Erfan) some approx of image size -> considering dimensions, mipLevels, arrayLayers, samples, minImageGranularity and texelBlockInfo (see ImageUploadUtility)
```
devshgraphicsprogramming, line 152: we should have some utility in `asset` namespace to setup the Regions array (calculate offsets into a CPUBuffer with provided alignment restrictions) for a blank ICPUImage (loaders, and other utilities would benefit).

Then we could reuse here (simply by iterating the regions array, finding the one with highest offset+size and rounding that up to whatever alignment is necessary).

src/nbl/video/COpenGL_LogicalDevice.h
```diff
@@ -135,7 +135,33 @@ class COpenGL_LogicalDevice : public IOpenGL_LogicalDevice
...
+        reqParams.mreqs.memoryTypeBits = m_physicalDevice->getDeviceLocalMemoryTypeBits();
+        reqParams.mreqs.alignmentLog2 = 0u; // TODO(Erfan) Alignment previously was 0u in getXXXMemoryRequirementsOnDedMem(). what to set here? get minXXXOffsetAlignment from physical device and deduce from usage?
```
devshgraphicsprogramming, line 154: doesn't really matter, you can either set 8 (max GL_PACK/UNPACK pixel buffer alignments) or 256, doesn't really matter

src/nbl/video/COpenGL_LogicalDevice.h
```diff
@@ -229,14 +255,35 @@ class COpenGL_LogicalDevice : public IOpenGL_LogicalDevice
...
+        SRequestAllocate reqParams;
+        reqParams.allocateInfo = info;
+        reqParams.memoryTypeIndexFlags = m_physicalDevice->getMemoryProperties().memoryTypes[info.memoryTypeIndex].propertyFlags;
+
+        SMemoryOffset output;
+        auto& req = m_threadHandler.request(std::move(reqParams),&output);
+        m_masterContextCallsInvoked++;
+        m_threadHandler.template waitForRequestCompletion<SRequestAllocate>(req);
+
+        return output;
```
devshgraphicsprogramming, line 267: bro, OpenGL can't allocate any memory, just return nullptr/invalid

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -81,6 +83,7 @@ class IOpenGL_LogicalDeviceBase
             ERT_INVALIDATE_MAPPED_MEMORY_RANGES,
             ERT_MAP_BUFFER_RANGE,
             ERT_UNMAP_BUFFER,
+            ERT_ALLOCATE,
```
devshgraphicsprogramming, line 86: OpenGL can't allocate shit, no request needed

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -230,6 +247,13 @@ class IOpenGL_LogicalDeviceBase
 
             core::smart_refctd_ptr<IDriverMemoryAllocation> buf;
         };
+        struct SRequestAllocate
+        {
+            static inline constexpr E_REQUEST_TYPE type = ERT_ALLOCATE;
+            using retval_t = IDeviceMemoryAllocator::SMemoryOffset;
+            IDeviceMemoryAllocator::SAllocateInfo allocateInfo;
+            core::bitflag<IPhysicalDevice::E_MEMORY_PROPERTY_FLAGS> memoryTypeIndexFlags;
+        };
```
devshgraphicsprogramming, line 256: no implementation necessary

devshgraphicsprogramming, line 256: I see what you did, but its best to "allocate" inside createBuffer/createImage on GL, because you want to ensure `glStorageBuffer` and `glStorageImage` run before the user has a chance to bind the generated handles to anything

devshgraphicsprogramming: so we shouldn't have an allocate command

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -625,6 +666,27 @@ class IOpenGL_LogicalDevice : public ILogicalDevice, protected impl::IOpenGL_Log
...
+                auto& p = std::get<SRequestAllocate>(req.params_variant);
+                IDeviceMemoryAllocator::SMemoryOffset* pretval = reinterpret_cast<IDeviceMemoryAllocator::SMemoryOffset*>(req.pretval);
+                IDeviceMemoryAllocator::SMemoryOffset& retval = *pretval;
+                if(p.allocateInfo.dedication)
+                {
+                    // TODO(Erfan): Use COpenGLMemoryAllocation to include COpenGLImage too and static cast to that and use the virtual funtion initMemory
+                    COpenGLBuffer* glBuffer = static_cast<COpenGLBuffer*>(p.allocateInfo.dedication);
+                    glBuffer->initMemory(&gl, p.memoryTypeIndexFlags);
+                    retval.memory = core::smart_refctd_ptr<COpenGLBuffer>(glBuffer);
+                    retval.offset = 0ull;
```
devshgraphicsprogramming, line 680: you can set the memory type bits, etc in the `allocate` directly

Just make sure `initMemory` does no OpenGL calls, and you don't need to do them on a thread

src/nbl/video/IOpenGLMemoryAllocation.h
```diff
@@ -0,0 +1,58 @@
...
+        : IDriverMemoryAllocation(dev, IDriverMemoryAllocation::EMAF_NONE, IDriverMemoryAllocation::EMPF_NONE), initialized(false)
+    {}
+
+    ~IOpenGLMemoryAllocation() = default;
+    
+    //! Whether the allocation was made for a specific resource and is supposed to only be bound to that resource.
+    bool isDedicated() const override { return true; }
+    
+    //! Returns the size of the memory allocation
+    size_t getAllocationSize() const override { return 0ull; }
```
devshgraphicsprogramming, line 26: this might fuck with some of our cross-platform memory mapping code, its best if you return the buffer size for an allocation backing a buffer

(basically make `COpenGLBuffer` override it and return the same as `getSize()`)

Erfan-Ahmadi: It already does, but since it's a pure virtual function in IDriverMemoryAllocation I had to add this in the middle. but at the end COpenGLImage/Buffer override and implement it

src/nbl/video/IOpenGL_LogicalDevice.h
Erfan-Ahmadi: discussed in discord

src/nbl/video/IOpenGLMemoryAllocation.h
Erfan-Ahmadi: But I think It was dumb, because we don't construct this class and hence I can just get rid of this

src/nbl/video/COpenGL_LogicalDevice.h
```diff
@@ -135,7 +135,33 @@ class COpenGL_LogicalDevice : public IOpenGL_LogicalDevice
...
+        reqParams.mreqs.size = 0u; // TODO(Erfan) some approx of image size -> considering dimensions, mipLevels, arrayLayers, samples, minImageGranularity and texelBlockInfo (see ImageUploadUtility)
+        reqParams.mreqs.memoryTypeBits = m_physicalDevice->getDeviceLocalMemoryTypeBits();
+        reqParams.mreqs.alignmentLog2 = 0u; // TODO(Erfan) Alignment previously was 0u in getXXXMemoryRequirementsOnDedMem(). what to set here? get minXXXOffsetAlignment from physical device and deduce from usage?
+        reqParams.mreqs.prefersDedicatedAllocation = true;
+        reqParams.mreqs.requiresDedicatedAllocation = true;
+
+        reqParams.creationParams = std::move(params);
+        auto& req = m_threadHandler.request(std::move(reqParams), &retval);
+        m_masterContextCallsInvoked++;
+        m_threadHandler.template waitForRequestCompletion<SRequestImageCreate>(req);
```
devshgraphicsprogramming, line 161: typo, should be 
```cpp
m_threadHandler.template waitForRequestCompletion<SRequestImageCreate2>(req);
```

request type sent must match the template param

devshgraphicsprogramming: tbf safest way to code it is
```cpp
m_threadHandler.template waitForRequestCompletion<decltype(reqParams)>(req);
```
P.S. I might be missing an `std::remove_reference_t<decltype(reqParams)>`

src/nbl/video/COpenGL_LogicalDevice.h
```diff
@@ -225,7 +252,65 @@ class COpenGL_LogicalDevice : public IOpenGL_LogicalDevice
...
+    }
+
+    core::smart_refctd_ptr<IGPUBuffer> createBuffer(const IGPUBuffer::SCreationParams& creationParams) override
+    {
+        SRequestBufferCreate2 reqParams;
+        reqParams.mreqs.size = creationParams.declaredSize;
+        reqParams.mreqs.memoryTypeBits = 0xffffffffu;
+        reqParams.mreqs.alignmentLog2 = 0u; // TODO(Erfan) Alignment previously was 0u in getXXXMemoryRequirementsOnDedMem(). what to set here? get minXXXOffsetAlignment from physical device and deduce from usage?
+        reqParams.mreqs.prefersDedicatedAllocation = true;
+        reqParams.mreqs.requiresDedicatedAllocation = true;
```
devshgraphicsprogramming, line 304: maybe don't send stuff you don't need XD

src/nbl/video/COpenGL_LogicalDevice.h
```diff
@@ -352,18 +437,17 @@ class COpenGL_LogicalDevice : public IOpenGL_LogicalDevice
         m_threadHandler.template waitForRequestCompletion<SRequestInvalidateMappedMemoryRanges>(req);
     }
 
-    void* mapMemory(const IDriverMemoryAllocation::MappedMemoryRange& memory, IDriverMemoryAllocation::E_MAPPING_CPU_ACCESS_FLAG access = IDriverMemoryAllocation::EMCAF_READ_AND_WRITE) override final
+    void* mapMemory(const IDriverMemoryAllocation::MappedMemoryRange& memory) override final
```
devshgraphicsprogramming, line 440: nope, even though buffer is from a memory type that supports RW, I still might want to map with a subset of usages.

so leave an extra parameters for narrowing down the mapping access (and bitwise AND it with the actual memory type capability so its a strict subset).

Write in the docs that that some APIs (like Vulkan ignore it) and we AND the caps anyway.

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -81,6 +83,7 @@ class IOpenGL_LogicalDeviceBase
             ERT_INVALIDATE_MAPPED_MEMORY_RANGES,
             ERT_MAP_BUFFER_RANGE,
             ERT_UNMAP_BUFFER,
+            ERT_ALLOCATE,
             //BIND_BUFFER_MEMORY,
```
devshgraphicsprogramming, line 87: remove `BIND_BUFFER_MEMORY`

src/nbl/video/IOpenGL_LogicalDevice.h
devshgraphicsprogramming, line 86: make sure `ERT_ALLOCATE` comes before `ERT_COMPUTE_PIPELINE_CREATE`

src/nbl/video/COpenGL_LogicalDevice.h
Erfan-Ahmadi: Yea I think It's better to write most of these stuff when processing the requests

include/nbl/video/IDeviceMemoryAllocator.h
devshgraphicsprogramming: I dont think these need to be virtual you could inline them and compare against `dereference()`

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -573,11 +572,19 @@ class IOpenGL_LogicalDevice : public ILogicalDevice, protected impl::IOpenGL_Log
             {
                 auto& p = std::get<SRequestBufferCreate2>(req.params_variant);
                 core::smart_refctd_ptr<IGPUBuffer>* pretval = reinterpret_cast<core::smart_refctd_ptr<IGPUBuffer>*>(req.pretval);
+
+                IDriverMemoryBacked::SDriverMemoryRequirements2 mreqs;
+                mreqs.size = p.creationParams.declaredSize;
+                mreqs.memoryTypeBits = 0xffffffffu;
+                mreqs.alignmentLog2 = 0u; // TODO(Erfan) Alignment previously was 0u in getXXXMemoryRequirementsOnDedMem(). what to set here? get minXXXOffsetAlignment from physical device and deduce from usage?
+                mreqs.prefersDedicatedAllocation = true;
+                mreqs.requiresDedicatedAllocation = true;
```
devshgraphicsprogramming, line 581: I'd rather move that setup into `COpenGLBuffer` constructor, and just take declaredSize as argument

src/nbl/video/IOpenGL_LogicalDevice.h
```diff
@@ -608,6 +615,13 @@ class IOpenGL_LogicalDevice : public ILogicalDevice, protected impl::IOpenGL_Log
                 GLenum target;
                 GLuint name;
 
+                IDriverMemoryBacked::SDriverMemoryRequirements2 mreqs;
+                mreqs.size = 0u; // TODO(Erfan) some approx of image size -> considering dimensions, mipLevels, arrayLayers, samples, minImageGranularity and texelBlockInfo (see ImageUploadUtility)
+                mreqs.memoryTypeBits = p.deviceLocalMemoryTypeBits;
+                mreqs.alignmentLog2 = 0u; // TODO(Erfan) Alignment previously was 0u in getXXXMemoryRequirementsOnDedMem(). what to set here? get minXXXOffsetAlignment from physical device and deduce from usage?
+                mreqs.prefersDedicatedAllocation = true;
+                mreqs.requiresDedicatedAllocation = true;
```
devshgraphicsprogramming, line 623: I'd rather move that setup into `COpenGLImage` constructor

include/nbl/video/IDeviceMemoryAllocator.h
Erfan-Ahmadi: oh yea :D
thanks for the tip

include/nbl/video/IPhysicalDevice.h
Erfan-Ahmadi: @devshgraphicsprogramming 

``dispatchBase`` exists on vulkan with no need for feature checking and is not valid to use in gl because we don't implement it.

https://github.com/Devsh-Graphics-Programming/Nabla/blob/fddcc6b81cfcb4229648043539c41ba3c2ba095c/src/nbl/video/COpenGLCommandBuffer.cpp#L833-L835

https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCmdDispatchBase.html

Erfan-Ahmadi: About `drawIndirectCount` I have no idea why it is a bool and in features but It's exposed under `maxDrawIndirectCount` in limits

devshgraphicsprogramming: > @devshgraphicsprogramming
> 
> `dispatchBase` exists on vulkan with no need for feature checking and is not valid to use in gl because we don't implement it.
> 
> https://github.com/Devsh-Graphics-Programming/Nabla/blob/fddcc6b81cfcb4229648043539c41ba3c2ba095c/src/nbl/video/COpenGLCommandBuffer.cpp#L833-L835
> 
> https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCmdDispatchBase.html

yes make it fail in GL

PR #362, opened 2022-06-05
Example fixes

Hazardu:
## Description
Necessary fixes for the newest buffer api


devshgraphicsprogramming:
@Hazardu when do I merge?

Issue #370, opened 2022-06-29
Text rendering plan


deprilula28:
## Description
Plan for the MSDF text rendering system

## Solution proposal

- Choice between MSDF & Vector Textures (Preference for MSDF; Should we want to change later, we need to change data structures for the glyphs & pixel shader that calculates coverage)

- Text rendering CPU side:
    - The text string is not GPU generated, so the glyph ID (index into a table of, texture atlas offsets [there are multiple due to LoD] and pixel size), local offset (kerning), and bounding box of whole string can be pre-computed on the CPU for each string
    - Pool the geometry (vertex data) for all text using SubAllocatedDataBuffer
    - Pool the strings (offset into geometry, aabb, transform matrix etc.) using CPropertyPool
    - **Expansion compute shader**: Given a list of visible text strings (other system will do frustum culling), perform prefix sum over glyph counts in string with CScanner (the indirect variant), and have the last invocation set up a DrawIndirect struct to render all glyphs.
    - Use DrawIndirect to draw all text and pass the glyph data from a UTB (no HW vertex attributes) via custom indexing scheme: `globalGlyphID = gl_VertexIndex >> 2`.
    - Add an option for rendering without an index buffer as well (we can live with 50% more vx shader invocations) then the custom indexing scheme is: `globalGlyphID = gl_VertexIndex / 6`.
    - Use `upper_bound(stringLengthPrefixSum,stringLengthPrefixSum+stringCount,globalGlyphID)-1P`  to find the localGlyphID within the string being processed, as well as the visible stringID. This will enable getting all string and glyph properties and working out the vertices of the string, as well as passing the texture offset data to frag shader.

Vertex Shader pseudocode:
```
layout() uniform usamplerBuffer nbl_glsl_ext_MSDFTextRenderer_StringGeometryPool; // r8_uint or r16_uint
layout() uniform samplerBuffer nbl_glsl_ext_MSDFTextRenderer_VisibleStringGeometryOffset
{
  uint data[];
} nbl_glsl_ext_MSDFTextRenderer_VisibleStringGeometryOffset;
layout() readonly buffer VisibleStringPrefixSum
{
  uint data[];
} nbl_glsl_ext_MSDFTextRenderer_VisibleStringPrefixSum;
layout() readonly buffer VisibleStringMVP
{
  mat4 data[];
} nbl_glsl_ext_MSDFTextRenderer_VisibleStringMVP;

[...]

PR #371, opened 2022-06-29
Text rendering mockup

deprilula28:
## Description
More information about the text rendering system plan: #370 

## TODO list
- [ ] Review API
- [ ] Implement everything
- [ ] Write an example/test (?)
- [ ] Test everything


include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,88 @@
...
+	);
+
+	// - Expansion compute shader: Prefix sum over glyph counts & Output draw args
+	// - Indirect draw for text
+	void drawText(core::smart_refctd_ptr<video::IGPUCommandBuffer> cmdbuf);
+
+private:
+
+private:
+	ILogicalDevice* m_device;
```
devshgraphicsprogramming, line 76: smartpointer needed

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,88 @@
...
+	int minX, minY, 
+		maxX, maxY;
+};
+
+class NBL_API TextRenderer
+{
+public:
+	TextRenderer(ILogicalDevice* device, uint32_t maxGlyphCount, uint32_t maxStringCount);
+
+	bool poolString(
```
devshgraphicsprogramming, line 43: `allocateString` and `freeString` plz

devshgraphicsprogramming: return a `uint32_t` offset (handle) to the string in the string pool

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,88 @@
...
+{
+namespace TextRendering
+{
+
+struct SGlyphData 
+{
+	// Offset in pixels from the start of the string
+	int offsetX, offsetY;
+	uint32_t textureAtlasGlyphIndex;
+};
```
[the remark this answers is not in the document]
deprilula28: uint16_t for the glyph index then?

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,97 @@
...
+
+	bool allocateString(
+		// Offset in pixels from top left in the screen
+		int offsetX, int offsetY,
+		uint32_t glyphCount,
+		SGlyphData const* glyphs,
+		// Offset into allocated parts of m_stringDataPropertyPool & m_geomDataBuffer
+		uint32_t* outStringOffset,
+		uint32_t* outGeometryOffset
+	);
```
devshgraphicsprogramming, line 60: have this signature instead
```cpp
using glyph_geometry_pool_t = video::SubAllocatedDataBufferST<>;
using string_pool_t = video::CPropertyPool<core::allocator,glyph_geometry_pool_t::size_type,StringBoundingBox>;

struct string_handle_t
{
    string_pool_t::size_type stringAddr = string_pool_t::invalid_address;
    glyph_geometry_pool_t::size_type glyphDataAddr = glyph_geometry_pool_t::invalid_address;
    glyph_geometry_pool_t::size_type glyphCount = 0u;
};

// how many were successfully allocated, `handles` will be properly filled between [0,retval)
uint32_t allocateStrings(
    const std::chrono::time_point<Clock>& maxWaitPoint, // lets have also a version without this (default)
    const uint32_t count, // how many strings
    string_handle_t* handles, // output handles, if `glyphDataAddr` was not primed with invalid_address, allocation will not happen, likewise for `stringDataAddr`
    const char* const* stringData,
    const StringBoundingBox* wrappingBoxes = nullptr // optional, to wrap paragraphs
);

void freeStrings(
    const uint32_t count, // how many strings
    const string_handle_t* handles,
    core::smart_refctd_ptr<IGPUFence>&& fence // for a deferred free
);
```

Then if you could use your custom DeferredFree functor
https://github.com/Devsh-Graphics-Programming/Nabla/blob/518ef10fc445326c9b9274216bc5800545ab2a8d/include/nbl/video/alloc/SubAllocatedDataBuffer.h#L61
that would also free the string addresses

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,103 @@
...
+
+	// Creates the pipeline layout and the pipeline for rendering text
+	core::smart_refctd_ptr<video::IGPUGraphicsPipeline> createPipeline(
+		video::IGPUObjectFromAssetConverter::SParams& cpu2gpuParams,
+
+		// The data in these buffers is provided by the frustum culling system before calling drawText
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStrings,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringMvps,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringGlyphCount
+	);
```
devshgraphicsprogramming, line 76: look at all the other extensions and utilities, they create default layouts in the constructor whenever feasible (pipeline and DS) and return cached versions

there are always separate functions for creating default layout descriptor sets and DS writing functions

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,103 @@
...
+	uint32_t glyphCount;
+	SGlyphData const* glyphs;
+};
+
+class NBL_API TextRenderer
+{
+public:
+	typedef typename uint32_t size_type;
+
+	TextRenderer(core::smart_refctd_ptr<ILogicalDevice> device, uint32_t maxGlyphCount, uint32_t maxStringCount);
```
devshgraphicsprogramming, line 53: you need `maxGlyphsPerString` as well to size `m_glyphIndexBuffer`

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,103 @@
...
+
+		// The data in these buffers is provided by the frustum culling system before calling drawText
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStrings,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringMvps,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringGlyphCount
+	);
+
+	// - Expansion compute shader: Prefix sum over glyph counts & Output draw args
+	// - Indirect draw for text
+	void drawText(core::smart_refctd_ptr<video::IGPUCommandBuffer> cmdbuf);
```
devshgraphicsprogramming, line 80: split into:
- prefixSumHelper
- drawText
- drawTextIndexed

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,103 @@
...
+private:
+
+private:
+	core::smart_refctd_ptr<ILogicalDevice> m_device;
+	
+	core::smart_refctd_ptr<video::CPropertyPool<core::allocator, uint32_t, StringBoundingBox>> m_stringDataPropertyPool;
+	core::smart_refctd_ptr<video::SubAllocatedDataBufferST<>> m_geomDataBuffer;
+
+	core::smart_refctd_ptr<video::IGPUBuffer> m_visibleStringGlyphCountPrefixSum;
+	core::smart_refctd_ptr<video::IGPUBuffer> m_indirectDrawArgs;
```
devshgraphicsprogramming, line 91: get the user to provide you with those (as DS+indirect buffer arg for the draw function), always

include/nbl/ext/TextRendering/TextRendering.h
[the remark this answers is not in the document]
devshgraphicsprogramming: > uint16_t for the glyph index then?

yeah, but we need to pad to 64bit anyway to be able to use TBO/BufferView (RBG16 formats aren't really supported), and we can't pack into 32bits

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,126 @@
...
+		core::smart_refctd_ptr<IGPUFence>&& fence // for a deferred free
+	);
+
+	// Creates the pipeline using the cached pipeline layout
+	core::smart_refctd_ptr<video::IGPUGraphicsPipeline> createPipeline(video::IGPUObjectFromAssetConverter::SParams& cpu2gpuParams);
+
+	// The data in these buffers is provided by the frustum culling system before calling drawText
+	void updateDescriptorSet(
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStrings,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringMvps,
```
devshgraphicsprogramming, line 92: let user bind their own DS for however they want to feed us the transforms.

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,126 @@
...
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringGlyphCount,
+		core::smart_refctd_ptr<video::IGPUBuffer> prefixSumOutput,
+		core::smart_refctd_ptr<video::IGPUBuffer> indirectDrawArgs
```
devshgraphicsprogramming, line 95: indirect draw is not bound in a DS for drawing.

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,126 @@
...
+
+	void drawTextIndexed(core::smart_refctd_ptr<video::IGPUCommandBuffer> cmdbuf, core::smart_refctd_ptr<video::IGPUBuffer> indirectDrawArgs);
+
+private:
+
+private:
+	core::smart_refctd_ptr<ILogicalDevice> m_device;
+
+	core::smart_refctd_ptr<video::IGPUPipelineLayout> m_pipelineLayout;
+	core::smart_refctd_ptr<video::IGPUDescriptorSet> m_descriptorSet;
```
devshgraphicsprogramming, line 108: only cache (keep global) the default layouts, indexed/non-indexed pipeline and descriptor set with the two pools.

Let user provide the buffer & descriptor set for the prefix summing / visible string list.

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,126 @@
...
+	core::smart_refctd_ptr<glyph_geometry_pool_t> m_geomDataBuffer;
+	core::smart_refctd_ptr<string_pool_t> m_stringDataPropertyPool;
+
+	// - 30 bits global glyph ID
+	// - 2 bits quad indices (0-3 per glyph)
+	core::smart_refctd_ptr<video::IGPUBuffer> m_glyphIndexBuffer;
+
+	core::smart_refctd_ptr<video::IGPUComputePipeline> m_expansionPipeline;
+
+	void prefixSumHelper(core::smart_refctd_ptr<video::IGPUCommandBuffer> cmdbuf);
```
devshgraphicsprogramming, line 119: make it public, I'd expect user to have to do
```cpp
textRenderer->prefixSumHelper(cmdbuf,visibleStringDS,indirectDrawDS);
cmdbuf->pipelineBarrier();
textRenderer->drawText(cmdbuf,visibleStringDS,indirectDrawArgs);
```

include/nbl/ext/TextRendering/TextRendering.h
devshgraphicsprogramming: actually dont cache the pipeline layout or pipeline (make create functions), user will need to override the shaders

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,126 @@
...
+
+	// Creates the pipeline using the cached pipeline layout
+	core::smart_refctd_ptr<video::IGPUGraphicsPipeline> createPipeline(video::IGPUObjectFromAssetConverter::SParams& cpu2gpuParams);
+
+	// The data in these buffers is provided by the frustum culling system before calling drawText
+	void updateDescriptorSet(
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStrings,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringMvps,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringGlyphCount,
+		core::smart_refctd_ptr<video::IGPUBuffer> prefixSumOutput,
```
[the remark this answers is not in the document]
devshgraphicsprogramming: yeah `cumulativeGlyphCount` works better than `prefixSumOutput`, agreed

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,232 @@
...
+
+struct SPixelCoord
+{
+	uint16_t x, y;
+};
+
+class NBL_API FontAtlas
+{
+public:
+	FontAtlas(IGPUQueue* queue, ILogicalDevice* device, const std::string& fontFilename, uint32_t glyphWidth, uint32_t glyphHeight, uint32_t charsPerRow);
```
Erfan-Ahmadi, line 39: would it be usefull to get E_FORMAT for image format here? 
Or are we always working with `asset::EF_R8G8B8A8_UNORM`?

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,232 @@
...
+private:
+	FT_Library library;
+	FT_Face face;
+
+	std::array<SPixelCoord, asciiAtlasCharacterCount> characterAtlasPosition;
```
Erfan-Ahmadi, line 44: Can it be defined mathematically? I think not because of msdfgen, I think we also need this as an `IGPUBufferView` later?

src/nbl/ext/TextRendering/TextRendering.cpp
```diff
@@ -0,0 +1,250 @@
...
+	{
+		uint32_t atlasGlyphX = 0;
+		uint32_t atlasRow = 0;
+
+		for (char k = ' '; k <= '~'; k++)
+		{
+			wchar_t unicode = wchar_t(k);
+			uint32_t glyphIndex = FT_Get_Char_Index(face, unicode);
+
+			// special case for space as it seems to break msdfgen
```
Erfan-Ahmadi, line 49: I'm no font expert but I think space is handled via `advanceX`, no glyph is needed. we can just simply advance the "pen position" and the next glyph would have a higher localXPosition or something.

Please do a little research to check if I didn't miss anything.

src/nbl/ext/TextRendering/TextRendering.cpp
```diff
@@ -0,0 +1,250 @@
...
+			msdfgen::Bitmap<float, 3> msdfMap(msdfWidth, msdfHeight);
+			msdfgen::generateMSDF(msdfMap, shape, 4.0, 1.0, { 4.0, 4.0 });
+
+			uint32_t rowLength = msdfWidth * 4;
+
+			video::IGPUBuffer::SCreationParams bufParams;
+			bufParams.size = rowLength * msdfHeight;
+			bufParams.usage = asset::IBuffer::EUF_TRANSFER_SRC_BIT;
+
+			auto data = device->createBuffer(std::move(bufParams));
```
Erfan-Ahmadi, line 71: TODO: Merge with `image_upload_utils` and use `uploadImageViaStagingBuffer` no need for this buffer anymore, use most explicit the variation where you create the fence, cmdbuffer and submit it yourself so you can record all the atlas copies and then submit at once.

(you would need to access an IUtilities object)

src/nbl/ext/TextRendering/TextRendering.cpp
deprilula28: it's not about layout, this if will skip characters that don't have any glyph associated with them, which would normally include space, but for some reason it looks like space is an empty/invalid glyph instead, which hits an assert in msdfgen when you feed it the space glyph

include/nbl/ext/TextRendering/TextRendering.h
deprilula28: no because not every character gets a glyph

src/nbl/ext/TextRendering/TextRendering.cpp
deprilula28: there is a single submit in the current implementation, im not too familiar with image_upload_utils to know what you mean

Erfan-Ahmadi: That's ok, just put the TODO.

See how I use it  [here](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/57a9f76a17d8659f5df471487c35771a27b098c3/09.ColorSpaceTest/main.cpp#L421)


that's the tool that helps you upload your CPU Buffer to a GPUImage region in chunks. It can also convert the format while copying.

src/nbl/ext/TextRendering/TextRendering.cpp
Erfan-Ahmadi: So you're saying msdfgen should skip and ignore "space" instead of hitting an assert?

Maybe that's expected because space shouldn't have any glyph associated with it?

include/nbl/ext/TextRendering/TextRendering.h
Erfan-Ahmadi: Seems like we went with msdf by default, so nvm

include/nbl/ext/TextRendering/TextRendering.h
Erfan-Ahmadi: Based on the pseudo-code fragment shader we need:
1. `IGPUBufferView/usamplerBuffer` with format `rg16_uint` for glyph pixel sizes.
2. IGPUImage for atlasImage which you already have,

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,232 @@
...
+
+	TextRenderer(core::smart_refctd_ptr<ILogicalDevice>&& device, uint32_t maxGlyphCount, uint32_t maxStringCount, uint32_t maxGlyphsPerString);
+
+	using pool_size_t = uint32_t;
+
+	using glyph_geometry_pool_t = video::SubAllocatedDataBufferST<pool_size_t>;
+	// Data stored as SoA in property pool:
+	// - Glyph offset
+	// - String bounding box
+	using string_pool_t = video::CPropertyPool<core::allocator, pool_size_t, StringBoundingBox>;
```
Erfan-Ahmadi, line 72: We need mvp matrix too

include/nbl/ext/TextRendering/TextRendering.h
Erfan-Ahmadi: Updated the vertex shader codelisting in the issue btw

include/nbl/ext/TextRendering/TextRendering.h
```diff
@@ -0,0 +1,232 @@
...
+
+	// Visible strings are provided by the user's culling system
+	void updateVisibleStringDS(
+		core::smart_refctd_ptr<video::IGPUDescriptorSet> visibleStringDS,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStrings,
+		core::smart_refctd_ptr<video::IGPUBuffer> visibleStringGlyphCounts,
+		core::smart_refctd_ptr<video::IGPUBuffer> cumulativeGlyphCount
+	);
+
+	void prefixSumHelper(
```
Erfan-Ahmadi, line 125: @devshgraphicsprogramming We don't need this anymore since user is appending + prefixSumming at the same time?
Or the signature should change to something like `appendAndPrefixSumHelper(...)`

PR #374, opened 2022-06-30
An alternative to the Indirect Scan

devshgraphicsprogramming:
Console devs can suck it with their `__XB_GdsOrderedCount`

Untitled


devshgraphicsprogramming:
Wonder if a "dynamically uniform" variant could be even faster 🤔 ?

maierfelix:
This is giga chad level

devshgraphicsprogramming:
> This is giga chad level

https://www.youtube.com/watch?v=8B5InWGdims

Issue #250, opened 2021-11-03
Image Format Promotion


Erfan-Ahmadi:
## Description
Format Promotion for image formats not supported in hardware.
Involved: @achalpandeyy @devshgraphicsprogramming @Erfan-Ahmadi 

## Rules for promotion:
1. we dont have a transcoder of block compressed formats, so you cannot promote to a compressed format
2. aspects must be preserved or added (changing depth or stencil into depth+stencil format is ok)
3. channel count must be preserved or increased
4. data range must be preserved or expanded in each channel
5. data precision must be preserved or improved in each channel (this usually means bit-depth when comparing non SRGB formats)
6. when there are multiple matches, you must choose the one with the smallest texel block
P.S. You need to be careful with usages, i.e. no storage image supports SRGB, but we usually get around that by creating UINT views and manually encoding.

## Converstation

> Erfan: Why not promote in the beginning of ICPUImage conversion instead of only promoting the ones used in ImageViews...

> Devsh: But the usage support sometimes depends on the views format

> Achal: Won't this promote the duplicated ones as well?
Ideally you want to put it after the elimateDuplicate calls
What is the point of first promoting then trashing the image?

devshgraphicsprogramming:
OK, so far we've agreed with @Erfan-Ahmadi to extend the `IUtilities` class with being able to upload regions to a texture while converting the format at the same time.

This will give us really easy texture format promotion.

The CPU2GPU converter will just need to decide whether and format to promote the `ICPUImage` and `ICPUImageView`

devshgraphicsprogramming:
but for any work on format promotion to start, @achalpandeyy needs to finish the format usage support reporting from `IPhysicalDevice`

devshgraphicsprogramming:
Supposedly finished
```cpp
        virtual const SFormatImageUsage& getImageFormatUsagesLinear(const asset::E_FORMAT format) = 0;
        virtual const SFormatImageUsage& getImageFormatUsagesOptimal(const asset::E_FORMAT format) = 0;
        virtual const SFormatBufferUsage& getBufferFormatUsages(const asset::E_FORMAT format) = 0;
 ```
 
 @achalpandeyy tell us of any remaining TODOs

achalpandeyy:
- Works on the Vulkan backend
[...]

PR #375, opened 2022-07-01
Add format promotion

deprilula28:
## Description
Format promotion (#250)

## Testing 
Output of every format attached to this comment. Code for that:
```cpp
auto physdev = logicalDevice->getPhysicalDevice();
		for (uint32_t format = 0; format < asset::E_FORMAT::EF_UNKNOWN; format++)
		{
			auto f = static_cast<asset::E_FORMAT>(format);
			for (uint32_t bufUsage = 0; bufUsage < bufUsageFlags.size(); bufUsage++)
			{
				video::IPhysicalDevice::FormatPromotionRequest<asset::IBuffer::E_USAGE_FLAGS> req{ f, bufUsageFlags[bufUsage] };
				auto res = physdev->promoteBufferFormat(req);
				auto fname = nameFormat(f);
				auto resname = nameFormat(res);
				std::cout << "Format " << fname << " Buffer usage " << bufUsageFlags[bufUsage].value << " Promoted to " << resname << "\n";
			}
			for (uint32_t imgUsage = 0; imgUsage < imgUsageFlags.size(); imgUsage++)
			{
				video::IPhysicalDevice::FormatPromotionRequest<asset::IImage::E_USAGE_FLAGS> req{ f, imgUsageFlags[imgUsage] };
				auto resl = physdev->promoteImageFormat(req, asset::IImage::ET_LINEAR);
				auto fname = nameFormat(f);
				auto resname = nameFormat(resl);
				std::cout << "Format " << fname << " Image usage " << imgUsageFlags[imgUsage].value << " (Linear) Promoted to " << resname << "\n";
			}
			for (uint32_t imgUsage = 0; imgUsage < imgUsageFlags.size(); imgUsage++)
			{
				video::IPhysicalDevice::FormatPromotionRequest<asset::IImage::E_USAGE_FLAGS> req{ f, imgUsageFlags[imgUsage] };
				auto rest = physdev->promoteImageFormat(req, asset::IImage::ET_OPTIMAL);
				auto fname = nameFormat(f);
				auto resname = nameFormat(rest);
				std::cout << "Format " << fname << " Image usage " << imgUsageFlags[imgUsage].value << " (Optimal) Promoted to " << resname << "\n";
			}
		}
``` 

## TODO list:

- [ ] PR reviews
- [ ] Proper way of handling precision
- [ ] Tighten up the code a little probably
- [ ] Handle some edge cases


include/nbl/asset/format/EFormat.h
```diff
@@ -484,6 +484,16 @@ const core::vector3du32_SIMD getBlockDimensions()
...
 
+inline uint32_t getMaxChannelBitDepth(asset::E_FORMAT _fmt)
+{
+#include "nbl/asset/format/impl/EFormat_getMaxChannelBitDepth.h"
+}
+template<asset::E_FORMAT _fmt>
+constexpr uint32_t getMaxChannelBitDepth()
+{
+#include "nbl/asset/format/impl/EFormat_getMaxChannelBitDepth.h"
+}
```
devshgraphicsprogramming, line 495: no, don't want such function.

Each channel's bit depth needs to be compared INDIVIDUALLY

include/nbl/asset/format/EFormat.h
```diff
@@ -1738,7 +1746,7 @@ inline value_type getFormatMinValue(E_FORMAT format, uint32_t channel)
 template <typename value_type>
 inline value_type getFormatPrecision(E_FORMAT format, uint32_t channel, value_type value)
 {
-    _NBL_DEBUG_BREAK_IF(isBlockCompressionFormat(format)); //????
+    //_NBL_DEBUG_BREAK_IF(isBlockCompressionFormat(format)); //????
```
devshgraphicsprogramming, line 1749: its hard to reason about the precision of a block compressed format.

This is because usually the compression is done by picking two end points in reduced prevision (RGB565) and then interpolating between them with reduced precision yet again.

So the "real" precision depends on the block's encoded values (to make this func work with block compressed, we'd need to pass a whole texel block data instead of channel's value)

Leave the assert here, never call this function for a block compressed format, and just point blank assume that the precision:
- is FLT_MAX for any channels not present
- 1.f for binary alpha formats (cut out transparency)
- highest it could possibly be (imagine whole tex block has the same value, in DXT this means 8bit SRGB prec, other block formats, no clue)

but don't implement that workaround here!

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+{
+    float min = std::numeric_limits<float>::infinity();
+
+    for (uint32_t i = 0; i < channels; i++)
+    {
+        min = std::min(min, asset::getFormatPrecision<float>(f, i, 0.0));
+    }
+
+    return min;
+}
```
devshgraphicsprogramming, line 116: no, compare each channel separately, only when precision is equal or greater for EACH channel invidually you can promote.

You're being overly conservative

include/nbl/asset/format/EFormat.h
deprilula28: oops, i didnt mean to commit that, i was just getting annoyed at the debug breaking there

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+    formatBufUsg.isInitialized = 1;
+    formatBufUsg.vertexAttribute = usages.hasFlags(asset::IBuffer::EUF_VERTEX_BUFFER_BIT);
+    formatBufUsg.bufferView = usages.hasFlags(asset::IBuffer::EUF_UNIFORM_TEXEL_BUFFER_BIT);
+    formatBufUsg.storageBufferView = usages.hasFlags(asset::IBuffer::EUF_STORAGE_TEXEL_BUFFER_BIT);
+    // Special flags for these?
+    formatBufUsg.accelerationStructureVertex = usages.hasFlags(asset::IBuffer::EUF_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT);
+    formatBufUsg.storageBufferViewAtomic = usages.hasFlags(asset::IBuffer::EUF_STORAGE_TEXEL_BUFFER_BIT);
+
+    return formatBufUsg;
+}
```
devshgraphicsprogramming, line 179: just make a constructor for `SFormatBufferUsage` that take the bitflag

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+{
+    auto buf_cache = this->m_formatPromotionCache.buffers;
+    auto cached = buf_cache.find(req);
+    if (cached != buf_cache.end())
+        return cached->second;
+
+    auto srcUsages = convBufferUsage(req.usages);
+
+    // Cache valid formats per usage?
+    core::bitflag<asset::E_FORMAT> validFormats;
```
devshgraphicsprogramming, line 191: nooo, `E_FORMAT` isnt a flag enum!!!!

you need a `unordered_set` !!!!

deprilula28: oops, thats a pretty bad one yeah

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+    default:
+        return asset::E_FORMAT::EF_UNKNOWN;
+    }
+    auto cached = cache.find(req);
+    if (cached != cache.end())
+        return cached->second;
+    auto srcUsages = convImageUsage(req.usages);
+
+    // Cache valid formats per usage?
+    core::bitflag<asset::E_FORMAT> validFormats;
```
devshgraphicsprogramming, line 248: https://github.com/Devsh-Graphics-Programming/Nabla/pull/375/files#r911994441

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+            break;
+        case asset::IImage::E_TILING::ET_OPTIMAL:
+            formatUsages = this->getImageFormatUsagesOptimal(f);
+            break;
+        default:
+            return asset::E_FORMAT::EF_UNKNOWN;
+        }
+        auto commonUsages = srcUsages & formatUsages;
+        commonUsages.log2MaxSamples = srcUsages.log2MaxSamples; // TODO: Handle sample counts
+        if (commonUsages == srcUsages)
```
devshgraphicsprogramming, line 267: implement a `<` operator for the usage struct

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+    auto srcUsages = convBufferUsage(req.usages);
+
+    // Cache valid formats per usage?
+    core::bitflag<asset::E_FORMAT> validFormats;
+
+    for (uint32_t format = 0; format < asset::E_FORMAT::EF_UNKNOWN; format++)
+    {
+        auto f = static_cast<asset::E_FORMAT>(format);
+        auto formatUsages = this->getBufferFormatUsages(f);
+        if ((srcUsages & formatUsages) == srcUsages)
```
devshgraphicsprogramming, line 197: implement a `<` operator for the usage struct

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+        {
+            validFormats |= f;
+        }
+    }
+
+    //if (validFormats.hasFlags(req.originalFormat)) return req.originalFormat;
+
+    return narrowDownFormatPromotion(validFormats, req.originalFormat);
```
devshgraphicsprogramming, line 205: I don't see you inserting into the cache!

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+        commonUsages.log2MaxSamples = srcUsages.log2MaxSamples; // TODO: Handle sample counts
+        if (commonUsages == srcUsages)
+        {
+            validFormats |= f;
+        }
+    }
+
+    //if (validFormats.hasFlags(req.originalFormat)) return req.originalFormat;
+
+    return narrowDownFormatPromotion(validFormats, req.originalFormat);
```
devshgraphicsprogramming, line 275: I don't see you inserting into the cache!

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+// - Data precision: Preserved or improved (per channel)
+//     - Bit depth when comparing non srgb
+// If there are multiple matches: Pick smallest texel block
+asset::E_FORMAT narrowDownFormatPromotion(core::bitflag<asset::E_FORMAT> validFormats, asset::E_FORMAT srcFormat)
+{
+    asset::E_FORMAT smallestTexelBlock = asset::E_FORMAT::EF_UNKNOWN;
+    uint32_t smallestTexelBlockSize = -1;
+
+    auto srcChannels = asset::getFormatChannelCount(srcFormat);
+    if (!srcChannels) return asset::EF_UNKNOWN;
```
devshgraphicsprogramming, line 132: assert this, this will never happen

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
 
+inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FORMAT _fmt)
+{
+    core::bitflag<asset::IImage::E_ASPECT_FLAGS> flags;
+    bool depthOrStencil = asset::isDepthOrStencilFormat(_fmt);
+    bool stencilOnly = asset::isStencilOnlyFormat(_fmt);
+    bool depthOnly = asset::isDepthOnlyFormat(_fmt);
+    if (depthOrStencil || depthOnly) flags |= asset::IImage::EAF_DEPTH_BIT;
+    if (depthOrStencil || stencilOnly) flags |= asset::IImage::EAF_STENCIL_BIT;
+    if (!depthOrStencil && !stencilOnly && !depthOnly) flags |= asset::IImage::EAF_COLOR_BIT;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: ah ok, I see.

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+// If there are multiple matches: Pick smallest texel block
+asset::E_FORMAT narrowDownFormatPromotion(core::bitflag<asset::E_FORMAT> validFormats, asset::E_FORMAT srcFormat)
+{
+    asset::E_FORMAT smallestTexelBlock = asset::E_FORMAT::EF_UNKNOWN;
+    uint32_t smallestTexelBlockSize = -1;
+
+    auto srcChannels = asset::getFormatChannelCount(srcFormat);
+    if (!srcChannels) return asset::EF_UNKNOWN;
+    auto srcAspects = getImageAspects(srcFormat);
+    auto srcBitDepth = asset::getMaxChannelBitDepth(srcFormat);
```
devshgraphicsprogramming, line 134: you don't need to compare bit depths, only precisions

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+        }
+
+        auto dstChannels = asset::getFormatChannelCount(f);
+        // Verify if promotion is valid from srcFormat -> format
+        bool promotionValid;
+        promotionValid = asset::isBlockCompressionFormat(f) ? f == srcFormat : true; // Can't transcode to compressed formats
+        promotionValid = promotionValid && dstChannels >= srcChannels; // Channel count
+        promotionValid = promotionValid && (getImageAspects(f) & srcAspects).value == srcAspects.value; // Aspects
+        promotionValid = promotionValid && asset::getMaxChannelBitDepth(f) >= srcBitDepth; // Data range
+        promotionValid = promotionValid && getFullPrecisionValue(f, dstChannels) >= srcPrecision; // Precision
```
devshgraphicsprogramming, line 154: legibility, use
```cpp
if (some_condition)
   continue;
```
for each rule separately

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+    auto srcPrecision = getFullPrecisionValue(srcFormat, srcChannels);
+
+    // Better way to iterate the bitflags here?
+    for (uint32_t format = 0; format < asset::E_FORMAT::EF_UNKNOWN; format++)
+    {
+        auto f = static_cast<asset::E_FORMAT>(format);
+        if (!validFormats.hasFlags(f))
+        {
+            continue;
+        }
```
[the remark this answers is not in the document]
devshgraphicsprogramming: assert that `format!=srcFormat`

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+
+// Rules for promotion:
+// - Cannot convert to block format
+// - Aspects: Preserved or added
+// - Channel count: Preserved or increased
+// - Data range: Preserved or increased (per channel)
+// - Data precision: Preserved or improved (per channel)
+//     - Bit depth when comparing non srgb
+// If there are multiple matches: Pick smallest texel block
+asset::E_FORMAT narrowDownFormatPromotion(core::bitflag<asset::E_FORMAT> validFormats, asset::E_FORMAT srcFormat)
```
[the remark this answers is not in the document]
devshgraphicsprogramming: add a condition that the `srcFormat` cannot be in `validFormats` you should never promote/change format if your format could be used

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+}
+
+asset::E_FORMAT IPhysicalDevice::promoteBufferFormat(const FormatPromotionRequest<asset::IBuffer::E_USAGE_FLAGS> req)
+{
+    auto buf_cache = this->m_formatPromotionCache.buffers;
+    auto cached = buf_cache.find(req);
+    if (cached != buf_cache.end())
+        return cached->second;
+
+    auto srcUsages = convBufferUsage(req.usages);
```
devshgraphicsprogramming, line 188: do an early out https://github.com/Devsh-Graphics-Programming/Nabla/pull/375/files#r912026997, if you can use the format right away
```cpp
if (req.canUseFormat(getBufferFormatUsages(srcFormat))
{
   buf_cache.insert(cached,{req,srcFormat});
   return srcFormat;
}
```

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+
+    // Cache valid formats per usage?
+    core::bitflag<asset::E_FORMAT> validFormats;
+
+    for (uint32_t format = 0; format < asset::E_FORMAT::EF_UNKNOWN; format++)
+    {
+        auto f = static_cast<asset::E_FORMAT>(format);
```
devshgraphicsprogramming, line 195: skip format equal to self, should have been handled earlier.

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+    case asset::IImage::E_TILING::ET_OPTIMAL:
+        cache = this->m_formatPromotionCache.optimalTilingImages;
+        break;
+    default:
+        return asset::E_FORMAT::EF_UNKNOWN;
+    }
+    auto cached = cache.find(req);
+    if (cached != cache.end())
+        return cached->second;
+    auto srcUsages = convImageUsage(req.usages);
```
devshgraphicsprogramming, line 245: do an early out https://github.com/Devsh-Graphics-Programming/Nabla/pull/375/files#r912026997, if you can use the format right away
```cpp
if (req.canUseFormat(getBufferFormatUsages(srcFormat))
{
   cache.insert(cached,{req,srcFormat});
   return srcFormat;
}
```

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,189 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+    // Special flags for these?
+    formatBufUsg.accelerationStructureVertex = usages.hasFlags(asset::IBuffer::EUF_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT);
+    formatBufUsg.storageBufferViewAtomic = usages.hasFlags(asset::IBuffer::EUF_STORAGE_TEXEL_BUFFER_BIT);
+
+    return formatBufUsg;
+}
+
+asset::E_FORMAT IPhysicalDevice::promoteBufferFormat(const FormatPromotionRequest<asset::IBuffer::E_USAGE_FLAGS> req)
+{
+    auto buf_cache = this->m_formatPromotionCache.buffers;
```
devshgraphicsprogramming, line 183: this needs to be `auto&`

src/nbl/video/IPhysicalDevice.cpp
devshgraphicsprogramming: so
```
// Can't transcode to compressed or planar formats
if (asset::isBlockCompressionFormat(f))
   continue;
if (asset::isPlanarFormat(f))
   continue;

// can use the format
if (req.canUseFormat(getBufferFormatUsages(f))
   continue;
   
// cannot have less channels
const dstChannels = getFormatChannelCount(f);
if (dstChannels<srcChannels)
   break;
// check the properties of the used channels
{
   auto c=0;
   while (c<srcChannels)
   {
      // none can have lower precision than source
      if (getFormatPrecision(f,c,0.f)<srcPrecision[c])
         break;
      // none can have less range than source
      if (getFormatMinValue(f,c)>srcMin[c] || getFormatMaxValue(f,c)<srcMax[c])
         break;
      c++;
   }
   if (c!=srcChannels)
      continue;
}
// same or more apsects
if ((getImageAspects(f) & srcAspects).value != srcAspects.value)
   continue;
```

devshgraphicsprogramming:
We should also care about the data types i.e.
- is Integer Format?
- is Unsigned Format?

My intuition is telling me this, because otherwise not only do you need to change the ImageView or Image format and convert on upload but also change the sampler type (`i`, `u` or no prefix) in the shader, which is too much work.

Same goes for vertex attributes.

deprilula28:
Here's how the promotion looks for all the formats so far: 
[new_promotion.txt](https://github.com/Devsh-Graphics-Programming/Nabla/files/9031472/new_promotion.txt)

devshgraphicsprogramming:
> Here's how the promotion looks for all the formats so far: [new_promotion.txt](https://github.com/Devsh-Graphics-Programming/Nabla/files/9031472/new_promotion.txt)

A) do not use it on PLANAR formats (we don't really support those)
B) I'm a bit concerned about 8bit formats getting promoted to 32bit
C) are you deleting things which could not be promoted? I don't see you testing every usage
D) how are you "tie-breaking", that is if multiple formats were found suitable and there's no "smallest" ?

include/nbl/video/IPhysicalDevice.h
```diff
@@ -328,6 +352,24 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
...
+                sampledImage(usages.hasFlags(asset::IImage::EUF_SAMPLED_BIT)),
+                storageImage(usages.hasFlags(asset::IImage::EUF_STORAGE_BIT)),
+                transferSrc(usages.hasFlags(asset::IImage::EUF_TRANSFER_SRC_BIT)),
+                transferDst(usages.hasFlags(asset::IImage::EUF_TRANSFER_DST_BIT)),
+                attachment((usages& core::bitflag<asset::IImage::E_USAGE_FLAGS>(asset::IImage::EUF_COLOR_ATTACHMENT_BIT | asset::IImage::EUF_DEPTH_STENCIL_ATTACHMENT_BIT)).value != 0),
+                // Other way of getting these flags?
+                blitSrc(usages.hasFlags(asset::IImage::EUF_TRANSFER_SRC_BIT)),
+                blitDst(usages.hasFlags(asset::IImage::EUF_TRANSFER_DST_BIT)),
+                storageImageAtomic(usages.hasFlags(asset::IImage::EUF_STORAGE_BIT)),
+                attachmentBlend((usages& core::bitflag<asset::IImage::E_USAGE_FLAGS>(asset::IImage::EUF_COLOR_ATTACHMENT_BIT | asset::IImage::EUF_DEPTH_STENCIL_ATTACHMENT_BIT)).value != 0)
```
devshgraphicsprogramming, line 370: only check color attachment bit, only colors can blend...

include/nbl/video/IPhysicalDevice.h
```diff
@@ -267,6 +267,20 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            {}
+
+            SFormatBufferUsage(core::bitflag<asset::IBuffer::E_USAGE_FLAGS> usages)
+                : isInitialized(1),
+                vertexAttribute(usages.hasFlags(asset::IBuffer::EUF_VERTEX_BUFFER_BIT)),
+                bufferView(usages.hasFlags(asset::IBuffer::EUF_UNIFORM_TEXEL_BUFFER_BIT)),
+                storageBufferView(usages.hasFlags(asset::IBuffer::EUF_STORAGE_TEXEL_BUFFER_BIT)),
+                // Other way of getting these flags?
+                accelerationStructureVertex(usages.hasFlags(asset::IBuffer::EUF_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT)),
+                storageBufferViewAtomic(usages.hasFlags(asset::IBuffer::EUF_STORAGE_TEXEL_BUFFER_BIT))
```
devshgraphicsprogramming, line 281: ok, this is gonna backfire, because every time you'll declare a `EUF_STORAGE_TEXEL_BUFFER_BIT` usage, it will think you're gonna do atomics, and then promote everything to R32_UINT or R32_INT

you need a `const bool useForAtomic` extra parameter for the ctor, and do an && with the `storageBufferView`

include/nbl/video/IPhysicalDevice.h
devshgraphicsprogramming: Ok same thing as https://github.com/Devsh-Graphics-Programming/Nabla/pull/375/commits/eda7c31cfebb608c84b3eec345f5ba89499ab82d#r912939223

Leave the things you cannot deduce 100% certain as false, let user patch them up later.

include/nbl/video/IPhysicalDevice.h
```diff
@@ -267,6 +267,20 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
...
+                : isInitialized(0)
+            {}
+
+            SFormatBufferUsage(core::bitflag<asset::IBuffer::E_USAGE_FLAGS> usages)
+                : isInitialized(1),
+                vertexAttribute(usages.hasFlags(asset::IBuffer::EUF_VERTEX_BUFFER_BIT)),
+                bufferView(usages.hasFlags(asset::IBuffer::EUF_UNIFORM_TEXEL_BUFFER_BIT)),
+                storageBufferView(usages.hasFlags(asset::IBuffer::EUF_STORAGE_TEXEL_BUFFER_BIT)),
+                // Other way of getting these flags?
+                accelerationStructureVertex(usages.hasFlags(asset::IBuffer::EUF_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT)),
```
devshgraphicsprogramming, line 280: I think that's correct @Erfan-Ahmadi, anything wrong?

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -103,58 +103,226 @@ inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FOR
...
+    case asset::EF_ETC2_R8G8B8A8_UNORM_BLOCK:
+    case asset::EF_ETC2_R8G8B8A8_SRGB_BLOCK:
+        rcpUnit = 1.0 / 31.0;
+        break;
+    case asset::EF_BC6H_UFLOAT_BLOCK:
+    case asset::EF_BC6H_SFLOAT_BLOCK:
+    {
+        // BC6 isn't really FP16, so this is an over-estimation
+        float f = 0.0;
+        uint16_t f16 = core::Float16Compressor::compress(f);
```
devshgraphicsprogramming, line 229: IEEE754 0.f always yields 0 bitpattern

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -103,58 +103,226 @@ inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FOR
...
+        uint16_t dir = core::Float16Compressor::compress(2.f * (f + 1.f));
```
devshgraphicsprogramming, line 230: IEEE754 **positive** floats are ordered same way as their bitrep, you can use `0x0001u`, unless you want to use minimum normalized value to account for "denormalized flush to 0 on GPUs"

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -103,58 +103,226 @@ inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FOR
...
+    case asset::EF_PVRTC1_4BPP_UNORM_BLOCK_IMG:
+    case asset::EF_PVRTC2_2BPP_UNORM_BLOCK_IMG:
+    case asset::EF_PVRTC2_4BPP_UNORM_BLOCK_IMG:
+    case asset::EF_PVRTC1_2BPP_SRGB_BLOCK_IMG:
+    case asset::EF_PVRTC1_4BPP_SRGB_BLOCK_IMG:
+    case asset::EF_PVRTC2_2BPP_SRGB_BLOCK_IMG:
+    case asset::EF_PVRTC2_4BPP_SRGB_BLOCK_IMG:
+        // TODO
+        // I couldn't find any documentation on how this works
+        return 0.0;
```
devshgraphicsprogramming, line 243: well by laws of entropy the precision cannot be higher than that of associated 8bit RGB or SRGB format

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -103,58 +103,226 @@ inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FOR
...
+    case asset::EF_ASTC_10x6_SRGB_BLOCK:
+    case asset::EF_ASTC_10x8_UNORM_BLOCK:
+    case asset::EF_ASTC_10x8_SRGB_BLOCK:
+    case asset::EF_ASTC_10x10_UNORM_BLOCK:
+    case asset::EF_ASTC_10x10_SRGB_BLOCK:
+    case asset::EF_ASTC_12x10_UNORM_BLOCK:
+    case asset::EF_ASTC_12x10_SRGB_BLOCK:
+    case asset::EF_ASTC_12x12_UNORM_BLOCK:
+    case asset::EF_ASTC_12x12_SRGB_BLOCK:
+        // (All of these could be using HDR. Take extra flag to assume FP16 precision?)
```
devshgraphicsprogramming, line 207: I guess so, but I thought the HDR ASTC formats were different enum values

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -103,58 +103,226 @@ inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FOR
...
 
-    for (uint32_t i = 0; i < channels; i++)
+    if (isSRGBFormat(format))
     {
-        min = std::min(min, asset::getFormatPrecision<float>(f, i, 0.0));
+        return core::srgb2lin(0.0 + rcpUnit) - core::srgb2lin(0.0);
     }
 
-    return min;
+    return rcpUnit;
```
devshgraphicsprogramming, line 251: I think you're messing up by reporting only the precision of the end-points (as seen in your BC1-5 cases)

Remember that the endpoint itself can be interpolated with interpolants.

Imagine an endpoint at `(0,0,0)` and `(nextUlpRed,nextUlpGreen,nextUlpBlue)`

Despite the fact that the endpoints can only be placed at 1/32 or 1/64 intervals, a single pixel has an interpolant which can interpolate the endpoints (in BC, IIRC its 2 bits, so your max precision is 1/128 or 1/256)

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -103,58 +103,226 @@ inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FOR
...
+    }
 
     // Better way to iterate the bitflags here?
     for (uint32_t format = 0; format < asset::E_FORMAT::EF_UNKNOWN; format++)
     {
         auto f = static_cast<asset::E_FORMAT>(format);
-        if (!validFormats.hasFlags(f))
-        {
+        if (!validFormats.contains(f))
             continue;
```
devshgraphicsprogramming, line 288: no, you shall `assert` that

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -103,58 +103,226 @@ inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FOR
...
+
+        {
+            auto c = 0;
+            while (c < srcChannels)
+            {
+                // Can't have less precision (higher precision value from getFormatPrecision) than source
+                if (asset::getFormatPrecision(f, c, 0.f) > srcPrecision[c])
+                    break;
+                // Can't have less range than source
+                if (asset::getFormatMinValue<float>(f, c) > srcMinVal[c] || asset::getFormatMaxValue<float>(f, c) < srcMaxVal[c])
```
devshgraphicsprogramming, line 309: `<double>` otherwise you'll mess up on the full 32-bit integer format comparisons

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -103,58 +103,226 @@ inline core::bitflag<asset::IImage::E_ASPECT_FLAGS> getImageAspects(asset::E_FOR
...
+// - Cannot convert to block or planar format
 // - Aspects: Preserved or added
 // - Channel count: Preserved or increased
 // - Data range: Preserved or increased (per channel)
 // - Data precision: Preserved or improved (per channel)
 //     - Bit depth when comparing non srgb
 // If there are multiple matches: Pick smallest texel block
-asset::E_FORMAT narrowDownFormatPromotion(core::bitflag<asset::E_FORMAT> validFormats, asset::E_FORMAT srcFormat)
+// srcFormat can't be in validFormats (no promotion should be made if the format itself is valid)
+asset::E_FORMAT narrowDownFormatPromotion(core::unordered_set<asset::E_FORMAT> validFormats, asset::E_FORMAT srcFormat)
```
devshgraphicsprogramming, line 263: take `const & validFormats`

include/nbl/video/IPhysicalDevice.h
Erfan-Ahmadi: It is fine,
although I will change all these structs later

include/nbl/video/IPhysicalDevice.h
```diff
@@ -333,6 +333,17 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
                     (storageBufferViewAtomic == other.storageBufferViewAtomic) &&
                     (accelerationStructureVertex == other.accelerationStructureVertex);
             }
+
+            // Probably not a great way to do this?
+            inline uint32_t value() const
```
devshgraphicsprogramming, line 338: declare `std::hash` overload

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,375 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+        if (texelBlockSize > smallestTexelBlockSize) {
+            continue;
+        }
+
+        if (texelBlockSize == smallestTexelBlockSize)
+        {
+            // Tie-breaking rules:
+            // - Precision
+            // - Value range
+            if (!higherPrecisionOrValueRange(smallestTexelBlock, f, srcChannels))
```
devshgraphicsprogramming, line 371: prefer current found fit if:
- current fit RGBA vs. BGRA matches the input format
- current fit has higher precision (but only in the source channels that are used)
- current fit has higher range (but only in the source channels that are used)

in this exact order

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,375 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+                continue;
+        }
+
+        smallestTexelBlockSize = texelBlockSize;
+        smallestTexelBlock = f;
+    }
+
+    return smallestTexelBlock;
```
devshgraphicsprogramming, line 379: `assert(smallestTexelBlock!=EF_UNKNOWN)`, do an early `if (validFormats.empty()) return EF_UNKNOWN;` at the top

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,375 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+            continue;
+    
+        // Can't promote between int and normalized/float/scaled formats
+        if (srcIntFormat != asset::isIntegerFormat(f))
+            continue;
+
+        // Can't have less channels
+        auto dstChannels = asset::getFormatChannelCount(f);
+        if (dstChannels < srcChannels)
+            continue;
```
devshgraphicsprogramming, line 336: could you hoist these checks and perform them before anything gets put into the `validFormats` set?

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,375 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+                c++;
+            }
+            // If we hit a break in the while loop
+            if (c != srcChannels)
+                continue;
+        }
+
+        // Can't have less aspects
+        if (!getImageAspects(f).hasFlags(srcAspects))
+            continue;
```
devshgraphicsprogramming, line 358: could you hoist these checks and perform them before anything gets put into the `validFormats` set?

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,375 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+
+        smallestTexelBlockSize = texelBlockSize;
+        smallestTexelBlock = f;
+    }
+
+    return smallestTexelBlock;
+}
+
+asset::E_FORMAT IPhysicalDevice::promoteBufferFormat(const FormatPromotionRequest<video::IPhysicalDevice::SFormatBufferUsage> req)
+{
```
devshgraphicsprogramming, line 383: before you even tap the cache, assert that the format is not block compressed, planar or has non-color aspects

All of those make no sense for a Buffer

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -253,14 +253,39 @@ float getBcFormatMaxPrecision(asset::E_FORMAT format, uint32_t channel)
...
+
+// Returns true if a has higher precision & value range than b
 bool higherPrecisionOrValueRange(asset::E_FORMAT a, asset::E_FORMAT b, uint32_t srcChannels)
 {
     auto c = 0;
     while (c < srcChannels)
     {
         // break if a has less precision (higher precision value from getFormatPrecision) than b
-        if (asset::getFormatPrecision(a, c, 0.f) > asset::getFormatPrecision(b, c, 0.f))
+        if (getMinFormatPrecision(a, c) > getMinFormatPrecision(b, c))
```
devshgraphicsprogramming: check precision at min, 0 and max

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -272,22 +272,38 @@ double getFormatPrecisionAt(asset::E_FORMAT format, uint32_t channel, double val
...
+    if (a == b)
+        return false;
+    // Can't transcode to BC or planar
+    if (asset::isBlockCompressionFormat(a))
+        return false;
+    if (asset::isPlanarFormat(a))
+        return false;
+    // Can't promote between int and normalized/float/scaled formats
+    if (asset::isIntegerFormat(a) != srcIntFormat)
+        return false;
```
devshgraphicsprogramming, line 288: also prevent signed becoming unsigned and vice versa

include/nbl/video/IPhysicalDevice.h
```diff
@@ -460,6 +518,28 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
...
+        {
+            asset::E_FORMAT originalFormat = asset::EF_UNKNOWN;
+            FORMAT_USAGE usages;
+
+            // pack into 64bit for easy hashing 
+            uint64_t operator()(const FormatPromotionRequest<FORMAT_USAGE>& r) const
+            {
+                uint64_t msb = uint64_t(std::hash<FORMAT_USAGE>()(r.usages));
+                return (msb << 32u) | r.originalFormat;
+            }
```
devshgraphicsprogramming, line 532: move to a custom hasher

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,440 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+    // Check value range
+    for (uint32_t c = 0; c < srcChannels; c++)
+    {
+        double mina = asset::getFormatMinValue<double>(a, c),
+            minb = asset::getFormatMinValue<double>(b, c),
+            maxa = asset::getFormatMaxValue<double>(a, c),
+            maxb = asset::getFormatMaxValue<double>(b, c);
+
+        // return true if a has better range than b
+        if (mina < minb && maxa > maxb)
```
devshgraphicsprogramming, line 354: use `<=` and >=`

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,440 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+            return true;
+    }
```
devshgraphicsprogramming, line 356: ok, what we want to avoid is using a format which has less "quantifiable values" in the source format's range,

so we should probably take the one that has the smallest "waste" given source input/output ranges
```
double waste = 0.0;
for (auto c=0; c<srcChannelCount; c++)
{
   // you can assert that the ranges of the dest are greater
   waste += (srcMin[c]-dstMin[c])/dstPrecision(srcMin[c],c);
   waste += (dstMax[c]-srcMax[c])/dstPrecision(srcMax[c],c);
}
```

Then tie break using the format with the lowest "waste"

src/nbl/video/IPhysicalDevice.cpp
```diff
@@ -90,4 +90,440 @@ bool IPhysicalDevice::validateLogicalDeviceCreation(const ILogicalDevice::SCreat
...
+    double srcMinVal[4];
+    double srcMaxVal[4];
+    for (uint32_t channel = 0; channel < srcChannels; channel++)
+    {
+        srcMinVal[channel] = asset::getFormatMinValue<double>(srcFormat, channel);
+        srcMaxVal[channel] = asset::getFormatMaxValue<double>(srcFormat, channel);
+    }
+
+    // Cache valid formats per usage?
+    core::unordered_set<asset::E_FORMAT> validFormats;
```
devshgraphicsprogramming, line 442: not a bad idea!

PR #393, opened 2022-07-20
GLSL define injection update for limits & features

deprilula28:
## Description
Described in issue #391 

## TODO list:
- [ ] Figure out how to get every GLSL extension which shaderc/glslang could use given device enabled features(?)


deprilula28:
wrong target branch

Issue #373, opened 2022-06-30
Command Pool Reset


Erfan-Ahmadi:
- [x] Add Atomic Reset Counter to Command Pool + cached counter in CommandBuffer
- [x] CommandBuffer::CheckForCommandPoolReset function

Command buffer Pool's atomic counter and if it was higher than it's cached value, it will change state to the "initial state"

Sane people don't reset command pool while recording command buffer so no need to CheckForCommandPoolReset in every function for now, but add later

Command Buffer

```c
public:
bool reset(flag)
{
  if (!canReset etc)
    return false;
  if(flag.hasValue(RELEASE_RESOURCES_BIT))
    releaseResourcesBackToPool(m_pool);
  resetCommon();
  return true;
}

protected:

uint32_t cachedPoolResetStamp = 0u;
inline bool checkForCommandPoolReset()
{
  if (m_pool->getResetCounter()<=cachedPoolResetStamp)
    return false;

  // no need to releaseResourcesBackToPool(m_pool); because pool already has claimed it's resources
  resetCommon();
  return true;
}
inline virtual void resetCommon()
{
  cachedPoolResetStamp = m_pool->getResetCounter();
}
```

devshgraphicsprogramming:
still not implemented https://github.com/Devsh-Graphics-Programming/Nabla/pull/345/files#r1040787374

devshgraphicsprogramming:
done

PR #408, opened 2022-08-04
Command pool

deprilula28:
## Description
Adds a command pool reset function (#373), i.e. ECF_RESET_COMMAND_BUFFER_BIT is not set when creating the command pool.

Implemented by using a reset counter that increases on each reset, then when `checkForCommandPoolReset` is called on command buffer (currently done within `begin`), that counter is checked against the counter of the command buffer, which will cause an "internal reset" if it is higher.

"Internal reset" consists of releasing references captured in VK & GL, and clearing commands on GL.

## Testing 
Tested using a ECF_NONE command pool for leaking references & commands on GL & VK.
- [ ] Test reset command buffer just after resetting the parent command pool (and do vice versa)
## TODO list:

- [ ] Reviews


include/nbl/video/IGPUCommandPool.h
```diff
@@ -31,11 +35,21 @@ class NBL_API IGPUCommandPool : public core::IReferenceCounted, public IBackendO
         // Vulkan: const VkCommandPool*
         virtual const void* getNativeHandle() const = 0;
 
+        virtual void reset()
+        {
+            assert(!getCreationFlags().hasFlags(ECF_RESET_COMMAND_BUFFER_BIT));
```
Erfan-Ahmadi, line 40: You can have `ECF_RESET_COMMAND_BUFFER_BIT` and reset command pool at the same time.
unless it's an internal reference holding reason I'm not aware of.
We should be able to both reset command buffers indvidually(if `ECF_RESET_COMMAND_BUFFER_BIT`) or reset the whole command pool.

include/nbl/video/IGPUCommandPool.h
```diff
@@ -31,11 +35,21 @@ class NBL_API IGPUCommandPool : public core::IReferenceCounted, public IBackendO
...
+        }
+
     protected:
         virtual ~IGPUCommandPool() = default;
 
         core::bitflag<E_CREATE_FLAGS> m_flags;
         uint32_t m_familyIx;
+        std::atomic_uint32_t m_resetCount;
+
+        virtual uint32_t getResetCounter() { return m_resetCount.load(); }
```
Erfan-Ahmadi, line 52: call `m_resetCount.load()` as late as possible when you need it. although this may be inlined and called when needed but let's be logical and return `std::atomic_uint32_t`
Also as I mentioned in the discord thread we could have even had normal uint instead of atomic. but since we decided to have atomics, we should return the counter as atomic and let user decide when to `.load` it

include/nbl/video/IGPUCommandBuffer.h
```diff
@@ -227,6 +227,25 @@ class NBL_API IGPUCommandBuffer :
             for (uint32_t i = _first + 1u; i < IGPUPipelineLayout::DESCRIPTOR_SET_COUNT; ++i)
                 _destPplnLayouts[i] = nullptr;
     }
+
+    uint32_t m_resetCount = 0;
```
Erfan-Ahmadi, line 231: `m_resetCounter`

src/nbl/video/COpenGLCommandPool.h
```diff
@@ -39,7 +39,13 @@ class COpenGLCommandPool final : public IGPUCommandPool
             std::unique_lock<std::mutex> lk(mutex);
             mempool.free_n<T>(const_cast<T*>(ptr), n);
         }
-        
+
+        void reset() override
+        {
+            // TODO reset some memory here?
```
Erfan-Ahmadi, line 45: It should act as if all the command buffer allocated from this pool were reset with `E_RESET_FLAGS::ERF_RELEASE_RESOURCES_BIT`

src/nbl/video/CVulkanCommandPool.cpp
```diff
@@ -12,6 +12,15 @@ CVulkanCommandPool::~CVulkanCommandPool()
     vk->vk.vkDestroyCommandPool(vulkanDevice->getInternalObject(), m_vkCommandPool, nullptr);
 }
 
+void CVulkanCommandPool::reset()
+{
+	const CVulkanLogicalDevice* vulkanDevice = static_cast<const CVulkanLogicalDevice*>(getOriginDevice());
+	auto* vk = vulkanDevice->getFunctionTable();
+	// TODO work out: always use release resources flag?
+	vk->vk.vkResetCommandPool(vulkanDevice->getInternalObject(), m_vkCommandPool, VK_COMMAND_POOL_RESET_RELEASE_RESOURCES_BIT);
```
Erfan-Ahmadi, line 20: Ok @deprilula28 I see that you got some enums a bit mixed up here. Let me explani why you don't want the `VK_COMMAND_POOL_RESET_RELEASE_RESOURCES_BIT` here.
1. CommandPool allocates from system memory
2. CommandBuffer allocates from command pool

So reseting the command pool with that flag recycles all of the resources from the command pool back to the system.
But reseting the command buffer with `VK_COMMAND_BUFFER_RESET_RELEASE_RESOURCES_BIT` specifies that most or all memory resources currently owned by the command buffer should be returned to the parent command pool. 

By this definitions reseting a command pool with no additional flags is like reseting every one of the command buffers with `VK_COMMAND_BUFFER_RESET_RELEASE_RESOURCES_BIT`, and that's what we want in the reset command pool. so that each of the  command buffer return their resources back to parent pool

Notice `VK_COMMAND_BUFFER_RESET_RELEASE_RESOURCES_BIT` is different from `VK_COMMAND_POOL_RESET_RELEASE_RESOURCES_BIT`

src/nbl/video/CVulkanCommandBuffer.h
```diff
@@ -39,8 +39,25 @@ class CVulkanCommandBuffer : public IGPUCommandBuffer
...
+    bool begin(core::bitflag<E_USAGE> recordingFlags, const SInheritanceInfo* inheritanceInfo = nullptr) override
+    {
+        // Resources are freed when the command buffer is reset, so:
+        // - When ECF_RESET_COMMAND_BUFFER_BIT is set on the command pool and we are resetting the command buffer
+        //     - Command buffer is reset when calling begin or reset (handled elsewhere)
+        // - When ECF_RESET_COMMAND_BUFFER_BIT is NOT set on the command pool, and the pool has been reset
+        if (m_cmdpool->getCreationFlags().hasFlags(video::IGPUCommandPool::ECF_RESET_COMMAND_BUFFER_BIT))
+        {
+            // "reset" the command buffer by removing existing resources
+            if (m_state != ES_PENDING) freeSpaceInCmdPool();
```
Erfan-Ahmadi, line 57: use `canReset` functions instead of these two if conditions it's exactly equivalent and makes sense

include/nbl/video/IGPUCommandBuffer.h
```diff
@@ -227,6 +227,25 @@ class NBL_API IGPUCommandBuffer :
...
+
+    virtual void resetCommon()
```
Erfan-Ahmadi, line 233: make this function non virtual
have another virtual function called `void releaseResourcesToPool`

src/nbl/video/COpenGLCommandBuffer.h
```diff
@@ -542,14 +542,28 @@ class COpenGLCommandBuffer final : public IGPUCommandBuffer
 
     COpenGLCommandBuffer(core::smart_refctd_ptr<const ILogicalDevice>&& dev, E_LEVEL lvl, core::smart_refctd_ptr<IGPUCommandPool>&& _cmdpool, system::logger_opt_smart_ptr&& logger, const COpenGLFeatureMap* _features);
 
+    void resetCommon() override
+    {
+        freeSpaceInCmdPool();
+        m_commands.clear();
```
Erfan-Ahmadi, line 548: these will be in `releaseResourcesToPool`

src/nbl/video/CVulkanCommandBuffer.h
```diff
@@ -39,8 +39,25 @@ class CVulkanCommandBuffer : public IGPUCommandBuffer
         freeSpaceInCmdPool();
     }
 
-    bool begin(core::bitflag<E_USAGE> recordingFlags, const SInheritanceInfo* inheritanceInfo=nullptr) override
+    void resetCommon() override
     {
+        freeSpaceInCmdPool();
```
Erfan-Ahmadi, line 44: this will be in `releaseResourcesToPool`

src/nbl/video/COpenGLCommandBuffer.h
Erfan-Ahmadi, line 548: this thing you're doing here is called `releaseResourcesBackToPool` and we don't need it in `resetCommon` and in fact we don't need resetCommon to be virtual.
because the pool has reclaimed all the resources already
See discussion: https://discord.com/channels/593902898015109131/888722959982813204/984443328680579082

src/nbl/video/COpenGLCommandBuffer.h
```diff
@@ -542,14 +542,28 @@ class COpenGLCommandBuffer final : public IGPUCommandBuffer
...
+        IGPUCommandBuffer::resetCommon();
+    }
+
     inline bool begin(core::bitflag<E_USAGE> _flags) override final
     {
-        reset(E_RESET_FLAGS::ERF_RELEASE_RESOURCES_BIT);
+        if (m_cmdpool->getCreationFlags().hasFlags(video::IGPUCommandPool::ECF_RESET_COMMAND_BUFFER_BIT))
```
Erfan-Ahmadi, line 554: change condition to `canReset`

src/nbl/video/COpenGLCommandBuffer.h
```diff
@@ -542,14 +542,28 @@ class COpenGLCommandBuffer final : public IGPUCommandBuffer
...
+            reset(E_RESET_FLAGS::ERF_RELEASE_RESOURCES_BIT);
+        else checkForCommandPoolReset();
+
         return IGPUCommandBuffer::begin(_flags);
     }
     
     inline bool begin(core::bitflag<E_USAGE> _flags, const SInheritanceInfo* inheritanceInfo = nullptr) override final
     {
+        if (m_cmdpool->getCreationFlags().hasFlags(video::IGPUCommandPool::ECF_RESET_COMMAND_BUFFER_BIT))
```
Erfan-Ahmadi, line 563: change condition to `canReset`

src/nbl/video/COpenGLCommandBuffer.h
Erfan-Ahmadi: Update: https://github.com/Devsh-Graphics-Programming/Nabla/pull/408/files#r938465420

src/nbl/video/COpenGLCommandPool.h
devshgraphicsprogramming: the reasoning is that whenever `checkForCommandPoolReset` returns true, it means all the data storage for the command "tapes" is already gone, so you'd need to do
```cpp
inline void IGPUCommandPool::reset()
{
   m_state = INITIAL;
   // all memory that backs references already gone and destructed
   if (checkForCommandPoolReset())
      return;

   // invalid usage
   if (!canReset())
   {
      NBL_DEBUG_BREAK(false);
      m_state = INVALID;
   }

   // TODO: iterate over the tapes and free their allocations
   reset_impl();
}
```

include/nbl/core/containers/CMemoryPool.h
```diff
@@ -80,6 +80,12 @@ class NBL_API CMemoryPool : public Uncopyable
         return free_n<T>(ptr, 1u);
     }
 
+    // TODO are we gonna keep this?
+    void reset()
+    {
+        m_alctr.reset();
+    }
+
```
devshgraphicsprogramming, line 88: yes but check/test `SimpleBlockBasedAllocator::reset` actually works

include/nbl/video/IGPUCommandBuffer.h
```diff
@@ -97,7 +97,7 @@ class NBL_API IGPUCommandBuffer :
             assert(false);
             return false;
         }
-        return base_t::reset(_flags);
+        return resetCommon();
```
devshgraphicsprogramming, line 100: check for commandpool reset first, then check valid usage, see my other comment

also call it `reset_impl` not `resetCommon`

include/nbl/video/IGPUCommandBuffer.h
devshgraphicsprogramming: I'd call it `m_resetCheckedStamp` to make it clear we're not counting resets on the commandbuffer but on the pool to which we're linked

include/nbl/video/IGPUCommandBuffer.h
```diff
@@ -227,6 +227,24 @@ class NBL_API IGPUCommandBuffer :
             for (uint32_t i = _first + 1u; i < IGPUPipelineLayout::DESCRIPTOR_SET_COUNT; ++i)
                 _destPplnLayouts[i] = nullptr;
     }
+
+    uint32_t m_resetCount = 0;
+
+    virtual bool resetCommon()
+    {
+        m_resetCount = m_cmdpool->getResetCounter().load();
```
devshgraphicsprogramming, line 235: no, you want to set the last-checked-stamp upon a `begin`, not before (you're only interested in transitions _into_ INITIAL or INVALID caused by pool reset)

devshgraphicsprogramming: If I have 4 cmdbuffers A,B,C,D and only set the stamp when I reset them manually, then I get a lot of false positives when calling `checkForCommandPoolReset`

include/nbl/video/IGPUCommandBuffer.h
devshgraphicsprogramming: furthermore, if the commandbuffer is not indiviually resettable it should go into INVALID instead of INITIAL (see my other comment)

include/nbl/video/IGPUCommandBuffer.h
```diff
@@ -227,6 +227,24 @@ class NBL_API IGPUCommandBuffer :
...
+    }
+
+    // This should be called on every command buffer method
+    // Only doing it in begin for now, we expect that the command pool won't be reset midway through recording a command buffer
+    inline bool checkForCommandPoolReset()
+    {
+        if (m_cmdpool->getResetCounter().load() <= m_resetCount)
+            return false;
+
+        return resetCommon();
```
devshgraphicsprogramming, line 246: `check` shouldn't call the reset_impl (or reset-common) since the only thing it needs to do is just transition the state into INITIAL

P.S. on another note it was a mistake to have a non-pure virtual `reset` in `ICommandBuffer`, unecessary complexity and stuff.

include/nbl/video/IGPUCommandPool.h
```diff
@@ -22,7 +26,7 @@ class NBL_API IGPUCommandPool : public core::IReferenceCounted, public IBackendO
             ECF_PROTECTED_BIT = 0x04
         };
 
-        IGPUCommandPool(core::smart_refctd_ptr<const ILogicalDevice>&& dev, core::bitflag<E_CREATE_FLAGS> _flags, uint32_t _familyIx) : IBackendObject(std::move(dev)), m_flags(_flags), m_familyIx(_familyIx) {}
+        IGPUCommandPool(core::smart_refctd_ptr<const ILogicalDevice>&& dev, core::bitflag<E_CREATE_FLAGS> _flags, uint32_t _familyIx) : IBackendObject(std::move(dev)), m_flags(_flags), m_familyIx(_familyIx), m_resetCount(0) {}
```
devshgraphicsprogramming, line 29: initialize `m_resetCount` with the current value of the pool's reset counter

include/nbl/video/IGPUCommandPool.h
```diff
@@ -31,11 +35,20 @@ class NBL_API IGPUCommandPool : public core::IReferenceCounted, public IBackendO
...
+        }
+
     protected:
         virtual ~IGPUCommandPool() = default;
 
         core::bitflag<E_CREATE_FLAGS> m_flags;
         uint32_t m_familyIx;
+        std::atomic_uint32_t m_resetCount;
+
+        std::atomic_uint32_t& getResetCounter() { return m_resetCount; }
```
devshgraphicsprogramming, line 51: you can make it public, so you don't need the friendship and forward declaration of `IGPUCommandBuffer`

Also its better to not return a reference and just "do the load"
```cpp
inline uint32_t getResetCount() const {return m_resetCount.load();}
```

src/nbl/video/COpenGLCommandBuffer.cpp
```diff
@@ -115,8 +115,7 @@ COpenGLCommandBuffer::~COpenGLCommandBuffer()
         if (!IGPUCommandBuffer::canReset())
             return false;
```
devshgraphicsprogramming, line 116: if we rewrite to a `reset_impl` then all these checks can be common to all backends.

src/nbl/video/COpenGLCommandBuffer.cpp
```diff
@@ -115,8 +115,7 @@ COpenGLCommandBuffer::~COpenGLCommandBuffer()
...
 
-        freeSpaceInCmdPool();
-        m_commands.clear();
+        releaseResourcesBackToPool();
         return IGPUCommandBuffer::reset(_flags);
```
devshgraphicsprogramming, line 119: if we rewrite to a `reset_impl` then the state transition can be common to all backends

src/nbl/video/COpenGLCommandBuffer.h
```diff
@@ -542,14 +542,40 @@ class COpenGLCommandBuffer final : public IGPUCommandBuffer
...
+
     inline bool begin(core::bitflag<E_USAGE> _flags) override final
     {
-        reset(E_RESET_FLAGS::ERF_RELEASE_RESOURCES_BIT);
+        checkForCommandPoolReset();
+        // skip reset if command pool was reset above and state is initial
+        if (canReset() && m_state != ES_INITIAL)
+            reset(E_RESET_FLAGS::ERF_RELEASE_RESOURCES_BIT);
+
         return IGPUCommandBuffer::begin(_flags);
```
devshgraphicsprogramming, line 568: hoist this stuff to `IGPUCommandBuffer::begin` and just leave an empty `virtual void begin_impl` for vulkan to override

src/nbl/video/COpenGLCommandBuffer.h
```diff
@@ -542,14 +542,40 @@ class COpenGLCommandBuffer final : public IGPUCommandBuffer
...
     inline bool begin(core::bitflag<E_USAGE> _flags, const SInheritanceInfo* inheritanceInfo = nullptr) override final
     {
+        checkForCommandPoolReset();
+        // skip reset if command pool was reset above and state is initial
+        if (canReset() && m_state != ES_INITIAL)
+            reset(E_RESET_FLAGS::ERF_RELEASE_RESOURCES_BIT);
+
+
         return IGPUCommandBuffer::begin(_flags, inheritanceInfo);
     }
```
devshgraphicsprogramming, line 580: same as https://github.com/Devsh-Graphics-Programming/Nabla/pull/408/files#r940648517

src/nbl/video/COpenGLCommandBuffer.h
devshgraphicsprogramming: you always want to do `m_commands.clear()` 

BUT **do not do** a `freeSpaceInCmdPool` if the entire pool has been reset

src/nbl/video/COpenGLCommandBuffer.h
```diff
@@ -542,14 +542,40 @@ class COpenGLCommandBuffer final : public IGPUCommandBuffer
...
+
+    bool resetCommon() override
+    {
+        // TODO resources for all command buffers already got freed from the command pool,
+        // this function should just make sure the command buffer stops pointing to memory that has
+        // been freed from the pool
+        // (this currently causes a crash due to double free)
+        releaseResourcesBackToPool();
+        return IGPUCommandBuffer::resetCommon();
+    }
```
devshgraphicsprogramming, line 559: you're crashing because you shouldn't free resources of the pool if the pool has been reset

src/nbl/video/COpenGLCommandPool.h
```diff
@@ -39,7 +39,16 @@ class COpenGLCommandPool final : public IGPUCommandPool
             std::unique_lock<std::mutex> lk(mutex);
             mempool.free_n<T>(const_cast<T*>(ptr), n);
         }
-        
+
+        void reset() override
+        {
+            // Free everything in the memory pool from all command buffers
+            // TODO figure out if this is the best way to do it
+            mempool.reset();
```
devshgraphicsprogramming, line 47: Its probably best, just check it works :)

devshgraphicsprogramming:
A couple of issues.

# Fix my sphagetti code

The pattern of having to remember to call `IGPUCommandPool::something` from the override of `something` in the base class, is incredibly dumb, I apologize.

We should stick to 
```cpp
inline bool something()
{
   stuff;
   
   something_impl();
   
   return more stuff;
}

protected:
virtual void something_impl() = 0;
```

## Turn all methods of `ICommandBuffer` into pure virtuals

Push the logic into `IGPUCommandBuffer`, our future CPU commandbuffer will be much more lax.

# Hoist all State Transition and flag checking logic to IGPUCommandBuffer

Have the `_impl` just be void functions.

# Hoist `releaseResourcesBackToPool` to `IGPUCommandBuffer` and make it a protected inline + a `releaseResourcesBackToPool_impl`

ALWAYS make it drop reference count stuff (do through the destructors of the tape entries) regardless of the value of `releaseResources`

Have `releaseResources` only control whether to free the memory for the entries and whether to call `vkResetCommandBuffer` with the flag or not.

## Do not call `releaseReourcesBackToPool` if commandpool has been reset already

This will cause a double free, the entries pushed into tapes are already gone (also the head/tail pointers to tape slabs/blocks).

### But do `m_commands.clear()` the OpenGL Commandbuffer

@achalpandeyy will remove `m_commands` soon anyway, it will be replaced by tail/head pointers to a tape allocated from the command pool's memory pool so these will need to be nullified in the future instead.

## Check for commandpool reset BEFORE EVERY method!

Directly related to the preceeding issue.

# Set the Last-Reset-Stamp in the `begin()` not during the `resetCommon`

What is `resetCommon` now I'd want to fold into `releaseResources`, the setting of the stamp should be upon a begin, not upon a reset.

The stamp lets you "catch" pool-resets which have occured since the last time you set the stamp, you only care about new pool resets _since the last time you begun the commandbuffer.

src/nbl/video/COpenGLCommandPool.h
devshgraphicsprogramming: ok, it wont work, you need to be able to iterate every single struct in the `mempool` and run its destructor

src/nbl/video/CVulkanCommandPool.cpp
```diff
@@ -12,6 +12,18 @@ CVulkanCommandPool::~CVulkanCommandPool()
     vk->vk.vkDestroyCommandPool(vulkanDevice->getInternalObject(), m_vkCommandPool, nullptr);
 }
 
+void CVulkanCommandPool::reset()
+{
+	// Free everything in the memory pool from all command buffers
+	// TODO figure out if this is the best way to do it
```
devshgraphicsprogramming, line 18: not only do we need to free the mempool, we need to run the destructors of all objects on the tape :(

Related to https://github.com/Devsh-Graphics-Programming/Nabla/pull/408#discussion_r940693463

src/nbl/video/CVulkanCommandPool.cpp
```diff
@@ -12,6 +12,18 @@ CVulkanCommandPool::~CVulkanCommandPool()
...
+	mempool.reset();
+
+	const CVulkanLogicalDevice* vulkanDevice = static_cast<const CVulkanLogicalDevice*>(getOriginDevice());
+	auto* vk = vulkanDevice->getFunctionTable();
+	vk->vk.vkResetCommandPool(vulkanDevice->getInternalObject(), m_vkCommandPool, 0);
+	IGPUCommandPool::reset();
```
devshgraphicsprogramming, line 24: you want to increment the counter at the start

src/nbl/video/CVulkanCommandPool.cpp
devshgraphicsprogramming: also `mempool`, etc. should move to the base probably

src/nbl/video/COpenGLCommandPool.h
achalpandeyy: This is a good point. I remember asking myself "what is wrong here" but couldn't figure out even though I did something [similar to this](https://github.com/Devsh-Graphics-Programming/Nabla/blob/b80d42cc34bc9753fa5b52503843de474fb05e61/include/nbl/video/IGPUCommandBuffer.h#L301) before myself.

include/nbl/video/IGPUCommandBuffer.h
achalpandeyy: As things are now, I think `m_resetCount = m_cmdpool->getResetCounter()` should only ever happen in `IGPUCommandBuffer::begin` and perhaps in `IGPUCommandBuffer::reset`?
If we do it in `checkForCommandPoolReset` (according to #373) it will end up getting done in every command buffer method since every command buffer method should call `checkForCommandPoolReset` and if that happens, something like the following could create problems:

```C++
cmdbuf->begin();
cmdpool->reset();
cmdbuf->beginRenderPass(); // detects command pool reset and fails, BUT updates the command buffer last-checked-stamp
cmdbuf->draw(); // doesn't detect command pool reset and tries to do work, and crashes
..
```

We could also solve this by doing state validation in the command buffer methods, something like:
```C++
if (m_state != ES_RECORDING)
    return false;
```

src/nbl/video/COpenGLCommandPool.h
achalpandeyy: Note: This is most likely `IGPUCommandBuffer::reset`.

achalpandeyy: I also want to point out that transitioning the state of a non-resettable command buffer to `INVALID` because it fails the `reset` call, is not consistent with the Vulkan spec.

include/nbl/video/IGPUCommandBuffer.h
achalpandeyy: Resolved.

src/nbl/video/COpenGLCommandPool.h
achalpandeyy: Resolved.

src/nbl/video/COpenGLCommandBuffer.h
achalpandeyy: Resolved.

include/nbl/core/containers/CMemoryPool.h
achalpandeyy: Is this still relevant?

include/nbl/video/IGPUCommandBuffer.h
achalpandeyy: Resolved.

include/nbl/video/IGPUCommandBuffer.h
achalpandeyy: I think I did what you asked for here, but it would be nice if you still answer the question above.

include/nbl/video/IGPUCommandBuffer.h
achalpandeyy: Resolved.

include/nbl/video/IGPUCommandPool.h
achalpandeyy: Resolved.

src/nbl/video/COpenGLCommandBuffer.cpp
achalpandeyy: Resolved.

src/nbl/video/COpenGLCommandBuffer.cpp
achalpandeyy: Resolved.

src/nbl/video/COpenGLCommandBuffer.h
achalpandeyy: Resolved.

src/nbl/video/COpenGLCommandBuffer.h
achalpandeyy: Resolved.

src/nbl/video/COpenGLCommandBuffer.h
achalpandeyy: Resolved.

src/nbl/video/CVulkanCommandPool.cpp
achalpandeyy: Resolved.

include/nbl/video/IGPUCommandBuffer.h
devshgraphicsprogramming: ok so we definitely need to set the timestamp in begin()
you shouldn't update the timestamp after a "successful" checkForParentPoolReset, to make sure it always returns true if there was a reset since last begin()

> We could also solve this by doing state validation in the command buffer methods, something like:

great idea, we could roll it into a single function

```cpp
bool checkStateBeforeRecording()
{
    if (m_state != ES_RECORDING)
    {
        m_logger.log("Failed to record into command buffer: not in RECORDING state.", system::ILogger::ELL_ERROR);
        return false;
    }
    if (checkForParentPoolReset())
    {
        m_logger.log("Failed to record into command buffer: pool was reset since the recording begin() call.", system::ILogger::ELL_ERROR);
        return false;
    }
    return true;
}
```

and use that as the prelude to any IGPUCommandBuffer method (including `end()`)

include/nbl/core/containers/CMemoryPool.h
devshgraphicsprogramming: well you still need to check the `CMemoryPool` reset works, but obviously it wont call the dtors of anything you allocated over it.

include/nbl/video/IGPUCommandBuffer.h
achalpandeyy: Resolved.

PR #421, opened 2022-09-20
Update setWindowSize & other window manager functions to ignore callback

deprilula28:
## Description
Currently, when resizing the window using the API call (`setWindowSize`) calls the event callback, but the result is unused as it doesn't stop the resize; Then, the event callback is called again, potentially with a new size including window borders and other possible calculations from Windows.

The maximum size returned by the surface capabilities should be taken into account when recreating the swapchain, as is done in the sister PR.

Sister PR: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/19

## Testing 
Example 09 was tested to not log the resizes twice.

### DIFF
--- a/include/nbl/ui/CWindowManagerWin32.h
+++ b/include/nbl/ui/CWindowManagerWin32.h
@@ -62,8 +62,22 @@ class CWindowManagerWin32 : public IWindowManager
 
 		inline bool setWindowSize_impl(IWindow* window, const uint32_t width, const uint32_t height) override
 		{
+			// Calculate real window size based on client size
+			RECT clientSize;
+			clientSize.left = 0;
+			clientSize.top = 0;
+			clientSize.right = width;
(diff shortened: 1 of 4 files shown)


include/nbl/ui/CWindowManagerWin32.h
```diff
@@ -256,6 +270,43 @@ class CWindowManagerWin32 : public IWindowManager
 				{
 				}
 
+				// TODO where should this be
+				static DWORD getWindowStyle(IWindow::E_CREATE_FLAGS flags)
```
Erfan-Ahmadi, line 274: @devshgraphicsprogramming

Erfan-Ahmadi: Devsh said: You could move it into Win32 window class

PR #425, opened 2022-10-17
Utilitiy interface revise

Erfan-Ahmadi:


include/nbl/video/IGPUQueue.h
```diff
@@ -27,6 +27,17 @@ class NBL_API IGPUQueue : public core::Interface, public core::Unmovable
             IGPUSemaphore*const * pSignalSemaphores = nullptr;
             uint32_t commandBufferCount = 0u;
             IGPUCommandBuffer*const * commandBuffers = nullptr;
+
+            inline bool isValid()
+            {
+                if(waitSemaphoreCount > 0u && (pWaitSemaphores == nullptr || pWaitDstStageMask == nullptr))
+                    return false;
+                if(signalSemaphoreCount > 0u && pSignalSemaphores == nullptr)
+                    return false;
```
devshgraphicsprogramming, line 36: its valid to not wait or not signal any semaphores.

For example if you split some work into 5 submits in a row (due to buffering), you're guaranteed that they'll execute in submission order therefore they don't need semaphores between them.

Erfan-Ahmadi: Yea, but this is basically saying if the `Count` is more than zero then the pointer shouldn't be null :D
If the count is zero it's ok

Erfan-Ahmadi: don't worry, It was just a misread :D

src/nbl/video/utilities/IUtilities.cpp
```diff
@@ -76,24 +76,21 @@ void IUtilities::updateImageViaStagingBuffer(
         {
             // but first sumbit the already buffered up copies
             cmdbuf->end();
-            IGPUQueue::SSubmitInfo submit;
+            IGPUQueue::SSubmitInfo submit = submitInfoBeforeOverwrite;
             submit.commandBufferCount = 1u;
             submit.commandBuffers = &cmdbuf;
```
devshgraphicsprogramming, line 81: you do not want without changing the commandbuffer count and commandbuffers to be submitted, because now you take a whole `intendedSubmitOfPreceedingWork`

src/nbl/video/utilities/IUtilities.cpp
```diff
@@ -76,24 +76,21 @@ void IUtilities::updateImageViaStagingBuffer(
...
-            m_device->blockForFences(1u, &fence);
-            waitSemaphoreCount = 0u;
-            semaphoresToWaitBeforeOverwrite = nullptr;
-            stagesToWaitForPerSemaphore = nullptr;
+            assert(submit.isValid());
+            submissionQueue->submit(1u, &submit, submissionFence);
+            m_device->blockForFences(1u, &submissionFence);
+            submitInfoBeforeOverwrite.waitSemaphoreCount = 0u;
+            submitInfoBeforeOverwrite.pWaitSemaphores = nullptr;
+            submitInfoBeforeOverwrite.pWaitDstStageMask = nullptr;
```
devshgraphicsprogramming, line 89: it is after the submission that you want to do
```cpp
            intendedSubmitOfPreceedingWork.waitSemaphoreCount = 0u;
            intendedSubmitOfPreceedingWork.pWaitSemaphores = nullptr;
            intendedSubmitOfPreceedingWork.pWaitDstStageMask = nullptr;
            intendedSubmitOfPreceedingWork.commandBufferCount = 1u;
            intendedSubmitOfPreceedingWork.commandBuffers = &cmdbuf;
```

src/nbl/video/utilities/IUtilities.cpp
```diff
@@ -135,45 +132,36 @@ void IUtilities::updateImageViaStagingBuffer(
         }
 
         // this doesn't actually free the memory, the memory is queued up to be freed only after the GPU fence/event is signalled
-        m_defaultUploadBuffer.get()->multi_deallocate(1u, &localOffset, &allocationSize, core::smart_refctd_ptr<IGPUFence>(fence), &cmdbuf); // can queue with a reset but not yet pending fence, just fine
+        m_defaultUploadBuffer.get()->multi_deallocate(1u, &localOffset, &allocationSize, core::smart_refctd_ptr<IGPUFence>(submissionFence), &cmdbuf); // can queue with a reset but not yet pending fence, just fine
     }
```
devshgraphicsprogramming, line 136: return `intendedSubmitOfPreceedingWork` at the end

devshgraphicsprogramming:
I'm impressed, you got good at designing APIs. Very cool change, cleans up my sphagetti a lot.

Just everything that uses this pattern of "submit on overflow" will need to be changed, not only image upload, but buffer upload and other utilities like `PropertyPoolHandler`

src/nbl/video/utilities/IUtilities.cpp
```diff
@@ -135,45 +152,35 @@ void IUtilities::updateImageViaStagingBuffer(
...
-    submit.pWaitSemaphores = semaphoresToWaitBeforeOverwrite;
-    submit.pWaitDstStageMask = stagesToWaitForPerSemaphore;
-    queue->submit(1u,&submit,fence);
+    if (noCommandBufferInNextSubmit)
+    {
+        cmdbuf->end();
+        assert(intendedNextSubmit.isValid());
+        submissionQueue->submit(1u,&intendedNextSubmit,submissionFence);
+        intendedNextSubmit.commandBufferCount = 0u;
+        intendedNextSubmit.commandBuffers = nullptr;
```
devshgraphicsprogramming, line 164: nullify everything, with a `= {}`

src/nbl/video/utilities/IUtilities.cpp
```diff
@@ -4,40 +4,60 @@
...
 {
+    if(!intendedNextSubmit.isValid())
+    {
+        // TODO: log error
+        assert(false);
+        return intendedNextSubmit;
+    }
+
+    bool noCommandBufferInNextSubmit = (intendedNextSubmit.commandBufferCount == 0);
+    if (noCommandBufferInNextSubmit)
```
devshgraphicsprogramming, line 19: I could base the detection on whether the last commandbuffer is resettable, and still recording and one time submit

If not then end the last commandbuffer (if still recording) and create a new one like here, plus copy all the old commandbuffers pointers to a unique_ptr new array + the new commandbuffer

PR #428, opened 2022-11-04
HLSL Rewrite

devshgraphicsprogramming:
### Basics

I'd start with the easy and simple shader headers, so files ending in `.glsl` that don't have any macros of use any resource types like `sampler2D`.

Later on we'll deal with actual shaders with stages, or ones that make heavy use of preprocessor directives (they will need to be turned into templates).

I took a few GLSL headers and rewrote them to HLSL.
 
Try to follow the same pattern:
- update any copyright comment to current year
- replace all `GLSL` or `glsl` with `HLSL` and `hlsl` respectively
- Header guards should have the form of `_NBL_HLSL_${ADDITIONALNAMESPACE}${_OPTIONAL EXTRA NESTED NAMESPACES}_${FILENAME}_INCLUDED_`
- remove prefixes and make them namespaces (remember that HLSL2021 is not C++11 and nested namespace declaration shortcut is not available)
- for anything after `nbl_glsl_` like `barycentric_` you need to make an informed deduction whether the next identifier was a namespace or a struct
- `vecN` turns into `floatN`
- whenever you see a matrix type like `matCxR`, turn it into a `floatRxC` if there's multiplications going on later on, otherwise keep as a `floatRxC` if there's use of the operator[] like in the `barycentril/util.hlsl` example, if unsure ask or comment
- matrix multiplications need to be replaced with `mul()` as `*` in HLSL just multiplies all elements pointwise (DO NOT WANT)
- fold free/global functions that take a struct as the first parameter back into non-static methods (**careful** if the struct was `in` then method needs to be const, if it was `out` then you should probably turn it into a static method)
- fold free/global functions whose last parf of the prefix matches a class name into static methods

### A word about matrix multiplication

GLSL is utterly fucked.

In textbook mathematics, MxN matrix, means "M rows" (dimensionality of a column) and "N columns" (dimensionality of a row).
Also when you multiply a MxN matrix A with OxP matrix B to get matrix C, like this C = A B, then:
- N and O need to match, so really A is a MxN and B must be a NxP
- C is going to be a MxP matrix

GLSL fucks this up because a mat4x3 is actually a matrix with 3 rows and 4 columns.
But at least this is preserved:
```glsl
matNxM A;
matPxN B;
matPxM C = A*B;
```
and you get the same result as in a textbook by computing the elements of C as dot products between rows of A and columns of B (this is why their dimensionalities have to match).

HLSL at matches the textbook and float3x4 is a matrix with 3 rows and 4 columns.
```hlsl
floatNxM A;
floatPxN B;
floatPxM C = mul(A,B);
```
So for multiplications using `floatMxN` to replace a GLSL `matNxM` is fine.

#### HOWEVER they differ in what `operator[]` returns

GLSL a `vecA matAxB::operator[](int i)` will return the i-th column 
HLSL a `vecB floatAxB::operator[](int i)` will return the i-th row

This a problem that can't be easily side-stepped in a scenario where `[]` is used explicitly, like for example when I construct a matrix from 2 of  3-dimensional vectors like with the barycentrics.

Then the `matNxM` has to become a `floatNxM` and not a `floatMxN` and you need to correct the rest of the code for the fact that the matrix is now a transpose.

This is why @deprilula28 will review and you can ask me and @Erfan-Ahmadi about these cases.

If unsure, leave a comment in the code.

### How do I know whether _prefix is a namespace or a struct?

Basically if you see a GLSL function that takes the same prefix as the first parameter, you know it was supposed to be a method.

For example look at `nbl_glsl_shapes_AABB_t` https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/glsl/shapes/aabb.glsl and its corresponding HLSL I made.
 
## Testing 

You can get a prebuilt DXC [here](https://github.com/microsoft/DirectXShaderCompiler/releases/download/v1.7.2207/dxc_2022_07_18.zip) for some reason `-HV 2021` does't work with that exe, so you can't do:
- templates 
- bitfields
- `using namespace nbl::hlsl::shapes`
- `using` instead of a `typedef`
yet, but @AnastaZIuk is working on giving you a functional HLSL2021 compiler.

For now make a huge `.bat` file that compiles a bunch of HLSL, I shouldn't be getting any `error` when I run it or `warning`s other than the one warning about PDB not being separate.

Later @AnastaZIuk will make that into a HLSL_Test CMake target.

### Validity of Syntax in "header" `.hlsl` files

`dxc.exe -HV 2018 -T lib_6_6 -IpathToNablaRoot/include/ -Zi -Fo test.bin pathToNablaRoot/include/nbl/builtin/hlsl/restOfThePathToMyHeader.hlsl`

### Validity of Syntax in "complete shader" `.stage.hlsl` files

`dxc.exe -HV 2018 -T ${shader_stage}_6_6 -IpathToNablaRoot/include/ -IpathToNablaRoot/src/ -spirv -Zi -Fo test.bin pathToNablaRoot/include/nbl/builtin/hlsl/restOfThePathToMyShader.${stage}.hlsl`

### [Much Later TODO] Validity of Syntax for "implementation" `.hlsl` files

`dxc.exe -HV 2018 -T lib_6_6 -IpathToNablaRoot/include/ -IpathToNablaRoot/src/ -Zi -Fo test.bin pathToNablaRoot/src/nbl/builtin/hlsl/restOfThePathToMyHeader.hlsl`

## TODO list:

- [ ] @AnastaZIuk get DXC building as a static lib (to link into Nabla) and an EXE (to build HLSL into SPV) which understands `-HV 2021` option
- [ ] @Erfan-Ahmadi `asset::CHLSLCompiler` so we can compile HLSL at runtime 
- [ ] @AnastaZIuk create CMake utility function to make `.stage.spv` from **full** (when you actually know the shader stage and entry point) HLSL shaders
- [ ] (Seprate PR) @AnastaZIuk improve the resource embedding CMake and scripts, so it can be used not only to embed resources in C++ using the `nbl::builtin` namespace but in external projects with custom namespaces, also handle the embedding of resources that will be created at a later date (like compiled)
- [ ] @Erfan-Ahmadi replace the preprocessor used to resolveIncludes, etc. with our fork of https://github.com/bnoazx005/tcpp/blob/master/source/tcppLibrary.hpp and refactor `IIncluder` etc.

The order of capabilities I want to get:
1. the ability to just compile HLSL at runtime like I do with HLSL (with proper include resolution and define injection)
2. the ability to precompile full shaders into SPIR-V and embed that as a builtin

The first one is enough for me to get unblocked and make a start on `ext::PathTracer`.

Suggested tasks for Shader Compiler:
1. Maybe we should load to `ICPUBuffer` instead of a C-array directly and create the `ICPUShader` from that?
2. Create a `CHLSLLoader` similar to `CGLSLLoader` but:
2a. Only file extensions it associates with are `*.h` and `*.hlsl`
2b. Check loadability by checking that file is under 64MB and regexping the file for any reserved keywords https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-keywords
2c. it returns a `ICPUShader` not an `ICPUSpecializedShader`


devshgraphicsprogramming:
# HLSL 2021

## New features

You can now use:

### Bitfields

Often in the code we have stuff like 
```glsl
uint varA_varB_varC; // N bits varA, M bits varB
```
You can now do
```hlsl
uint varA : N;
uint varB : M;
uint varC ; P;
```
where `P=32-M-N`

### The `using` keyword

Using has two uses

#### Using Namespace

Basically saves you writing out `namespace::after::common::ancestor::identifier` and you can simply put `using namespace::after::common::ancestor::identifier`.

You need to put it in the global scope (not inside any struct of function definition body).

Careful to never put it outside of a `namespace {}` block, cause then it pollutes all code that includes.

#### Using as an alias

Instead of writing `typedef Type Alias;` you can write `using Alias = Type;` is just a bit saner.

### Templates

Basically same as C++03, however there's no explicit instantiation or `extern template` so all templated functions (or methods of templated classes) have to be declared inline.

And you should probably stay away from SFINAE.

devshgraphicsprogramming:
# How to rewrite our awful GLSL macros into HLSL templates

Lets take `algorithm.glsl` for a spin.

## See the example codegen

First you want to see what is the code the Preprocessor would generate for an example instatiation.

For this you can run `gcc -E` over the code containing an instantiation, and you will see what the preprocessor spits out.

You can even do this with [Compiler Explorer](https://compiler-explorer.com/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAOykrgAyeAyYAHI%2BAEaYxCAArNIADqgKhE4MHt6%2BAaSp6Y4CoeFRLLHxSbaY9kUMQgRMxATZPn6BdpgOmQ1NBCWRMXGJ0gqNza25HeP9YYPlw0kAlLaoXsTI7BzmAMx4VAxYVADUAPoRFsGnFnIAksEAKrcRpwASwUJXAILBAOIA8gAlW4PV4AWVOzzcwTkABFsLDTiYNF9dkc5mcLlcbvcni93p9Tj8AcDQRCoTD4YjkaiUbt9odMCdzpdrndHs9Tr8PlcwV83ID/kJIRFoXCEUi6WYdujwpjWTiOS9uYS%2BQKhSKxVTJbTUdLZZhjliuTzTtgAGo/CAADSWx2tNMdUplTIxxpVV2wAEU5FbraQAJp2iDu00Wv1LEw7WFR2Ghwnh4IQIOR50Go2sj3nf4PM0%2Bv2B4Pxz2WpO2sBgWPFs2l5NLVO6tGuuXV4LYIRCG2F44hzNh2u2qNuauJusNps0Ft9wm/QHYL4PbCArtBnsjgeRnbYddWlNOxv65uG6u3YURHOnZQ5iAANyL05LVrv5gAbL2rlnR3eALRcesV2MNEdQ9J2PB9IWFK8HlvYMKxPM8LygmCGwAegAKjQ4CXVAjMP1NQU5AiRE5GUS9rzvHsuC8Y5rBohI3BpKAmOYijv2OKi7RMfw3B7KBWPYrxOK3KNsGODRVC4DRBJo7i6IYlFjkUpSmP4jiZJ4lS7TYtSRJEsSJKkoTsD08SNAADkEzjZJMeiaSUxSWK0gSrI0vinJ04St30yTpK4nibPkr57Ictzjm03zZM0sLnJozzRPEnyjJM1RzMs2LjK80zJEsly5LskLHOinTItC8KksygyIv82yFOCqKyvU3iYKK3y4u8wz0uS1L60agL8uU%2BqYr8prVNajL4sq3K%2Btq%2BzBuK1zmoa3SKsSzqKu68qJo0bL602/SNDMHLepqoLmMKpaSsWoa2oSjrhumoK6tKobLtGvbbrGrqLJ6%2B6Tqe575pG9zPpWu7rL%2B/63sauaQYm1blq2773pSnafvBwKBvOl6FqhhH2qqvKZsx3HXuB5H4ZulKkeOjHZph3L6bWuGweq2m6auwHGbxj7kY2pn9tR3nDt2/nTI43aaQwlD9wnN1wLcf5RQXSEwWUYIzBXO1rRo6VdmOAMsPTasFaV3NblVstu2NxW3GV821Y1/0g0No9cJNQkTdtxcIgXbBHat%2BWbbti333dx8yxWHckz3NNXet03t19nYV1IAAtO1469xPFw1zPfZ9nPNdIKOIHT8cQLlvCPaD73fckFPU9IAB1DPA4Tgu/dDrNPfz33/aDYu26zjuNcblvx1jnDqy%2BIjNbXcDR0HMwXxLmOD2wyuw6JWFYTnrv%2BwjaxV/rF2p/AoQ5AsPeS9tb9j4nr4UJQ44EAIAhkgUEAn%2BAQgEC8aIAB0aAWAoWAKgVAwB6AoQUAgJgWANgoTwAoBQXhMAKBQlwLgCQEg0ifrLKcVdeRfGtHvKAN8T5bn3gmWsKYhIADF56EJrBGJYlB76oWlhXAhW8wTPFIVQ8ONoKFuAEcw6OIsowMNEYvesbCF40IlnSZ0rh9gyy4WBJh8JoRfDnKcYI/wm5LmuP8QiiIFaqwgDowEXwAznC%2BGCbApAHgBmUI48xyhW5MJ7rXRc9cvBhAIMcWgqAADucRTjRDWIcU4pArE2LsQ40gMTWztiEEsCA/jBDHFiD/BgpBjhhGOJkwJrh8mFOca444N4xCoM3FYJReoN7cKzFo4IOjsCnBIq4wExjTGnHcZYwE1jbE%2B0SRUtx/xVaeK3t47O2A/EBKCaE8JkSvDRNiUM%2BJozHHJPAm2Ds6TinZMwLkspDAimLNKQU854yqk1MwHUmWk9N4tOwNo3R%2BjDE9IsCY2ecSRn2MceM6Zrz3kdM%2BUYn5fSBn/ISUClxjiUkHMJjPIevdC5HOCWE4gESonoBibC7ZhzFk5LCGci5WSrnlIRXc7wDyZIWGIJgAg6xzl5x8fMiAWKVl4oJZsgFiTdlMP2WkiApK8mlOqXSx5/gYwNPwRorerT2mdOUN03pfz%2BVwqcQikFpplW6K6ZC35ZjJnKEGcM7V4zB7CtSZxE67K5kayOV4ZIyQeVrPxRsy1RKMkkpOWS65FKSmHHJbcqVtSGVMpZcQNlaKOV%2BLdR69ZhLAVJJtVvEV6TxVBFDRGh5UYrCyqeY0o24F4R0OeB0qFRF%2BlmpVmrCAdDCJuG1amsZCLSDuL1dXdudc/VZObaKNtWrtnpu7QOwJOag1HKpTcml%2Bb7WBS4hYFFyIACcmLGCxVhMcVw35xWFrXRodd%2BxGFb1PNmXMSF6AMB%2BidDdK7j3ro3WgBgYxg3HHCCE4I27YzeS8EOIck50BgiEFfW9dTn2voEB%2Bo5%2BAqAnH/be7837f0MELSWp%2Bxw0BeFoOgIpDBiCeFoMcHYxwSDHBYCQQ0jg2AKFPi8/V2BK0RHBQY410KzUWq2Wm613a3avNY9Wk1dbVYNqTNynFqyU2jr452gTD1H0nvFTu44AyS7tuwJGBIFg0OMBsrCUgi7YoMIQ1QEAUkj1KZPRu29an9MYZ2KumzL7ZXPpCQgOghouWMAAtGO6D7bP%2BBc0Fl9KmA3nOsP%2BjT8irRaZ0xYQ9lhfNOfGrGDihnjP3PoUExglnAPOegxF3JOsrDRnU9xzTcmHGJeSxBgzcUMuCSyyZyReW/BWaK65riu7XPRtZcc3J1gICxaYaOBLNkkuRdazl0zAkCtQaU8W%2BV6jBPMeE3ojj3yTU8YFfC1xPargVqrVtr5GrTUWK0zq1xGasxZrUU0xVQnTtGp21xq7NWDsTKmetwkJ22MXbE6Re2SZXXuuk7y71vGO23cU2FjdqmYvcfzXdg%2BSZJu6cc4Z3LUkQDmeswjk99n/3Y%2B6wj9zrnPPeZ7Le/zsJAvLuC6Fpn4X11I8sMjixqPqs%2BsBXVmbKXb0I2azp2EuOvAgGF%2BT1niPItla5%2BanncWMdfYF8NoXjX0vRkywkcX828fS5Z06NzfWwsDdjUNsII2BnK/G7WTH03ht64lxIFrMvHQrfXmWzRLHXtqs45qvnsPtN/eO37wHb2gcwq%2BzdxFey7X7mrADjpUea273%2BXwYJTACAP2YPR5ITBNhfuiLQT39Svj57QYX4vCBaAKDL3SELNIq8KBr4aBgpfy%2Be7laiL3vX9zoUwiiB4/xYT/E/jSV4eBgAIDiMcZQxAnBL4IAATxADSNir936fyftR%2BBRhkheAAT/Agf9AFOGgSQAgmB0BKCaMgBAADX4sEb18JuHfMC3%2BOEQKjTAADWhoTANEL4GgnetApwkCDepwISJA/%2BwAxGrqQg1%2BEAkYoBxwVAay3QAgxwTACguBRSjgtAhAq%2BP%2BqAGBRedAhAOehohAACm%2BL8b8H8X8KE%2B%2BcQh%2Bx%2Bp%2B5%2BACl%2B2e6A0QYghgmwz%2BBAr%2BNIwQoSC%2BS%2BJAJBG%2BKIW%2BTBu%2BrBGA7BwAR%2BJ%2Bv8/8PBqAKE0QeG/%2BYQSgzQIhYh8hjBO%2BLBbBsaahnBmhF%2BOhbAxAwAmAJhpGksaE0sq2OwKiVAHAKwtAnACQvAfgHAWgpAqAnADEnOlgxwCgawGwhouwPApABAmgfhKw/%2BIAkgZgACZgXAL42R/gBRGgWCBRCQ%2BgnAkgwRaR4RnAvAn8GgKRaRKwcAsASAwCyQ3mZAFAEAHRXRKABgRgpwTKVAtQmQpAWAN4eAmw5oeAmAIS/w7qIRyRNAtAN%2BxAn8YqNR0QYQTQ6%2B3AvAwCbAgg/wDAtA%2BxYRWALAhgkCWwlxeATK3QN4aCNRmAqgXQXgN%2BNRASNQNRxB0QxAexHgWANRBAS%2BLAnAPAKwoxTAwACgsx8xix%2BWBxMgggIgYg7AUgqJ8gSgagNRugXA%2BgNxKA0Wlg%2BgeA0Qn8kAKwqAyQdQn8HA342A9RNQXQdQLghwkwfghJIQcwZQFQegBQGQAgXJgpaQwpDAAw/JwwhJnQ2B9QMwopsprJ8pvQzQUpQw8QspipngbQegYwfQGpCwWpKwcR6wmwEg/hgR1RoRvAERHAxwqgZkL434BROGgxwAPYIxYxAgwYuAhAlGSRSwvAqRtpLRSAaw78nx5AlAfR9AxAEQrAWwTpLpbpyAHpXpTIPpd6vAt%2BAZS%2B6Aeg/AaJog4gWJxZOJKg6gtppABJpAISgJyQkJfhFRHAQRpAIRYR9p/wnxR%2BgSqAJwKZrpkg7pNxmZoxbJmQwYHgLAnR8ZOsOwf4IZzRGRiQ/gAC/gZkOwCQ/g0oCQ66Zkkg255RARHAVRpAEJkkjRnZdpdRtglmTRYZVpHAZgNpXZd5oZWg9YxmcQ6QzgkgQAA%3D%3D)

Here's what we have for a `NBL_GLSL_DEFINE_UPPER_BOUND(ARRAY,float)`
```glsl
uint upper_bound_ARRAY_NBL_GLSL_LESS(uint begin, in uint end, in float value)
{
   uint len = end-begin;
   if ((len&(len-1))!=0)
   {
       const uint newLen = 0x1u<<findMSB(len);
       const uint diff = len-newLen;
       begin = (value<ARRAY[newLen]) ? 0u:diff; len = newLen;
   }
   while (len!=0u)
   {
       begin += (value<ARRAY[begin+(len>>=1u)]) ? 0u:len;
       begin += (value<ARRAY[begin+(len>>=1u)]) ? 0u:len;
    }
    return begin+((value<ARRAY[begin]) ? 0u:1u);
}
```

PR #432, opened 2022-11-09
ShaderCompiler cleanup

Erfan-Ahmadi:
## Description
Examples PR:  https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/20

- [x] getExtraGLSLDefines() to getExtraShaderDefines() and rename the device to be language independant
- [x] remove assumptions of glsl in IShader functions
  - [x] asset::IShader::insertDefines assumes there is a `#version` directive
- [x] IShaderCompiler
- [x] Cleanup IncludeHandler related files
- [x] Interface for HLSLCompiler ~~and Linker~~

---

- [x] Compiler Options Struct

HLSL has a bunch of options (GLSL too, but we're removing that and don't care much about supporting) like:
- optimization level
- debug info
etc.
obviously some stuff doesn't make sense to expose like the choice to do SPIR-V gen, Shader Model target (we want SPIR-V target version instead)

> So instead of making our function signature obnoxiously big, lets wrap all this shit up in a simple struct, maybe even add some other stuff thats already in the signatures like self-inclusion limit, etc.

> Also DXC takes compiler options as strings, which is fucking retarded and error prone (but forward compatible), so lets help ourselves here. 

_-- devsh, not Erfan, Erfan is a mature engineer and doesn't use profanity at work_

---

- [x] Ability to pass our own CIncludeHandler

Basically if I want different search paths in my compilations, I need to mess with ICompiler's default include handler, which is meh.

So if we could have an optional (default value nullptr) field in the compiler options struct above, then we use the provided IncludeHandler, else use the compiler's default.


include/nbl/asset/ICPUShader.h
```diff
@@ -70,26 +56,32 @@ class NBL_API ICPUShader : public IAsset, public IShader
...
 		}
 
-		const ICPUBuffer* getSPVorGLSL() const { return m_code.get(); };
-		bool containsGLSL() const { return m_containsGLSL; }
-		
+		const ICPUBuffer* getContent() const { return m_code.get(); };
+
+		inline E_CONTENT_TYPE getContentType() const { return m_contentType; }
+		// TODO: find better function name
+		inline bool isContentRawCode() const
```
devshgraphicsprogramming, line 80: `isContentHighLevelLanguage`

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,138 @@
+// Copyright (C) 2018-2022 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+#ifndef __NBL_ASSET_I_SHADER_COMPILER_H_INCLUDED__
+#define __NBL_ASSET_I_SHADER_COMPILER_H_INCLUDED__
```
devshgraphicsprogramming, line 6: things to fix going forward, remove `__` from header guards, its reserved for system/compiler preprocessor defines apparently

Erfan-Ahmadi: So it will be `NBL_ASSET_I_SHADER_COMPILER_H_INCLUDED`?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,138 @@
...
+
+		Internally function "disables" all preprocessor directives (so that they're not processed by preprocessor) except `#include` (and also `#version` and `#pragma shader_stage`).
+		Note that among the directives there may be include guards. Because of that, _maxSelfInclusionCnt parameter is provided.
+
+		@param _maxSelfInclusionCnt Max self-inclusion count of possible file being #include'd. If no self-inclusions are allowed, should be set to 0.
+
+		@param _originFilepath Path to not necesarilly existing file whose directory will be base for relative (""-type) top-level #include's resolution.
+			If _originFilepath is non-path-like string (e.g. "whatever" - no slashes), the base directory is assumed to be "." (working directory of your executable). It's important for it to be unique.
+
+		@returns Shader containing logically same GLSL code as input but with #include directives resolved.
```
devshgraphicsprogramming, line 40: update comment, GLSL-> High Level

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,138 @@
...
+#include "nbl/asset/utils/ISPIRVOptimizer.h"
+
+namespace nbl::asset
+{
+
+class NBL_API IShaderCompiler : public core::IReferenceCounted
+{
+	public:
+
+		IShaderCompiler(system::ISystem* _s);
```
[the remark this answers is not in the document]
Erfan-Ahmadi: I think it was because of circular references? I thought about this too but If System also holds a IGLSLCompiler then it'll be a circular ref

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,138 @@
...
+			system::IFile* _sourcefile,
+			IShader::E_SHADER_STAGE _stage,
+			const char* _originFilepath,
+			uint32_t _maxSelfInclusionCnt = 4u,
+			system::logger_opt_ptr logger = nullptr) const;
+		
+		/*
+			Creates a formatted copy of the original
+
+			@param original An original glsl shader (must contain glsl and not be a dummy object of be a nullptr).
```
devshgraphicsprogramming, line 59: glsl -> high level

had a type "of be a nullptr", should be "or be a nullptr"

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,138 @@
...
+			@param fmt A string with c-like format, which will be filled with data from ...args
+			@param ...args Data to fill fmt with
+			@returns shader containing fmt filled with ...args, placed before the original code.
+
+			If original == nullptr, the output buffer will only contain the data from fmt. If original code contains #version specifier,
+			then the filled fmt will be placed onto the next line after #version in the output buffer. If not, fmt will be placed into the
+			beginning of the output buffer.
```
devshgraphicsprogramming, line 66: to make this work both for HLSL and GLSL, we need to expose the position where we'll paste the formatted string and give GLSL compiler a function to find the first line after the `#version` directive and an overload of this static function which misses the position parameter

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,138 @@
...
+			std::string_view origCode;
+			auto outCode = reinterpret_cast<char*>(outBuffer->getPointer());
+			if (original!=nullptr)
+			{
+				origCode = std::string_view(reinterpret_cast<const char*>(original->getContent()->getPointer()),origLen);
+				auto start = origCode.find("#version");
+				auto end = origCode.find("\n",start);
+				if (end!=std::string_view::npos)
+					versionDirectiveLength = end+1u;
+			}
```
devshgraphicsprogramming, line 111: this is GLSL specific, the position for inserting the formatted string needs to come from the outside

include/nbl/asset/utils/IShaderCompiler.h
Erfan-Ahmadi: I'll leave this for the next PR for whoever works on integrating DXC and compiling HLSL
Note that this PR is only to have IShaderCompiler cleaned up and nice + IncludeHandler related files
and only the interface for HLSLCompiler not the implementation. then I'll be done. (I'll update the PR notes)

include/nbl/asset/utils/IShaderCompiler.h
Erfan-Ahmadi: Ah ok, I guess I'll do it in this PR

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: no, `_NBL_ASSET_I_SHADER_COMPILER_H_INCLUDED_`

include/nbl/asset/utils/IShaderCompiler.h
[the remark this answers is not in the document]
Erfan-Ahmadi: then I'll change those to smart_refctd

include/nbl/asset/IShader.h
```diff
@@ -53,8 +53,13 @@ class NBL_API IShader : public virtual core::IReferenceCounted
 			ESS_ALL = 0xffffffff
 		};
 
-		struct buffer_contains_glsl_t {};
-		_NBL_STATIC_INLINE const buffer_contains_glsl_t buffer_contains_glsl = {};
+		enum E_CONTENT_TYPE : uint8_t
```
devshgraphicsprogramming, line 56: `enum class` for this enum and `E_SHADER_STAGE` if possible

src/nbl/asset/interchange/CGLSLLoader.cpp
```diff
@@ -45,7 +45,7 @@ SAssetBundle CGLSLLoader::loadAsset(system::IFile* _file, const IAssetLoader::SA
 		return {};
 	}
 
-	auto shader = core::make_smart_refctd_ptr<ICPUShader>(reinterpret_cast<char*>(source), found->second, filename.string());
+	auto shader = core::make_smart_refctd_ptr<ICPUShader>(reinterpret_cast<char*>(source), found->second, IShader::ECT_GLSL, filename.string());
```
devshgraphicsprogramming: Tasks for @deprilula28 :
1. Maybe we should load to `ICPUBuffer` directly and create the `ICPUShader` from that?
2. Create a `CHLSLLoader` similar to `CGLSLLoader` but:
2a. Only file extensions it associates with are `*.h` and `*.hlsl`
2b. Check loadability by checking that file is under 64MB and regexping the file for any reserved keywords https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-appendix-keywords
2c. it returns a `ICPUShader` not an `ICPUSpecializedShader`

Then I want/need a SPIR-V Loader and DXIL Loader, these will require using SPIR-V Tools and DXC to check loadability (validitity) and check integrity before returnin the asset.

@Erfan-Ahmadi we need a `ESS_LIBRARY` to deal with HLSL as there's no way to infer the stage, unless we start using file extensions like `.vert.hlsl`. Even then we'd still need `ESS_LIBRARY` to deal with stageless bits of HLSL we want to link later.

src/nbl/video/COpenGLShader.h
```diff
@@ -17,26 +17,20 @@ namespace nbl::video
 class COpenGLShader : public IGPUShader
```
devshgraphicsprogramming, line 17: nuke OpenGL!

Erfan-Ahmadi: Oh no, everything?! Can I do it later? I'm scared :))

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -530,7 +530,7 @@ class CVulkanLogicalDevice final : public ILogicalDevice
                     shaderStage, cpushader->getFilepathHint().c_str(), 4u, logger);
 
             spirv = m_physicalDevice->getGLSLCompiler()->compileSPIRVFromGLSL(
```
devshgraphicsprogramming, line 532: everything up until here would be the same for HLSL and GLSL, also would be to take a SPIR-V optimizer + compiler options struct or something

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -539,10 +539,14 @@ class CVulkanLogicalDevice final : public ILogicalDevice
                 logger,
                 m_physicalDevice->getLimits().spirvVersion);
         }
-        else
+        else if(cpushader->getContentType() == asset::ICPUShader::E_CONTENT_TYPE::ECT_SPIRV)
         {
             spirv = core::smart_refctd_ptr_static_cast<asset::ICPUBuffer>(source->clone(1u));
```
devshgraphicsprogramming, line 544: why does `CVulkanShader` need to hold onto a copy of the SPIR-V >

src/nbl/video/COpenGLShader.h
devshgraphicsprogramming: just the implementations of the functions, slap `assert(false)` in them

src/nbl/video/CVulkanLogicalDevice.h
Erfan-Ahmadi: I think you asked @achalpandeyy a similar question a long time ago in an old PR and I can't recall the answer but I remember you were satisfied with it.
apparently in CVulkanShader:
```c
    // We want to use the SPIR-V optimizer later.
    core::smart_refctd_ptr<asset::ICPUBuffer> m_spirv;
```

Erfan-Ahmadi: But if we use the SPIRV optimizer sooner here then there is no need I think to hold this in CVulkanShader

src/nbl/video/CVulkanLogicalDevice.h
devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/432/files#r1017795668 dont take a SPIR-V optimizer anymore for `createSpecializedShader`, take it here

src/nbl/asset/utils/CShaderIntrospector.cpp
```diff
@@ -119,8 +119,10 @@ const CIntrospectionData* CShaderIntrospector::introspect(const ICPUShader* _sha
 
         return introspectSPV(spvShader.get());
     }
-    else
+    else if (_shader->getContentType() == ICPUShader::E_CONTENT_TYPE::ECT_SPIRV)
         return introspectSPV(_shader);
+    else
+        return nullptr; // TODO[ShaderCompiler] ?
```
devshgraphicsprogramming, line 125: else its a serious bug

src/nbl/asset/utils/CShaderIntrospector.cpp
```diff
@@ -101,15 +101,15 @@ const CIntrospectionData* CShaderIntrospector::introspect(const ICPUShader* _sha
...
+        auto end = begin+_shader->getContent()->getSize();
         std::string glsl(begin,end);
         ICPUShader::insertDefines(glsl,_params.extraDefines);
         auto glslShader_woIncludes = m_glslCompiler->resolveIncludeDirectives(glsl.c_str(), _shader->getStage(), _shader->getFilepathHint().c_str());
         auto spvShader = m_glslCompiler->createSPIRVFromGLSL(
-            reinterpret_cast<const char*>(glslShader_woIncludes->getSPVorGLSL()->getPointer()),
+            reinterpret_cast<const char*>(glslShader_woIncludes->getContent()->getPointer()),
             glslShader_woIncludes->getStage(),
             _params.entryPoint,
             glslShader_woIncludes->getFilepathHint().c_str()
```
devshgraphicsprogramming, line 115: ok you need to be able to compile the HLSL on-demand as well

src/nbl/asset/utils/CShaderIntrospector.cpp
Erfan-Ahmadi: wdym?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -125,12 +123,228 @@ class NBL_API IShaderCompiler : public core::IReferenceCounted
...
 
-		IIncludeHandler* getIncludeHandler() { return m_inclHandler.get(); }
-		const IIncludeHandler* getIncludeHandler() const { return m_inclHandler.get(); }
+		class NBL_API IIncludeLoader : public core::IReferenceCounted
+		{
+		public:
+			virtual std::string getInclude(const system::path& searchPath, const std::string& includeName) const = 0;
+		};
+
+		class NBL_API IIncludeGenerator : public core::IReferenceCounted
```
devshgraphicsprogramming, line 132: btw you can delete any `NBL_API` in the files you're retouching, only `NBL_API2` actually does anything for DLLs

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -125,12 +123,228 @@ class NBL_API IShaderCompiler : public core::IReferenceCounted
...
+			{
+				core::vector<std::string> args;
+
+				std::stringstream ss{ _path };
+				std::string arg;
+				while (std::getline(ss, arg, '/'))
+					args.push_back(std::move(arg));
+
+				return args;
+			}
```
devshgraphicsprogramming, line 169: you can kick the inline function implementations to `IShaderCompiler.cpp`

devshgraphicsprogramming: don't forget the NBL_API2 though

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -125,12 +123,228 @@ class NBL_API IShaderCompiler : public core::IReferenceCounted
...
+				const size_t size = f->getSize();
+
+				std::string contents(size, '\0');
+				system::IFile::success_t succ;
+				f->read(succ, contents.data(), 0, size);
+				const bool success = bool(succ);
+				assert(success);
+
+				return contents;
+			}
```
devshgraphicsprogramming, line 201: kick to CPP file

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -125,12 +123,228 @@ class NBL_API IShaderCompiler : public core::IReferenceCounted
...
+
+		protected:
+			core::smart_refctd_ptr<system::ISystem> m_system;
+		};
+
+		class NBL_API CIncludeFinder : public core::IReferenceCounted
```
devshgraphicsprogramming, line 207: mark as `NBL_API2` and boot implementations of everything to cpp file

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -110,27 +111,17 @@ namespace nbl::asset::impl
...
-            const bool reqBuiltin = asset::IIncludeHandler::isBuiltinPath(_requested_source);
-            if (!reqFromBuiltin && !reqBuiltin)
-            {
-                //While #includ'ing a builtin, one must specify its full path (starting with "nbl/builtin" or "/nbl/builtin").
-                //  This rule applies also while a builtin is #includ`ing another builtin.
-                //While including a filesystem file it must be either absolute path (or relative to any search dir added to asset::iIncludeHandler; <>-type),
-                //  or path relative to executable's working directory (""-type).
-                relDir = std::filesystem::path(_requesting_source).parent_path();
-            }
-
```
devshgraphicsprogramming: this needs to stay, it basically prevents you from breaking out of `nbl/builtin` once you're inside it (makes sure our "system headers" can't include anything from the outside, which saves us a bunch of bugs)

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -110,27 +111,17 @@ namespace nbl::asset::impl
...
+            std::filesystem::path relDir = std::filesystem::path(_requesting_source).parent_path();
             std::filesystem::path name = (_type == shaderc_include_type_relative) ? (relDir/_requested_source) : (_requested_source);
 
-            if (!reqBuiltin)
+            // TODO: did I change this correctly?!
+            if (std::filesystem::exists(name))
```
devshgraphicsprogramming, line 118: probably yes

devshgraphicsprogramming: even if the above rel-dir check stays

src/nbl/video/COpenGLShader.h
Erfan-Ahmadi: Can you get someone else to do this? I don't want to.

include/nbl/video/SPhysicalDeviceLimits.h
```diff
@@ -2,7 +2,7 @@
 #define _NBL_VIDEO_S_PHYSICAL_DEVICE_LIMITS_H_INCLUDED_
 
 #include <type_traits>
-#include "nbl/asset/utils/IGLSLCompiler.h" // asset::IGLSLCompiler::E_SPIRV_VERSION
+#include "nbl/asset/utils/CGLSLCompiler.h" // asset::CGLSLCompiler::E_SPIRV_VERSION
```
devshgraphicsprogramming, line 5: `IShaderCompiler` or `CompilerSet` now?

include/nbl/video/IPhysicalDevice.h
```diff
@@ -549,7 +549,7 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
 
         //
         inline system::ISystem* getSystem() const {return m_system.get();}
-        inline asset::IGLSLCompiler* getGLSLCompiler() const {return m_GLSLCompiler.get();}
+        inline asset::CGLSLCompiler* getGLSLCompiler() const {return m_GLSLCompiler.get();}
```
devshgraphicsprogramming, line 552: lets make it hold the new compilerSet you've introduced

include/nbl/asset/utils/CShaderIntrospector.h
```diff
@@ -156,7 +156,7 @@ class NBL_API CShaderIntrospector : public core::Uncopyable
 		size_t calcBytesizeforType(spirv_cross::Compiler& _comp, const spirv_cross::SPIRType& _type) const;
 
 	private:
-		const IGLSLCompiler* m_glslCompiler;
+		const CGLSLCompiler* m_glslCompiler;
```
devshgraphicsprogramming, line 159: make it hold a smartpointer to a compiler set?

include/nbl/asset/ICPUShader.h
```diff
@@ -29,35 +29,21 @@ class NBL_API ICPUShader : public IAsset, public IShader
...
 		ICPUShader(
-			const char* _glsl,
+			const char* code,
 			const E_SHADER_STAGE stage,
-			std::string&& filepathHint)
-			: ICPUShader(core::make_smart_refctd_ptr<ICPUBuffer>(strlen(_glsl) + 1u), true, 
-				stage, std::move(filepathHint))
+			E_CONTENT_TYPE contentType,
+			std::string&& filepathHint) 
+			: ICPUShader(core::make_smart_refctd_ptr<ICPUBuffer>(strlen(code) + 1u), stage, contentType, std::move(filepathHint))
```
devshgraphicsprogramming, line 44: assert that content is not SPIR-V, otherwise strlen goes boom

include/nbl/asset/utils/CGLSLCompiler.h
```diff
@@ -0,0 +1,174 @@
...
+		CGLSLCompiler(core::smart_refctd_ptr<system::ISystem>&& system);
+
+		struct SOptions : IShaderCompiler::SOptions
+		{
+			void setCommonData(const IShaderCompiler::SOptions& opt)
+			{
+				static_cast<IShaderCompiler::SOptions&>(*this) = opt;
+			}
+
+			virtual IShader::E_CONTENT_TYPE getCodeContentType() const override { return IShader::E_CONTENT_TYPE::ECT_GLSL; };
```
devshgraphicsprogramming, line 28: the final (which you can add to this struct btw) derivation doesn't need `virtual`

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,253 @@
...
+			@maxSelfInclusionCount used only when includeFinder is not nullptr
+			@genDebugInfo Requests compiler to generate debug info (most importantly objects' names).
+				The engine, while running on OpenGL, won't be able to set push constants for shaders loaded as SPIR-V without debug info.
+		*/
+		struct SOptions
+		{
+			IShader::E_SHADER_STAGE stage = IShader::E_SHADER_STAGE::ESS_UNKNOWN;
+			E_SPIRV_VERSION targetSpirvVersion = E_SPIRV_VERSION::ESV_1_6;
+			const char* entryPoint = nullptr;
+			const char* sourceIdentifier = nullptr;
```
devshgraphicsprogramming, line 128: could be `std::string_view`s

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,253 @@
...
+			std::string* outAssembly = nullptr;
```
devshgraphicsprogramming, line 129: btw what did we use this for ?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -0,0 +1,253 @@
...
+			@targetSpirvVersion spirv version
+			@entryPoint entryPoint
+			@sourceIdentifier String that will be printed along with possible errors as source identifier, and used as include's requestingSrc
+			@outAssembly Optional parameter; if not nullptr, SPIR-V assembly is saved in there.
+			@spirvOptimizer Optional parameter;
+			@logger Optional parameter; used for logging errors/info
+			@includeFinder Optional parameter; if not nullptr, it will resolve the includes in the code
+			@maxSelfInclusionCount used only when includeFinder is not nullptr
+			@genDebugInfo Requests compiler to generate debug info (most importantly objects' names).
+				The engine, while running on OpenGL, won't be able to set push constants for shaders loaded as SPIR-V without debug info.
```
devshgraphicsprogramming, line 121: anything non-vulkan, basically you can't recover the names of original variables with `CShaderIntrospector` without debug info

By variables we mean names of PC/SSBO/UBO blocks, as they're essentially instantiations of structs with custom packing.

src/nbl/asset/utils/CShaderIntrospector.cpp
```diff
@@ -101,25 +101,31 @@ const CIntrospectionData* CShaderIntrospector::introspect(const ICPUShader* _sha
...
 
-    if (_shader->containsGLSL())
+    auto begin = reinterpret_cast<const char*>(_shader->getContent()->getPointer());
+    auto end = begin + _shader->getContent()->getSize();
+    std::string code(begin, end);
+    if (_shader->getContentType() != ICPUShader::E_CONTENT_TYPE::ECT_SPIRV)
+        ICPUShader::insertDefines(code, _params.extraDefines);
+
+    // TODO:
+    core::smart_refctd_ptr<asset::CCompilerSet> compilerSet;
```
devshgraphicsprogramming, line 111: ask for the compiler set and options  in `SIntrospectionParams`

then patch up any options that don't make sense, like `genDebugInfo`, `optimizer`, `sourceIdentifier`, `includeFinder`

include/nbl/asset/utils/CGLSLCompiler.h
```diff
@@ -0,0 +1,174 @@
...
+		core::smart_refctd_ptr<ICPUShader> createSPIRVFromGLSL(
+			system::IFile* _sourcefile,
+			IShader::E_SHADER_STAGE _stage,
+			const char* _entryPoint,
+			const char* _compilationId,
+			const ISPIRVOptimizer* _opt = nullptr,
+			bool _genDebugInfo = true,
+			std::string* _outAssembly = nullptr,
+			system::logger_opt_ptr logger = nullptr,
+			const E_SPIRV_VERSION targetSpirvVersion = E_SPIRV_VERSION::ESV_1_6) const;
```
devshgraphicsprogramming, line 89: nuke them while you're at it, only used by OpenGL backend?

include/nbl/asset/utils/CGLSLCompiler.h
```diff
@@ -0,0 +1,174 @@
...
+
+		@params code high level code
+		@params options see IShaderCompiler::SOptions
+			- entryPoint Must be "main" since shaderc does not allow other entry points for GLSL. Kept with hope that shaderc will drop that requirement.
+
+		@returns Shader containing SPIR-V bytecode.
+		*/
+		core::smart_refctd_ptr<ICPUBuffer> compileToSPIRV(const char* code, const CGLSLCompiler::SOptions& options) const;
+
+		core::smart_refctd_ptr<ICPUShader> createSPIRVShader(const char* code, const CGLSLCompiler::SOptions& options) const;
```
devshgraphicsprogramming, line 52: why do we have one for the buffer and one for the shader output?

include/nbl/asset/utils/CHLSLCompiler.h
```diff
@@ -0,0 +1,59 @@
...
+		CHLSLCompiler(core::smart_refctd_ptr<system::ISystem>&& system);
+
+		struct SOptions : IShaderCompiler::SOptions
+		{
+			// TODO: Add extra dxc options
+
+			void setCommonData(const IShaderCompiler::SOptions& opt)
+			{
+				static_cast<IShaderCompiler::SOptions&>(*this) = opt;
+			}
```
devshgraphicsprogramming, line 28: you can hoist this to `IShaderCompiler::SOptions` btw, a derived `SOptions` will downcast itself just fine

include/nbl/asset/utils/CHLSLCompiler.h
```diff
@@ -0,0 +1,59 @@
...
+		core::smart_refctd_ptr<ICPUBuffer> compileToSPIRV(const char* code, const CHLSLCompiler::SOptions& options) const;
+
+		core::smart_refctd_ptr<ICPUShader> createSPIRVShader(const char* code, const CHLSLCompiler::SOptions& options) const;
+
+		core::smart_refctd_ptr<ICPUShader> createSPIRVShader(system::IFile* sourceFile, const CHLSLCompiler::SOptions& options) const;
+
+		/*
+		 If original code contains #version specifier,
+			then the filled fmt will be placed onto the next line after #version in the output buffer. If not, fmt will be placed into the
+			beginning of the output buffer.
```
devshgraphicsprogramming, line 42: fix up the comment, I guess you're providing the same signature as the GLSL one so we don't need to care about specifying the position?

include/nbl/asset/utils/CHLSLCompiler.h
```diff
@@ -0,0 +1,59 @@
...
+		static core::smart_refctd_ptr<ICPUShader> createOverridenCopy(const ICPUShader* original, const char* fmt, Args... args)
+		{
+			return IShaderCompiler::createOverridenCopy(original, 0u, fmt, args...);
+		}
+
+		static inline const char* getStorageImageFormatQualifier(const asset::E_FORMAT format)
+		{
+			// TODO
+			return "";
+		}
```
devshgraphicsprogramming, line 54: comment out for now, it will be @achalpandeyy 's worry when he ports the Bloom or Convolution Blit to HLSL

src/nbl/asset/utils/CGLSLCompiler.cpp
```diff
@@ -0,0 +1,155 @@
...
+    size_t fileSize = sourceFile->getSize();
+    std::string code(fileSize, '\0');
+
+    system::IFile::success_t success;
+    sourceFile->read(success, code.data(), 0, fileSize);
+    if (success)
+        return createSPIRVShader(code.c_str(), options);
+    else
+        return nullptr;
+}
```
devshgraphicsprogramming, line 105: hmm, could be hoisted to `IShaderCompiler` methinks

src/nbl/asset/utils/CGLSLCompiler.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+core::smart_refctd_ptr<ICPUShader> CGLSLCompiler::createSPIRVShader(const char* code, const CGLSLCompiler::SOptions& options) const
+{
+    auto spirv = compileToSPIRV(code, options);
+    if (spirv)
+        return core::make_smart_refctd_ptr<asset::ICPUShader>(std::move(spirv), options.stage, IShader::E_CONTENT_TYPE::ECT_GLSL, options.sourceIdentifier);
+    else
+        return nullptr;
+}
```
devshgraphicsprogramming, line 92: not sure why don't go straight to `ICPUShader` always and there is an `ICPUBuffer`, but this should be hoisted to `IShaderCompiler`

src/nbl/asset/utils/CGLSLCompiler.cpp
```diff
@@ -0,0 +1,155 @@
...
+        else
+        {
+            options.logger.log("shaderc requires entry point to be \"main\" in GLSL", system::ILogger::ELL_ERROR);
+        }
+    }
+    else
+    {
+        options.logger.log("code is nullptr", system::ILogger::ELL_ERROR);
+    }
+    return outSpirv;
```
devshgraphicsprogramming, line 82: minor stylistic issue, to avoid "Python Levels of Indentation" esp if a function returns a pointer (dumb or smart) its better to do 
```cpp
if (failCondition)
{
   log("the what and why", system::ILogger::ELL_ERROR);
   return nullptr;
}
   
// rest of code
```

src/nbl/asset/utils/CGLSLCompiler.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+            shaderc::AssemblyCompilationResult asm_res;
+            shaderc::SpvCompilationResult bin_res;
+            if (options.outAssembly)
+            {
+                asm_res = comp.CompileGlslToSpvAssembly(code, glsl_len, stage, options.sourceIdentifier ? options.sourceIdentifier : "", shadercOptions);
+                options.outAssembly->resize(std::distance(asm_res.cbegin(), asm_res.cend()));
+                memcpy(options.outAssembly->data(), asm_res.cbegin(), options.outAssembly->size());
+                bin_res = comp.AssembleToSpv(options.outAssembly->data(), options.outAssembly->size(), shadercOptions);
+            }
```
devshgraphicsprogramming, line 54: @Crisspl why did we need to keep the assembly around?

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -0,0 +1,384 @@
...
+        return;
+
+    auto itr = m_generators.begin();
+    for (; itr != m_generators.end(); ++itr)
+    {
+        auto str = (*itr)->getPrefix();
+        if (str.compare(generator->getPrefix()) <= 0) // Reverse Lexicographic Order
+            break;
+    }
+    m_generators.insert(itr, generator);
```
devshgraphicsprogramming, line 318: `lower_bound` takes a comparator functor argument btw, you could make this binary-insertion sorted

devshgraphicsprogramming:
Shouldn't the target branch be `master` instead of `hlsl` ?

include/nbl/asset/utils/CGLSLCompiler.h
Erfan-Ahmadi: yea, I leaked an old habbit of mine into the code.
my logic was even if it's a derivation it's a virtual function.
will fix it

include/nbl/asset/utils/CGLSLCompiler.h
Erfan-Ahmadi: these functions map to previous `compileSPIRVFromGLSL` and `createSPIRVFromGLSL`
Sometimes I think sometimes we use ICPUShader and ICPUBuffer interchangabily. I had plans in mind to go over all these and see if I can clean things up.
(but sometimes ICPUShader is returned because stage will be inferred from the code using pragmas, but not in this case)

include/nbl/asset/utils/CGLSLCompiler.h
Erfan-Ahmadi: no it was used by both backends but I didn't want to change/fix opengl 
I will most probably remove this and comment the places used to ignore compile errors

include/nbl/asset/utils/CHLSLCompiler.h
Erfan-Ahmadi: yea good point 😄

src/nbl/video/IOpenGL_PhysicalDeviceBase.h
```diff
@@ -28,7 +28,7 @@ class IOpenGLPhysicalDeviceBase : public IPhysicalDevice
 
 		IOpenGLPhysicalDeviceBase(
 			core::smart_refctd_ptr<system::ISystem>&& s,
-			core::smart_refctd_ptr<asset::IGLSLCompiler>&& glslc,
+			core::smart_refctd_ptr<asset::CGLSLCompiler>&& glslc,
```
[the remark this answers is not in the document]
Erfan-Ahmadi: not yet, but in my plan ;)
We need an `m_api` in IPhysicalDevice so anytime we need the compiler: logicalDevice -> physicalDevice -> apiConnection -> smart_refctd<CCompilerSet> 
and if that's too many indirections, we can just hold the refctd inside the logical device.

include/nbl/asset/ICPUShader.h
Erfan-Ahmadi: ok, but why?

include/nbl/asset/utils/IShaderCompiler.h
Erfan-Ahmadi: I thought about it, I'm just not used to the idea :D

include/nbl/asset/utils/IShaderCompiler.h
Erfan-Ahmadi: outAssembly? I have no idea where it's used but it's implemented for glsl compiler. (if it's not nullptr it will first compile to asm then from asm to spirv)

src/nbl/asset/utils/CGLSLCompiler.cpp
Erfan-Ahmadi: no because the compile function is not virtual. If it was it had to take `IShaderCompiler::SOptions` instead of `HLSL/GLSLCompiler::SOptions` then we would've needed to do the conversions inside the compile functions instead of `CCompilerSet` meaning get the `getType` of Options and cast them to appropriate struct.

I can make them virtual and do the second option. I just had to make a choice and It was weird to me to do those casts and checks inside the compile functions.

src/nbl/asset/utils/CGLSLCompiler.cpp
Erfan-Ahmadi: yea I hoped you wouldn't notice that, cause it's a syntax I use on my own.
I prefer the this way cause all the failure points can be viewed next to each other and there is only a single point of return which can be inspected while debugging.
I'll change the style to your preference though.

src/nbl/asset/utils/IShaderCompiler.cpp
Erfan-Ahmadi: ok, good idea, I'll use lower_bound instead of my custom for

src/nbl/asset/utils/CGLSLCompiler.cpp
Erfan-Ahmadi: I can't find any other reference to `outAssembly` outside of `IGLSLCompiler` even in the examples.
NUKE IT?!

include/nbl/asset/utils/CGLSLCompiler.h
Erfan-Ahmadi: Anyways, 99% of the codebase uses the Buffer one so If I merge them into one function that returns ICPUShader I'll have to touch up many places

include/nbl/asset/ICPUShader.h
devshgraphicsprogramming: SPIR-V is not null terminated until a source string

src/nbl/asset/utils/CGLSLCompiler.cpp
devshgraphicsprogramming: image

Yes but actually do it, don't do blabbing on about it unlike some other people.

src/nbl/asset/utils/CGLSLCompiler.cpp
devshgraphicsprogramming: the code is identical

you can make the `createSPIRVShader(, const IShaderCompiler::SOptions&)` a protected virtual so no one abuses it :)

And then have the ones that take the actual proper otpion structs non virtual and call the virtual one

include/nbl/asset/utils/CGLSLCompiler.h
Erfan-Ahmadi: but now I investigated a bit and It seems I can merge them into one to return ICPUShader no problem

devshgraphicsprogramming: then do that

src/nbl/asset/utils/CGLSLCompiler.cpp
Erfan-Ahmadi: hmmm, I don't like this. We'll be casting back and forth multiple times. CCompilerSet forces the struct to be `CHLSLCompiler::SOptions` so then It calls `CHLSLCompiler::createSPIRVShader`  --calls-->  `createSPIRVShader(, const IShaderCompiler::SOptions&)` which may cast AGAIN to `CHLSLCompiler::SOptions` inside and do the same checks we do in CCompilerSet.

If you wanna go this route I suggest just having a public virtual `createSPIRVShader(, const IShaderCompiler::SOptions&)` and do checks for type and cast inside the function

devshgraphicsprogramming: > If you wanna go this route I suggest just having a public virtual createSPIRVShader(, const IShaderCompiler::SOptions&) and do checks for type and cast inside the function

yes but that opens the door to abuse, which is to pass the wrong option struct.

Erfan-Ahmadi: Seem like this is the only way if you want to make this function that takes a file virtual.

but how are you going to prevent people from abusing this `createSPIRVShader(system::IFile* sourceFile, const IShaderCompiler::SOptions& options)`?

When we take a file we take IShaderCompiler::SOptions in the virtual function but when we want to give the source directly we have to use CHLSL/GLSLCompiler::SOptions in the non-virtual functions. that's just inconsistent

src/nbl/asset/utils/CShaderIntrospector.cpp
Erfan-Ahmadi: the only thing that makes sense is entry point, you want me to make `SIntrospectionParams` hold a CompilerOption and patch it instead of just getting an `entryPoint`? doesn't make sense to me

devshgraphicsprogramming: you can make the introspection method take `CCompulerSet*` and `IShaderCompiler::SOptions`

then you can ignore options that get in the way of introspection, such as an `optimizer` or a lack of debug info.

Futhermore you can substitute `sourceIndentier` and `includeFinder` if they're invalid

Erfan-Ahmadi: The entry point in `SIntrospectionParams` is also a member in `IShaderCompiler::SOptions`, I think I will remove that but IntrospectionParams will be only this .
```c
struct SIntrospectionParams
{
	// const char* entryPoint;
	// if the Shader is already compiled to SPV, it will ignore this member
	core::SRange<const char* const> extraDefines;
};
```

There is also a Comparator thing which I can't make sense of. https://github.com/Devsh-Graphics-Programming/Nabla/blob/compiler_cleanup/include/nbl/asset/utils/CShaderIntrospector.h#L180
I only know that it's being used to cache introspections

So I need to also account for the CompileOptions in the caching as well. because some language specific compile options might affect how the bindings or resources are positioned, and then It becomes a pain in the ass cause we'll have to cache the `CHLSL/CGLSLCompiler::SOptions` as well, maybe in a union?!

But at the end of the day I think we'll have our compile options inside the `IntrospectionParams`; but the CCompilerSet would be retrieved via function params as follows:

```c
CShaderIntrospector::introspect(const ICPUShader* _shader, CCompilerSet* compilerSet, const SIntrospectionParams& _params)
```

TLDR; I'm mind-f@!ked and this seems non-trivial to fix

devshgraphicsprogramming: > There is also a Comparator thing which I can't make sense of. https://github.com/Devsh-Graphics-Programming/Nabla/blob/compiler_cleanup/include/nbl/asset/utils/CShaderIntrospector.h#L180
I only know that it's being used to cache introspections

That's because the SPIR-V produced by a compilation differs on the entry point and the defines you have before.

So if you want to cache introspections for a given `ICPUShader` then you need to make sure that the entry point and defines were identical, that's all.

Erfan-Ahmadi: ok, but entry point in the params is duplicated (once in `SIntrospectionParams` and once in IShaderCompiler::Options)
Another thing: this is based on the assumption the compiler options taken in the params doesn't affect the introspection (even if it's patched), otherwise it should be cached

src/nbl/video/IOpenGL_PhysicalDeviceBase.h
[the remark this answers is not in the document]
devshgraphicsprogramming: APIConnection holds physical devices, that's what I meant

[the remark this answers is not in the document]
Erfan-Ahmadi: I just found out I refactored all this shit in my other branch. that's one of the reasons I had to finalize that first, I moved the compile to api connection because it was getting created for every physicalDevice.
https://github.com/Devsh-Graphics-Programming/Nabla/blob/erfan_image_upload_utils/src/nbl/video/IOpenGL_PhysicalDeviceBase.h#L29
After I merge that I want to make a few changes:
- Even remove the compiler from apiconnection (doesn't make much sense to me)
- The user will create a compilerSet and gives a `smart_refctd<CCompilerSet>` that to ILogicalDevice::SCreationParams, If it was nullptr, LogicalDevice will create one (we've discussed this before that logical device takes a compiler set to enable sharing)
- PhysicalDevice and APIConnection then have nothing to do with compilers and I will regain my sanity

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -0,0 +1,173 @@
+// Copyright (C) 2018-2020 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+#ifndef __NBL_ASSET_C_SHADER_INTROSPECTOR_H_INCLUDED__
+#define __NBL_ASSET_C_SHADER_INTROSPECTOR_H_INCLUDED__
```
devshgraphicsprogramming, line 6: update the header guard

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -0,0 +1,173 @@
...
+{
+    class ParsedIR;
+    class Compiler;
+    struct SPIRType;
+}
+
+namespace nbl::asset
+{
+
+class NBL_API CIntrospectionData : public core::IReferenceCounted
```
devshgraphicsprogramming, line 34: can we make it a nested class while we're at it?

PR #434, opened 2022-11-13
First hlsl port for Scan

kpentaris:
## Description
First commit for porting macros.glsl to hlsl. Renamed the file to binops.glsl, contains functors for and, or, xor etc

## Testing 
N/A

## TODO list:
Answer all TODO comments


AnastaZIuk:

include/nbl/builtin/hlsl/binops.hlsl
```diff
@@ -0,0 +1,72 @@
+// Copyright (C) 2022 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+#ifndef _NBL_BUILTIN_HLSL_BINOPS_INCLUDED_
+#define _NBL_BUILTIN_HLSL_BINOPS_INCLUDED_
+// TODO (PentaKon): Should this file be renamed to binops.hlsl ?
```
devshgraphicsprogramming, line 6: well its already named this

include/nbl/builtin/hlsl/binops.hlsl
```diff
@@ -0,0 +1,72 @@
...
+// TODO (PentaKon): Should we move isNPoT from algorithm.hlsl to this file ?
```
devshgraphicsprogramming, line 7: no cause its not a binop, but please move the less-than comparator (also add greater-than, and the 2 than-equal versions)

include/nbl/builtin/hlsl/binops.hlsl
```diff
@@ -0,0 +1,72 @@
...
+namespace nbl
+{
+namespace hlsl
+{
+// TODO (PentaKon): NBL_GLSL_EVAL? Keep as macro or don't use at all?
```
devshgraphicsprogramming, line 12: that thing is only to force the stringification of a define token, you dont need it here

include/nbl/builtin/hlsl/binops.hlsl
```diff
@@ -0,0 +1,72 @@
...
+};
+
+template<typename T>
+struct xor
+{
+    T operator()(const T lhs, const T rhs)
+    {
+        return lhs^rhs;
+    }
+};
```
devshgraphicsprogramming, line 40: you need to prefix their names with `bitwise_` to disambiguate

include/nbl/builtin/hlsl/binops.hlsl
```diff
@@ -0,0 +1,72 @@
...
+        return lhs>rhs ? rhs : lhs;
+    }
+};
+
+template<typename T>
+struct max
+{
+    T operator()(const T lhs, const T rhs)
+    {
+        return lhs>rhs ? lhs : rhs;
```
devshgraphicsprogramming, line 65: usually all comparators in C++ land are implement in the terms of the `<` operator

you should have
```cpp
template<typename T, class StatelessComparator>
struct min
{
    T operator()(const T lhs, const T rhs)
    {
       return StatelessComparator(lhs<rhs) ? lhs:rhs;
    }
};

// default
template<typename T>
struct min : min<T,less_than<T>> {};
```

devshgraphicsprogramming: if someone complains about multiple definitions, you can also try
```cpp
template<typename T, class Comparator>
struct min
{
    T operator()(const T lhs, const T rhs, inout Comparator comp)
    {
       return comp(lhs<rhs) ? lhs:rhs;
    }
};

// default
template<typename T>
struct min
{
    T operator()(const T lhs, const T rhs)
    {
       return min<T,less_than<T>>()(lhs,rhs,less_than<T>());
    }
};
```

devshgraphicsprogramming:
# How to port a subgroup op

```cpp
#ifndef NBL_GL_KHR_shader_subgroup_arithmetic
#include <nbl/builtin/hlsl/subgroup/shared_arithmetic_portability.glsl>
#endif

namespace nbl
{
namespace hlsl
{
namespace subgroup
{

#ifdef NBL_GL_KHR_shader_subgroup_arithmetic
namespace native
{

template<class Binop>
struct reduction;
template<class Binop>
struct exclusive_scan;
template<class Binop>
struct inclusive_scan;

template<>
struct reduction<binops::and>
{
    template<typename T>
    T operator()(const T x)
    {
        return subgroupAnd(x); // whatever the HLSL WaveIntrinsic equivalent is
    }
};
template<>
struct exclusive_scan<binops::and>
{
    template<typename T>
    T operator()(const T x)
    {
        return subgroupExclusiveAnd(x); // whatever the HLSL WaveIntrinsic equivalent is
    }
};
template<>
struct inclusive_scan<binops::and>
{
    template<typename T>
    T operator()(const T x)
    {
        return subgroupInclusiveAnd(x); // whatever the HLSL WaveIntrinsic equivalent is
    }
};

// TODO similar specializations for And, Xor, Or, Add, Mul, Min, Max
}
#endif


template<class Binop>
struct reduction
{
    template<typename T>
    T operator()(inout ScratchAccessor scratch, const T x)
    {
    #ifdef NBL_GL_KHR_shader_subgroup_arithmetic
        return native::reduction<Binop>()(x);
    #else
        return portability::reduction<Binop,ScratchAccessor>(scratch)(x);
    #endif
    }
};

template<class Binop>
struct exclusive_scan
{
    template<typename T>
    T operator()(inout ScratchAccessor scratch, const T x)
    {
    #ifdef NBL_GL_KHR_shader_subgroup_arithmetic
        return native::exclusive_scan<Binop>()(x);
    #else
        return portability::exclusive_scan<Binop,ScratchAccessor>(scratch)(x);
    #endif
    }
};

template<class Binop>
struct inclusive_scan
{
    template<typename T>
    T operator()(inout ScratchAccessor scratch, const T x)
    {
    #ifdef NBL_GL_KHR_shader_subgroup_arithmetic
        return native::inclusive_scan<Binop>()(x);
    #else
        return portability::inclusive_scan<Binop,ScratchAccessor>(scratch)(x);
    #endif
    }
};

}
}
}
```

include/nbl/builtin/hlsl/binops.hlsl
```diff
@@ -49,7 +49,7 @@ struct mul
 template<typename T, class Comparator>
 struct min
 {
-    T operator()(const T lhs, const T rhs, in Comparator comp)
+    T operator()(const T lhs, const T rhs, inout Comparator comp)
```
devshgraphicsprogramming, line 52: whatabout the `max` ?

kpentaris: ffs, sloppy morning commit

include/nbl/builtin/hlsl/macros.hlsl
```diff
@@ -0,0 +1,37 @@
...
+#define NBL_HLSL_XOR(X,Y) (NBL_HLSL_EVAL(X)^NBL_HLSL_EVAL(Y))
+
+#define NBL_HLSL_ADD(X,Y) (NBL_HLSL_EVAL(X)+NBL_HLSL_EVAL(Y))
+#define NBL_HLSL_SUB(X,Y) (NBL_HLSL_EVAL(X)-NBL_HLSL_EVAL(Y))
+#define NBL_HLSL_MUL(X,Y) (NBL_HLSL_EVAL(X)*NBL_HLSL_EVAL(Y))
+
+// https://github.com/google/shaderc/issues/1155
+// This issue still seems to exist
+// #define NBL_HLSL_MAX(X,Y) (((NBL_HLSL_EVAL(X))>(NBL_HLSL_EVAL(Y))) ? (NBL_HLSL_EVAL(X)):(NBL_HLSL_EVAL(Y)))
+// #define NBL_HLSL_MIN(X,Y) (((NBL_HLSL_EVAL(X))<(NBL_HLSL_EVAL(Y))) ? (NBL_HLSL_EVAL(X)):(NBL_HLSL_EVAL(Y)))
```
devshgraphicsprogramming, line 35: can you check if it works if you slap a big `NBL_HLSL_EVAL` around the `< ? :`

kpentaris: Tried this in compiler explorer

`int a = NBL_HLSL_EVAL(NBL_HLSL_MAX(5,4));`

and it preprocesses to

`int a = (((5)>(4)) ? (5):(4));`  👎

devshgraphicsprogramming: can you try and see if DXC will complain about `uint someArray[NBL_HLSL_MAX(5,4)]` ?

kpentaris: Doesn't complain and it actually seems to work. The DXIL IR contains the following
```
!9 = !DIGlobalVariable(name: "someArray", linkageName: "\01?someArray@@3QBIB", scope: !0, file: !1, line: 39, type: !10, isLocal: false, isDefinition: true)
!10 = !DICompositeType(tag: DW_TAG_array_type, baseType: !11, **size: 160**, align: 32, elements: !14)
```
The `size:160` matches the 5 uints the array is supposed to be.

devshgraphicsprogramming: but 5 uints is 20 bytes ?

devshgraphicsprogramming: ah okay, its in bits :D

devshgraphicsprogramming: we can uncomment NBL_HLSL_MIN and MAX then

kpentaris:
Documentation on HLSL Shader Model (for supported Wave Intrinsics)

[https://microsoft.github.io/DirectX-Specs/](url)

include/nbl/builtin/hlsl/binops.hlsl
```diff
@@ -37,6 +37,15 @@ struct bitwise_xor
     }
 };
 
+template<typename T>
+struct bitwise_add
```
devshgraphicsprogramming, line 41: `bitwise` add XD ?

mul, min?

 wtf!?

kpentaris: Don't we need them? Aren't these supposed to be used by reduce and scan ops? Or is it the prefix that's bad?

include/nbl/builtin/hlsl/subgroup/arithmetic_portability.hlsl
```diff
@@ -0,0 +1,95 @@
+#ifndef _NBL_BUILTIN_HLSL_SUBGROUP_ARITHMETIC_PORTABILITY_INCLUDED_
+#define _NBL_BUILTIN_HLSL_SUBGROUP_ARITHMETIC_PORTABILITY_INCLUDED_
+
+#ifndef NBL_GL_KHR_shader_subgroup_arithmetic
+#include <nbl/builtin/hlsl/subgroup/basic_portability.hlsl>
+#endif
+
+static uint4 WHOLE_WAVE = ~0;
```
devshgraphicsprogramming, line 8: `const` also why do you define it here?

include/nbl/builtin/hlsl/subgroup/arithmetic_portability_impl.hlsl
```diff
@@ -0,0 +1,435 @@
+#ifndef _NBL_BUILTIN_HLSL_SUBGROUP_ARITHMETIC_PORTABILITY_IMPL_INCLUDED_
+#define _NBL_BUILTIN_HLSL_SUBGROUP_ARITHMETIC_PORTABILITY_IMPL_INCLUDED_
+
+uint threadIndex : SV_GroupIndex; // Use HLSL terminology or follow GLSL (i.e. call them invocations instead of threads)?
```
devshgraphicsprogramming, line 4: follow GLSL, also dont define global variables like this

include/nbl/builtin/hlsl/subgroup/arithmetic_portability_impl.hlsl
```diff
@@ -0,0 +1,435 @@
...
+        return WaveMultiPrefixAnd(x, WHOLE_WAVE);
+    }
+};
+template<>
+struct inclusive_scan<binops::bitwise_and>
+{
+    template<typename T>
+    T operator()(const T x)
+    {
+        return WaveMultiPrefixAnd(x, WHOLE_WAVE) & x; // TODO (PentaKon): Should this use the binops::bitwise_and functor?
```
devshgraphicsprogramming, line 42: hmm interesting approach

include/nbl/builtin/hlsl/subgroup/arithmetic_portability.hlsl
kpentaris: Hmm I'm not really familiar with the HLSL compiler tricks yet so I'm not sure if something like `~0` should be defined once as static or be inlined was kinda waiting for a confirmation. Additionally my understanding is that with HLSL we want to avoid #defines whenever we can and replace them with static values.

include/nbl/builtin/hlsl/subgroup/arithmetic_portability_impl.hlsl
devshgraphicsprogramming: no don't use functor

include/nbl/builtin/hlsl/subgroup/arithmetic_portability_impl.hlsl
kpentaris: From my tests, HLSL doesn't allow defining SV_* variables inside a function scope. Do you have a suggestion?

include/nbl/builtin/hlsl/subgroup/basic_portability.hlsl
```diff
@@ -0,0 +1,29 @@
+#ifndef _NBL_BUILTIN_HLSL_SUBGROUP_BASIC_PORTABILITY_INCLUDED_
+#define _NBL_BUILTIN_HLSL_SUBGROUP_BASIC_PORTABILITY_INCLUDED_
+
+namespace nbl 
+{
+namespace hlsl
+{
+	// USE GLSL OR HLSL TERMINOLOGY? (Workgroup vs Threadgroup, Subgroup vs Wave, Invocation vs Thread)
```
devshgraphicsprogramming, line 8: GLSL terminology

include/nbl/builtin/hlsl/subgroup/basic_portability.hlsl
```diff
@@ -0,0 +1,29 @@
...
+	static const uint MinSubgroupSizeLog2 = 2;
+	static const uint MinSubgroupSize = 0x1 << MinSubgroupSizeLog2;
+	
+#ifdef NBL_HLSL_IMPL_GL_NV_shader_thread_group
+	static const uint MaxSubgroupSizeLog2 = 5;
+#elif defined(NBL_HLSL_IMPL_GL_AMD_gcn_shader)||defined(NBL_HLSL_IMPL_GL_ARB_shader_ballot)
+	static const uint MaxSubgroupSizeLog2 = 6;
+#else
+	static const uint MaxSubgroupSizeLog2 = 7;
+#endif
```
devshgraphicsprogramming, line 20: from `NBL_GL_` limits injected by compiler

kpentaris: I'm guessing you mean `NBL_GLSL_LIMIT_` defines that are located in src/nbl/video/IPhysicalDevice.cpp ? Not sure these are injected.

include/nbl/builtin/hlsl/subgroup/arithmetic_portability.hlsl
devshgraphicsprogramming: #define and #undef at the end after you don't need it

include/nbl/builtin/hlsl/subgroup/arithmetic_portability_impl.hlsl
devshgraphicsprogramming: IIRC you can just `#define gl_ SV_`

kpentaris: > IIRC you can just `#define gl_ SV_`

What I mean is, in HLSL you can't do something like
```
void myFun() {
  uint threadIdx: SV_GroupIndex;
}
```
Instead it has to be in the form of a parameter or global, or part of a struct.
For example:
```
void myFun(uint threadIdx: SV_GroupIndex) {
  ...
}
```

devshgraphicsprogramming: then do 
```hlsl
uint gl_LocalInvocationIndex : VS_GroupIndex;
```
or this stuff
https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#builtin-variables

PR #435, opened 2022-11-28
Double Convolution

achalpandeyy:
Currently we have:

the following kernels:
- Dirac (added in this PR)
- Box
- Triangle
- Gaussian
- Mitchell
- Kaiser

Kernel modifiers allow you to create a new kernel as a modifications of the above kernels.
Currently we have the following 4 kernel modifiers:
- Scaling & Stretching of a given kernel (removed in this PR)
- Different kernels per channel
- Derivative of a given kernel
- Convolution of two kernels

After removing `CScaledImageFilterKernel`, we will be eliminating the first one.

To eliminate the Scaling & Stretching modifier, means to make scaling and stretching an inherent property of the filter kernels.

### TODO
- [x] Implement numerical integration for "double" convolution.
- [x] Implement specializations for certain reconstruction and resampling kernel combinations.
- [x] Fixes for #436.
- [x] Make filters work with BC formats.
    - [x] Fix BC1 and BC3 decoding.
- [x] Add a `concept` for blittable kernels.
- [x] Remove `CScaledImageFilterKernel`.
- [x] Make kernel supports run-time changeable (affecting Mitchell and Kaiser).
- [x] Cherry-pick some fixes from the `ditt` branch. [Check this private Discord link.](https://discord.com/channels/593902898015109131/1043939034776473710/1062726446256234577)
- [x] Add derivative and smoothness in kernel weights.


include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -440,19 +441,22 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
 			const core::vectorSIMDf fOutExtent(outExtentLayerCount);
 			const auto fScale = fInExtent.preciseDivision(fOutExtent);
 			const auto halfTexelOffset = fScale*0.5f-core::vectorSIMDf(0.f,0.f,0.f,0.5f);
-			const auto startCoord =  [&halfTexelOffset,&scaledKernelX,&scaledKernelY,&scaledKernelZ]() -> core::vectorSIMDi32
+			const auto startCoord =  [&halfTexelOffset,state]() -> core::vectorSIMDi32
```
devshgraphicsprogramming, line 444: why are you passing `state` by value instead of by reference?

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+		}
+		return 0.0;
+	}
+};
+
+// Truncated Gaussian function, with stddev = 1.0, if you want a different stddev then you need to scale it.
+struct SGaussianFunction
+{
+	constexpr static inline float min_support = -3.f;
+	constexpr static inline float max_support = +3.f;
```
devshgraphicsprogramming, line 92: were the supports (truncation) hardcoded previously?

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+				const double normalizationFactor = core::inversesqrt(2.0 * core::PI<double>()) / std::erff(core::sqrt<double>(2.0) * double(min_support));
+				return normalizationFactor * exp2(-0.72134752 * x * x);
+			}
+			else if constexpr (derivative == 1)
+			{
+				return -x * SGaussianFunction::weight(x);
+			}
+			else if constexpr (derivative == 2)
+			{
+				return x * (x + 1.0) * SGaussianFunction::weight(x);
```
devshgraphicsprogramming, line 111: I think product rule dictates this should be `(x*x-1)*Gauss` so you're wrong here

it would be safest for you to define higher order derivatives "compile time recursively"
```cpp
template<uint32_t order>
class polynomial_t final : public std::array<int64_t,order+1>
{
        using base_t = std::array<int64_t,order+1>;

    public:
        constexpr polynomial_t() : base_t() {}
        constexpr polynomial_t(base_t _list) : base_t(_list) {}

        template<typename T>
        constexpr T evaluate(const T x) const
        {
            T result = T(base_t::back());
            for (int64_t i=order-1; i>=0ll; i--)
            {
                result *= x;
                result += base_t::operator[](i);
            }
            return result;
        }

        constexpr auto differentiate() const
        {
            polynomial_t<order-1> f_prime;
            for (uint32_t power=1; power<=order; power++)
                f_prime[power-1] = base_t::operator[](power)*power;
            return f_prime;
        }
};
```

then in your gaussian's weight you could have
```cpp
private:
    template<int32_t derivative>
    constexpr static polynomial_t<derivative> differentialPolynomialFactor()
    {
        static_assert(derivative>0,"DIFFERENTIATION ONLY!");
        if constexpr (derivative!=1)
        {
            constexpr auto f = differentialPolynomialFactor<derivative-1>();
            constexpr auto f_prime = f.differentiate();
            polynomial_t<derivative> retval;
            for (uint32_t power=0; power<f_prime.size(); power++)
                retval[power] = f_prime[power];
            for (uint32_t power=0; power<f.size(); power++)
                retval[power+1] -= f[power];
            return retval;
        }
        else
            return std::array<int64_t,2>{0,-1};
    }
```
and implement your `derivative>0` as
```cpp
else if constexpr (derivative>0)
   return differentialPolynomialFactor<derivative>().evaluate(x)*SGaussianFunction::weight(x);
```

devshgraphicsprogramming: https://godbolt.org/z/YchrEj1Tj

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+			}
+			else if constexpr (derivative == 3)
+			{
+				retval = core::mix(6.f * p3, 6.f * q3, x >= 1.0);
+			}
+			else
+			{
+				static_assert(false);
+				return core::nan<double>();
+			}
```
devshgraphicsprogramming, line 164: I think that higher derivatives would just be 0

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+	float m_maxSupport = Function1D::max_support;
+	float m_invStretch = 1.f;
+	float m_totalScale = 1.f;
+};
+
+namespace impl
+{
+
+template<class WeightFunction1D>
+struct weight_function_value_type : protected WeightFunction1D
```
devshgraphicsprogramming, line 297: you don't need to inherit to access `weight` via declval

unless some weight function's `weight()` is protected?

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+	constexpr static inline uint32_t k_smoothness = std::numeric_limits<uint32_t>::max();
+
+	template <int32_t derivative = 0>
+	static inline double weight(float x)
+	{
+		if (x >= min_support && x < max_support)
+		{
+			if constexpr (derivative == 0)
+			{
+				const double normalizationFactor = core::inversesqrt(2.0 * core::PI<double>()) / std::erff(core::sqrt<double>(2.0) * double(min_support));
```
devshgraphicsprogramming, line 102: the denominator is wrong.

Its supposed to be `std::erff(core::inversesqrt(2.0)*double(max_support))`

because CDF (which also is the integral from -INF to x) is `0.5+0.5*erf(x/sqrt(2))`

so to get definite integral, you have `0.5(erf(x/sqrt(2))-erf(-x/sqrt(2)))` which due to erf identity becomes `erf(x/sqrt(2))`

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+	{ t.stretchAndScale(x) }	-> std::same_as<void>;
+};
+
+template <typename Function1D, int32_t derivative = 0>
+class CWeightFunction1D /*final*/ // TODO(achal): Cannot make it final just yet because `weight_function_value_type` inherits from this.
+{
+	static_assert(derivative <= Function1D::k_smoothness);
+public:
+	constexpr static inline uint32_t k_smoothness = Function1D::k_smoothness-derivative;
+	constexpr static inline float k_energy[4] = { 0.f, 0.f, 0.f, 0.f }; // TODO(achal): Implement.
```
devshgraphicsprogramming, line 245: link me the discord convo about this

include/nbl/asset/filters/kernels/WeightFunctions.h
achalpandeyy: Yep, look here in `master` https://github.com/Devsh-Graphics-Programming/Nabla/blob/258f3642c22e31bed6c0114418190ebf16db2d14/include/nbl/asset/filters/kernels/CGaussianImageFilterKernel.h#L19

devshgraphicsprogramming: they weren't hardcoded, they were compile time tweakable

do compile time tweak via template OR runtime (if you can)

include/nbl/asset/filters/NormalizationStates.h
```diff
@@ -45,7 +45,7 @@ class CGlobalNormalizationState
 				const auto val = encodeBuffer[channel];
 				if constexpr (std::is_floating_point_v<Tenc>)
 				{
-					core::atomic_fetch_max(oldMinValue+channel,val);
+					core::atomic_fetch_min(oldMinValue+channel,val);
```
devshgraphicsprogramming, line 48: Note to self, I better fix this on the `ditt` branch XD

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -78,47 +76,58 @@ class CBlitImageFilterBase : public impl::CSwizzleableAndDitherableFilterBase<Sw
...
+	bool Clamp						= true,
+	typename BlitUtilities			= CBlitUtilities<>,
+	typename LutDataType			= float>
+class CBlitImageFilter :
+	public CImageFilter<CBlitImageFilter<Swizzle, Dither, Normalization, Clamp, BlitUtilities, LutDataType>>,
+	public CBlitImageFilterBase<Swizzle, Dither, Normalization, Clamp>
 {
 	public:
-		using blit_utils_t = CBlitUtilities<KernelX, KernelY, KernelZ>;
+		using blit_utils_t = BlitUtilities;
```
devshgraphicsprogramming, line 91: stick a `static_assert(std::is_base_of_v<IBlitUtilities,blit_utils_t>,"Only template instantiations of `CBlitUtilities` are allowed as the `BlitUtilities` template argument!");` below

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
```diff
@@ -0,0 +1,175 @@
...
+#include "nbl/asset/filters/kernels/WeightFunctions.h"
+
+namespace nbl::asset
+{
+
+// If we allow the user to specify a derivative of CConvolutionWeightFunction1D then there will be the following problem:
+// There will be no way for us to evaluate the constituent functions (m_funcA, m_funcB) on an arbitrary derivative because
+// `derivative` is a class template member not a function template member (done to handle chain rule) of CWeightFunction1D.
+// So, we would have to create new `CWeightFunction1D`s entirely and use them (not just here, but in the various specializations
+// of this function). Not sure if its worth it, both in terms of performance and code complexity.
```
devshgraphicsprogramming, line 16: explain what you mean?

include/nbl/asset/filters/CMipMapGenerationImageFilter.h
```diff
@@ -23,19 +23,15 @@ namespace asset
...
 // (actually in the case of using a Gaussian for both resampling and reconstruction, this is equivalent to using a single kernel of 3,3,5,9,..)
 
-template<typename Swizzle=VoidSwizzle, typename Dither=IdentityDither/*TODO: WhiteNoiseDither*/, typename Normalization=void, bool Clamp=false, class ResamplingKernelX = CKaiserImageFilterKernel<>, class ReconstructionKernelX = CMitchellImageFilterKernel<>, class ResamplingKernelY = ResamplingKernelX, class ReconstructionKernelY = ReconstructionKernelX, class ResamplingKernelZ = ResamplingKernelY, class ReconstructionKernelZ = ReconstructionKernelY>
-class CMipMapGenerationImageFilter : public CImageFilter<CMipMapGenerationImageFilter<Swizzle,Dither,Normalization,Clamp, ResamplingKernelX,ReconstructionKernelX, ResamplingKernelY,ReconstructionKernelY, ResamplingKernelZ,ReconstructionKernelZ> >, public CBasicImageFilterCommon
+template<typename Swizzle=VoidSwizzle, typename Dither=IdentityDither/*TODO: WhiteNoiseDither*/, typename Normalization=void, bool Clamp=false, typename BlitUtilities = CBlitUtilities<SMitchellFunction<>>>
```
devshgraphicsprogramming, line 26: make the default Clamp true and the default for the BlitUtils a Convolution of Kaiser with Mitchell

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: I see you copied my original code, seems it had a bug with the squareroot inside the erf not being a reciprocal :(

include/nbl/asset/utils/ICPUVirtualTexture.h
```diff
@@ -132,8 +132,12 @@ class ICPUVirtualTexture final : public IVirtualTexture<ICPUImageView, ICPUSampl
 
         upscaled_img->setBufferAndRegions(std::move(buf), std::move(regions));
 
-        using blit_filter_t = asset::CBlitImageFilter<asset::VoidSwizzle,asset::IdentityDither/*TODO: White Noise*/,void,false,asset::CMitchellImageFilterKernel<>>;
-        blit_filter_t::state_type blit;
+        using blit_filter_t = asset::CBlitImageFilter<asset::VoidSwizzle,asset::IdentityDither/*TODO: White Noise*/,void,false, asset::CBlitUtilities<asset::CWeightFunction1D<asset::SMitchellFunction<>>>>;
```
devshgraphicsprogramming, line 135: style nitpick, you can remove `asset::` from all of the types, you're already in this namespace

include/nbl/asset/utils/ICPUVirtualTexture.h
```diff
@@ -132,8 +132,12 @@ class ICPUVirtualTexture final : public IVirtualTexture<ICPUImageView, ICPUSampl
...
+        auto convolutionKernels = blit_filter_t::blit_utils_t::getConvolutionKernels(
```
devshgraphicsprogramming, line 136: the resampling filter should be Mitchell and the reconstruction should be Linear or Mitchell as well

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
```diff
@@ -0,0 +1,175 @@
...
+		if constexpr (std::is_same_v<WeightFunction1DB, CWeightFunction1D<SDiracFunction>>)
+		{
+			return m_funcA.weight(x, channel);
+		}
+		else if (std::is_same_v<WeightFunction1DA, CWeightFunction1D<SDiracFunction>>)
+		{
+			return m_funcB.weight(x, channel);
+		}
+		else
+		{
```
devshgraphicsprogramming, line 88: hoist this to `weight`

include/nbl/asset/filters/kernels/CChannelIndependentWeightFunction.h
```diff
@@ -0,0 +1,136 @@
...
+
+#include <type_traits>
+#include <tuple>
+
+
+namespace nbl::asset
+{
+
+// we always invoke the first channel of the kernel for each kernel assigned to a channel
+template <class FirstWeightFunction1D, class... OtherWeightFunctions>
```
devshgraphicsprogramming, line 18: why aren't you constraining to the `KernelWeightFunction1D` concept?

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: which it isn't

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+			return core::nan<double>();
+		}
+	}
+};
+
+// Standard Box function, symmetric, value in the support is 1, integral is 1, so support must be [-1/2,1/2)
+struct SBoxFunction
+{
+	constexpr static inline float min_support = -0.5f;
+	constexpr static inline float max_support = +0.5f;
```
devshgraphicsprogramming, line 39: ok, if you're not making the box filter slightly large here, then you need to apply a slight stretch in `CBlitUtilities` when you detect it being used

https://discord.com/channels/593902898015109131/1043939034776473710/1062733502266744842

Something like `1.f+1.f/8192.f` stretch

devshgraphicsprogramming: ACtually not important if the default is double convolution.

src/nbl/asset/utils/CDerivativeMapCreator.cpp
```diff
@@ -1,154 +1,49 @@
...
 
-	using ReconstructionKernel = CGaussianImageFilterKernel<>; // or Mitchell
-	using DerivKernel_ = CDerivativeImageFilterKernel<ReconstructionKernel>;
-	using DerivKernel = MyKernel<DerivKernel_>;
-	using XDerivKernel_ = CChannelIndependentImageFilterKernel<DerivKernel, CBoxImageFilterKernel>;
-	using YDerivKernel_ = CChannelIndependentImageFilterKernel<CBoxImageFilterKernel, DerivKernel>;
-	using XDerivKernel = SeparateOutXAxisKernel<XDerivKernel_>;
-	using YDerivKernel = SeparateOutXAxisKernel<YDerivKernel_>;
+	// or Mitchell
+	using ReconstructionFunction = CWeightFunction1D<SGaussianFunction>;
```
devshgraphicsprogramming, line 16: An additional Gaussian/Mitchell will needlessly smooth our input and attenuate high frequencies, basically we'll get the same derivative map "as if" we made it from Mip Level 1.

As you are sure that the outputted derivative map will be same resolution you can make the Reconstruction Function a Dirac.

This is because even if you made it a perfect Sinc (which doesn't do anything to the signal - in theory), it would behave like a Dirac because the phase is always 0 and all zeros of the Sinc function align with neighbouring sample locations, and the middle/center one gets a weight of 1.

src/nbl/asset/utils/CDerivativeMapCreator.cpp
```diff
@@ -187,6 +82,9 @@ core::smart_refctd_ptr<ICPUImage> CDerivativeMapCreator::createDerivativeMapFrom
 	state.scratchMemoryByteSize = DerivativeMapFilter::getRequiredScratchByteSize(&state);
 	state.scratchMemory = reinterpret_cast<uint8_t*>(_NBL_ALIGNED_MALLOC(state.scratchMemoryByteSize, _NBL_SIMD_ALIGNMENT));
 
+	if (!DerivativeMapFilter::blit_utils_t:: template computeScaledKernelPhasedLUT<float>(state.scratchMemory + DerivativeMapFilter::getScratchOffset(&state, DerivativeMapFilter::ESU_SCALED_KERNEL_PHASED_LUT), state.inExtentLayerCount, state.outExtentLayerCount, state.inImage->getCreationParameters().type, convolutionKernels))
+		return nullptr;
```
devshgraphicsprogramming, line 86: isn't there a cleaner API to call to populate the LUT?

I mean most of the arguments you're passing are already in your `state`

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
devshgraphicsprogramming: with convolution you have the derivative property
Derivative of convolution is the same as convolution where the first function is a derivative
https://discord.com/channels/593902898015109131/1043939034776473710/1064514311822774383

devshgraphicsprogramming: > explain what you mean? I don't get the problem.

Ok I get it now

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+	static inline constexpr double q2 = (6.0 * b + 30.0 * c) / 6.0;
+	static inline constexpr double q3 = (-b - 6.0 * c) / 6.0;
+};
+
+// Kaiser filter, basically a windowed sinc.
+struct SKaiserFunction
+{
+	constexpr static inline float min_support = -3.f;
+	constexpr static inline float max_support = +3.f;
+	constexpr static inline uint32_t k_smoothness = 3; // TODO(achal): Figure out what this should be
```
devshgraphicsprogramming, line 188: if you're gonna only implement 1 derivative, then 1

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
```diff
@@ -0,0 +1,175 @@
...
+
+	inline void stretchAndScale(const float stretchFactor)
+	{
+		stretch(stretchFactor);
+		scale(1.f / stretchFactor);
+	}
+
+	inline double weight(const float x, const uint8_t channel, const uint32_t sampleCount = 64u) const
+	{
+		return static_cast<double>(m_totalScale * weight_impl(x * m_invStretch, channel, sampleCount));
```
devshgraphicsprogramming, line 61: you want to cast the `m_totalScale` before the multiply

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
devshgraphicsprogramming: Amend the comment and tell the user about the convolution property.

Basically that if they want "Nth order derivative of `Conv<CWeightFunction<A>,CWeightFunction<B>>`" then they should create "`Conv<CWeightFunction<A,i>,CWeightFunction<B,j>`" where `i<A::k_smoothness && j<B::k_smoothness && (i+j)==n`

include/nbl/asset/filters/kernels/CFloatingPointSeparableImageFilterKernel.h
```diff
@@ -0,0 +1,82 @@
...
+// For conditions of distribution and use, see copyright notice in nabla.h
+#ifndef _NBL_ASSET_C_FLOATING_POINT_SEPARABLE_IMAGE_FILTER_KERNEL_H_INCLUDED_
+#define _NBL_ASSET_C_FLOATING_POINT_SEPARABLE_IMAGE_FILTER_KERNEL_H_INCLUDED_
+
+namespace nbl::asset
+{
+
+// kernels that requires pixels and arithmetic to be done in precise floats AND are separable AND have the same kernel function in each dimension AND have a rational support
+template<KernelWeightFunction1D WeightFunction1D>
+class CFloatingPointSeparableImageFilterKernel : public impl::weight_function_value_type_t<WeightFunction1D>
```
devshgraphicsprogramming, line 12: whats with the weird inheritance?

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -78,47 +76,58 @@ class CBlitImageFilterBase : public impl::CSwizzleableAndDitherableFilterBase<Sw
...
 // copy while filtering the input into the output, a rare filter where the input and output extents can be different, still works one mip level at a time
-template<typename Swizzle=DefaultSwizzle, typename Dither=CWhiteNoiseDither, typename Normalization=void, bool Clamp=true, class KernelX=CBoxImageFilterKernel, class KernelY=KernelX, class KernelZ=KernelX, typename lut_value_type = typename KernelX::value_type>
-class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Normalization,Clamp,KernelX,KernelX,KernelX>>, public CBlitImageFilterBase<typename KernelY::value_type,Swizzle,Dither,Normalization,Clamp>
+template<
+	typename Swizzle				= DefaultSwizzle,
+	typename Dither					= CWhiteNoiseDither,
+	typename Normalization			= void,
+	bool Clamp						= true,
+	typename BlitUtilities			= CBlitUtilities<>,
+	typename LutDataType			= float>
```
devshgraphicsprogramming, line 85: eh, lets stick the LUT data type into `CBlitUtilities`, will de-template a lot of member functions.

include/nbl/asset/filters/CBlitUtilities.h
```diff
@@ -24,50 +27,67 @@ class IBlitUtilities
...
-		return ((phaseCount[0] * scaledKernelX.getWindowSize().x) + (phaseCount[1] * scaledKernelY.getWindowSize().y) + (phaseCount[2] * scaledKernelZ.getWindowSize().z)) * sizeof(lut_value_type) * MaxChannels;
+		return ((phaseCount.x * windowSize.x) + (phaseCount.y * windowSize.y) + (phaseCount.z * windowSize.z)) * sizeof(LutDataType) * MaxChannels;
 	}
 
-	template <typename lut_value_type = typename KernelX::value_type>
-	static bool computeScaledKernelPhasedLUT(void* outKernelWeights, const core::vectorSIMDu32& inExtent, const core::vectorSIMDu32& outExtent, const asset::IImage::E_TYPE inImageType,
-		const KernelX& kernelX, const KernelY& kernelY, const KernelZ& kernelZ)
+	template <typename LutDataType>
+	static bool computeScaledKernelPhasedLUT(
+		void*							outKernelWeights,
```
devshgraphicsprogramming, line 86: The method needs a `const double normalizeWeightsTo=1.0` argument

You need to compute the normalization factor if `!core::isnan(normalizeWeightsTo)` which would be the `normalizeWeightsTo` divided by the sum of all weights for a given axis & phase.

This is why:
https://discord.com/channels/593902898015109131/1043939034776473710/1062723347835793428
https://discord.com/channels/593902898015109131/1043939034776473710/1062724151699308655
https://discord.com/channels/593902898015109131/1043939034776473710/1067676795391967323

Basically even having a reconstruction filter won't help you against value fluctuations (ringing) like these:
image

Why?

Imagine you're using the convolution of two gaussians or two sinc filters for your kernel.

In the gaussian case you'll have a variance thats a sum of the two input gaussians, in the sinc you'll be left with the wider one.

Now imagine you're upscaling a constant value image, so you have many many phases and a really wide kernel in input coordinates.

**Even though the kernel resulting from the convolution integrates to a constant, you won't get a constant image!**

Because there's a finite amount of input pixels within the kernel's support, its not a real "full" integration, and therefore its like doing a low sample count Riemann Sum with slightly shifted (by the phase) sample points for each output pixel.

Since Riemann Sum != The True Integral if the sample count does not increase to infinity, you'll get under or oversampling.

devshgraphicsprogramming: Btw you can't quantize/compress the weights to `float` or `float16_t` before you're done normalizing them, so you should always require `outKernelWeights` be `sizeof(double)` sized

include/nbl/asset/filters/CBlitUtilities.h
```diff
@@ -77,120 +97,103 @@ class CBlitUtilities : public IBlitUtilities
...
+		if (inImageType == IImage::ET_3D)
+			windowSize.z = std::get<2>(kernels).getWindowSize();
+
+		return windowSize;
 	}
 
-	template <typename lut_value_type = typename KernelX::value_type>
-	static inline core::vectorSIMDu32 getScaledKernelPhasedLUTAxisOffsets(const core::vectorSIMDu32& phaseCount, const core::vectorSIMDi32& real_window_size)
+	template <typename LutDataType>
+	static inline core::vectorSIMDu32 getScaledKernelPhasedLUTAxisOffsets(const core::vectorSIMDu32& phaseCount, const core::vectorSIMDi32& windowSize)
```
devshgraphicsprogramming, line 187: this is fine to keep as is, because weights eventually get compressed

So this has no bearing
https://github.com/Devsh-Graphics-Programming/Nabla/pull/435/files#r1196525176

(you can compress "in place" like a memmove)

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Because `state` is just a raw pointer (and we don't mutate the pointer itself, inside the lambda)
https://github.com/Devsh-Graphics-Programming/Nabla/blob/75a7a200fadfb04397b86c631f0d3a085e10b86e/include/nbl/asset/filters/CBlitImageFilter.h#L292

include/nbl/asset/filters/kernels/WeightFunctions.h
achalpandeyy: Ah you're right.. they weren't hardcoded.

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: ah ok, wasn't apparent without Intellisense XD

include/nbl/asset/filters/kernels/WeightFunctions.h
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	// This method will keep the integral of the weight function constant.
+	inline void stretchAndScale(const float stretchFactor)
+	{
+		stretch(stretchFactor);
+		scale(1.f / stretchFactor);
+	}
+
+	inline double weight(const float x, const uint8_t channel) const
```
devshgraphicsprogramming, line 276: ok, we should probably stick to Single Responsibility Principle and not bother with worrying about channels in `CWeightFunction1D` and delegate that to a function later down the road

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: I'd get rid of this whole namespace and declare a 
```cpp
public:
using value_type = decltype(std::declval<Function1D>().weight(0.f));
```
inside of the `CWeightFunction1D`

and then use that to express stretch, scale, `k_energy` and the return value type

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: Also due to https://github.com/Devsh-Graphics-Programming/Nabla/pull/435/files#r1196619001 it should be single valued and thats it.

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: solved it myself.

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: solved it myself

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: solved it myself.

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: done by myself.

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: fixed it myself.

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
devshgraphicsprogramming: fixed it myself.

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: done it myself

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
devshgraphicsprogramming: done it myself

include/nbl/asset/filters/kernels/WeightFunctions.h
devshgraphicsprogramming: done it myself.

include/nbl/asset/filters/kernels/CFloatingPointSeparableImageFilterKernel.h
devshgraphicsprogramming: removed the whole class

include/nbl/asset/filters/kernels/CFloatingPointSeparableImageFilterKernel.h
```diff
@@ -0,0 +1,82 @@
...
+		const auto& inParams = inImage->getCreationParameters();
+		const auto& outParams = inImage->getCreationParameters();
+		return !(isIntegerFormat(inParams.format)||isIntegerFormat(outParams.format));
+	}
+
+	//template<typename... Args> // TODO: later if needed
+	inline value_type weight(const float x, const uint32_t channel) const
+	{
+		if ((x >= getMinSupport()) && (x < getMaxSupport()))
+			return func.weight(x, channel);
```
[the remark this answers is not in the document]
devshgraphicsprogramming: removed whole function and class

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
```diff
@@ -0,0 +1,175 @@
...
+		{
+			return m_funcB.weight(x, channel);
+		}
+		else
+		{
+			auto [minIntegrationLimit, maxIntegrationLimit] = getIntegrationDomain(x);
+
+			// TODO(achal): what ????
+			// if this happens, it means that `m_ratio=INF` and it degenerated into a dirac delta
+			if (minIntegrationLimit == maxIntegrationLimit)
```
devshgraphicsprogramming, line 93: This was a comment for the original case of convolving `f*g` where `f` was supposed to be unscaled and only `g` was scaled by `m_ratio` .

Now obviously everything changed, and you take two full functions with independent scales (great design choice and simplification of the API btw)

So this was supposed to be in the case of when `m_funcB.m_invStretch > FLT_MAX` or `m_funcA.m_invStretch < FLT_MIN` (or some other epsilon)

This makes the function B degenerate to a single point (dirac delta with multiplier equal to total energy under the curve) or function A  has stretched to what is basically flat line with value `f_a(0)`

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
```diff
@@ -0,0 +1,175 @@
...
+		{
+			return m_funcA.weight(x, channel);
+		}
+		else if (std::is_same_v<WeightFunction1DA, CWeightFunction1D<SDiracFunction>>)
+		{
+			return m_funcB.weight(x, channel);
+		}
+		else
+		{
+			auto [minIntegrationLimit, maxIntegrationLimit] = getIntegrationDomain(x);
```
devshgraphicsprogramming, line 89: WIP

devshgraphicsprogramming: Basicallly handle that before going to compute `dtau` and doing the numerical integration

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
devshgraphicsprogramming: did it myself

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
devshgraphicsprogramming: implemented it myself.

src/nbl/asset/utils/CDerivativeMapCreator.cpp
devshgraphicsprogramming: I addded one as part of #499

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
devshgraphicsprogramming: @achalpandeyy mark as resolved when read

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
```diff
@@ -0,0 +1,175 @@
...
+				assert(false);
+				assert(WeightFunction1DB::k_energy[channel] != 0.f); // TODO(achal): Remove.
+				return m_funcA.weight(x, channel) * WeightFunction1DB::k_energy[channel];
+			}
+
+			const double dtau = (maxIntegrationLimit - minIntegrationLimit) / sampleCount;
+
+			// TODO(achal): what ???
+			// if this happened then `m_ratio=0` and we have infinite domain, this is not a problem
+			if (core::isnan<double>(dtau))
```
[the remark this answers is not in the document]
devshgraphicsprogramming: @achalpandeyy mark as resolved when read

include/nbl/asset/utils/ICPUVirtualTexture.h
achalpandeyy: Resolved.

include/nbl/asset/utils/ICPUVirtualTexture.h
achalpandeyy: Both reconstruction and resampling are Mitchell
Resolved.

src/nbl/asset/utils/CDerivativeMapCreator.cpp
achalpandeyy: Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
achalpandeyy: Resolved.

include/nbl/asset/filters/CBlitUtilities.h
```diff
@@ -24,50 +27,67 @@ class IBlitUtilities
...
 	{
-		const auto scaledKernelX = constructScaledKernel(kernelX, inExtent, outExtent);
-		const auto scaledKernelY = constructScaledKernel(kernelY, inExtent, outExtent);
-		const auto scaledKernelZ = constructScaledKernel(kernelZ, inExtent, outExtent);
+		const auto windowSize = getWindowSize(inImageType, kernels);
+		return getScaledKernelPhasedLUTSize<LutDataType>(inExtent, outExtent, inImageType, windowSize);
+	}
 
+	template <typename LutDataType>
+	static inline size_t getScaledKernelPhasedLUTSize(
```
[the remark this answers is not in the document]
achalpandeyy: Resolved.

include/nbl/asset/filters/CBlitUtilities.h
achalpandeyy: Resolved

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
achalpandeyy: Can we mark it resolved then?

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
```diff
@@ -0,0 +1,175 @@
...
+template <>
+double CConvolutionWeightFunction1D<CWeightFunction1D<SBoxFunction>, CWeightFunction1D<SBoxFunction>>::weight_impl(const float x, const uint32_t channel, const uint32_t) const;
+
+template <>
+double CConvolutionWeightFunction1D<CWeightFunction1D<SGaussianFunction>, CWeightFunction1D<SGaussianFunction>>::weight_impl(const float x, const uint32_t channel, const uint32_t) const;
+
+// TODO: Specialization: CConvolutionWeightFunction1D<CWeightFunction1D<STriangleFunction>, CWeightFunction1D<STriangleFunction>> = this is tricky but feasible
+
+template <>
+double CConvolutionWeightFunction1D<CWeightFunction1D<SKaiserFunction>, CWeightFunction1D<SKaiserFunction>>::weight_impl(const float x, const uint32_t channel, const uint32_t) const;
```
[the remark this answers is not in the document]
achalpandeyy: For Box it would be zero everywhere except where it is not defined.

For Kaiser it doesn't seem very straight forward to integrate the first order derivative of the bessel function thingy.

Not sure about Gaussian, but seems doable at least.

src/nbl/asset/utils/CDerivativeMapCreator.cpp
achalpandeyy: Yep, incorporated it.
Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -536,15 +536,15 @@ class CBlitImageFilter :
 								if (!srcPix[0])
 									continue;
 
-								auto sample = lineBuffer+i*MaxChannels;
-								value_t swizzledSample[MaxChannels];
+								// TODO: make sure there is no leak due to ChannelCount!
```
devshgraphicsprogramming, line 539: @achalpandeyy what's with the TODO?

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -431,10 +431,10 @@ class CBlitImageFilter :
...
 					const core::vectorSIMDu32 localOutPos = writeBlockPos - outOffsetLayer;
 
-					value_t sample[MaxChannels];
+					value_t sample[ChannelCount];
 					const size_t offset = IImage::SBufferCopy::getLocalByteOffset(localOutPos, intermediateStrides[axis]);
 					const auto* first = intermediateStorage[axis]+offset;
-					std::copy(first,first+MaxChannels,sample);
+					std::copy(first,first+ChannelCount,sample);
 
 					sample[alphaChannel] *= coverageScale;
```
devshgraphicsprogramming, line 439: do you have anything in `validate()` that checks/ensures `alphaChannel<=ChannelCount` ?

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: your `onDecode` uses `decodePixelsRuntime` which needs a place to  store all 4 channels, so `preSwizzleSample` needs to be MaxChannels or 4 elements in size

devshgraphicsprogramming: the swizzle probably as well, basically unless you make `onDecode` take a `channelCount` and copy stuff over, you need `MaxChannelCount` on all arrays

Actually thinking of this, its best you address the TODO here:
https://github.com/Devsh-Graphics-Programming/Nabla/blob/fb457e9d39fede6d125ab75e3e823dfd4ab2f3bc/include/nbl/asset/filters/CSwizzleableAndDitherableFilterBase.h#L356

declare a `Tdec tmp[4];` to store the swizzle results to, then copy `channelCount` values of `tmp` into `decodeBuffer`

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
devshgraphicsprogramming: yep, I just wanted you to read this.

include/nbl/asset/filters/CBlitUtilities.h
```diff
@@ -129,19 +158,58 @@ class CBlitUtilities : public IBlitUtilities
...
+							if constexpr (std::is_same_v<LutDataType, uint16_t>)
+								outKernelWeightsPixel[(i * _windowSize + j) * ChannelCount + ch] = core::Float16Compressor::compress(float(weight));
+							else
+								outKernelWeightsPixel[(i * _windowSize + j) * ChannelCount + ch] = LutDataType(weight);
+						}
 						else
-							outKernelWeightsPixel[(i * _windowSize + j) * MaxChannels + ch] = LutDataType(weight);
+						{
+							accum[ch] += weight;
+							outKernelWeightsPixel_f64[(i * _windowSize + j) * ChannelCount + ch] = weight;
```
devshgraphicsprogramming: minor stylistic nitpick, can you hoist the calculation of the index `(i * _windowSize + j) * ChannelCount + ch` outside the conditional?

include/nbl/asset/filters/CBlitUtilities.h
```diff
@@ -129,19 +158,58 @@ class CBlitUtilities : public IBlitUtilities
...
+						}
 					}
 					
 					relativePos -= 1.f;
 				}
+
+				if (shouldNormalize)
+				{
+					double normalizationFactor[ChannelCount] = { };
```
devshgraphicsprogramming, line 186: does this notation always initalize whole array to 0? I forgot my C++ basics.

include/nbl/asset/filters/CBlitUtilities.h
```diff
@@ -129,19 +158,58 @@ class CBlitUtilities : public IBlitUtilities
...
+					std::fill(normalizationFactor, normalizationFactor + ChannelCount, 1.0);
```
devshgraphicsprogramming, line 187: there's a better way, if all weights are close to 0 and their sum was supposed to equal `normalizeWeightsTo`

if you detect `accum[ch]<threshold` then fill all weights with `normalizeWeightsTo/double(_windowSize)`

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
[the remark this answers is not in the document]
devshgraphicsprogramming: Leave it for later/when needed I guess.

include/nbl/asset/filters/kernels/CConvolutionWeightFunction.h
```diff
@@ -84,7 +84,7 @@ class CConvolutionWeightFunction1D final : public impl::IWeightFunction1D<typena
 		const WeightFunction1DA m_funcA;
 		const WeightFunction1DB m_funcB;
 
-		value_t weight_impl(const float x, const uint32_t sampleCount) const
+		value_t NBL_API2 weight_impl(const float x, const uint32_t sampleCount) const
```
devshgraphicsprogramming, line 87: funny it needs an NBL_API2, I thought it would be inline, can you explicitly declare it inline instead and see what happens?

include/nbl/asset/filters/CBlitImageFilter.h
AnastaZIuk: in base_t validation there is 

```cpp
if (state->alphaChannel>=4)
	return false;
```

but cannot see that check above anywhere so adding it

include/nbl/asset/filters/CBlitUtilities.h
```diff
@@ -129,19 +158,58 @@ class CBlitUtilities : public IBlitUtilities
...
+
+							LutDataType compressed;
+							{
+								if constexpr (std::is_same_v<LutDataType, uint16_t>)
+									compressed = core::Float16Compressor::compress(float(normalized));
+								else
+									compressed = LutDataType(normalized);
+							}
+
+							memcpy(&outKernelWeightsPixel[idx], &compressed, sizeof(LutDataType));
```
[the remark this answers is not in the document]
AnastaZIuk: resolved in following commit

include/nbl/asset/filters/CBlitUtilities.h
AnastaZIuk: resolved in following commit

include/nbl/asset/filters/CBlitImageFilter.h
AnastaZIuk: resolved in following commit

include/nbl/asset/filters/CBlitImageFilter.h
AnastaZIuk: resolved in following commit, we are doing it in place now like @achalpandeyy suggested in his TODO

include/nbl/asset/filters/CBlitUtilities.h
AnastaZIuk: I'm leaving it to @achalpandeyy

achalpandeyy: Resolved.

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: and it will break because swizzles permute RGBA channels around and you need a auxiliary scratch array!

include/nbl/asset/filters/CSwizzleAndConvertImageFilter.h
```diff
@@ -69,13 +69,13 @@ class CSwizzleAndConvertImageFilterBase : public CSwizzleableAndDitherableFilter
...
 
 							if constexpr (kInFormat!=EF_UNKNOWN)
-								base_t::template onDecode<kInFormat,encodeBufferType>(state, srcPix, decodeBuffer, encodeBuffer, blockX, blockY);
+								base_t::template onDecode<kInFormat>(state, srcPix, decodeBuffer, blockX, blockY);
 							else
-								base_t::template onDecode<encodeBufferType>(rInFormat, state, srcPix, decodeBuffer, encodeBuffer, blockX, blockY);
-							state->normalization.prepass(encodeBuffer,readBlockPos*blockDims+commonExecuteData.offsetDifferenceInTexels,blockX,blockY,4u/*TODO: figure this out*/);
+								base_t::template onDecode<encodeBufferType>(rInFormat, state, srcPix, decodeBuffer, blockX, blockY);
+
+							state->normalization.prepass(decodeBuffer,readBlockPos*blockDims+commonExecuteData.offsetDifferenceInTexels,blockX,blockY,4u/*TODO: figure this out*/);
```
devshgraphicsprogramming: all changes to this file from this commit break it: https://github.com/Devsh-Graphics-Programming/Nabla/pull/435#discussion_r1278978388

devshgraphicsprogramming: ok you've moved the temp

include/nbl/asset/filters/CSwizzleableAndDitherableFilterBase.h
```diff
@@ -62,29 +62,33 @@ class CSwizzleableAndDitherableFilterBase
...
 		*/
-		template<E_FORMAT inFormat, typename Tdec, typename Tenc>
-		static void onDecode(state_type* state, const void* srcPix[4], Tdec* decodeBuffer, Tenc* encodeBuffer, uint32_t blockX, uint32_t blockY)
+
+		template<E_FORMAT inFormat, typename Tdec>
+		static void onDecode(state_type* state, const void* srcPix[4], Tdec* decodeBuffer, uint32_t blockX, uint32_t blockY)
 		{
 			static_assert(sizeof(Tdec)==8u, "Encode/Decode types must be double, int64_t or uint64_t!");
-			static_assert(sizeof(Tenc)==8u, "Encode/Decode types must be double, int64_t or uint64_t!");
 			asset::decodePixels<inFormat>(srcPix, decodeBuffer, blockX, blockY);
```
devshgraphicsprogramming, line 70: decode to `tmp/swizzled` **you have no guarantee that `decodeBuffer` is 4 elements long

include/nbl/asset/filters/CBlitImageFilter.h
devshgraphicsprogramming: ok I see the temp has moved to `onDecode`, but issues persist

PR #345, opened 2022-04-14
Descriptor lifetime tracking & Command Pool revamp

achalpandeyy:
## TODO list:
- [x] OpenGL command buffer execution (`COpenGLCommandBuffer::executeAll`).
- [x] Merge `master` and get things working.
- [x] `calc_size` method is not needed to be defined explicitly for "fixed size commands". By "fixed size commands", we mean the commands whose size is know at compile-time. For example: `CDrawIndexedIndirectCountCmd` and `CBeginRenderPassCmd` (among many others) are fixed size commands, whereas `CPipelineBarrierCmd` is not --because the number of resources the command operates on, is not known at compile-time.
- [x] Fix the inversion of control happening between `IGPUCommandPool` and `IGPUCommandBuffer`. [Related comment.](https://github.com/Devsh-Graphics-Programming/Nabla/pull/345#issuecomment-1193632400)
- [x] Have two separate segmented lists for non Vulkan-like backends i.e. those rendering backends which need to hold on to some data other than `smart_refctd_ptr` of descriptors to execute its commands, such as the OpenGL backend. Where the first segmented list will hold the `smart_refctd_ptr` of descriptors (common to both Vulkan-like and non Vulkan-like backends) and the second one will hold on to other data. [More details in this comment.](https://github.com/Devsh-Graphics-Programming/Nabla/pull/345#issuecomment-1116194117)
- [x] [Command Pool Resetting](https://github.com/Devsh-Graphics-Programming/Nabla/pull/408#issuecomment-1208613540).
    - [x] Turn all `ICommandBuffer` methods into pure `virutal`s.
    - [x] Hoist all state transition and validation to `IGPUCommandBuffer`. Backend specific `_impl` functions would return `void`.
    - [x] Detect if the command pool of a command buffer has been already reset with the help of counters in `IGPUCommandPool` and `IGPUCommandBuffer` as mentioned in #373 and as already being done in #408 (works like a versioning system), and do **NOT** release resources in this case.
    - [x] Check for command pool reset before every command buffer method.
    - [x] Hoist `releaseResourcesBackToPool` to `IGPUCommandBuffer`. Context: #408.
    - [x] `IGPUCommandBuffer`'s last-reset-stamp/counter should be set in `begin` and not in `resetCommon` as currently being done in #408.
    - [x] Figure out a way to release the command segment list memory on pool reset without keeping track of child command buffer heads.
[...]

PR #444, opened 2022-12-31
Add opeator| for E_SHADER_STAGE

sunho:
## Description
Add oprator| for E_SHADER_STAGE so that we can do something like `asset::IShader::ESS_VERTEX | asset::IShader::ESS_FRAGMENT`.

### DIFF
--- a/include/nbl/asset/IShader.h
+++ b/include/nbl/asset/IShader.h
@@ -75,6 +75,11 @@ protected:
 	std::string m_filepathHint;
 };
 
+inline IShader::E_SHADER_STAGE operator|(IShader::E_SHADER_STAGE lhs,
+                                         IShader::E_SHADER_STAGE rhs) {
+  return static_cast<IShader::E_SHADER_STAGE>(static_cast<uint32_t>(lhs) |
+                                     static_cast<uint32_t>(rhs));
+}
 }
 
 #endif


include/nbl/asset/IShader.h
```diff
@@ -75,6 +75,11 @@ class NBL_API IShader : public virtual core::IReferenceCounted
 	std::string m_filepathHint;
 };
 
+inline IShader::E_SHADER_STAGE operator|(IShader::E_SHADER_STAGE lhs,
+                                         IShader::E_SHADER_STAGE rhs) {
+  return static_cast<IShader::E_SHADER_STAGE>(static_cast<uint32_t>(lhs) |
+                                     static_cast<uint32_t>(rhs));
+}
```
devshgraphicsprogramming, line 82: @Erfan-Ahmadi @achalpandeyy don't we have a macro for that?

devshgraphicsprogramming: > Not yet

should I merge then?

Erfan-Ahmadi: NO.
Oh this PR is only this?
do this manually for now. (Portability Use bitflags like previously)
I'll add it soon.

achalpandeyy: > (Portability Use bitflags like previously)

I second this.

Am I missing some context? The intended usage should be `core::bitflag<IShader::E_SHADER_STAGE>`
In #345 we do the same.

Erfan-Ahmadi: No it's just about being able to do bitflag(A|B|C) instead of bitflag(A)|B|C
Also for implicit conversions you can just do A|B|C which is much nicer.

But I'm gonna have this for our bitflag enums via a define instead of manual implementation for each of them.

Erfan-Ahmadi: It's just a convenience thing.

PR #453, opened 2023-02-17
Expose Nabla APIs

deprilula28:
## Description
Exposes some Nabla APIs using NABLA_API

### DIFF
--- a/include/nbl/asset/utils/CHLSLCompiler.h
+++ b/include/nbl/asset/utils/CHLSLCompiler.h
@@ -9,11 +9,10 @@
 #include "nbl/asset/utils/IShaderCompiler.h"
 
 #include <wrl.h>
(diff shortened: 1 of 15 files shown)


include/nbl/asset/utils/CHLSLCompiler.h
```diff
@@ -9,12 +9,11 @@
 #include "nbl/asset/utils/IShaderCompiler.h"
 
 #include <wrl.h>
+#include <combaseapi.h>
+#include <dxc/dxc/include/dxc/dxcapi.h>
 
 using Microsoft::WRL::ComPtr;
```
devshgraphicsprogramming, line 15: @Erfan-Ahmadi @deprilula28 can you guys hide this so it only appears in `/src/nbl` files?

PR #454, opened 2023-02-18
add missing includes for non PCH build

pollend:
## Description
this attempts to address some of the build problems partially on linux.  

## Testing

verify this changes compile for windows should be sufficient.

### DIFF
--- a/include/nbl/system/IFileArchive.h
+++ b/include/nbl/system/IFileArchive.h
@@ -9,6 +9,7 @@
 #include "nbl/system/path.h"
 #include "nbl/system/ILogger.h"
 #include "nbl/system/IFileBase.h"
+#include "nbl/core/SRange.h"
 
 #include <string_view>
 #include <algorithm>
--- a/include/nbl/system/IFileArchive.h
+++ b/include/nbl/system/IFileArchive.h
@@ -6,10 +6,11 @@
 #ifndef _NBL_SYSTEM_I_FILE_ARCHIVE_H_INCLUDED_
 #define _NBL_SYSTEM_I_FILE_ARCHIVE_H_INCLUDED_
 
+#include "nbl/core/SRange.h"
+
 #include "nbl/system/path.h"
 #include "nbl/system/ILogger.h"
 #include "nbl/system/IFileBase.h"
-#include "nbl/core/SRange.h"
 
 #include <string_view>
 #include <algorithm>
--- a/include/nbl/system/ISystem.h
+++ b/include/nbl/system/ISystem.h
@@ -38,6 +38,10 @@ class NBL_API2 ISystem : public core::IReferenceCounted
         {
             static inline constexpr E_REQUEST_TYPE type = RT;
         };
+
+        struct SRequestParams_EMPTY
+        {
+        };
         struct SRequestParams_CREATE_FILE : SRequestParamsBase<ERT_CREATE_FILE>
         {
             char filename[MAX_FILENAME_LENGTH] {};
@@ -61,10 +65,11 @@ class NBL_API2 ISystem : public core::IReferenceCounted
         {
             E_REQUEST_TYPE type;
             std::variant<
+                SRequestParams_EMPTY,
                 SRequestParams_CREATE_FILE,
                 SRequestParams_READ,
                 SRequestParams_WRITE
-            > params;
+            > params = SRequestParams_EMPTY();
         };
         static inline constexpr uint32_t CircularBufferSize = 256u;
         class CAsyncQueue : public ICancellableAsyncQueueDispatcher<CAsyncQueue,SRequestType,CircularBufferSize>
--- a/include/nbl/system/ISystemPOSIX.h
+++ b/include/nbl/system/ISystemPOSIX.h
@@ -1,8 +1,9 @@
 #ifndef _NBL_SYSTEM_C_SYSTEM_CALLER_POSIX_INCLUDED_
 #define _NBL_SYSTEM_C_SYSTEM_CALLER_POSIX_INCLUDED_
 
-#include "nbl/system/ISystem.h"
 #include "nbl/core/util/bitflag.h"
+
+#include "nbl/system/ISystem.h"
 #include "nbl/system/IFile.h"
 
 namespace nbl::system
--- a/include/nbl/system/ISystemPOSIX.h
+++ b/include/nbl/system/ISystemPOSIX.h
@@ -2,6 +2,8 @@
 #define _NBL_SYSTEM_C_SYSTEM_CALLER_POSIX_INCLUDED_
 
 #include "nbl/system/ISystem.h"
+#include "nbl/core/util/bitflag.h"
+#include "nbl/system/IFile.h"
 
 namespace nbl::system
 {
--- a/include/nbl/type_traits.h
+++ b/include/nbl/type_traits.h
@@ -19,7 +19,7 @@ template <bool... Vals>
 using bool_sequence = std::integer_sequence<bool, Vals...>;
 
 template <typename T, typename... Us>
-_INLINE_VAR constexpr bool is_any_of_v = (... || std::is_same<T, Us>::value);
+inline constexpr bool is_any_of_v = (... || std::is_same<T, Us>::value);
 
 template<typename T, typename... Us>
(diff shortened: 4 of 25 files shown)


include/nbl/system/ISystem.h
```diff
@@ -61,10 +64,11 @@ class NBL_API2 ISystem : public core::IReferenceCounted
         {
             E_REQUEST_TYPE type;
             std::variant<
+                SRequestParams_EMPTY,
                 SRequestParams_CREATE_FILE,
                 SRequestParams_READ,
                 SRequestParams_WRITE
-            > params;
+            > params = SRequestParams_EMPTY();
```
pollend, line 71: this doesn't have a default constructor so its breaks a couple of classes that have a default construtor. I guess this is related to gnu implementation of STL 


I guess SRequestParams_CREATE_FILE does not have a default constructor? 

https://en.cppreference.com/w/cpp/utility/variant/variant

```
1) Default constructor. Constructs a variant holding the [value-initialized](https://en.cppreference.com/w/cpp/language/value_initialization) value of the first alternative ([index()](https://en.cppreference.com/w/cpp/utility/variant/index) is zero).

    This constructor is constexpr if and only if the value initialization of the alternative type T_0 would satisfy the requirements for a [constexpr function](https://en.cppreference.com/w/cpp/language/constexpr).
    This overload participates in overload resolution only if [std::is_default_constructible_v](http://en.cppreference.com/w/cpp/types/is_default_constructible)<T_0> is true.
```

include/nbl/type_traits.h
```diff
@@ -19,7 +19,7 @@ template <bool... Vals>
 using bool_sequence = std::integer_sequence<bool, Vals...>;
 
 template <typename T, typename... Us>
-_INLINE_VAR constexpr bool is_any_of_v = (... || std::is_same<T, Us>::value);
```
pollend, line 22: this is a window specific flag 

https://github.com/microsoft/STL/blob/main/stl/inc/type_traits

src/nbl/asset/utils/CMeshManipulator.cpp
```diff
@@ -275,7 +276,9 @@ core::smart_refctd_ptr<ICPUMeshBuffer> IMeshManipulator::createMeshBufferUniqueP
     constexpr uint32_t NEW_VTX_BUF_BINDING = 0u;
 
     auto pipeline = core::smart_refctd_ptr_static_cast<asset::ICPURenderpassIndependentPipeline>(oldPipeline->clone(0u));
-    auto& vtxParams = pipeline->getVertexInputParams() = {};
+    auto& vtxParams = pipeline->getVertexInputParams();
+    vtxParams = SVertexInputParams();
```
pollend, line 280:
```
[build] /home/.../projects/Nabla/src/nbl/asset/utils/CMeshManipulator.cpp:280:18: error: no match for ‘operator=’ (operand types are ‘nbl::asset::SVertexInputParams’ and ‘<brace-enclosed initializer list>’)
[build]   280 |     vtxParams = {};
[build]       |                  ^
[build] In file included from /home/.../projects/Nabla/include/nbl/asset/ICPURenderpassIndependentPipeline.h:8,
[build]                  from /home/.../projects/Nabla/include/nbl/asset/utils/CSPIRVIntrospector.h:16,
[build]                  from /home/.../projects/Nabla/include/nbl/asset/asset.h:47,
[build]                  from /home/.../projects/Nabla/src/nbl/asset/pch_asset.h:8,
[build]                  from /home/.../projects/Nabla/src/nbl/pch.h:12,
[build]                  from /home/.../projects/Nabla/build/src/nbl/CMakeFiles/Nabla.dir/cmake_pch.hxx:5,
[build]                  from <command-line>:
```

include/nbl/system/ISystem.h
```diff
@@ -61,10 +65,11 @@ class NBL_API2 ISystem : public core::IReferenceCounted
         {
             E_REQUEST_TYPE type;
             std::variant<
```
pollend, line 67: any reason why you need a type because std::variant provides something of the sort.

include/nbl/system/ISystemPOSIX.h
```diff
@@ -2,6 +2,8 @@
 #define _NBL_SYSTEM_C_SYSTEM_CALLER_POSIX_INCLUDED_
 
 #include "nbl/system/ISystem.h"
+#include "nbl/core/util/bitflag.h"
```
devshgraphicsprogramming, line 5: you want to include `core` headers before any of the `system` headers

include/nbl/type_traits.h
devshgraphicsprogramming: thx, I think someone just copied it from the STL verbatim for some other `is___v`

include/nbl/video/IGPUShader.h
```diff
@@ -5,11 +5,12 @@
 #ifndef __NBL_VIDEO_I_GPU_SHADER_H_INCLUDED__
 #define __NBL_VIDEO_I_GPU_SHADER_H_INCLUDED__
 
+#include "nbl/asset/IShader.h"
```
devshgraphicsprogramming, line 8: the correct order is
1. core
2. system
3. asset
4. ui
5. video
6. scene

include/nbl/video/IPhysicalDevice.h
```diff
@@ -19,6 +19,7 @@
 
 #include "SPhysicalDeviceLimits.h"
 #include "SPhysicalDeviceFeatures.h"
+#include "nbl/core/util/bitflag.h"
```
devshgraphicsprogramming, line 22: again, include ordering

src/nbl/video/CVulkanPhysicalDevice.h
```diff
@@ -1150,9 +1150,11 @@ class CVulkanPhysicalDevice final : public IPhysicalDevice
             m_properties.limits.shaderStencilExport = isExtensionSupported(VK_EXT_SHADER_STENCIL_EXPORT_EXTENSION_NAME);
             m_properties.limits.decorateString = isExtensionSupported(VK_GOOGLE_DECORATE_STRING_EXTENSION_NAME);
 
+#ifdef _NBL_WINDOWS_API_
             m_properties.limits.externalFence = isExtensionSupported(VK_KHR_EXTERNAL_FENCE_FD_EXTENSION_NAME) || isExtensionSupported(VK_KHR_EXTERNAL_FENCE_WIN32_EXTENSION_NAME);
             m_properties.limits.externalMemory = isExtensionSupported(VK_KHR_EXTERNAL_MEMORY_FD_EXTENSION_NAME) || isExtensionSupported(VK_KHR_EXTERNAL_MEMORY_WIN32_EXTENSION_NAME);
             m_properties.limits.externalSemaphore = isExtensionSupported(VK_KHR_EXTERNAL_SEMAPHORE_FD_EXTENSION_NAME) || isExtensionSupported(VK_KHR_EXTERNAL_SEMAPHORE_WIN32_EXTENSION_NAME);
+#endif
```
devshgraphicsprogramming, line 1157: thats not the proper way to resolve it, the FD extension name is always available, and its the Linux one

so only the `||` part with `WIN32` needs to be ifdeffed

devshgraphicsprogramming: same on the later `insertExtensionIfAvailable` thing later on

include/nbl/system/IFileArchive.h
```diff
@@ -9,6 +9,7 @@
 #include "nbl/system/path.h"
 #include "nbl/system/ILogger.h"
 #include "nbl/system/IFileBase.h"
+#include "nbl/core/SRange.h"
```
devshgraphicsprogramming, line 12: core comes before anything else

include/nbl/system/ISystem.h
devshgraphicsprogramming: just for this: https://github.com/Devsh-Graphics-Programming/Nabla/blob/61fa8cdedd3868f8e22c8790874eddcf72e96d57/src/nbl/system/ISystem.cpp#L311

pollend: I know that you can try to resolve the variant for each type i mean.

https://en.cppreference.com/w/cpp/utility/variant/get_if

devshgraphicsprogramming:
@pollend mind merging #457 into this first?

PR #462, opened 2023-02-21
feature: implement xcb for windowing

pollend:
## Description
WIP support that replaces x11 with xcb


src/nbl/CMakeLists.txt
```diff
@@ -559,6 +561,7 @@ endif()
 if (UNIX AND NOT ANDROID AND NOT APPLE)
 	target_include_directories(Nabla PUBLIC 
 		${X11_INCLUDE_DIR} 
+		${X11_xcb_INCLUDE_PATH}
 		X11_Xrandr_INCLUDE_PATH
 		${X11_xf86vmode_INCLUDE_PATH}
```
devshgraphicsprogramming, line 566: so we need any of the other include paths?

pollend: Probably not.

devshgraphicsprogramming: then remove

include/nbl/ui/CClipboardManagerXcb.h
```diff
@@ -0,0 +1,63 @@
...
+		XcbConnection::XCBAtomToken<core::StringLiteral("INCR")> m_INCR;
+		
+
+		XcbConnection::XCBAtomToken<core::StringLiteral("UTF8_STRING")> m_formatUTF8_0;
+		XcbConnection::XCBAtomToken<core::StringLiteral("text/plain;charset=utf-8")> m_formatUTF8_1;
+		XcbConnection::XCBAtomToken<core::StringLiteral("text/plain;charset=UTF-8")> m_formatUTF8_2;
+		XcbConnection::XCBAtomToken<core::StringLiteral("GTK_TEXT_BUFFER_CONTENTS")> m_formatGTK;
+		XcbConnection::XCBAtomToken<core::StringLiteral("STRING")> m_formatString;
+		XcbConnection::XCBAtomToken<core::StringLiteral("TEXT")> m_formatText;
+		XcbConnection::XCBAtomToken<core::StringLiteral("text/plain")> m_formatTextPlain;
```
devshgraphicsprogramming, line 54: can you split this class into `IClipboardManagerXCB` and `CClipboardManagerXCB : public IClipboardManagerXCB` which would be in `src/nbl/ui`

I'm about to refactor all this `nbl::ui` crap so it doesn't leak system headers into includes in `include/nbl/ui`

devshgraphicsprogramming:
Btw please decide on `XCB` vs `Xcb` in the class names and document that decision.

devshgraphicsprogramming:
I'm currently rewriting the `future_t` and `AsyncQueueDispatcher` because of deadlock bugs.

This will need to be brought in line with the changes later.

pollend:
ref: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/32

include/nbl/ui/CWindowManagerXCB.h
```diff
@@ -0,0 +1,94 @@
...
+    xcb_get_selection_owner
+);
+
+NBL_SYSTEM_DECLARE_DYNAMIC_FUNCTION_CALLER_CLASS(XcbIcccm, system::DefaultFuncPtrLoader,
+    xcb_icccm_set_wm_hints,
+    xcb_icccm_size_hints_set_size,
+    xcb_icccm_size_hints_set_min_size,
+    xcb_icccm_size_hints_set_max_size,
+    xcb_icccm_set_wm_normal_hints
+);
```
devshgraphicsprogramming, line 60: you can actually move these inside `CWindowManagerXCB` to make a nested class, so this stuff isn't freely handing in the namespace

include/nbl/ui/CWindowManagerXCB.h
```diff
@@ -0,0 +1,94 @@
...
+	inline SDisplayInfo getPrimaryDisplayInfo() const override final {
+		return SDisplayInfo();
+	}
+
+    CWindowManagerXCB();
+    ~CWindowManagerXCB() override = default;
+
+	virtual core::smart_refctd_ptr<IWindow> createWindow(IWindow::SCreationParams&& creationParams) override;
+
+	virtual void destroyWindow(IWindow* wnd) override final {}
```
devshgraphicsprogramming, line 81: make the class `final` and remove any `virtual` keywords from the overrides

include/nbl/ui/XCBConnection.h
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    const Xcb& getXcbFunctionTable() const { return m_windowManager->getXcbFunctionTable(); }
+    const XcbIcccm& getXcbIcccmFunctionTable() const { return m_windowManager->getXcbIcccmFunctionTable(); }
+
+    const xcb_screen_t* primaryScreen();
+
+private:
+    core::smart_refctd_ptr<CWindowManagerXCB> m_windowManager;
+    xcb_connection_t* m_connection = nullptr;
```
devshgraphicsprogramming, line 101: what is an XCB Connection and why is it holding onto the window manager?

include/nbl/ui/IWindowXCB.h
```diff
@@ -0,0 +1,88 @@
...
+
+            XCBConnection::MotifWmHints hints;
+            hints.flags = motifFlags.value;
+            hints.functions = motifFunctions.value;
+            hints.decorations = motifDecorations.value;
+            hints.input_mode = 0;
+            hints.status = 0;
+            return hints;
+
+        }
```
devshgraphicsprogramming, line 80: all this stuff you want to push into `CWindowXCB`, so that you may remove any XCB and X11 header includes from headers in `include/nbl/ui`

src/nbl/video/CSurfaceVulkan.cpp
```diff
@@ -143,5 +143,29 @@ namespace nbl::video
 			return nullptr;
 		}
 	}
+#elif defined(_NBL_PLATFORM_LINUX_)
+	
+	#include <xcb/xcb.h>
```
devshgraphicsprogramming, line 148: for this include to work you need to remove the enclosing `namespace nbl::video {` and do 
```cpp
using namespace nbl;
using namespace nbl::video;
```
at the top instead

include/nbl/ui/IWindowXCB.h
```diff
@@ -0,0 +1,88 @@
...
+    protected:
+        virtual ~IWindowXCB() = default;
+        inline IWindowXCB(SCreationParams&& params) : IWindow(std::move(params)) {}
+
+    public:
+        using IWindow::IWindow;
+
+        const void* getNativeHandle() const { return nullptr; }
+        virtual xcb_window_t getXcbWindow() const = 0;
+        virtual xcb_connection_t* getXcbConnection() const = 0;
```
devshgraphicsprogramming, line 27: I suggest forward declaring outside the namespace
```cpp
using xcb_window_t = uint32_t;
struct xcb_connection_t;
```
and then following our convention which is
```cpp
struct native_handle_t
{
    xcb_window_t window;
    xcb_connection_t* connection;
};
const native_handle_t& getNativeHandle() const =0;
```

include/nbl/ui/IClipboardManagerXCB.h
```diff
@@ -0,0 +1,28 @@
...
+        virtual std::string getClipboardText() = 0;
+        virtual bool setClipboardText(const std::string_view& data) = 0;
+        virtual void process(const IWindowXCB* window, xcb_generic_event_t* event) = 0;
+};
+
+}
+
+#endif
+
+#endif
```
devshgraphicsprogramming, line 28: you probably don't need an `IClip` in header dir and `CClip` in source, see if you can go with `CClip` in source dir right away

include/nbl/ui/CWindowManagerXCB.h
```diff
@@ -0,0 +1,94 @@
+#ifndef C_WINDOW_MANAGER_XCB
+#define C_WINDOW_MANAGER_XCB
```
devshgraphicsprogramming, line 2: you want to start and end your headerguard with `_` and include full namespace and the `INCLUDED` suffix

include/nbl/ui/CCursorControlXCB.h
```diff
@@ -0,0 +1,34 @@
+#ifndef __NBL_SYSTEM_C_CURSOR_CONTROL_XCB_H_INCLUDED__
+#define __NBL_SYSTEM_C_CURSOR_CONTROL_XCB_H_INCLUDED__
```
devshgraphicsprogramming, line 2: wrong namespace, also one too many `_` on the beginning and end

include/nbl/ui/CCursorControlXCB.h
```diff
@@ -0,0 +1,34 @@
...
+
+        void setPosition(SPosition pos) override;
+        void setRelativePosition(IWindow* window, SRelativePosition pos) override;
+
+        SPosition getPosition() override;
+        SRelativePosition getRelativePosition(IWindow* window) override;
+    private: 
+        core::smart_refctd_ptr<XCBConnection> m_connection;
+};
+}
```
devshgraphicsprogramming, line 30: I dunno if its a good idea in your case, but for the Win32 stuff (#476) it ended up making more sense to have the manager inherit both from `ICursorControl` and `IWindowManager` because of the global nature of the cursor on Windows.

Don't know if this is applicable here

include/nbl/ui/CClipboardManagerXCB.h
```diff
@@ -0,0 +1,52 @@
...
+        XCBConnection::XCBAtomToken<core::StringLiteral("STRING")> m_formatString;
+        XCBConnection::XCBAtomToken<core::StringLiteral("TEXT")> m_formatText;
+        XCBConnection::XCBAtomToken<core::StringLiteral("text/plain")> m_formatTextPlain;
+};
+
+}
+
+#endif
+
+#endif
```
devshgraphicsprogramming, line 52: this header should probably be in `src/nbl/ui` don't think anyone outside the engine needs to know about it

include/nbl/ui/CWindowManagerXCB.h
```diff
@@ -0,0 +1,94 @@
...
+
+	virtual bool setWindowSize_impl(IWindow* window, uint32_t width, uint32_t height) override;
+	virtual bool setWindowPosition_impl(IWindow* window, int32_t x, int32_t y) override;
+	virtual bool setWindowRotation_impl(IWindow* window, bool landscape) override;
+	virtual bool setWindowVisible_impl(IWindow* window, bool visible) override;
+	virtual bool setWindowMaximized_impl(IWindow* window, bool maximized) override;
+
+	inline SDisplayInfo getPrimaryDisplayInfo() const override final {
+		return SDisplayInfo();
+	}
```
devshgraphicsprogramming, line 74: TODO

include/nbl/ui/CWindowManagerXCB.h
```diff
@@ -0,0 +1,94 @@
...
+#include "nbl/ui/IWindow.h"
+#include "nbl/ui/IWindowManager.h"
+
+#include <functional>
+#include <memory>
+#include <string>
+
+#include <xcb/xcb.h>
+#include <xcb/xcb_icccm.h>
+#include <xcb/xproto.h>
```
devshgraphicsprogramming, line 18: it would be decent to follow the Win32 pattern in #476 and create a `IWindowManagerXCB` as a public header with a single method
```cpp
static core::smart_refctd_ptr<IWindowManagerXCB> create();
```
then hide all these implementation details in `src`

include/nbl/ui/CWindowXCB.h
```diff
@@ -0,0 +1,85 @@
+#ifndef __C_WINDOW_XCB_H_INCLUDED__
+#define __C_WINDOW_XCB_H_INCLUDED__
```
devshgraphicsprogramming, line 2: header guard needs improving, also `CWindowXCB` needs to be in `src`

src/nbl/ui/CClipboardManagerXCB.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <xcb/xproto.h>
+
+namespace nbl::ui
+{
+    std::string CClipboardManagerXCB::getClipboardText() {
+        {
+            std::unique_lock<std::mutex> lk(m_clipboardMutex);
+            m_clipboardResponseCV.wait_until(lk, std::chrono::system_clock::now() + std::chrono::seconds(1));
+        }
+        std::lock_guard lk(m_clipboardMutex);
```
devshgraphicsprogramming, line 18: what on earth is going on here?

src/nbl/ui/CClipboardManagerXCB.cpp
```diff
@@ -0,0 +1,147 @@
...
+        std::string response = std::move(m_clipboardResponse);
+        m_clipboardResponse = std::string();
+        return response;
+    }
+
+
+    bool CClipboardManagerXCB::setClipboardText(const std::string_view& data) {
+        std::lock_guard lk(m_clipboardMutex);
+        m_savedClipboard = data;
```
devshgraphicsprogramming, line 27: arent you missing some event you should send or something?

src/nbl/ui/CClipboardManagerXCB.cpp
```diff
@@ -0,0 +1,147 @@
...
+    void CClipboardManagerXCB::process(const IWindowXCB* window, xcb_generic_event_t* event) {
+        const auto& xcb = m_connection->getXcbFunctionTable();
+        
+        auto TARGETS = m_connection->resolveAtom(m_TARGETS);
+
+        switch(event->response_type & ~0x80) {
+            // XCB_ATOM
+            // Somone is requesting the clipboard data
+            case XCB_SELECTION_REQUEST: {
+                auto* sne = reinterpret_cast<xcb_selection_request_event_t*>(event);
```
devshgraphicsprogramming, line 40: seems like this is a Window event, ergo `CWindowXCB` should probably inherit from `IClipboardManager` and you should have no `ClipboardManagerXCB`

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+
+void CWindowXCB::CDispatchThread::init()
+{
+
+}
+
+void CWindowXCB::CDispatchThread::exit()
+{
+}
```
devshgraphicsprogramming, line 107: btw these methods are wholly optional IIRC (if you don't define them, everything will work anyway

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+    auto wnd = static_cast<IWindowXCB*>(window);
+    wnd->setWindowVisible_impl(visible);
+    return true;
+}
+
+bool CWindowManagerXCB::setWindowMaximized_impl(IWindow* window, bool maximized) {
+    auto wnd = static_cast<IWindowXCB*>(window);
+    wnd->setWindowMaximized_impl(maximized);
+    return true;
+}
```
devshgraphicsprogramming, line 148: why isn't this in its own `.cpp` file?

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+CWindowXCB::CWindowXCB(core::smart_refctd_ptr<CWindowManagerXCB>&& winManager, SCreationParams&& params):
+    IWindowXCB(std::move(params)),
+    m_windowManager(winManager),
+    m_connection(core::make_smart_refctd_ptr<XCBConnection>(core::smart_refctd_ptr<CWindowManagerXCB>(m_windowManager))),
+    m_cursorControl(core::make_smart_refctd_ptr<CCursorControlXCB>(core::smart_refctd_ptr<XCBConnection>(m_connection))),
```
devshgraphicsprogramming, line 154: you could probably fold these into the window manager itself.

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+bool CWindowManagerXCB::setWindowMaximized_impl(IWindow* window, bool maximized) {
+    auto wnd = static_cast<IWindowXCB*>(window);
+    wnd->setWindowMaximized_impl(maximized);
+    return true;
+}
+
+CWindowXCB::CWindowXCB(core::smart_refctd_ptr<CWindowManagerXCB>&& winManager, SCreationParams&& params):
+    IWindowXCB(std::move(params)),
+    m_windowManager(winManager),
```
devshgraphicsprogramming, line 152: you want to `std::move()` is helps to skip an atomic counter op

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+    
+    if(isAlwaysOnTop()) {
+        XCBConnection::XCBAtomToken<core::StringLiteral("NET_WM_STATE_ABOVE")> NET_WM_STATE_ABOVE;
+        m_connection->setNetMWState(
+            primaryScreen->root,
+            m_xcbWindow, false, m_connection->resolveAtom(NET_WM_STATE_ABOVE));
+    }
+
+    xcb.pxcb_map_window(m_connection->getRawConnection(), m_xcbWindow);
+    xcb.pxcb_flush(m_connection->getRawConnection());
```
devshgraphicsprogramming, line 206: can any of this fail?

if yes, can you do all this inside the `CWindowManagerXCB::createWindow` and pass `xcbWindow` and any other variables we need to remembers straight to the constructor?

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+bool CWindowXCB::setWindowPosition_impl(int32_t x, int32_t y) {
+    auto& xcb = m_windowManager->getXcbFunctionTable();
+
+    const int32_t values[] = { x, y };
+    auto cookie = xcb.pxcb_configure_window_checked(m_connection->getRawConnection(), m_xcbWindow, XCB_CONFIG_WINDOW_X | XCB_CONFIG_WINDOW_Y, values);
+    bool check = checkXcbCookie(xcb, m_connection->getRawConnection(), cookie);
+    xcb.pxcb_flush(m_connection->getRawConnection());
+    assert(check);
+    return true;
```
devshgraphicsprogramming, line 249: do not asert, this is why the function returns a bool, so it can handle failures

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+    auto& xcb = m_windowManager->getXcbFunctionTable();
+    auto& xcbIccm = m_windowManager->getXcbIcccmFunctionTable();
+
+    xcb_size_hints_t hints = {0};
+
+    xcbIccm.pxcb_icccm_size_hints_set_size(&hints, true, width, height);
+    if(!isResizable()) {
+        xcbIccm.pxcb_icccm_size_hints_set_min_size(&hints, width, height);
+        xcbIccm.pxcb_icccm_size_hints_set_max_size(&hints, width, height);
+    }
```
devshgraphicsprogramming, line 236: if you look at `IWindow::setWindowSize` we shouldn't be able to get here if the window is not resizable

src/nbl/ui/CWindowXCB.cpp
devshgraphicsprogramming: `init` isn't but exit is

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+    if(isAlwaysOnTop()) {
+        XCBConnection::XCBAtomToken<core::StringLiteral("NET_WM_STATE_ABOVE")> NET_WM_STATE_ABOVE;
+        m_connection->setNetMWState(
+            primaryScreen->root,
+            m_xcbWindow, false, m_connection->resolveAtom(NET_WM_STATE_ABOVE));
+    }
+
+    xcb.pxcb_map_window(m_connection->getRawConnection(), m_xcbWindow);
+    xcb.pxcb_flush(m_connection->getRawConnection());
+    m_dispatcher.start();
```
devshgraphicsprogramming, line 207: there's a fun trick, `IThreadHandler` has a special constructor with a tag parameter which tells it to start the thread on construction

src/nbl/ui/CWindowXCB.cpp
```diff
@@ -0,0 +1,296 @@
...
+    if(m_quit) {
+        return; 
+    }
+    auto& xcb = m_window.m_windowManager->getXcbFunctionTable();
+    auto& connection = m_window.m_connection;
+
+    auto MW_DELETE_WINDOW = connection->resolveAtom(m_window.m_WM_DELETE_WINDOW);
+    auto NET_WM_PING = connection->resolveAtom(m_window.m_NET_WM_PING);
+
+    if(auto event = xcb.pxcb_wait_for_event(connection->getRawConnection())) {
```
devshgraphicsprogramming, line 45: this puts the thread to sleep if there's no events, correct?

pollend: yep, xcb implements it own dispatch queue.

src/nbl/ui/CWindowXCB.cpp
pollend: I think I was copying the windowmanager, not sure why I followed along.

include/nbl/ui/XCBConnection.h
pollend: I need it for xcb and xcbIccm

include/nbl/ui/CClipboardManagerXcb.h
devshgraphicsprogramming: Refactor is done, merge latest `master` to yourself.

src/nbl/ui/CClipboardManagerXCB.cpp
pollend: the clipboard is a mess ... I need to rethink this

src/nbl/ui/CWindowXCB.cpp
pollend: umm I'll have to explore that

src/nbl/ui/CClipboardManagerXCB.cpp
pollend: why didn't I do that originally ...

include/nbl/ui/CCursorControlXCB.h
pollend: not sure how that works in terms of xcb I know x11 supports more then one cursor but applications tends to catch fire if you try to set it up like that ...

PR #465, opened 2023-02-24
bugfix: address build issues with embedded assets for clang and gnu compiler

pollend:
## Description

this addresses build problems with clang and g++ 

- remove builtinResources.h from builtinResourceData.cpp
- remove typename from  before NBL_CORE_UNIQUE_STRING_LITERAL_TYPE


devshgraphicsprogramming:
@AnastaZIuk make sure to incorporate this into your Built-in resources rework.

PR #466, opened 2023-02-24
bugfix: address crash releated to clang when compiling promoteBufferFormat or promoteImageFormat

pollend:
## Description

this is a workaround to a crash observed and tested against clang-14,15,16 for linux. the workaround is just to expand the templates themselves so its quite a bit more verbose then the original code. it doesn't seem too bad though, made sure to keep the API basically the same. 

ref: https://github.com/llvm/llvm-project/issues/60938#issuecomment-1442706682

### DIFF
--- a/include/nbl/video/IPhysicalDevice.h
+++ b/include/nbl/video/IPhysicalDevice.h
@@ -581,33 +581,15 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
             );
         }
 
-        template<typename FORMAT_USAGE>
-        struct FormatPromotionRequest
-        {
+        struct SBufferFormatPromotionRequest {
             asset::E_FORMAT originalFormat = asset::EF_UNKNOWN;
-            FORMAT_USAGE usages = FORMAT_USAGE();
-
-            struct hash
-            {
-                // pack into 64bit for easy hashing 
-                uint64_t operator()(const FormatPromotionRequest<FORMAT_USAGE>& r) const
-                {
-                    uint64_t msb = uint64_t(std::hash<FORMAT_USAGE>()(r.usages));
-                    return (msb << 32u) | r.originalFormat;
-                }
-            };
-
-            struct equal_to
-            {
-                bool operator()(const FormatPromotionRequest<FORMAT_USAGE>& l, const FormatPromotionRequest<FORMAT_USAGE>& r) const
-                {
-                    return l.originalFormat == r.originalFormat && l.usages == r.usages;
-                }
-            };
+            SFormatBufferUsages::SUsage usages = SFormatBufferUsages::SUsage();
         };
 
-        using SBufferFormatPromotionRequest = FormatPromotionRequest<IPhysicalDevice::SFormatBufferUsages::SUsage>;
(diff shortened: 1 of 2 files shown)


include/nbl/video/IPhysicalDevice.h
```diff
@@ -581,33 +581,15 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
             );
         }
 
-        template<typename FORMAT_USAGE>
-        struct FormatPromotionRequest
-        {
+        struct SBufferFormatPromotionRequest {
             asset::E_FORMAT originalFormat = asset::EF_UNKNOWN;
-            FORMAT_USAGE usages = FORMAT_USAGE();
```
devshgraphicsprogramming, line 588: @Erfan-Ahmadi remind me why FORMAT_USAGE was a template parameter?

include/nbl/video/IPhysicalDevice.h
```diff
@@ -755,4 +766,28 @@ namespace std
...
+
+    inline bool IPhysicalDevice::SBufferFormatPromotionRequestEqualTo::operator()(const SBufferFormatPromotionRequest& l, const SBufferFormatPromotionRequest& r) const
+    {
+        return l.originalFormat == r.originalFormat && l.usages == r.usages;
+    }
+
+     inline bool IPhysicalDevice::SImageFormatPromotionRequestEqualTo::operator()(const SImageFormatPromotionRequest& l, const SImageFormatPromotionRequest& r) const
+    {
+        return l.originalFormat == r.originalFormat && l.usages == r.usages;
+    }
```
devshgraphicsprogramming, line 789: is there any problems with having these inline where the functions are declared?

pollend: std::hash is declared above. its why I've declared it after that.

devshgraphicsprogramming: ok I guess that solves some issues :D

pollend: how does it work in a template though :?

include/nbl/video/IPhysicalDevice.h
pollend: pretty sure it was to have equal and hash defined.

include/nbl/video/IPhysicalDevice.h
devshgraphicsprogramming: what you did is correct otherwise we'd have to "foward" declare the std::hash template which would have been hard

include/nbl/video/IPhysicalDevice.h
devshgraphicsprogramming: hmm ok, because if it wasn't template dependant then if would complain about the lack of definition for `std::hash`

devshgraphicsprogramming:
@Erfan-Ahmadi I delegate the power and responsibility for pressing the Big Green Button to you.

include/nbl/video/IPhysicalDevice.h
pollend: just curious how you declare something ahead of std::hash in a template but if its written out then the compiler will complain about it

include/nbl/video/IPhysicalDevice.h
pollend: that's new behavior for me.

PR #470, opened 2023-03-03
refactor: minor rework ILogicalDevice with span

pollend:
- flushMappedMemoryRanges
- invalidateMappedMemoryRanges

## Description
this is a test branch that explores replacing SRange with span for ILogicalDevice.

### DIFF
--- a/include/nbl/core/SRange.h
+++ b/include/nbl/core/SRange.h
@@ -11,6 +11,7 @@
 #include <type_traits>
 #include <utility>
 #include <iterator>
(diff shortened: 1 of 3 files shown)


devshgraphicsprogramming:
Ok it seems that `std::span` can only replace `core::SRange` when the iterators are default (pointers).

Also it seems that most places where we allow some other iterators we could just use the C++20 concept of a `view`

We'll address this in the near future when we'll work on the WASM port.

PR #480, opened 2023-03-18
bugfix: address memory assertions for command pool

pollend:
## Description
I was observing an assertion on IGPUCommandPool.
```
Alloc'ed 0x5555573aec80 with 56
Alloc'ed 0x5555573749c0 with 56
Alloc'ed 0x555556e9b230 with 56
Alloc'ed 0x555556e9c4d0 with 56
Alloc'ed 0x5555571e8410 with 128
Alloc'ed 0x555556ea5610 with 128
Alloc'ed 0x555556eaabc0 with 128
Alloc'ed 0x5555573fd0c0 with 128
Alloc'ed 0x55555734a090 with 128
Alloc'ed 0x7fffddfff010 with 33556544
helloworld_d: /home/michaelpollind/projects/Nabla/include/nbl/video/IGPUCommandPool.h:66: nbl::video::IGPUCommandPool::ICommand::ICommand(uint32_t): Assertion `ptrdiff_t(this) % alignof(ICommand) == 0' failed.

Thread 1 "helloworld_d" received signal SIGABRT, Aborted.
```

by default the pool allocates on a 16 bit boundary so 0x7fffddfff010 is a multiple of 16 but not 64.

### DIFF
--- a/include/nbl/core/memory/memory.h
+++ b/include/nbl/core/memory/memory.h
@@ -49,7 +49,7 @@ namespace impl
         if (alignment < MIN_ALIGN) alignment = MIN_ALIGN;
         if (size == 0) return nullptr;
         void* p;
-        if (::posix_memalign(&p, alignment>alignof(std::max_align_t) ? alignof(std::max_align_t):alignment, size) != 0) p = nullptr;
+        if (::posix_memalign(&p, alignment, size) != 0) p = nullptr;
         return p;
     }
 }
--- a/include/nbl/video/IGPUCommandPool.h
+++ b/include/nbl/video/IGPUCommandPool.h
@@ -344,7 +344,10 @@ private:
         }
 
         CCommandSegment* m_head = nullptr;
-        core::CMemoryPool<core::PoolAddressAllocator<uint32_t>, core::default_aligned_allocator, false, uint32_t> m_pool;
+        
+        template <typename T>
+        using pool_alignment = core::aligned_allocator<T,COMMAND_SEGMENT_ALIGNMENT>;
+        core::CMemoryPool<core::PoolAddressAllocator<uint32_t>, pool_alignment, false, uint32_t> m_pool;
     };
 
     friend class IGPUCommandBuffer;


include/nbl/core/memory/memory.h
```diff
@@ -49,7 +49,7 @@ namespace impl
         if (alignment < MIN_ALIGN) alignment = MIN_ALIGN;
         if (size == 0) return nullptr;
         void* p;
-        if (::posix_memalign(&p, alignment>alignof(std::max_align_t) ? alignof(std::max_align_t):alignment, size) != 0) p = nullptr;
```
pollend, line 52: I couldn't figure out why this originally sets an upper bound of 16 here. 

I was looking at the man pages for https://linux.die.net/man/3/posix_memalign the only restriction is that is had to be a power of two.

```
EINVAL

The alignment argument was not a power of two, or was not a multiple of sizeof(void *). 
```

devshgraphicsprogramming: ah, `max_align_t` is a constraint on your stack https://en.cppreference.com/w/cpp/types/max_align_t

> Pointers returned by allocation functions such as [std::malloc](https://en.cppreference.com/w/cpp/memory/c/malloc) are suitably aligned for any object, which means they are aligned at least as strictly as std::max_align_t.

Basically its not really valid to declare structs or classes `alignas(A) T` with `A>alignof(std::max_align_t)` or stuff like this will fail:
```cpp
T t;
assert((ptrdiff_t(&t)%alignof(T))==0);
```

But yeah you're right, for the heap it shouldn't matter

include/nbl/video/IGPUCommandPool.h
```diff
@@ -344,7 +344,10 @@ class IGPUCommandPool : public core::IReferenceCounted, public IBackendObject
         }
 
         CCommandSegment* m_head = nullptr;
-        core::CMemoryPool<core::PoolAddressAllocator<uint32_t>, core::default_aligned_allocator, false, uint32_t> m_pool;
+        
+        template <typename T>
+        using pool_alignment = core::aligned_allocator<T,COMMAND_SEGMENT_ALIGNMENT>;
+        core::CMemoryPool<core::PoolAddressAllocator<uint32_t>, pool_alignment, false, uint32_t> m_pool;
```
devshgraphicsprogramming, line 350: can you explain this change?

pollend: sets the alignment to 64.

devshgraphicsprogramming: but shouldn't the `m_pool.emplace` take into account the alignment of what its emplacing?
https://github.com/Devsh-Graphics-Programming/Nabla/blob/6a17dc80e50be0fee6114e2efd2408fbb223968a/include/nbl/core/containers/CMemoryPool.h#LL43C19-L43C20

pollend: the default alignment is 16
0x7fffddfff010 is a multiple of 16 but not 64

pollend: > but shouldn't the `m_pool.emplace` take into account the alignment of what its emplacing? https://github.com/Devsh-Graphics-Programming/Nabla/blob/6a17dc80e50be0fee6114e2efd2408fbb223968a/include/nbl/core/containers/CMemoryPool.h#LL43C19-L43C20

the memory where its getting allocated is aligned to 16.

devshgraphicsprogramming:
```cpp
    class alignas(COMMAND_ALIGNMENT) ICommand
    {
```

you can clearly see it being aligned.

> the memory where its getting allocated is aligned to 16.

this is a problem that needs fixing, 

are you getting 64 here and if not why not?
https://github.com/Devsh-Graphics-Programming/Nabla/blob/6a17dc80e50be0fee6114e2efd2408fbb223968a/include/nbl/core/containers/CMemoryPool.h#LL43C19-L43C20

pollend: when you malloc memory there isn't any guarantee that it will be on a 64 bit boundary? it doesn't know anything abotu the memory requriments for the struct right?

image

devshgraphicsprogramming: > when you malloc memory there isn't any guarantee that it will be on a 64 bit boundary? 

`malloc` doesn't, but `core::aligned_allocator::allocate` asks for the alignment explicitly.

I'm asking you why isn't `CMemoryPool::allocate` being called with `a=64` when emplacing the `CommandSegment` ?

pollend: you can look at aligned_allocator i think there are two methods allocate and allocate with an alignment. memory pool calls allocated so it defaults to what you pass into the second template parameter. in this case I've passed in COMMAND_SEGMENT_ALIGNMENT which is 64.

devshgraphicsprogramming: See whether this function is being called with alignment=64
https://github.com/Devsh-Graphics-Programming/Nabla/blob/c6d0dde3f7b5e4bde019290562997c8dfad28268/include/nbl/core/alloc/SimpleBlockBasedAllocator.h#L104

if it is then the only reason why it would fail is because the Block itself is not aligned to 64
https://github.com/Devsh-Graphics-Programming/Nabla/blob/c6d0dde3f7b5e4bde019290562997c8dfad28268/include/nbl/core/alloc/SimpleBlockBasedAllocator.h#L45

which would require `meta_alignment<64`  here
https://github.com/Devsh-Graphics-Programming/Nabla/blob/c6d0dde3f7b5e4bde019290562997c8dfad28268/include/nbl/core/alloc/SimpleBlockBasedAllocator.h#L196

just wondering what fk-ed up along the way, cause this seems like an important bug to diagnose.

PR #482, opened 2023-03-21
Proposal of interface rewrite

devshgraphicsprogramming:


include/nbl/asset/filters/kernels/CDiracImageFilterKernel.h
```diff
@@ -1,41 +1,32 @@
...
-
-public:
-	CDiracImageFilterKernel() : Base(0.f) {}
+	constexpr static inline uint32_t k_smoothness = 0;
+	constexpr static inline float min_support = std::nextafter<float>(0.f,-1.f);
+	constexpr static inline float max_support = std::nextafter<float>(0.f,+1.f);
 
-	template <uint32_t derivative = 0>
-	inline float weight(float x, int32_t channel) const
+	template<int32_t derivative=0>
```
achalpandeyy, line 15: I don't get the reason of making it `int32_t` from `uint32_t`. It doesn't matter for positive values because we probably won't have so much higher order derivatives, but do you expect `derivative` to be negative?

include/nbl/asset/filters/kernels/CDiracImageFilterKernel.h
```diff
@@ -1,41 +1,32 @@
...
+	inline float operator(float x, uint32_t channel) const
```
achalpandeyy, line 16: You made an inline method named `operator`?

include/nbl/asset/filters/kernels/CDiracImageFilterKernel.h
devshgraphicsprogramming: well if its negative you'd integrate, but yeah it doesn't matter much

PR #499, opened 2023-05-30
Fix Validation Errors and Bumpmaps

### LINKED ISSUE #411: Example 06.Meshloaders black objects on DebugMode
## Describe the bug
objects display as all black in Debug Mode.
RenderDoc Capture shows the problem is with the 1024x1024 bumpmap being all black or 0,0 for those objects but they should be all 1,1 (yellow in RG16)

Problem may be with the loaders and the origin may be the recent changes made to the sponza zip file that caused this.

## Steps to Reproduce
Run example 06 on master branch on debug mode
The correct version is rendered in release mode.

You should see something like this

### LINKED ISSUE #472: Fix Threadsafeness of Refcounting on Descriptor Pools and Sets
## Describe the bug
The current design fails to account that a call to `~IGPUDescriptorSet` or `~IDescriptorPool` can come from any thread and we can't impose external synchronization constraints on them.

A similar issue may affect CommandBuffers and CommandPools.

We decided to not tackle that as part of #345 

## Steps to Reproduce
Just think about it.

## Expected vs observed behavior
Destructors should be thread safe, and anything needed to ensure that should propagate to other functions.

However other methods of DescriptorPool and Set should retain their external synchronisation obligation.

## Environment
Everywhere

devshgraphicsprogramming:
## Description

Fixes:
- querying for device properties and limits for extensions that are not supported (dynamic pNext chain construction)
- querying format usages for formats which are not supported (no extension or vulkan instance version)
- not generating mip-maps for images that already have defined mip pyramids (like KTX/DDS)
- only make the mip-pyramid full if there was no mip-pyramid to start with in the first place
- fix #411 
- upstream a bunch of fixes from `ditt` branch
- minor issues with compilation of certain examples

## Testing 

Just running examples, nothing complex like comparing device props and limit structs before and after implementing changes.

## TODO list:

Actually fix #472

### DIFF
--- a/include/nbl/asset/filters/CBlitImageFilter.h
+++ b/include/nbl/asset/filters/CBlitImageFilter.h
@@ -627,7 +627,10 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
 					});
 					// we'll only get here if we have to do coverage adjustment
 					if (needsNormalization && lastPass)
+					{
+						state->normalization.finalize<value_type>();
 						storeToImage(core::rational<int64_t>(cvg_num,cvg_den),axis,outOffsetLayer);
+					}
 				};
 				// filter in X-axis
 				filterAxis(IImage::ET_1D,scaledKernelX);
--- a/include/nbl/asset/filters/CBlitImageFilter.h
+++ b/include/nbl/asset/filters/CBlitImageFilter.h
@@ -87,8 +87,9 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
 	private:
 		using value_type = typename KernelX::value_type;
 		using base_t = CBlitImageFilterBase<value_type,Swizzle,Dither,Normalization,Clamp>;
+		//using this_t = CBlitImageFilter<Swizzle,Dither,Normalization,Clamp,KernelX,KernelX,KernelX,lut_value_type>;
 
-		_NBL_STATIC_INLINE_CONSTEXPR auto MaxChannels = blit_utils_t::MaxChannels;
+		static inline constexpr auto MaxChannels = blit_utils_t::MaxChannels;
 
 	public:
 		// we'll probably never remove this requirement
@@ -96,6 +97,19 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
 
 		virtual ~CBlitImageFilter() {}
 
+		enum E_SCRATCH_USAGE
+		{
+			ESU_SCALED_KERNEL_PHASED_LUT = 0,
+			ESU_DECODE_WRITE = 1,
+			ESU_BLIT_X_AXIS_READ = ESU_DECODE_WRITE,
+			ESU_BLIT_X_AXIS_WRITE = 2,
+			ESU_BLIT_Y_AXIS_READ = ESU_BLIT_X_AXIS_WRITE,
+			ESU_BLIT_Y_AXIS_WRITE = 3,
+			ESU_BLIT_Z_AXIS_READ = ESU_BLIT_Y_AXIS_WRITE,
+			ESU_BLIT_Z_AXIS_WRITE = 4,
+			ESU_ALPHA_HISTOGRAM = 5,
+			ESU_COUNT
+		};
 		class CState : public IImageFilter::IState, public base_t::CStateBase
 		{
 			public:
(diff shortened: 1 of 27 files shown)


include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -1036,7 +1041,10 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
 
         if (needToCompMipsForThisImg(cpuimg))
         {
-            params.usage |= asset::IImage::EUF_TRANSFER_SRC_BIT;
+            params.usage |= asset::IImage::EUF_TRANSFER_SRC_BIT; // this is for blit
+            // I'm already adding usage flags for mip-mapping compute shader
+            params.usage |= asset::IImage::EUF_SAMPLED_BIT; // to read source mips
+            params.usage |= asset::IImage::EUF_STORAGE_BIT; // to write dst mips
```
devshgraphicsprogramming, line 1047: am I likely to fuckup the format promotion with this?

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -1025,8 +1030,8 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
         promotionRequest.originalFormat = params.format;
         promotionRequest.usages = {};
 
-        const bool integerFmt = asset::isIntegerFormat(params.format);
-        if (!integerFmt)
+        // override the mip-count if its not an integer format and there was no mip-pyramid specified 
+        if (params.mipLevels==1u && !asset::isIntegerFormat(params.format))
```
devshgraphicsprogramming, line 1034: I hope this makes sense @Erfan-Ahmadi

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -847,12 +847,17 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
             return false;
         auto format = img->getCreationParameters().format;
         if (asset::isIntegerFormat(format) || asset::isBlockCompressionFormat(format))
             return false;
+        // its enough to define a single mipmap region above the base level to prevent automatic computation
+        for (auto& region : img->getRegions())
+        if (region.imageSubresource.mipLevel)
+            return false;
         return true;
     };
```
devshgraphicsprogramming, line 862: just give this a look @Erfan-Ahmadi

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
devshgraphicsprogramming: ok answered my own question with some testing, doing this causes an SRGB image to be promoted to 16bit UNORM

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -1025,36 +1030,60 @@ auto IGPUObjectFromAssetConverter::create(const asset::ICPUImage** const _begin,
...
+            // but image views with formats that are store-able can be created
+            if (!newFormatIsStorable)
+            {
+                params.flags |= asset::IImage::ECF_MUTABLE_FORMAT_BIT;
+                params.flags |= asset::IImage::ECF_EXTENDED_USAGE_BIT;
+                if (asset::isBlockCompressionFormat(newFormat))
+                    params.flags |= asset::IImage::ECF_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT;
+            }
+        }
+#endif
```
devshgraphicsprogramming, line 1082: aaaand it seems we have a blocker @Erfan-Ahmadi

devshgraphicsprogramming: basically without the list of all the formats we'll use, this will never pass validation :(

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
Erfan-Ahmadi: yea it makes sense, althought I would use  `> 0u` instead of treating it like a bool :D

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
Erfan-Ahmadi: yea it does

include/nbl/video/utilities/IGPUObjectFromAssetConverter.h
```diff
@@ -1525,20 +1549,55 @@ inline created_gpu_object_array<asset::ICPUImageView> IGPUObjectFromAssetConvert
...
             params.image = (*gpuDeps)[redirs[i]];
-            params.viewType = static_cast<IGPUImageView::E_TYPE>(cpuparams.viewType);
-            params.format = params.image->getCreationParameters().format;
-            memcpy(&params.components, &cpuparams.components, sizeof(params.components));
-            params.subresourceRange = cpuparams.subresourceRange;
-            params.subresourceRange.levelCount = (*gpuDeps)[redirs[i]]->getCreationParameters().mipLevels - params.subresourceRange.baseMipLevel;
+            const auto& gpuImgParams = params.image->getCreationParameters();
+            // override the view's format if the source image got promoted, a bit crude, but don't want to scratch my head about how to promote the views and guess semantics
+            const bool formatGotPromoted = asset::getFormatClass(cpuParams.format)!=asset::getFormatClass(gpuImgParams.format);
+            params.format = formatGotPromoted ? gpuImgParams.format:cpuParams.format;
```
devshgraphicsprogramming, line 1568: @Erfan-Ahmadi corrected the thing that would wipe away all views' formats when using Asset Converter

include/nbl/asset/filters/CBlitImageFilter.h
```diff
@@ -541,10 +552,10 @@ class CBlitImageFilter : public CImageFilter<CBlitImageFilter<Swizzle,Dither,Nor
 									continue;
 
 								auto sample = lineBuffer+i*MaxChannels;
-								value_type swizzledSample[MaxChannels];
+								value_type preSwizzleSample[MaxChannels];
 
 								// TODO: make sure there is no leak due to MaxChannels!
-								base_t::template onDecode(inFormat, state, srcPix, sample, swizzledSample, inBlockCoord.x, inBlockCoord.y);
+								base_t::template onDecode(inFormat, state, srcPix, preSwizzleSample, sample, inBlockCoord.x, inBlockCoord.y);
```
devshgraphicsprogramming, line 558: @achalpandeyy before the `swizzledSample` wouldn't get used at all which would mess with Derivative Map from Heightmap generation

PR #500, opened 2023-05-31
Make SHARED builtin-resources builds work

AnastaZIuk:
Require Nabla to be linked into BRs libraries due to some codebase symbols, add SHARED builtin-resources test to examples 52. Apply some fixes and mute IntelliSense in builtin-resources headers


src/nbl/builtin/builtinHeaderGen.py
```diff
@@ -28,6 +28,13 @@
 
     outp.write("#ifndef _" + guardSuffix + "_BUILTINRESOURCEDATA_H_\n")
     outp.write("#define _" + guardSuffix + "_BUILTINRESOURCEDATA_H_\n")
+  
+    outp.write("#ifdef _WIN32 // Visual Studio define\n")
```
devshgraphicsprogramming, line 32: use `__INTELLISENSE__`

AnastaZIuk: resolved

PR #502, opened 2023-05-31
Mount directories fix

Hazardu:
## Description
In this PR `IFileArchive::listAssets(const path& pathRelativeToArchive)` method was changed to no longer be virtual, to take a parameter that is not const and its implementation was changed to return lexicographically in range matches.
Because the match range's bounds are not easy to imagine, i wrote a long comment explaining what is happening.

`IFileArchive::m_items` became private, and can be assigned via protected method

### DIFF
--- a/src/nbl/system/IFileArchive.cpp
+++ b/src/nbl/system/IFileArchive.cpp
@@ -13,7 +13,15 @@ IFileArchive::SFileList IFileArchive::listAssets(const path& pathRelativeToArchi
 			auto begin = trimmedList.m_data->begin();
 			auto end = trimmedList.m_data->end();
 			const SFileList::SEntry itemToFind = { pathRelativeToArchive };
-			trimmedList.m_range = { &(*std::lower_bound(begin,end,itemToFind)),&(*std::upper_bound(begin,end,itemToFind)) };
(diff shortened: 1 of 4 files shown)


src/nbl/system/IFileArchive.cpp
```diff
@@ -6,29 +6,44 @@ using namespace nbl;
...
+			//
+			// lower bound
+			// by appending slash and char with code 1, we essentially make sure that all matches will be lexicographically greater, and start with pathRelativeToArchive/
+			auto lower = std::lower_bound(begin,end, SFileList::SEntry{pathRelativeToArchive/"\1"});
+			if (lower!=end)
+			{
+				// upper bound
+				// we cannot do std::upper_bound, because any value longer than pathRelativeToArchive is greater, thus the only matches between std::lower_bound and std::upper_bound would be ones exactly matching pathRelativeToArchive
+				// by appending char with code 1 and taking the lower bound, we can find the first element with prefix greater than pathRelativeToArchive, thus the upper bound
+				auto upper = std::lower_bound(lower,end,SFileList::SEntry{pathRelativeToArchive.string()+"\1"});
```
devshgraphicsprogramming, line 40: I think we messed up, our first lower bound is looking for `path/` but our next should be looking for `path[next char after /]`

right now our upper bound can come out lower than out lower bound

PR #509, opened 2023-06-05
Add ECF_CANNOT_MAXIMIZE to window flags

deprilula28:
## Description
Adds an option to window flags: ECF_CANNOT_MAXIMIZE. When set in Win32, disables the maximize button as below.
plystldemo_rwdi_wNhQ5IrD48

## Testing 
Behaviour with the flag enabled & disabled was tested.


devshgraphicsprogramming:
@deprilula28 can you invert the behaviour, as in call the flag `ECF_MAXIMIZABLE` (so that default behaviour is for maximization to be disabled?

devshgraphicsprogramming:
Another question, now that I locked down the window, can I do something that will cause the swapchain to become invalid?

because before a maximize would imply a resize, I'm just wondering now if moving the window between screens or minimizing it and popping it up again can trigger a similar mess?

deprilula28:
> Another question, now that I locked down the window, can I do something that will cause the swapchain to become invalid?
> 
> because before a maximize would imply a resize, I'm just wondering now if moving the window between screens or minimizing it and popping it up again can trigger a similar mess?

you can still resize it by dragging the edge of the window, as well as minimizing as you said

deprilula28:
> > Another question, now that I locked down the window, can I do something that will cause the swapchain to become invalid?
> > because before a maximize would imply a resize, I'm just wondering now if moving the window between screens or minimizing it and popping it up again can trigger a similar mess?
> 
> you can still resize it by dragging the edge of the window, as well as minimizing as you said

i've patched ECF_RESIZABLE so that it doesn't allow you to resize if it's off

PR #517, opened 2023-07-13
Cuda interop

devshgraphicsprogramming:


include/nbl/video/CCUDADevice.h
```diff
@@ -179,16 +179,51 @@ class CCUDADevice : public core::IReferenceCounted
...
+		{
+			size_t size;
+			CUdeviceptr ptr;
+			CUmemGenericAllocationHandle memory;
+			void* osHandle;
+		};
+
+		core::smart_refctd_ptr<IGPUBuffer> exportGPUBuffer(SSharedCUDAMemory mem, ILogicalDevice* device);
+
+		CUresult importGPUBuffer(IGPUBuffer* buf, SSharedCUDAMemory* outPtr);
```
devshgraphicsprogramming, line 195: shouldn't these be named the other way round?

Basically from the point of view of Nabla, so:
- export = give already created Nabla buffer to other API
- import = take already created external API buffer to Nabla

devshgraphicsprogramming: btw now that we decided to only have "import" we'd only leave this
```cpp
core::smart_refctd_ptr<IGPUBuffer> exportGPUBuffer(const SSharedCUDAMemory& mem, ILogicalDevice* device);
```

include/nbl/video/CCUDADevice.h
```diff
@@ -179,16 +179,51 @@ class CCUDADevice : public core::IReferenceCounted
 		static CUresult acquireAndGetMipmappedArray(GraphicsAPIObjLink<video::IGPUImage>* linksBegin, GraphicsAPIObjLink<video::IGPUImage>* linksEnd, CUstream stream);
 		static CUresult acquireAndGetArray(GraphicsAPIObjLink<video::IGPUImage>* linksBegin, GraphicsAPIObjLink<video::IGPUImage>* linksEnd, uint32_t* arrayIndices, uint32_t* mipLevels, CUstream stream);
 #endif
```
devshgraphicsprogramming, line 181: btw feel free to nuke the ifdef-ed out code after replacement with Vulkan equivalent.

include/nbl/video/CCUDADevice.h
```diff
@@ -179,16 +179,51 @@ class CCUDADevice : public core::IReferenceCounted
...
+		};
+
+		core::smart_refctd_ptr<IGPUBuffer> exportGPUBuffer(SSharedCUDAMemory mem, ILogicalDevice* device);
+
+		CUresult importGPUBuffer(IGPUBuffer* buf, SSharedCUDAMemory* outPtr);
+		CUresult createExportableMemory(size_t size, size_t alignment, SSharedCUDAMemory* outMem);
+		CUresult releaseExportableMemory(SSharedCUDAMemory mem);
 	protected:
+
+		struct SCUDACleaner : video::ICleanup, SSharedCUDAMemory
```
devshgraphicsprogramming, line 200: please use composition over inheritance, there's no reason to inherit from `SSharedCUDAMemoery`

include/nbl/video/CCUDADevice.h
```diff
@@ -179,16 +179,51 @@ class CCUDADevice : public core::IReferenceCounted
...
+			size_t size;
+			CUdeviceptr ptr;
+			CUmemGenericAllocationHandle memory;
+			void* osHandle;
+		};
+
+		core::smart_refctd_ptr<IGPUBuffer> exportGPUBuffer(SSharedCUDAMemory mem, ILogicalDevice* device);
+
+		CUresult importGPUBuffer(IGPUBuffer* buf, SSharedCUDAMemory* outPtr);
+		CUresult createExportableMemory(size_t size, size_t alignment, SSharedCUDAMemory* outMem);
```
devshgraphicsprogramming, line 196: btw our coding convention is to put `out` variables first

include/nbl/video/CCUDADevice.h
```diff
@@ -179,16 +179,74 @@ class CCUDADevice : public core::IReferenceCounted
 		static CUresult acquireAndGetMipmappedArray(GraphicsAPIObjLink<video::IGPUImage>* linksBegin, GraphicsAPIObjLink<video::IGPUImage>* linksEnd, CUstream stream);
 		static CUresult acquireAndGetArray(GraphicsAPIObjLink<video::IGPUImage>* linksBegin, GraphicsAPIObjLink<video::IGPUImage>* linksEnd, uint32_t* arrayIndices, uint32_t* mipLevels, CUstream stream);
 #endif
+		CUdevice getInternalObject() const { return m_handle; }
+		const CCUDAHandler* getHandler() const { return m_handler.get();  }
 
+		struct SSharedCUDAMemory : core::IReferenceCounted
```
devshgraphicsprogramming, line 185: when you inherit from IRferenceCounted you need to make it a `class` and encapsulate all fields (also making as many as possible constant) and **make sure ctor and dtor are protected** (you can declare a friendship with the factory that makes them though)

include/nbl/video/IDeviceMemoryBacked.h
```diff
@@ -13,12 +13,38 @@ namespace nbl::video
...
+            EHT_OPAQUE_WIN32 = 0x00000002,
+            EHT_OPAQUE_WIN32_KMT = 0x00000004,
+            EHT_D3D11_TEXTURE = 0x00000008,
+            EHT_D3D11_TEXTURE_KMT = 0x00000010,
+            EHT_D3D12_HEAP = 0x00000020,
+            EHT_D3D12_RESOURCE = 0x00000040,
+            EHT_DMA_BUF = 0x00000200,
+            EHT_HOST_ALLOCATION = 0x00000080,
+            EHT_HOST_MAPPED_FOREIGN_MEMORY = 0x00000100,
+            EHT_RDMA_ADDRES = 0x00001000,
```
devshgraphicsprogramming, line 45: only leave uncommented the ones you'll implement/add a codepath for

include/nbl/video/IDeviceMemoryBacked.h
```diff
@@ -30,7 +56,11 @@ class IDeviceMemoryBacked : public virtual core::IReferenceCounted
             uint8_t queueFamilyIndexCount = 0u;
             // Thus the destructor will skip the call to `vkDestroy` or `glDelete` on the handle, this is only useful for "imported" objects
             bool skipHandleDestroy = false;
-
+            // Handle Type for external resources
+            core::bitflag<E_EXTERNAL_HANDLE_TYPE> externalHandleType = EHT_NONE;
+            //! Imports the given handle  if externalHandle != nullptr && externalMemoryHandleType != EHT_NONE
+            //! Creates exportable memory if externalHandle == nullptr && externalMemoryHandleType != EHT_NONE
+            void* externalHandle = nullptr;
```
devshgraphicsprogramming, line 63: nice and clean API, me like

include/nbl/video/IDeviceMemoryBacked.h
```diff
@@ -84,7 +117,12 @@ class IDeviceMemoryBacked : public virtual core::IReferenceCounted
         {
             const uint32_t* queueFamilyIndices = nullptr;
         };
-
+        
+        void chainPreDestroyCleanup(std::unique_ptr<ICleanup> first)
+        {
+            ICleanup::chain(m_cachedCreationParams.preDestroyCleanup, std::move(first));
+        }
```
devshgraphicsprogramming, line 124: the cleanups are not meant to be dynamic, all the cleanups you'll ever need are required to be specified in the SCreationParams

devshgraphicsprogramming: the fact you've added this method tells me one thing, you're attempting to use the system to make the Nabla object hold onto something _after it has already been created_, probably the CUDA resource that you've created from it. Esp because you're hooking into the pre-destroy cleanup.

Whereas the intended use is for Nabla object to hold onto some external resoruce it has been created from.

In the case where you have a Nabla object, owned and created by Nabla, being exported, which created an object in some other API that needs to be destroyed before the Nabla object...

The correct way is for the other API object wrapper to hold a `smart_Refctd_ptr` to the Nabla object it was imported from.

include/nbl/video/IDeviceMemoryBacked.h
```diff
@@ -13,12 +13,38 @@ namespace nbl::video
...
     virtual ~ICleanup() = 0;
+
+    std::unique_ptr<ICleanup> next;
+
+    static void chain(std::unique_ptr<ICleanup>& first, std::unique_ptr<ICleanup>&& next)
+    {
+        if (first)
+            return chain(first->next, std::move(next));
+        first = std::move(next);
+    }
```
devshgraphicsprogramming, line 24: don't like it being the default, if you want to chain up cleanups the proper way would be to either:
- have a "metacleanup" that inherits from multiple cleanups (and each of them virtually from ICleanup)
- have this same chain, but in a `CLinkedListCleanup final`

Anyway this is probably not needed for anything due to what I think is the misuse of the API https://github.com/Devsh-Graphics-Programming/Nabla/pull/517/files#r1264588641

include/nbl/video/IDeviceMemoryBacked.h
devshgraphicsprogramming: btw IDeviceMemoryBacked is not inherited from by semaphores or device memory allocations, are you sure all these enums are applicable to buffers and images?

include/nbl/video/ILogicalDevice.h
```diff
@@ -423,6 +423,8 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
         //
         virtual void unmapMemory(IDeviceMemoryAllocation* memory) = 0;
 
+        virtual void* getExternalMemoryHandle(IDeviceMemoryBacked* obj) const = 0;
```
devshgraphicsprogramming, line 426: do we need it though?

We already have `getExternalMemoryHandle` on the `IDeviceMemoryBacked` itself

src/nbl/video/CVulkanBuffer.h
```diff
@@ -39,6 +39,24 @@ class CVulkanBuffer : public IGPUBuffer
...
+
+            auto& ccp = getCachedCreationParams();
+
+            if (ccp.externalHandleType.value)
+            {
+                if (ccp.externalHandle)
+                    return m_cachedExternalHandle = ccp.externalHandle;
+
+                return m_cachedExternalHandle = getOriginDevice()->getExternalMemoryHandle(this);
+            }
```
devshgraphicsprogramming, line 55: will this handle ever change?

if not, then query and cache it in the constructor and always return `m_cachedExternalHandle` (don't even make this method virtual)

src/nbl/video/CVulkanBuffer.cpp
```diff
@@ -8,7 +8,7 @@ namespace nbl::video
 CVulkanBuffer::~CVulkanBuffer()
 {
 	preDestroyStep();
-	if (m_cachedCreationParams.skipHandleDestroy)
+	if (!m_cachedCreationParams.skipHandleDestroy)
```
devshgraphicsprogramming, line 11: what a typo! did I mess up the same thing anywhere else?

src/nbl/video/CCUDAHandler.cpp
```diff
@@ -410,7 +411,7 @@ core::smart_refctd_ptr<CCUDAHandler> CCUDAHandler::create(system::ISystem* syste
 	NVRTC nvrtc = {};
 	#if defined(_NBL_WINDOWS_API_)
 	// Perpetual TODO: any new CUDA releases we need to account for?
-	const char* nvrtc64_versions[] = { "nvrtc64_111","nvrtc64_110","nvrtc64_102","nvrtc64_101","nvrtc64_100","nvrtc64_92","nvrtc64_91","nvrtc64_90","nvrtc64_80","nvrtc64_75","nvrtc64_70",nullptr };
+	const char* nvrtc64_versions[] = { "nvrtc64_120", "nvrtc64_111","nvrtc64_110","nvrtc64_102","nvrtc64_101","nvrtc64_100","nvrtc64_92","nvrtc64_91","nvrtc64_90","nvrtc64_80","nvrtc64_75","nvrtc64_70",nullptr };
 	const char* nvrtc64_suffices[] = {"","_","_0","_1","_2",nullptr};
```
devshgraphicsprogramming, line 415: btw CUDA 11 had 8 minor versions, and 12 has 3 minor versions, so you're probably missing a few in the version table

src/nbl/video/CVulkanLogicalDevice.h
```diff
@@ -826,6 +767,27 @@ class CVulkanLogicalDevice final : public ILogicalDevice
...
+        auto vkHandle = *static_cast<const VkDeviceMemory*>(mem->getNativeHandle());
+        if (!vkHandle)
+            return nullptr;
+        VkMemoryGetWin32HandleInfoKHR getHandleInfo = {
+            .sType = VK_STRUCTURE_TYPE_MEMORY_GET_WIN32_HANDLE_INFO_KHR,
+            .memory = vkHandle,
+            .handleType = VkExternalMemoryHandleTypeFlagBits(obj->getCachedCreationParams().externalHandleType.value),
+        };
+        void* handle = 0;
+        if(VK_SUCCESS == m_devf.vk.vkGetMemoryWin32HandleKHR(m_vkdev, &getHandleInfo, &handle))
```
devshgraphicsprogramming, line 786: please guard Windows specific code with the appropriate `#ifdef NBL_....._PLATFORM_....`

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -9,6 +9,37 @@
 namespace nbl::video
 {
 
+core::smart_refctd_ptr<IGPUSemaphore> CVulkanLogicalDevice::createSemaphore(IGPUSemaphore::SCreationParams&& params)
+{
+    VkImportSemaphoreWin32HandleInfoKHR importInfo = { VK_STRUCTURE_TYPE_IMPORT_SEMAPHORE_WIN32_HANDLE_INFO_KHR };
```
devshgraphicsprogramming, line 14: this is currently an unused variable correct?

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -9,6 +9,37 @@
...
+        return nullptr;
+
+    if (params.externalHandleType.value)
+    {
+        VkSemaphoreGetWin32HandleInfoKHR props = {
+            .sType = VK_STRUCTURE_TYPE_SEMAPHORE_GET_WIN32_HANDLE_INFO_KHR,
+            .semaphore = semaphore,
+            .handleType = static_cast<VkExternalSemaphoreHandleTypeFlagBits>(params.externalHandleType.value),
+        };
+        if (VK_SUCCESS != m_devf.vk.vkGetSemaphoreWin32HandleKHR(m_vkdev, &props, &params.externalHandle))
```
devshgraphicsprogramming, line 33: I'd make sure you follow the same pattern as you do with Buffers/Images, which is not to hotpatch the creation parameters (because we can no longer tell if the thing was created from an import or created for export)

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -86,16 +121,37 @@ IDeviceMemoryAllocator::SMemoryOffset CVulkanLogicalDevice::allocate(const SAllo
...
+                importInfo.handleType = VkExternalMemoryHandleTypeFlagBits(ccParams.externalHandleType.value);
+                importInfo.handle = ccParams.externalHandle;
+                vk_dedicatedInfo.pNext = &importInfo;
+            }
+            else // Exporting
+            {
+                exportInfo.handleTypes = VkExternalMemoryHandleTypeFlags(ccParams.externalHandleType.value);
+                vk_dedicatedInfo.pNext = &exportInfo;
+            }
+        }
```
devshgraphicsprogramming, line 153: the limit about external memory needing to be dedicated is dictated by Vulkan/other APIs not because of how we've structured pre-existing Nabla code, right?

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -785,4 +841,78 @@ bool CVulkanLogicalDevice::getQueryPoolResults(IQueryPool* queryPool, uint32_t f
...
+    vk_createInfo.usage = getVkBufferUsageFlagsFromBufferUsageFlags(creationParams.usage);
+    vk_createInfo.sharingMode = creationParams.isConcurrentSharing() ? VK_SHARING_MODE_CONCURRENT : VK_SHARING_MODE_EXCLUSIVE;
+    vk_createInfo.queueFamilyIndexCount = creationParams.queueFamilyIndexCount;
+    vk_createInfo.pQueueFamilyIndices = creationParams.queueFamilyIndices;
+
+    if (external)
+    {
+        VkPhysicalDeviceExternalBufferInfo bufferInfo = {
+            .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTERNAL_BUFFER_INFO,
+            .usage = vk_createInfo.usage,
```
devshgraphicsprogramming, line 871: whats the deal with a double `usage` field?

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -785,4 +841,78 @@ bool CVulkanLogicalDevice::getQueryPoolResults(IQueryPool* queryPool, uint32_t f
...
+            .handleType = VkExternalMemoryHandleTypeFlagBits(creationParams.externalHandleType.value),
+        };
+
+        VkExternalBufferProperties externalProps = { VK_STRUCTURE_TYPE_EXTERNAL_BUFFER_PROPERTIES };
+        VkPhysicalDevice vk_physicalDevice = static_cast<const CVulkanPhysicalDevice*>(m_physicalDevice)->getInternalObject();
+        vkGetPhysicalDeviceExternalBufferProperties(vk_physicalDevice, &bufferInfo, &externalProps);
```
devshgraphicsprogramming, line 877: dont query at runtime, query during physical device Limit polling and expose as an `[HANDLE_TYPE_COUNT][USAGES]` array in `SPhysicalDeviceLimits` unless handle type are a bitfield and joining them post-facto is not valid

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -785,4 +841,78 @@ bool CVulkanLogicalDevice::getQueryPoolResults(IQueryPool* queryPool, uint32_t f
...
+        VkMemoryRequirements2 vk_memoryRequirements = { VK_STRUCTURE_TYPE_MEMORY_REQUIREMENTS_2 };
+        vk_memoryRequirements.pNext = &vk_dedicatedMemoryRequirements;
+        m_devf.vk.vkGetBufferMemoryRequirements2(m_vkdev, &vk_memoryRequirementsInfo, &vk_memoryRequirements);
+
+        IDeviceMemoryBacked::SDeviceMemoryRequirements bufferMemoryReqs = {};
+        bufferMemoryReqs.size = vk_memoryRequirements.memoryRequirements.size;
+        bufferMemoryReqs.memoryTypeBits = vk_memoryRequirements.memoryRequirements.memoryTypeBits;
+        bufferMemoryReqs.alignmentLog2 = std::log2(vk_memoryRequirements.memoryRequirements.alignment);
+        bufferMemoryReqs.prefersDedicatedAllocation = vk_dedicatedMemoryRequirements.prefersDedicatedAllocation;
+        bufferMemoryReqs.requiresDedicatedAllocation = vk_dedicatedMemoryRequirements.requiresDedicatedAllocation;
```
devshgraphicsprogramming, line 903: override these two with `true` if `external` or assert that they were true, unless your requirements are not true here
https://github.com/Devsh-Graphics-Programming/Nabla/pull/517/files#r1264590615

include/nbl/video/CVulkanImage.h
```diff
@@ -36,6 +36,24 @@ class CVulkanImage : public IGPUImage
...
+
+			auto& ccp = getCachedCreationParams();
+
+			if (ccp.externalHandleType.value)
+			{
+				if (ccp.externalHandle)
+					return m_cachedExternalHandle = ccp.externalHandle;
+
+				return m_cachedExternalHandle = getOriginDevice()->getExternalMemoryHandle(this);
+			}
```
devshgraphicsprogramming, line 52: same as what I commented about the buffer

src/nbl/video/CVulkanLogicalDevice.cpp
atkurtul: Correct

src/nbl/video/CVulkanLogicalDevice.cpp
atkurtul: It does seem unnecessary. No idea why it is needed.

include/nbl/video/ILogicalDevice.h
atkurtul: IDeviceMemoryBacked calls this one and caches the returned handle. This could be removed. Also, I think these handles need to be released by platform specific functions such as CloseHandle on win32 so we might be needing to track them.

src/nbl/video/CVulkanBuffer.cpp
atkurtul: Other dtors were fine

src/nbl/video/CVulkanBuffer.h
atkurtul: Handle will not change, even across processes it's the same.

src/nbl/video/CVulkanLogicalDevice.cpp
atkurtul: Correct

src/nbl/video/CVulkanLogicalDevice.cpp
atkurtul: Yeah, I'm aware this broke the pattern with the external resources. Although atm CUDA can only import semaphores so we don't need to deal with imports on nabla side.

src/nbl/video/CVulkanLogicalDevice.cpp
atkurtul: Just to make it clear, unlike textures it is not required for external buffers to have dedicated allocations. It's just easier for us to make them dedicated.

include/nbl/video/IDeviceMemoryBacked.h
atkurtul: Semaphores have seperate flags. Buffers, images and memory allocations use the same bitflags.

src/nbl/video/CVulkanLogicalDevice.cpp
atkurtul: We would have to query each handle type bit for every combination of BuferUsage.

src/nbl/video/CVulkanLogicalDevice.cpp
devshgraphicsprogramming: If I were to guess, maybe the export usages can be subsets of the actual usages?

include/nbl/asset/IBuffer.h
```diff
@@ -19,31 +19,48 @@ class IBuffer : public core::IBuffer, public IDescriptor
...
+			//EUF_TRANSFORM_FEEDBACK_COUNTER_BUFFER_BIT_EXT = 0x00001000,
+			EUF_CONDITIONAL_RENDERING_BIT_EXT = 0x00000200,
+			EUF_ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_BIT = 0x00080000,
+			EUF_ACCELERATION_STRUCTURE_STORAGE_BIT = 0x00100000,
+			EUF_SHADER_BINDING_TABLE_BIT = 0x00000400,
 			//! synthetic Nabla inventions
 			// whether `IGPUCommandBuffer::updateBuffer` can be used on this buffer
 			EUF_INLINE_UPDATE_VIA_CMDBUF = 0x80000000u,
+
+			EUF_ANY = EUF_TRANSFER_SRC_BIT
```
devshgraphicsprogramming, line 46: rename to ALL, however you'll get huge validation and creation problems if you actually try to use it, because if you don't enable or have acceleration structure/raytracing them it will fail to create the buffer

What do you need it for?

include/nbl/video/CCUDASharedMemory.h
```diff
@@ -0,0 +1,69 @@
...
+        {
+            void* osHandle;
+            int fd;
+        };
+    };
+
+    const SCreationParams& getCreationParams() const { return m_params; }
+
+    core::smart_refctd_ptr<IGPUBuffer> exportAsBuffer(ILogicalDevice* device, core::bitflag<asset::IBuffer::E_USAGE_FLAGS> usage) const;
+    core::smart_refctd_ptr<IGPUImage>  exportAsImage(ILogicalDevice* device, asset::IImage::SCreationParams&& params) const;
```
devshgraphicsprogramming, line 51: what guarantee do you have that the images will be the same as a CUDA image?

wouldn't it be better to export an `IGPUImage` from something we know is a CUDA image (and we also can query all its stuff to fill `IGPUImage::SCreationParams`)?

devshgraphicsprogramming, line 51: Also don't use `asset::IResource` use `IGPUResource` even if its a typedef for the exact same thing

include/nbl/video/CCUDASharedSemaphore.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "cuda.h"
+#include "nvrtc.h"
+#if CUDA_VERSION < 9000
+	#error "Need CUDA 9.0 SDK or higher."
+#endif
+
+// useful includes in the future
+//#include "cudaEGL.h"
+//#include "cudaVDPAU.h"
```
devshgraphicsprogramming, line 17: definitely not anymore!

include/nbl/video/CVulkanImage.h
```diff
@@ -36,6 +36,24 @@ class CVulkanImage : public IGPUImage
...
+
+			auto& ccp = getCachedCreationParams();
+
+			if (ccp.externalHandleTypes.value)
+			{
+				if (ccp.externalHandle)
+					return m_cachedExternalHandle = ccp.externalHandle;
+
+				return m_cachedExternalHandle = getOriginDevice()->getExternalHandle(this);
+			}
```
devshgraphicsprogramming, line 52: I think you can more this to `IDeviceMemoryBacked` and query it at `bind()` time

include/nbl/video/IDeviceMemoryBacked.h
```diff
@@ -19,6 +19,59 @@ struct NBL_API2 ICleanup
...
+            EHT_D3D11_TEXTURE_KMT = 0x00000010,
+            EHT_D3D12_HEAP = 0x00000020,
+            EHT_D3D12_RESOURCE = 0x00000040,
+            EHT_HOST_ALLOCATION_BIT = 0x00000080,
+            EHT_HOST_MAPPED_FOREIGN_MEMORY = 0x00000100,
+            EHT_DMA_BUF = 0x00000200,
+            EHT_ANDROID_HARDWARE_BUFFER = 0x00000400,
+            EHT_ZIRCON_VMO = 0x00000800,
+            EHT_RDMA_ADDRESS = 0x00001000,
+            EHT_LAST = EHT_RDMA_ADDRESS,
```
devshgraphicsprogramming, line 39: don't expose enums we don't need/support

include/nbl/video/IGPUImage.h
```diff
@@ -29,6 +29,7 @@ class IGPUImage : public asset::IImage, public IDeviceMemoryBacked, public IBack
 		struct SCreationParams : asset::IImage::SCreationParams, IDeviceMemoryBacked::SCreationParams
 		{
 			E_TILING tiling = ET_OPTIMAL;
+			//! Ignored if external
```
devshgraphicsprogramming, line 32: even if creates EXPORTABLE (not imports) ?

include/nbl/video/ILogicalDevice.h
```diff
@@ -423,6 +423,8 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
         //
         virtual void unmapMemory(IDeviceMemoryAllocation* memory) = 0;
 
+        virtual void* getExternalHandle(IDeviceMemoryBacked* obj) const = 0;
```
devshgraphicsprogramming, line 426: erm, should only be a method of `IDeviceMemoryBacked`

include/nbl/video/IPhysicalDevice.h
```diff
@@ -606,7 +606,23 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
             return createLogicalDevice_impl(std::move(params));
         }
 
+
+        //! Not thread safe
```
devshgraphicsprogramming, line 610: then pair a mutex with the cache!

include/nbl/video/CCUDASharedMemory.h
devshgraphicsprogramming: ok so my source of confusion was the naming, as the function actually exports a memory allocation and creates an image over it.

src/nbl/video/CVulkanLogicalDevice.cpp
devshgraphicsprogramming: with refcounted osHandles we will no longer need to know/care if object was created from import or to export

src/nbl/video/CVulkanLogicalDevice.cpp
devshgraphicsprogramming: yea that'd be dumb, you're right

PR #534, opened 2023-09-07
glm compat headers

devshgraphicsprogramming:

PR #542, opened 2023-09-13
Matrix dimension conversion

atkurtul:
## Description

Addressed comments

https://github.com/Devsh-Graphics-Programming/Nabla/pull/534#discussion_r1318800332
https://github.com/Devsh-Graphics-Programming/Nabla/pull/534#discussion_r1318800332
https://github.com/Devsh-Graphics-Programming/Nabla/pull/534#discussion_r1319056609

## TODO list:

Get started with type traits for HLSL


include/nbl/builtin/hlsl/cpp_compat/matrix.h
```diff
@@ -22,42 +22,35 @@ struct matrix final : private glm::mat<N,M,T>
     using Base::operator=;
     using Base::operator[];
 
-    template<uint16_t X, uint16_t Y> matrix(matrix<T, X, Y> const&) = delete;
+    template<uint16_t X, uint16_t Y, std::enable_if<!(X == N && Y == M) && X <= N && Y <= M>>
+    matrix(matrix<T, X, Y> const& m) : Base(m)
```
devshgraphicsprogramming, line 26: explicit

devshgraphicsprogramming:
Also can you make sure the extension on the HLSL headers is `.hlsl` not `.h` ?

and move `cpp_compat/cpp_compat.h` to just `cpp_compat.hlsl`

include/nbl/builtin/hlsl/cpp_compat/matrix.h
```diff
@@ -22,42 +22,35 @@ struct matrix final : private glm::mat<N,M,T>
...
+    {
+    }
 
     matrix(matrix const&) = default;
     explicit matrix(Base const& base) : Base(base) {}
```
devshgraphicsprogramming, line 31: having both this and `using Base::Base` makes no sense

Btw won't `using Base::Base` and `Base::operator=` give you public constructors and assignment operators that accept `Base` and not `matrix`? (A definite DO NOT WANT)

include/nbl/builtin/hlsl/cpp_compat/matrix.h
devshgraphicsprogramming: you don't want implicit casts of `float4x4` to `float3x3`

include/nbl/builtin/hlsl/cpp_compat/matrix.h
atkurtul: Well if we dont use Base::Base we would have to write out every single constructor manually. And by declaring the matrix constructor that takes Base we force it to be explicit. However, instead of Base::operator= we could default the assignment operator probably.

devshgraphicsprogramming: ok do that, make sure we can never implicitly convert `matrix` to `Base`, neither when assigning or passing as function arg or in some operator.

include/nbl/builtin/hlsl/cpp_compat/matrix.hlsl
```diff
@@ -23,7 +23,7 @@ struct matrix final : private glm::mat<N,M,T>
     using Base::operator[];
```
devshgraphicsprogramming, line 23: explain to me how this doesn't produce an
```cpp
glm::mat<N,M,T> operator=(const glm::mat<N,M,T>&);
```

AnastaZIuk:
approved

PR #548, opened 2023-09-20
Update tcpp version to include fixes from the original repository.

deprilula28:
## Testing 

Checked that behaviour matched what it was previously, and that things were being invoked.

## TODO list:

- [ ] run tcpp unit tests
- [ ] verify again that shaders compile

### DIFF
--- a/src/nbl/asset/utils/CHLSLCompiler.cpp
+++ b/src/nbl/asset/utils/CHLSLCompiler.cpp
@@ -61,7 +61,7 @@ CHLSLCompiler::~CHLSLCompiler()
     delete m_dxcCompilerTypes;
 }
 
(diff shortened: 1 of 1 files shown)


src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -112,6 +112,7 @@ static tcpp::IInputStream* getInputStreamInclude(
         re.append(requestedSource);
         re.append(" not found");
         includeStack.push_back(includeStack.back());
+        return tcpp::TInputStreamUniquePtr(new tcpp::StringInputStream(re));
         return new tcpp::StringInputStream(re);
```
atkurtul, line 116: Halts compilation

PR #546, opened 2023-09-19
hlsl type_traits

atkurtul:
## Testing 
Tested on godbolt
https://godbolt.org/z/cerxYW4vP

## TODO list:

- meta_math
- std-compliant aliases for comparison operators
- `is_pot` equivalents for some of the old GLSL macros
- macro unification
- `std::numerical_limits`
- `NBL_ASSERT(COND)`
- `#define static_assert(...) _Static_assert(__VA_ARGS__)`

PR #519, opened 2023-07-15
HLSL migration of all shaders related to example 48.ArithmeticUnitTest

kpentaris:
## Description
Migration to HLSL of all shaders related to subgroup and workgroup reduction, inclusive and exclusive scan operations.

## Testing
Used example 48 ArithmeticUnitTest

## TODO list:
Every comment that starts with // REVIEW should be checked and the related code reviewed or the question answered.

PR #556, opened 2023-09-28
First draft of Blit GLSL to HLSL migration.

theoreticalphysicsftw:
First draft of the CComputeBlit port from GLSL to HLSL.

## Description
This commit implicitly assumes both  https://github.com/Devsh-Graphics-Programming/Nabla/pull/546 and https://github.com/Devsh-Graphics-Programming/Nabla/pull/519 have been finalized. Furthermore it assumes `-enable-16bit-types` is passed to DXC in `nbl/asset/utils/CHLSLCompiler.cpp`  so that those PRs can compile successfully. It also has few other assumptions about how the final versions of these PRs are going to look like which are commented in the code itself.

## Testing 

This was tested on a working branch with both of those PRs merged + few extra edits to get some of the not yet implemented functionality. Since `nbl/builtin/hlsl/workgroup/*` seems to have compilation issues and the normalization stage of `CComputeBlit` requires `inclusive_scan` I need to first get this working before figuring out the issues with that code.

On the other hand the main blit shader causes DXC to ICE. I will need some more time to figure out why.

## TODO list:

I have to figure out why the ICE happens, then compile/debug and make sure that the FilterTest example works when the
`UseGLSL` template parameter of `CComputeBlit` is set to `false`.


devshgraphicsprogramming:
Good first try

devshgraphicsprogramming:
>  Furthermore it assumes -enable-16bit-types is passed to DXC in nbl/asset/utils/CHLSLCompiler.cpp so that those PRs can compile successfully.

We'll try to get #546 merged tomorrow, btw `-enable-16bit-types` and friends are already on `master`

> On the other hand the main blit shader causes DXC to ICE. I will need some more time to figure out why.

if you follow this and make me a godbolt, then I can investigate myself.
https://www.youtube.com/watch?v=sup-_MR686o

include/nbl/builtin/hlsl/blit/parameters.hlsl
```diff
@@ -0,0 +1,69 @@
...
+namespace nbl
+{
+namespace hlsl
+{
+namespace blit
+{
+
+struct parameters_t
+{
+	uint32_t3 dims; // input dimensions in lower 16 bits, output dimensions in higher 16 bits
```
devshgraphicsprogramming, line 17: you can get rid of this hack, as we always compile with `-enable-16bit-types` and `-fvk-use-scalar-layout` on master

include/nbl/builtin/hlsl/blit/parameters.hlsl
```diff
@@ -0,0 +1,69 @@
...
+	//! Offset into the shared memory array which tells us from where the second buffer of shared memory begins
+	//! Given by max(memory_for_preload_region, memory_for_result_of_y_pass)
+	uint32_t secondScratchOffset;
+	uint32_t3 iterationRegionXPrefixProducts;
+	float referenceAlpha;
+	float3 fScale;
+	uint32_t inPixelCount;
+	float3 negativeSupport;
+	uint32_t outPixelCount;
+	uint32_t3 windowDimPhaseCount; // windowDim in lower 16 bits, phaseCount in higher 16 bits
```
devshgraphicsprogramming, line 27: same stuff as https://github.com/Devsh-Graphics-Programming/Nabla/pull/556/files#r1344290771

include/nbl/builtin/hlsl/blit/parameters.hlsl
```diff
@@ -0,0 +1,69 @@
...
+{
+	uint32_t3 dims; // input dimensions in lower 16 bits, output dimensions in higher 16 bits
+	//! Offset into the shared memory array which tells us from where the second buffer of shared memory begins
+	//! Given by max(memory_for_preload_region, memory_for_result_of_y_pass)
+	uint32_t secondScratchOffset;
+	uint32_t3 iterationRegionXPrefixProducts;
+	float referenceAlpha;
+	float3 fScale;
+	uint32_t inPixelCount;
+	float3 negativeSupport;
```
devshgraphicsprogramming, line 25: slight nitpick, use `float32_t` instead of plain `float` going foward

include/nbl/builtin/hlsl/blit/parameters.hlsl
```diff
@@ -0,0 +1,69 @@
...
+	uint32_t3 getWindowDimensions()
+	{
+		NBL_CONSTEXPR uint32_t mask = ((1u << 16) - 1);
+		return windowDimPhaseCount & mask;
+	}
+
+	uint32_t3 getPhaseCount()
+	{
+		return windowDimPhaseCount >> 16;
+	}
```
devshgraphicsprogramming, line 55: these getters can disappear thanks to https://github.com/Devsh-Graphics-Programming/Nabla/pull/556/files#r1344290771

include/nbl/builtin/hlsl/blit/parameters.hlsl
```diff
@@ -0,0 +1,69 @@
...
+	uint32_t inPixelCount;
+	float3 negativeSupport;
+	uint32_t outPixelCount;
+	uint32_t3 windowDimPhaseCount; // windowDim in lower 16 bits, phaseCount in higher 16 bits
+	uint32_t kernelWeightsOffsetY;
+	uint32_t3 iterationRegionYPrefixProducts;
+	uint32_t kernelWeightsOffsetZ;
+	uint32_t3 iterationRegionZPrefixProducts;
+	uint32_t outputTexelsPerWGZ;
+	uint32_t3 preloadRegion;
```
devshgraphicsprogramming, line 33: check the logic if these can ever be greater than `uint16_t` (hint: max image size is 2^15 on all platforms for all image types), if no then make these into 16bit integers

include/nbl/video/utilities/CComputeBlit.h
```diff
@@ -4,6 +4,7 @@
 
 namespace nbl::video
 {
+template <bool UseGLSL = true>
```
devshgraphicsprogramming, line 7: don't modify the existing class, rename this one `CComputeBlitGLSL` and copy all the code to `CComputeBlit`

this is because we'll eventually delete the GLSL version and only keep the HLSL one

devshgraphicsprogramming: I'm telling you to do this because in the HLSL version we can get rid of silly things like `#include "nbl/builtin/glsl/blit/parameters.glsl"` in the middle of this class, and just include it at the top so parameters are `nbl::hlsl::blit::parameters_t` and not `nbl::video::CComputeBlit<>::parameters_t` in C++

plus a whole bunchof other braindead patterns GLSL forced us into

devshgraphicsprogramming: for example the 
```cpp
template <typename BlitUtilities>
	inline void buildParameters
```
method can move to become a `static parameters_t create` pseudo-constructor defined directly in the HLSL

include/nbl/builtin/hlsl/multi_dimensional_array_addressing/multi_dimensional_array_addressing.hlsl
```diff
@@ -0,0 +1,42 @@
...
+#ifndef _NBL_NBL_HLSL_MULTI_DIMENSIONAL_ARRAY_ADDRESSING_INCLUDED_
+#define _NBL_NBL_HLSL_MULTI_DIMENSIONAL_ARRAY_ADDRESSING_INCLUDED_
+
+#include <nbl/builtin/hlsl/cpp_compat/cpp_compat.hlsl>
+
+namespace nbl
+{
+namespace hlsl
+{
+namespace multi_dimensional_array_addressing
```
devshgraphicsprogramming, line 13: ok while we're at it I'd rename the namespace to `nd_array_addressing`

include/nbl/builtin/hlsl/blit/formats_encode.hlsl
```diff
@@ -0,0 +1,45 @@
...
+namespace nbl
+{
+namespace hlsl
+{
+namespace blit
+{
+
+template <E_FORMAT format>
+struct encode_output { using type = float4; };
+template<> struct encode_output<EF_R32G32_SFLOAT> { using type = uint32_t4; };
```
devshgraphicsprogramming, line 21: > VK_FORMAT_R32G32_UINT, VK_FORMAT_R32G32_SINT, and VK_FORMAT_R32G32_SFLOAT from StorageImageExtendedFormats SPIR-V capability, are already covered by core Vulkan [mandatory format support](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#formats-mandatory-features-32bit).

include/nbl/builtin/hlsl/blit/formats_encode.hlsl
```diff
@@ -0,0 +1,45 @@
...
+template<> struct encode_output<EF_B10G11R11_UFLOAT_PACK32> { using type = uint32_t4; };
```
devshgraphicsprogramming, line 22: I think this entire header and `encode` functions can be nuked, because both formats you have special paths for here MUST be supported as storage images if `shaderStorageImageExtendedFormats` feature is enabled (which I'll soon require and you can force-enable)

devshgraphicsprogramming: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#formats-mandatory-features-64bit

include/nbl/builtin/hlsl/blit/descriptors.hlsl
```diff
@@ -0,0 +1,30 @@
...
+typedef type_traits::integral_constant<uint32_t, 2> StatisticsBinding;
+
+typedef type_traits::integral_constant<uint32_t, 1> KernelWeightsDescriptorSet;
+typedef type_traits::integral_constant<uint32_t, 0> KernelWeightsBinding;
+
+}
+}
+}
+
+#endif
```
devshgraphicsprogramming, line 30: headers called `descriptors` which used to declare default descriptors and binding numbers are no longer needed and are to be deleted

include/nbl/builtin/hlsl/blit/default_compute_alpha_test.comp
```diff
@@ -0,0 +1,25 @@
...
+nbl::hlsl::blit::parameters_t params;
+
+
+[numthreads(nbl::hlsl::blit::WorkgroupSizeX::value, nbl::hlsl::blit::WorkgroupSizeY::value, nbl::hlsl::blit::WorkgroupSizeZ::value)]
+void main(uint32_t3 dispatchThreadID : SV_DispatchThreadID, uint32_t3 groupID : SV_GroupID)
+{
+
+	nbl::hlsl::blit::alpha_test<nbl::hlsl::blit::BlitDimCount::value>(statistics, inTexture, params, dispatchThreadID, groupID);
+}
+
```
devshgraphicsprogramming, line 25: we no longer require this header, everything is so simple and overridable now, that the `CComputeBlit` can "synthesise" the entry point and compute shader itself.

include/nbl/builtin/hlsl/blit/default_compute_blit.comp
```diff
@@ -0,0 +1,48 @@
...
+		nbl::hlsl::blit::SoftwareEncodeFormat::value>(
+			inTexture,
+			outTexture,
+			kernelWeights,
+			statistics,
+			sharedAccessor,
+			params,
+			groupID,
+			groupIndex);
+}
```
devshgraphicsprogramming, line 48: https://github.com/Devsh-Graphics-Programming/Nabla/pull/556/files#r1344341741

include/nbl/builtin/hlsl/blit/default_compute_common.comp
```diff
@@ -0,0 +1,34 @@
...
+typedef type_traits::integral_constant<uint32_t, _NBL_WORKGROUP_SIZE_X_> WorkgroupSizeX;
+typedef type_traits::integral_constant<uint32_t, _NBL_WORKGROUP_SIZE_Y_> WorkgroupSizeY;
+typedef type_traits::integral_constant<uint32_t, _NBL_WORKGROUP_SIZE_Z_> WorkgroupSizeZ;
+typedef type_traits::integral_constant<uint32_t, WorkgroupSizeX::value * WorkgroupSizeY::value * WorkgroupSizeZ::value> WorkgroupSize;
+typedef type_traits::integral_constant<uint32_t, _NBL_BLIT_DIM_COUNT_> BlitDimCount;
+typedef type_traits::integral_constant<uint32_t, _NBL_BLIT_ALPHA_BIN_COUNT_> AlphaBinCount;
+typedef type_traits::integral_constant<E_FORMAT, (E_FORMAT)_NBL_BLIT_SOFTWARE_ENCODE_FORMAT_> SoftwareEncodeFormat;
+
+typedef dim_to_in_texture<BlitDimCount::value>::type InTextureType;
+typedef dim_and_software_encode_to_out_texture<BlitDimCount::value, SoftwareEncodeFormat::value != E_FORMAT::EF_UNKNOWN>::type OutTextureType;
```
devshgraphicsprogramming, line 28: such globals or typedefs are unportable, the correct way to resolve this would be a 

```cpp
namespace blit
{

template<
   uint16_t _WorkgroupSizeX,
   uint16_t _WorkgroupSizeY,
   uint16_t _WorkgroupSizeZ,
   uint16_t _BlitDimCount,
   uint16_t _AlphaBinCount,
> struct consteval_parameters_t
{
    NBL_CONSTEXPR_STATIC_INLINE  uint32_t WorkgroupSizeX;
    ....
};

}
```

devshgraphicsprogramming: P.S. thanks to extendedShaderImageStorageFormats feature you no longer need to care about the software encode format

include/nbl/builtin/hlsl/blit/common.hlsl
```diff
@@ -0,0 +1,147 @@
...
+{
+	using type = RWTexture2DArray<float4>;
+};
+
+
+template <>
+struct dim_and_software_encode_to_out_texture<3, false>
+{
+	using type = RWTexture3D<float4>;
+};
```
devshgraphicsprogramming, line 78: thanks to shaderStorageImageExtendedFormats you only need to care about `float4`, so you can use a simpler way to do it just as you did with `dim_to_texture`

include/nbl/builtin/hlsl/blit/common.hlsl
```diff
@@ -0,0 +1,147 @@
...
+struct dim_to_in_texture<1> { using type = Texture1DArray<float4>; };
+
+template <>
+struct dim_to_in_texture<2> { using type = Texture2DArray<float4>; };
+
+template <>
+struct dim_to_in_texture<3> { using type = Texture3D<float4>; };
+
+
+template<uint32_t Dimension, bool RequiresSoftwareEncode>
```
devshgraphicsprogramming, line 37: hide this in impl namespace

include/nbl/builtin/hlsl/blit/common.hlsl
```diff
@@ -0,0 +1,147 @@
...
+{
+namespace blit
+{
+
+template <uint32_t AlphaBinCount>
+struct AlphaStatistics
+{
+	uint32_t passedPixelCount;
+	uint32_t histogram [AlphaBinCount];
+};
```
devshgraphicsprogramming, line 22: you simply cannot define this, it needs to be provided by the user because the values need to be `groupshared` and either global or `static` which in turn makes them consume shared memory and reduce occupancy for no reason if you don't use them.

This is why Pentakon's PR #519  requires them to be "provided from the outside" and there's a set "interface/concept" these shared acessors need to conform to, like providing a `uint32_t` set/get plus workgroup execution barriers.

devshgraphicsprogramming: P.S. separate `REF` accessor per member, so in this case, one for the histogram and one for the passed pixel count

include/nbl/builtin/hlsl/blit/common.hlsl
```diff
@@ -0,0 +1,147 @@
...
+
+template <typename T, uint32_t N>
+struct SharedAccessor
+{
+	T data[N];
+
+	void set(uint32_t idx, NBL_CONST_REF_ARG(T) val) { data[idx] = val; }
+	T get(uint32_t idx) { return data[idx]; }
+	uint32_t size() { return N; }
+};
```
devshgraphicsprogramming, line 140: same stuff as https://github.com/Devsh-Graphics-Programming/Nabla/pull/556/files#r1344392263

btw is it `T` because the data can be vector typed?

include/nbl/builtin/hlsl/blit/normalization.hlsl
```diff
@@ -0,0 +1,108 @@
...
+{
+
+
+//! WARNING: ONLY WORKS FOR `dividendMsb<=2^23` DUE TO FP32 ABUSE !!!
+uint32_t integerDivide_64_32_32(in uint32_t dividendMsb, in uint32_t dividendLsb, in uint32_t divisor)
+{
+	const float msbRatio = float(dividendMsb) / float(divisor);
+	const uint32_t quotient = uint32_t((msbRatio * (~0u)) + msbRatio) + dividendLsb / divisor;
+	return quotient;
+}
```
devshgraphicsprogramming, line 25: move to `math`, call it `fastDivide_int23_t`

include/nbl/builtin/hlsl/blit/normalization.hlsl
```diff
@@ -0,0 +1,108 @@
...
+
+struct OpAdd
+{
+	uint operator()(uint v1, uint v2) { return v1 + v2; }
+};
```
devshgraphicsprogramming, line 30: we have this in `binops` , lets wait on the normalization review till #519 gets merged

devshgraphicsprogramming:
\execute order 0x45ull

include/nbl/builtin/hlsl/blit/blit.hlsl
```diff
@@ -0,0 +1,174 @@
...
+
+	const uint32_t3 outputTexelsPerWG = params.getOutputTexelsPerWG();
+
+
+	const float3 scale = params.fScale;
+	const float3 halfScale = scale * float3(0.5f, 0.5f, 0.5f);
+
+	const uint32_t3 minOutputPixel = groupID * outputTexelsPerWG;
+	const float3 minOutputPixelCenterOfWG = float3(minOutputPixel)*scale + halfScale;
+	const int32_t3 regionStartCoord = getMinKernelWindowCoord(minOutputPixelCenterOfWG, params.negativeSupport); // this can be negative, in which case HW sampler takes care of wrapping for us
```
devshgraphicsprogramming, line 63: I'd wrap these in a `derived_parameters_t` struct and compute them inside a `static derived_parameters_t create(NBL_REF_ARG(parameters_t) prime)`

devshgraphicsprogramming: this will also help you get rid of the free floating `getMinKernelWindowCoord` function

include/nbl/builtin/hlsl/blit/blit.hlsl
```diff
@@ -0,0 +1,174 @@
...
+namespace blit
+{
+
+// p in input space
+int32_t3 getMinKernelWindowCoord(NBL_CONST_REF_ARG(int32_t3) p, NBL_CONST_REF_ARG(float3) minSupport)
+{
+	return int32_t3(ceil(p - float3(0.5f, 0.5f, 0.5f) + minSupport));
+}
+
+template <
```
devshgraphicsprogramming, line 27: 1. its not nice to have an identifier that shadows a namespace name, so lets name this `compute_blit`

2. something to improve usability and style, change the `blit` from a function to a `struct` with a `template<> execute(...)` then you can spec certain template parameters and arguments on the `struct` and its `static compute_blit create(...)` methods while others on the `execute(...)` like the accessors used

include/nbl/builtin/hlsl/blit/blit.hlsl
```diff
@@ -0,0 +1,174 @@
...
+namespace hlsl
+{
+namespace blit
+{
+
+// p in input space
+int32_t3 getMinKernelWindowCoord(NBL_CONST_REF_ARG(int32_t3) p, NBL_CONST_REF_ARG(float3) minSupport)
+{
+	return int32_t3(ceil(p - float3(0.5f, 0.5f, 0.5f) + minSupport));
+}
```
devshgraphicsprogramming, line 25: do we use this function anywhere else, and if no why not fold it into the function below

include/nbl/builtin/hlsl/blit/blit.hlsl
```diff
@@ -0,0 +1,174 @@
...
+	const float3 scale = params.fScale;
+	const float3 halfScale = scale * float3(0.5f, 0.5f, 0.5f);
+
+	const uint32_t3 minOutputPixel = groupID * outputTexelsPerWG;
+	const float3 minOutputPixelCenterOfWG = float3(minOutputPixel)*scale + halfScale;
+	const int32_t3 regionStartCoord = getMinKernelWindowCoord(minOutputPixelCenterOfWG, params.negativeSupport); // this can be negative, in which case HW sampler takes care of wrapping for us
+
+	const uint32_t3 preloadRegion = params.preloadRegion;
+
+	const uint32_t virtualInvocations = preloadRegion.x * preloadRegion.y * preloadRegion.z;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: stuff below would go into `execute` and I'll review the rest later

include/nbl/builtin/hlsl/blit/blit.hlsl
devshgraphicsprogramming: btw again its a bit weird that theres a `blit.hlsl` inside a `blit` folder in a `blit` namespace, so change the path to `nbl/builtin/hlsl/blit/compute_blit.hlsl` as well

include/nbl/builtin/hlsl/blit/alpha_test.hlsl
```diff
@@ -0,0 +1,41 @@
...
+namespace nbl
+{
+namespace hlsl
+{
+namespace blit
+{
+
+template <uint32_t BlitDimCount, typename StatisticsBuffer, typename InTexture>
+inline void alpha_test(
+	NBL_REF_ARG(StatisticsBuffer) statistics,
```
devshgraphicsprogramming, line 20: take an accessor instead of a descriptor resource, this will allow for more flexibility

devshgraphicsprogramming: same for the input texture, use an accessor

include/nbl/builtin/hlsl/blit/alpha_test.hlsl
```diff
@@ -0,0 +1,41 @@
...
+	NBL_CONST_REF_ARG(InTexture) inTexture,
+	NBL_CONST_REF_ARG(parameters_t) params,
+	NBL_CONST_REF_ARG(uint32_t3) dispatchThreadID,
+	NBL_CONST_REF_ARG(uint32_t3) groupID)
```
devshgraphicsprogramming, line 24: I'd stick to GLSL terminology and `globalInvocationID` instead of `dispatchThreadID`

include/nbl/builtin/hlsl/blit/alpha_test.hlsl
```diff
@@ -0,0 +1,41 @@
...
+{
+	const uint32_t3 inDim = params.getInputImageDimensions();
+
+	if (all(dispatchThreadID < inDim))
+	{
+		const float alpha = getData<BlitDimCount>(inTexture, dispatchThreadID, groupID.z).a;
+		if (alpha > params.referenceAlpha)
+			InterlockedAdd(statistics[groupID.z].passedPixelCount, uint32_t(1));
```
devshgraphicsprogramming, line 32: with an accessor we'd probably want a `statsAccessor.vote(const uint16_t layer, const uint16_t ballot)`

include/nbl/builtin/hlsl/blit/alpha_test.hlsl
```diff
@@ -0,0 +1,41 @@
...
+namespace hlsl
+{
+namespace blit
+{
+
+template <uint32_t BlitDimCount, typename StatisticsBuffer, typename InTexture>
+inline void alpha_test(
+	NBL_REF_ARG(StatisticsBuffer) statistics,
+	NBL_CONST_REF_ARG(InTexture) inTexture,
+	NBL_CONST_REF_ARG(parameters_t) params,
```
devshgraphicsprogramming, line 22: btw you only use `params` for inputImageDims and reference alpha, maybe just pass them normally

include/nbl/builtin/hlsl/blit/alpha_test.hlsl
devshgraphicsprogramming: also you can make them `uint16_t3` because we're processing textures/images

devshgraphicsprogramming:
Time to merge `master` to this, `type_traits` just landed with juicy things like `decltype`

include/nbl/builtin/hlsl/blit/common.hlsl
devshgraphicsprogramming: Updated the comment, I forgot this was never in shared memory.

theoreticalphysicsftw:
The correctness issues are fixed now and all the filter tests pass like in the GLSL version.

include/nbl/builtin/hlsl/blit/temp.hlsl
```diff
@@ -12,8 +12,10 @@ namespace hlsl
 
 namespace spirv
 {
+// DXC fails with uint32_t2 result, so we have to use a struct.
+struct umul_result_t { uint32_t lsb; uint32_t msb; };
 [[vk::ext_instruction(/* OpUMulExtended */ 151)]]
-uint32_t2 umulExtended(uint32_t v0, uint32_t v1);
+umul_result_t umulExtended(uint32_t v0, uint32_t v1);
```
devshgraphicsprogramming: doesn't HLSL have a builtin function for this?

theoreticalphysicsftw: I also thought that it has because DXBC has such instruction from long time ago, but it seems they never exposed it to HLSL:
https://github.com/microsoft/DirectXShaderCompiler/issues/2821

devshgraphicsprogramming: hmm, btw we now require `64bit` integers, so is that an issue any more?

theoreticalphysicsftw: For this particular use case it's not a problem since the operation can safely be replaced by normal multiplication of two `uint64_t`s. I can use 64-bit ints to avoid having to do this, I left this because I wanted to be as close to the GLSL code as possible. It's only a problem if it's in performance critical code where the difference between `OpIMul` on 64 bit integers and `OpUMulExtended` might matter or if one actually needs the msb of a 64 bit multiplication. In this case this only one multiplication is tiny part of the computational cost of the whole shader...

devshgraphicsprogramming: yes, it was a hack to not require 64bit integer support, but after upgrading our minimum to Vulkan 1.3 and supported devices (those that still get drivers) + requiring BDA, we've basically had to require 64bit int support too

devshgraphicsprogramming:
Ok turns out that due to some error in my NV driver, some of the `uint16_t` usages have to be replaced by `uint32_t`

PR #563, opened 2023-10-04
numeric_limits

devshgraphicsprogramming:
## Description

Implement `std::numeric_limits` in HLSL

## TODO list:

- [ ] merge `master` into this after `type_traits` is in


include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
...
+#include <type_traits.hlsl>
+
+using half = double;
+
+// C++ headers
+#ifndef __HLSL_VERSION
+#include <limits>
+#endif
+
+namespace nbl::hlsl::numeric_limits
```
devshgraphicsprogramming, line 16: no special namespace, just `nbl::hlsl`, `numeric_limits` is a struct in std

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
+// Copyright (C) 2022 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+#ifndef _NBL_BUILTIN_HLSL_NUMERIC_LIMITS_INCLUDED_
+#define _NBL_BUILTIN_HLSL_NUMERIC_LIMITS_INCLUDED_
+
+#include <type_traits.hlsl>
+
+using half = double;
```
devshgraphicsprogramming, line 9: full `nbl/builtin/hlsl` always needed

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
+// Copyright (C) 2022 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+#ifndef _NBL_BUILTIN_HLSL_NUMERIC_LIMITS_INCLUDED_
+#define _NBL_BUILTIN_HLSL_NUMERIC_LIMITS_INCLUDED_
+
+#include <type_traits.hlsl>
```
devshgraphicsprogramming, line 7: "full" path with `nbl/builtin/hlsl` needed otherwise the embeddings might not work nicely

include/nbl/builtin/hlsl/numeric_limits.hlsl
devshgraphicsprogramming, line 9: this needs to go into the `ifndef` portion of `cpp_compat` or similar, and use `float16_t` instead of `half`

Then we either use `std::float16_t` from `<stdfloat>` if @AnastaZIuk can bump Nabla to C++23 or you need to switch between compilers:
- Clang & GCC `_Float16`
- MSVC add a git mirror of https://half.sourceforge.net/ as a submodule

Or alterantively take a dive inside our OpenEXR submodule, update it to latest, see if the EXR/image tests still run
https://github.com/AcademySoftwareFoundation/openexr/blob/159e57120dae5e587454e0d4bd916c792831a8ce/website/PortingGuide.rst#L7

You'll need help from @AnastaZIuk to upgrade the CMake build system to add the spliced out `IlmBase` submodule to NAbla's 3rdparty

Then there's `IlmBase/Half/half.h` inside which looks more or less C++23 compliant and even has a std::numeric_limits specialization for C++

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
+// Copyright (C) 2022 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+#ifndef _NBL_BUILTIN_HLSL_NUMERIC_LIMITS_INCLUDED_
+#define _NBL_BUILTIN_HLSL_NUMERIC_LIMITS_INCLUDED_
```
devshgraphicsprogramming, line 5: we should call it `limits.hlsl` probably

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
...
+    round_to_nearest          = 1,
+    round_toward_infinity     = 2,
+    round_toward_neg_infinity = 3
+};
+
+template<class T> struct num_base;
+
+template<class T> struct num_traits
+{
+    static const T MIN        = nbl::hlsl::type_traits::is_signed<T>::value ? ((T(1)<<num_base<T>::digits)-T(1)) : T(0);
```
devshgraphicsprogramming, line 39: you're inside `nbl::hlsl` already, dtop the extra stuff

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
...
+
+};
+
+
+using limits = std::numeric_limits<int>;
+#else
+
+
+
+using limits = std::numeric_limits<float>;
```
devshgraphicsprogramming, line 197: I think you should do 
```cpp
template<typename T>
struct numeric_limits : std::numeric_limits<T> {};
```
instead of an alias.

In case we ever want to use the "trait forwarder" to "easily" implement the traits for `vec` and `matrix`.

Not entirely convinced its a good idea, but at least it leaves the door open

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
...
+
+template<class T>
+struct num_base
+{
+    static const int size = sizeof(T);
+
+    static const int F16 = size / 2;
+    static const int F32 = size / 4;
+    static const int F64 = size / 8;
+
```
devshgraphicsprogramming, line 77: please use `NBL_CONSTEXPR` instead of  `static const`, I don't want to spend time fixing up C++03 to C++11 when DXC finally gets its stuff together

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
...
+    static T min() { return num_definitions<T>::MIN; }
+    static T max() { return num_definitions<T>::MAX; }
+    static T lowest() { return -max(); }
+    static T denorm_min() { return num_definitions<T>::DENORM_MIN }
+    
+    static T epsilon() { return num_epsilon; }
+    static T round_error() { return T(0.5); }
+    static T infinity() { return T(0); }
+    static T quiet_NaN() { return T(0); }
+    static T signaling_NaN() { return T(0); }
```
devshgraphicsprogramming, line 182: because of DXC not having constexpr functions, we need to break with STL and declate the functions as variables

include/nbl/builtin/hlsl/numeric_limits.hlsl
devshgraphicsprogramming: actually because of https://github.com/Devsh-Graphics-Programming/Nabla/pull/563/files#r1347172124 this will need to be
```cpp
template<typename T>
struct numeric_limits
{
   static constexpr inline T min =  :std::numeric_limits<T>::min();
   ...etc...
};
```
for C++

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
...
+{
+    static const double MAX         = 1.7976931348623158e+308;
+    static const double MIN         = 2.2250738585072014e-308;
+    static const double DENORM_MIN  = 4.9406564584124654e-324;
+};
+
+
+template<class T>
+struct num_base
+{
```
devshgraphicsprogramming, line 71: all this hackery is fine as long as you hide it in `impl` or `detail` extra namespace

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
...
+    static const int min_exponent10 = !is_integer ? float_min_decimal_exponent : 0;
+    // 	one more than the largest integer power of the radix that is a valid finite floating-point value
+    static const int max_exponent = !is_integer ? float_max_exponent : 0;
+    // the largest integer power of 10 that is a valid finite floating-point value
+    static const int max_exponent10 = !is_integer ? float_max_decimal_exponent : 0;
+
+    // identifies types which can cause arithmetic operations to trap
+    static const bool traps = false;
+    // identifies floating-point types that detect tinyness before rounding
+    static const bool tinyness_before = false;
```
devshgraphicsprogramming, line 171: I'd rather move this into `numeric_limits` instead of keeping it in `num_base` cause from what I see `num_base` just provides a bunch of hacks to compile all this stuff efficiently

Also moving it out will allow you to break the inheritance so `numric_limits` doesnt `: num_base<T>` but `using _impl_base = impl::num_base<T>` internally instead which is cleaner IMHO

include/nbl/builtin/hlsl/numeric_limits.hlsl
```diff
@@ -0,0 +1,202 @@
...
+    static T infinity() { return T(0); }
+    static T quiet_NaN() { return T(0); }
+    static T signaling_NaN() { return T(0); }
+};
+
+template<class T>
+struct numeric_limits : num_base<T>
+{
+
+};
```
devshgraphicsprogramming, line 189: I'd add a specialization for `bool`

include/nbl/builtin/hlsl/numeric_limits.hlsl
devshgraphicsprogramming: ok even with C++23, the support for individual `std::floatN_t` is optional, looks like Clang doesn't implement (but at least has a `_Float16` you can use), GCC does and waiting for confimarion about MSVC 19.38 and 19.39

include/nbl/builtin/hlsl/cpp_compat.hlsl
```diff
@@ -38,6 +45,60 @@ namespace nbl
...
 {
 
+namespace impl
+{
+    template<class T> float16_t asfloat16_t(T val) { return asfloat16(val); }
+    template<class T> float32_t asfloat32_t(T val) { return asfloat(val); }
+    template<class T> int16_t   asint16_t(T val) { return asint16(val); }
+    template<class T> int32_t   asint32_t(T val) { return asint(val); }
+    template<class T> uint16_t  asuint16_t(T val) { return asuint16(val); }
+    template<class T> uint32_t  asuint32_t(T val) { return asuint(val); }
```
devshgraphicsprogramming, line 55: you don't need to make a named and explicitly sized `asuint`, `asint` or `asfloat`, because a X-bit float can only convert to an X-bit signed/unsigned integer

atkurtul: I dont entirely understand what you suggest

devshgraphicsprogramming: why on earth would you allow any `T` to enter these functions? constrain a little bit

include/nbl/builtin/hlsl/cpp_compat.hlsl
```diff
@@ -38,6 +45,60 @@ namespace nbl
...
+    template<class T> uint64_t asuint64_t(T val) {  return val; }
+    template<class T> int64_t asint64_t(T val) {  return asuint64_t(val); }
+    
+    template<>
+    uint64_t asuint64_t<float64_t>(float64_t val) 
+    { 
+        uint32_t lo, hi;
+        asuint(val, lo, hi);
+        return (uint64_t(hi) << 32u) | uint64_t(lo);
+    }
```
devshgraphicsprogramming: this is really disgusting, it seems that HLSL is a bit dumb when it comes to converting doubles.

I'd take a look at #519 and the `spirv` header (copy it over), then make intrinsics for `OpConvertFtoU`, etc.
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_conversion_instructions

also use our type-traits to constrain the `T` a little bit

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,206 @@
...
+
+template<> 
+struct num_traits<float64_t>
+{
+    NBL_CONSTEXPR_STATIC_INLINE float64_t MAX           = 1.7976931348623158e+308;
+    NBL_CONSTEXPR_STATIC_INLINE float64_t MIN           = 2.2250738585072014e-308;
+    NBL_CONSTEXPR_STATIC_INLINE float64_t DENORM_MIN    = 4.9406564584124654e-324;
+    NBL_CONSTEXPR_STATIC_INLINE uint64_t  QUIET_NAN     = 0x7FF8000000000000ull;
+    NBL_CONSTEXPR_STATIC_INLINE uint64_t  SIGNALING_NAN = 0x7FF0000000000001ull;
+};
```
devshgraphicsprogramming: you want to hide this in `impl`

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,206 @@
...
+    denorm_present       = 1
+};
+
+enum float_round_style {
+    round_indeterminate       = -1,
+    round_toward_zero         = 0,
+    round_to_nearest          = 1,
+    round_toward_infinity     = 2,
+    round_toward_neg_infinity = 3
+};
```
devshgraphicsprogramming, line 35: if these exist in `std` then we need an alias in the C++ side, or we need to define them outside

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,206 @@
...
+    NBL_CONSTEXPR_STATIC_INLINE float64_t MAX           = 1.7976931348623158e+308;
+    NBL_CONSTEXPR_STATIC_INLINE float64_t MIN           = 2.2250738585072014e-308;
+    NBL_CONSTEXPR_STATIC_INLINE float64_t DENORM_MIN    = 4.9406564584124654e-324;
+    NBL_CONSTEXPR_STATIC_INLINE uint64_t  QUIET_NAN     = 0x7FF8000000000000ull;
+    NBL_CONSTEXPR_STATIC_INLINE uint64_t  SIGNALING_NAN = 0x7FF0000000000001ull;
+};
+
+template<class T>
+struct numeric_limits
+{
```
devshgraphicsprogramming, line 90: hmm I'd actually define this in `impl` outside of `__HLSL_VERSION`, simply so that we can add a test full of `static_assert` on the CPU that `nbl::hlsl::impl::numeric_limits==nbl::hlsl::numeric_limits`

P.s. obviously `nbl::hlsl::numeric_limits` would still be an alias for `std::numeric_limits`

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,206 @@
...
+};
+
+#ifdef NBL_HALF_DEFINED
+template<> 
+struct num_traits<half>
+{
+    NBL_CONSTEXPR_STATIC_INLINE half MIN         = 6.103515e-05;
+    NBL_CONSTEXPR_STATIC_INLINE half MAX         = 65504;
+    NBL_CONSTEXPR_STATIC_INLINE half DENORM_MIN  = 5.96046448e-08;
+};
```
devshgraphicsprogramming, line 62: `half` also has NaNs

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,206 @@
...
+    NBL_CONSTEXPR_STATIC_INLINE bool tinyness_before = false;
+
+    static T min() { return num_traits<T>::MIN; }
+    static T max() { return num_traits<T>::MAX; }
+    static T lowest() { return is_integer ? min() : -max(); }
+    static T denorm_min() { return num_traits<T>::DENORM_MIN; }
+    
+    static T epsilon() { return num_epsilon; }
+    static T round_error() { return T(0.5); }
+    static T infinity() { return T(1e+300 * 1e+300); }
```
devshgraphicsprogramming: 1. this is gonna return garbage for ints (should be 0)
2. just use `1.0/0.0`

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,206 @@
...
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t  SIGNALING_NAN = 0x7FC00001;
+};
+
+template<> 
+struct num_traits<float64_t>
+{
+    NBL_CONSTEXPR_STATIC_INLINE float64_t MAX           = 1.7976931348623158e+308;
+    NBL_CONSTEXPR_STATIC_INLINE float64_t MIN           = 2.2250738585072014e-308;
+    NBL_CONSTEXPR_STATIC_INLINE float64_t DENORM_MIN    = 4.9406564584124654e-324;
+    NBL_CONSTEXPR_STATIC_INLINE uint64_t  QUIET_NAN     = 0x7FF8000000000000ull;
```
devshgraphicsprogramming, line 81: btw isn't `0.0/0.0` always quiet NaN in HLSL with SPIR-V out
(unless in C++ with `--ffast-math` which flushes all Inf,Nan to 0)

btw we can always add runtime tests that the bitpattern is indeed that.

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,206 @@
...
+    static T infinity() { return T(1e+300 * 1e+300); }
+    static T quiet_NaN() { return bit_cast<T>(num_traits<T>::QUIET_NAN); }
+    static T signaling_NaN() { return bit_cast<T>(num_traits<T>::SIGNALING_NAN); }
+};
+
+
+#else
+
+template<class T>
+using numeric_limits = std::numeric_limits<T>;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: I already mentioned this: https://github.com/Devsh-Graphics-Programming/Nabla/pull/563#discussion_r1347164805

include/nbl/builtin/hlsl/mpl.hlsl
```diff
@@ -55,9 +81,40 @@ struct log2
...
+
+#else
+
+template<class T, class U>
+constexpr T bit_cast(const U& val)
+{
+    return std::bit_cast<T, U>(val);
 }
+
+#endif
```
devshgraphicsprogramming, line 116: why is bitcast in `mpl`? it should be in `bit.hlsl` and not inside any namespace!

include/nbl/builtin/hlsl/cpp_compat/vector.hlsl
```diff
@@ -7,6 +7,7 @@
 #include <glm/glm.hpp>
 #include <glm/detail/_swizzle.hpp>
 #include <stdint.h>
+#include <half.h>
```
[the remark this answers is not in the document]
devshgraphicsprogramming: this is the proper way to do it
https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/src/nbl/asset/interchange/CImageLoaderOpenEXR.cpp#L34

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,223 @@
...
+    // identifies floating-point types that detect tinyness before rounding
+    NBL_CONSTEXPR_STATIC_INLINE bool tinyness_before = false;
+    
+    NBL_CONSTEXPR_STATIC_INLINE T min = impl::num_traits<T>::MIN;
+    NBL_CONSTEXPR_STATIC_INLINE T max = impl::num_traits<T>::MAX;
+    NBL_CONSTEXPR_STATIC_INLINE T lowest = is_integer ? min : -max;
+    NBL_CONSTEXPR_STATIC_INLINE T denorm_min = impl::num_traits<T>::DENORM_MIN;
+    NBL_CONSTEXPR_STATIC_INLINE T epsilon = is_integer ? 0 : (T(1) / T(1ull<<(float_digits-1)));
+    NBL_CONSTEXPR_STATIC_INLINE T round_error = is_same<T, bool>::value ? 0 : T(0.5);
+    NBL_CONSTEXPR_STATIC_INLINE T infinity = is_same<T, bool>::value ? 0 : T(1e+300 * 1e+300);
```
devshgraphicsprogramming, line 198: just use `1.0/0.0` its more sane than relying on some huge constant getting converted to INF (also huge constant is just as susceptible to `--ffast-math` on the CPU)

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,223 @@
...
+struct numeric_limits : std::numeric_limits<T> 
+{
+    using Base = std::numeric_limits<T>;
+    NBL_CONSTEXPR_STATIC_INLINE T min = Base::min();
+    NBL_CONSTEXPR_STATIC_INLINE T max = Base::max();
+    NBL_CONSTEXPR_STATIC_INLINE T lowest = Base::lowest();
+    NBL_CONSTEXPR_STATIC_INLINE T denorm_min = Base::denorm_min();
+    NBL_CONSTEXPR_STATIC_INLINE T epsilon = Base::epsilon();
+    NBL_CONSTEXPR_STATIC_INLINE T round_error = Base::round_error();
+    NBL_CONSTEXPR_STATIC_INLINE T infinity = Base::infinity();
```
devshgraphicsprogramming, line 215: hmm because of `--fast-math` on the CPU, maybe for both HLSL and C++ we should provide/store the `inf` and NaN (quiet/signalling) bitpatterns as `uintN_t` ?

Then this works around C++ flushing out Inf/NaN patterns to 0 behind our back, and HLSL not supporting bitcasts as constexpr functions!
(also allows us to implement `isnan` and `isinf` which are not prone to get fucked up with `--fast-math`, which we already have in `nbl::core::`)

include/nbl/builtin/hlsl/mpl.hlsl
```diff
@@ -69,47 +54,72 @@ uint64_t asuint64_t<float64_t>(float64_t val)
...
+
+template<class T, T N>
+struct countl_zero : impl::countl_zero<uint64_t(N), (sizeof(T) * 8)>
+{
+#ifdef __cplusplus
+    static_assert
 #else
-      std::countl_zero(N);
+    _Static_assert
 #endif
```
devshgraphicsprogramming, line 83: just make an `#define static_assert(...) _Static_assert(__VA_ARGS__)` in `macros.h` for the __HLSL_VERSION path

include/nbl/builtin/hlsl/mpl.hlsl
```diff
@@ -69,47 +54,72 @@ uint64_t asuint64_t<float64_t>(float64_t val)
...
+// #undef NBL_DECLARE_BIT_CAST_TYPES
+// #undef NBL_DECLARE_BIT_CAST_BITS
 
-#undef NBL_DECLARE_BIT_CAST
-#undef NBL_DECLARE_BIT_CAST_TYPES
-#undef NBL_DECLARE_BIT_CAST_BITS
+template<class T, class U>
+T bit_cast(U val)
+{
+    return spirv::bitcast<T, U>(val);
```
devshgraphicsprogramming, line 121: you probably want to static-assert `sizeof(T)<=sizeof(U)`
(but comment it out for now cause DXC )

include/nbl/builtin/hlsl/spirv_intrinsics/core.hlsl
```diff
@@ -15,6 +15,7 @@ namespace hlsl
 namespace spirv
 {
 
+#ifdef __HLSL_VERSION
```
devshgraphicsprogramming, line 18: you probably want to move this up so `nbl::hlsl::spirv` isn't even a namespace for C++

include/nbl/builtin/hlsl/limits.hlsl
atkurtul: it's uint now

include/nbl/builtin/hlsl/limits.hlsl
```diff
@@ -0,0 +1,286 @@
...
+
+template<> 
+struct num_traits<float16_t> : num_base<float16_t>
+{
+    using type = 
+#ifdef __cplusplus
+    float32_t
+#else
+    float16_t
+#endif
```
devshgraphicsprogramming, line 212: this needs to change to `float16_t`

include/nbl/builtin/hlsl/type_traits.hlsl
```diff
@@ -560,7 +559,7 @@ template<class T, unsigned I = 0>
 using extent = std::extent<T, I>;
 
 template<typename T>
-struct typeid_t : std::integral_constant<uint64_t,typeid(T).hash_code()> {};
+struct typeid_t : std::integral_constant<uint64_t,0> {};
```
devshgraphicsprogramming, line 562: ok we need to:
- get rid of a C++ `typeid_t` cause `typeid()` is a runtime thing in C++ 
- and move the HLSL `typeid_t` to `impl` namespace
- implement a `nbl::hlsl::type_info` for HLSL and alias it to `std::type_info`

atkurtul: Ok this is for another PR

devshgraphicsprogramming: these two I need before merging:
- get rid of a C++ typeid_t cause typeid() is a runtime thing in C++
- and move the HLSL typeid_t to impl namespace

PR #574, opened 2023-10-23
Fixed countl_zero_bit

Przemog1:


include/nbl/builtin/hlsl/bit.hlsl
```diff
@@ -61,7 +61,7 @@ template<typename T>
 uint16_t countl_zero(T n)
 {
     uint16_t result = 0u;
-    for(uint32_t bits_log2 = 6u; bits_log2 >= 0u; bits_log2--)
+    for(int32_t bits_log2 = 6u; bits_log2 >= 0u; bits_log2--)
```
devshgraphicsprogramming, line 64: you want to remove the `u` suffices from the literals as well

PR #588, opened 2023-10-30
TgMath

devshgraphicsprogramming:


include/nbl/builtin/hlsl/bit.hlsl
```diff
@@ -25,7 +30,7 @@ namespace hlsl
 template<class T, class U>
 T bit_cast(U val)
 {
-    static_assert(sizeof(T) <= sizeof(U));
+    static_assert(sizeof(T) <= sizeof(U), "destination type must be less than or equal to the source type in size");
```
devshgraphicsprogramming, line 33: btw I read the spec recently, and it looks like sizes must match

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,258 @@
...
+template<class T>
+T copysign(T x, T sign_)
+{
+    return sign_ < T(0) ? -x : x;
+}
+
+// TODO:
+// nan	Generate quiet NaN (function)
+// nextafter	Next representable value (function)
+// nexttoward	Next representable value toward precise value (function)
```
devshgraphicsprogramming, line 134: we had them:
- `nbl::core::nan()`
- the other two somewhere in the IEEE754 GLSL headers

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,258 @@
...
+{
+    return INTRINSIC_NAMESPACE(sqrt)(x*x+y*y);
+}
+
+// Floating-point manipulation functions
+
+template<class T>
+T copysign(T x, T sign_)
+{
+    return sign_ < T(0) ? -x : x;
```
devshgraphicsprogramming, line 128: use bitwise ops, bitcast to uint32_t, clear the high bit on x, clear mantissa on `sign_` and bitwise OR them together and bitcast back

P.S. I think I made a nice trait on `master` or `workgroup_scan` which gives you an N-byte unsiged int type

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,258 @@
...
+T expm1(T x) 
+{ 
+    return INTRINSIC_NAMESPACE(exp)(x) - T(1); 
+}
+
+template<class T> 
+T log1p(T x) 
+{ 
+    return INTRINSIC_NAMESPACE(log)(x + T(1)); 
+}
```
devshgraphicsprogramming, line 94: hehehe, no.

These are very special functions and @Erfan-Ahmadi will tell you why.

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,258 @@
...
+    
+    T t = x + T(6.5);
+    return c * sqrt2pi * INTRINSIC_NAMESPACE(pow)(t, (x-T(.5)))*INTRINSIC_NAMESPACE(exp)(-t)*q;
+}
+
+template<class T>
+T lgamma(T x)
+{
+    return INTRINSIC_NAMESPACE(log)(INTRINSIC_NAMESPACE(tgamma)(x));
+}
```
devshgraphicsprogramming, line 199: where did these implementations come from btw ?

include/nbl/builtin/hlsl/cpp_compat.hlsl
```diff
@@ -32,6 +32,8 @@ using add_pointer = std::add_pointer<T>;
 
 #define NBL_REF_ARG(T) typename nbl::hlsl::add_reference<T>::type
 #define NBL_CONST_REF_ARG(T) typename nbl::hlsl::add_reference<std::add_const_t<T>>::type
+#define NBL_ADDRESS_OF(x) &x
+#define NBL_LANG_SELECT(x, y) x
```
devshgraphicsprogramming, line 36: lets not have them around globally (yet)

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,258 @@
...
+#include <cmath>
+#endif
+
+namespace nbl
+{
+namespace hlsl
+{
+
+
+#ifdef __cplusplus
```
devshgraphicsprogramming, line 20: don't use `__cplusplus`, use `ifndef __HLSL_VERSION` (there might be a C++-like compiler emitting SPIR-V soon - either Shady or Clang-HLSL)

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,258 @@
+// Copyright (C) 2022 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+#ifndef _NBL_BUILTIN_HLSL_CMATH_INCLUDED_
+#define _NBL_BUILTIN_HLSL_CMATH_INCLUDED_
```
devshgraphicsprogramming, line 5: btw didn't we agree just to have `tgmath` instead?

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,258 @@
...
+#ifdef __cplusplus
+#define INTRINSIC_NAMESPACE(x) std::x
+#else
+#define INTRINSIC_NAMESPACE(x) x
+#endif
+
+#define NBL_ALIAS_BINARY_FUNCTION2(name,impl) template<class T> T name(T x, T y) { return INTRINSIC_NAMESPACE(impl)(x, y); }
+#define NBL_ALIAS_BINARY_FUNCTION(fn) NBL_ALIAS_BINARY_FUNCTION2(fn,fn)
+#define NBL_ALIAS_UNARY_FUNCTION2(name,impl)  template<class T> T name(T x) { return INTRINSIC_NAMESPACE(impl)(x); }
+#define NBL_ALIAS_UNARY_FUNCTION(fn)  NBL_ALIAS_UNARY_FUNCTION2(fn,fn)
```
devshgraphicsprogramming, line 29: life is never that easy, since we're bridging the gap between HLSL and STL and having a unified namespace for lang/target agnostic code

#### C++

In HLSL all those functions must be able to work on `vector`, so C++ needs a per-component "forwarder" to do the function on each component of a vector when `is_vector` is true.

Also C++ has templated functions in `tgmath` and I'd prefer you to use that rather than rely on overload resolution to hit the right one.

### HLSL

Check if the intrinsics don't have templated versions (cause apparently `ByteAddressBuffer` does and they're not in the documentation XD).

### Something to think about

#### How will this all with `complex`

Does `tgmath` need to include `complex` ? Should we forward declare the specializations?

Remember that `complex` will be templated not only on `float_t` but `fvec_t` and e.g. `complex<float32_t3>` is not a builtin type so will need to be forwarded component by component even in HLSL.

#### After OpenEXR upgrade: Use IlmMath's versions of all these functions if it has them

HLSL doesn't have all these functions for `half` types (only GLSL on AMD used to).

For now its okay to not support any of this for HLSL and C++.

include/nbl/builtin/hlsl/cmath.hlsl
Erfan-Ahmadi: I think we encountered it in bezier arclen computations. 
Anyways that's a special function that's supposed to be more accurate for values near 0 than log(1+x)

devshgraphicsprogramming: AFAIK both Boost and MUSL are permissively licensed (notices must be preserved) and Boost is more likely to have specialized implementations for different float sizes
https://www.boost.org/doc/libs/1_77_0/boost/math/special_functions/log1p.hpp

devshgraphicsprogramming: Hmm given that `log` functions on GPUs are somewhat accelerated, snooping the implementations of cmath (which OpenCL requires) from clvk, sycl, or LLVM is probably best
https://github.com/llvm-mirror/libclc/blob/master/generic/lib/math/log1p.cl

devshgraphicsprogramming:
Can steal from https://www.boost.org/doc/libs/1_83_0/libs/core/doc/html/core/cmath.html

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -1,11 +1,10 @@
 // Copyright (C) 2022 - DevSH Graphics Programming Sp. z O.O.
 // This file is part of the "Nabla Engine".
 // For conditions of distribution and use, see copyright notice in nabla.h
-#ifndef _NBL_BUILTIN_HLSL_CMATH_INCLUDED_
-#define _NBL_BUILTIN_HLSL_CMATH_INCLUDED_
+#ifndef _NBL_BUILTIN_HLSL_TGMATH_INCLUDED_
+#define _NBL_BUILTIN_HLSL_TGMATH_INCLUDED_
```
devshgraphicsprogramming, line 5: need to rename the file as well

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -125,12 +142,21 @@ T hypot(T x, T y)
...
+    const uint_type a = std::bit_cast<uint_type, T>(sign_) & m;
+    const uint_type b = std::bit_cast<uint_type, T>(x) & (m-1);
+    return std::bit_cast<T, uint_type>(a ^ b);
 }
 
+// Generate quiet NaN (function)
+template<class T>
+T nan()
+{
+    return nbl::core::nan<T>();
```
devshgraphicsprogramming, line 156: no `numeric_limits`

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -99,7 +116,7 @@ int32_t ilogb(T x)
     using uint_type = typename nbl::hlsl::numeric_limits<T>::uint_type;
     const int32_t shift = (impl::num_base<T>::float_digits-1);
     const uint_type mask = ~(((uint_type(1) << shift) - 1) | (uint_type(1)<<(sizeof(T)*8-1)));
-    int32_t bits = (bit_cast<uint_type, T>(x) & mask) >> shift;
+    int32_t bits = (std::bit_cast<uint_type, T>(x) & mask) >> shift; //nbl::hlsl::bit_cast sucks as it pulls the ambiguous one from intrinsics.h
```
devshgraphicsprogramming, line 119: explain

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -125,12 +142,21 @@ T hypot(T x, T y)
 template<class T>
 T copysign(T x, T sign_)
 {
-    return sign_ < T(0) ? -x : x;
+    using uint_type = uint32_t; //typename nbl::hlsl::numeric_limits<T>::uint_type;
```
devshgraphicsprogramming, line 145: what was the problem?

include/nbl/builtin/hlsl/cmath.hlsl
devshgraphicsprogramming: I want all this to be tested and compile as both C++ and HLSL, you obviously haven't tested

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -125,12 +142,21 @@ T hypot(T x, T y)
...
+    constexpr uint_type m = (1u << 31);
```
devshgraphicsprogramming, line 146: I want all this to be tested and compile as both C++ and HLSL, you obviously haven't tested

HLSL doesn't have constexpr

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,258 @@
...
+#define NBL_ALIAS_UNARY_FUNCTION(fn)  NBL_ALIAS_UNARY_FUNCTION2(fn,fn)
+
+#define NBL_ALIAS_FUNCTION_WITH_OUTPUT_PARAM(fn, out_type) \
+template<class T>  \
+T fn(T x, NBL_REF_ARG(out_type) y) { \
+    NBL_LANG_SELECT(out_type, T) out_; \
+    T ret = INTRINSIC_NAMESPACE(fn)(x, NBL_ADDRESS_OF(out_)); \
+    y = out_type(out_); \
+    return ret; \
+}
```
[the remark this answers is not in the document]
devshgraphicsprogramming: IMHO this macro is too ugly for the amount of function aliases it provides.

I'd rather have `frexp` and `modf` written out by hand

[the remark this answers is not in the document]
devshgraphicsprogramming: Also `NBL_ADDRESS_OF` and `LAND_SELECT` seems to be super weird, the purpose seems to be to silently use `out_type` and perform no conversions under HLSL and `T` under C++

HLSL can do conversions during copy-in and copy-out, so one could just always use `T` throughout and rely on this mechanic whenever HLSL want to use a different type

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,286 @@
...
+{
+namespace hlsl
+{
+
+#ifndef __HLSL_VERSION
+    #define NBL_ADDRESS_OF(x) &x
+    #define NBL_LANG_SELECT(x, y) x
+    #define INTRINSIC_NAMESPACE(x) std::x
+#else
+    #define INTRINSIC_NAMESPACE(x) x
```
devshgraphicsprogramming, line 23: `INTRINSIC_NAMESPACE(blah)` shouldn't be a macro, just a regular define `INTRINSIC_NAMESPACE`

also the HLSL version should actually fully write out the namespace, i.e. `nbl::hlsl`

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,286 @@
...
+NBL_ALIAS_UNARY_FUNCTION(tan)
+NBL_ALIAS_UNARY_FUNCTION(acos)
+NBL_ALIAS_UNARY_FUNCTION(asin)
+NBL_ALIAS_UNARY_FUNCTION(atan)
+NBL_ALIAS_BINARY_FUNCTION(atan2)
+
+// Hyperbolic functions
+NBL_ALIAS_UNARY_FUNCTION(cosh)
+NBL_ALIAS_UNARY_FUNCTION(sinh)
+NBL_ALIAS_UNARY_FUNCTION(tanh)
```
devshgraphicsprogramming, line 55: comment out the stuff thats not currently available with out spirv & glsl headers

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,286 @@
...
+// Exponential and logarithmic functions
+NBL_ALIAS_UNARY_FUNCTION(exp)
+NBL_ALIAS_FUNCTION_WITH_OUTPUT_PARAM(frexp, int32_t)
+NBL_ALIAS_BINARY_FUNCTION(ldexp)
+NBL_ALIAS_UNARY_FUNCTION(log)
+NBL_ALIAS_UNARY_FUNCTION(log10)
+NBL_ALIAS_FUNCTION_WITH_OUTPUT_PARAM(modf, T)
+NBL_ALIAS_UNARY_FUNCTION(exp2)
+NBL_ALIAS_UNARY_FUNCTION(log2)
+NBL_ALIAS_UNARY_FUNCTION2(logb,log)
```
devshgraphicsprogramming, line 85: https://github.com/Devsh-Graphics-Programming/Nabla/pull/588/files#r1554730455

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,286 @@
...
+template<class T> 
+T expm1(T x) 
+{ 
+    //wrong
+    // for better implementation, missing function for evaluate_polynomial
+    // https://www.boost.org/doc/libs/1_83_0/boost/math/special_functions/expm1.hpp
+
+    //or use std::expm1f, std::expm1l depending on size of T
+    return INTRINSIC_NAMESPACE(exp)(x) - T(1); 
+}
```
devshgraphicsprogramming, line 96: remove implementation

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,286 @@
...
+T asinh(T x)
+{
+    return INTRINSIC_NAMESPACE(log)(x + INTRINSIC_NAMESPACE(sqrt)(x*x + T(1)));
+}
+
+template<class T>
+T atanh(T x)
+{
+    return T(0.5) * INTRINSIC_NAMESPACE(log)((T(1)+x)/(T(1)-x));
+}
```
devshgraphicsprogramming, line 73: remove this and replace with a TODO comment to forward directly

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,286 @@
...
+template<class T> 
+T scalbn(T x, int32_t n) 
+{ 
+    return x * INTRINSIC_NAMESPACE(exp2)(n); 
+}
+
+// Power functions
+NBL_ALIAS_BINARY_FUNCTION(pow)
+NBL_ALIAS_UNARY_FUNCTION(sqrt)
+NBL_ALIAS_UNARY_FUNCTION(cbrt)
```
devshgraphicsprogramming, line 132: `cbrt` surely isn't available in SPIR-V, comment out, TODO

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,286 @@
...
+T erf(T x)
+{
+    // Bürmann series approximation  
+    // https://www.desmos.com/calculator/myf9ylguh1
+    // https://en.wikipedia.org/wiki/Error_function#Numerical_approximations
+    T E = INTRINSIC_NAMESPACE(exp)(-x*x);
+    T P = T(0.886226925453);
+    T re = INTRINSIC_NAMESPACE(sqrt)(T(1)-E)*(T(1)+T(.155)*E/P*(T(1)-T(.275)*E));
+    return INTRINSIC_NAMESPACE(copysign)(re, x);
+}
```
devshgraphicsprogramming, line 174: Remove all code/impl thats either subpar or we don't have a license/software-bill-of-materials for

include/nbl/builtin/hlsl/cmath.hlsl
devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/588/files#r1554730780

include/nbl/builtin/hlsl/cmath.hlsl
```diff
@@ -0,0 +1,286 @@
...
+
+// Minimum, maximum, difference functions
+
+NBL_ALIAS_BINARY_FUNCTION2(fmax, max)
+NBL_ALIAS_BINARY_FUNCTION2(fmin, min)
+template<class T>
+T fdim(T x, T y)
+{
+    return INTRINSIC_NAMESPACE(max)(T(0),x-y);
+}
```
devshgraphicsprogramming, line 271: comment out whatever doesn't work in HLSL

include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -0,0 +1,233 @@
...
+//T asinh(T x)
+//{
+//    return INTRINSIC_NAMESPACE(log)(x + INTRINSIC_NAMESPACE(sqrt)(x*x + T(1)));
+//}
+
+template<class T>
+T atanh(T x)
+{
+    return T(0.5) * INTRINSIC_NAMESPACE(log)((T(1)+x)/(T(1)-x));
+}
```
devshgraphicsprogramming, line 74: @keptsecret don't grab these implementations, should have been done in terms of `log1p`

See https://git.musl-libc.org/cgit/musl/tree/src/math/acosh.c

include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -0,0 +1,233 @@
...
+
+template<class T>
+int32_t ilogb(T x) 
+{ 
+    using uint_type = typename nbl::hlsl::numeric_limits<T>::uint_type;
+    const int32_t shift = (impl::num_base<T>::float_digits-1);
+    const uint_type mask = ~(((uint_type(1) << shift) - 1) | (uint_type(1)<<(sizeof(T)*8-1)));
+    int32_t bits = (INTRINSIC_NAMESPACE(bit_cast)<uint_type, T>(x) & mask) >> shift; 
+    return bits + impl::num_base<T>::min_exponent - 2;
+}
```
devshgraphicsprogramming, line 123: these need to get checked against musl and boost again

include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -0,0 +1,233 @@
...
+    return x;
+}
+
+// Generate quiet NaN (function)
+template<class T>
+T nan()
+{
+    using uint_type = typename nbl::hlsl::numeric_limits<T>::uint_type;
+    return INTRINSIC_NAMESPACE(bit_cast)<uint_type, T>(nbl::hlsl::numeric_limits<T>::quiet_NaN);
+}
```
devshgraphicsprogramming, line 158: @keptsecret @Przemog1's ieee stuff does it better

include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -0,0 +1,233 @@
...
+#define NBL_ALIAS_UNARY_FUNCTION(fn)  NBL_ALIAS_UNARY_FUNCTION2(fn,fn)
+
+#define NBL_ALIAS_FUNCTION_WITH_OUTPUT_PARAM(fn, out_type) \
+template<class T>  \
+T fn(T x, NBL_REF_ARG(out_type) y) { \
+    NBL_LANG_SELECT(out_type, T) out_; \
+    T ret = INTRINSIC_NAMESPACE(fn)(x, NBL_ADDRESS_OF(out_)); \
+    y = out_type(out_); \
+    return ret; \
+}
```
devshgraphicsprogramming, line 39: this has backfired for us spectacularly, we need to actually write them out by hand

devshgraphicsprogramming:
@keptsecret this isn't really salavageable, mayber `log1p` and friends but nothing else

include/nbl/builtin/hlsl/tgmath.hlsl
Przemog1: here: [ieee754.hlsl](https://github.com/Devsh-Graphics-Programming/Nabla/blob/cpp_intrinsics_improvements/include/nbl/builtin/hlsl/ieee754.hlsl)

Przemog1: @devshgraphicsprogramming maybe @keptsecret should merge [this pr?](https://github.com/Devsh-Graphics-Programming/Nabla/pull/804)

devshgraphicsprogramming: probably yes

PR #589, opened 2023-10-30
Nahim subgroup arithmetic

nahiim:
## Description
Use spv intrinsics in glsl_compat files


devshgraphicsprogramming:
you need to cherry pick your SPV commits and apply them at a branch that you create from `workgroup_scan` HEAD.

PR #590, opened 2023-10-30
member tester macros

atkurtul:


include/nbl/builtin/hlsl/member_test_macros.hlsl
```diff
@@ -0,0 +1,53 @@
...
+namespace hlsl
+{
+
+namespace impl
+{
+
+template<class T>
+struct valid_expression : true_type {};
+
+template<> struct valid_expression<void> : false_type {};
```
devshgraphicsprogramming, line 22: this is just `make_void` all over again?

include/nbl/builtin/hlsl/spirv_intrinsics/core.hlsl
```diff
@@ -110,10 +110,6 @@ void memoryBarrier(uint32_t memoryScope, uint32_t memorySemantics);
 #pragma region CONVERSION
 
 template<class T, class U>
-// [[vk::ext_extension("SPV_KHR_physical_storage_buffer")]]
-// [[vk::ext_capability(spv::CapabilityPhysicalStorageBufferAddresses)]]
-// [[vk::ext_capability(spv::CapabilityAddresses)]]
```
devshgraphicsprogramming, line 115: you'd probably be declaring those on a partial specialization at some point, useful to leave that as a comment

include/nbl/builtin/hlsl/member_test_macros.hlsl
atkurtul: No it's different.  
`make_void` maps everything to void. 
`valid_expression` maps void to false and everything else to true
Rather than doing
```cpp
template<class T,class=void>
struct spec{ };
template<class T>
struct spec<T, typename enable_if<!is_same<decltype(expr), void>::value, void>::type>
```
we can do
```cpp
template<class T,class=bool_constant<true>>
struct spec{ };
template<class T>
struct spec<T, bool_constant<valid_expression<decltype(expr)>::value>> {};
```

devshgraphicsprogramming: do you even need an extra `bool_constant` at that stage? wonder if:
```cpp
template<class T,bool=true>
struct spec{ };
template<class T>
struct spec<T,valid_expression<decltype(expr)>::value> {};
```
would just work?

atkurtul: it doesnt Dxc complains about partial spec being type dependent

PR #591, opened 2023-10-31
Nahim complex math

nahiim:
## Description
Start drafting complex math


devshgraphicsprogramming:
why on earth do I have so many commits!?

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -17,19 +14,21 @@ namespace hlsl
...
+
+
+template <typename T, uint16_t N>
+struct complex_t
 {
-    vector<T, N> real;
-    vector<T, N> imag;
+    T real[N];
+    T imag[N];
 };
```
atkurtul, line 24: Why does complex_t need a dimension?

devshgraphicsprogramming: it does not, the dimensionality is supposed to come from `T=vector<float_type,N`

PR #592, opened 2023-11-03
add complex.hlsl

nahiim:
## Description
Start drafting complex.hlsl


include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,77 @@
...
+complex_t<T> divide(const complex_t<T> a, const complex_t<T> b)
+{
+    complex_t<T> result;
+
+    T denominator = b.real * b.real + b.imag * b.imag;
+    result.real = (a.real * b.real + a.imag * b.imag) / denominator;
+    result.imag = (a.imag * b.real - a.real * b.imag) / denominator;
+
+    return result;
+}
```
devshgraphicsprogramming, line 69: HLSL supports these operators

the operators it doesn't support are `+=` and those you can put in `functional` as `plus_assign` (is to `+=`  what `plus` is to `+`)

nahiim: I don't really understand what you mean. I just implemented arithmetic operations since we hlsl doesn't allow operator overloading

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -24,7 +26,7 @@ struct complex_t
 };
 
 template<typename T>
-complex_t<T> add(const complex_t<T> a, const complex_t<T> b)
+complex_t<T> plus(const complex_t<T> a, const complex_t<T> b)
 {
     complex_t<T> result;
```
devshgraphicsprogramming: we already have struct functors for those in `nbl/builtin/hlsl/functional` you need to add an explicit specialization

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -14,6 +14,8 @@ namespace nbl
 {
 namespace hlsl
 {
+namespace complex
+{
```
devshgraphicsprogramming, line 18: no need for this extra namespace, see the specialization comment

include/nbl/builtin/hlsl/radix_fft.hlsl
```diff
@@ -0,0 +1,71 @@
...
+
+namespace nbl
+{
+namespace hlsl
+{
+
+
+// Fast Fourier Transform
+template<typename T, uint32_t N>
+void fft_radix2(complex::complex_t<T> data[N], bool is_inverse)
```
devshgraphicsprogramming, line 19: make the `is_inverse` a template parameter

`assert(N<gl_SubgroupSize)` 

btw the `data` need not be `data[N]`, each invocation comes with its own register value and you can cooperatively exchange via the use of `subgroupShuffleXor` (add it to https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/hlsl/glsl_compat/subgroup_shuffle.hlsl please).

I mean the original GLSL code is here
https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/glsl/subgroup/fft.glsl

include/nbl/builtin/hlsl/radix_fft.hlsl
```diff
@@ -0,0 +1,71 @@
...
+            complex::complex_t<T> temp = data[i];
+            data[i] = data[j];
+            data[j] = temp;
+        }
+    }
+
+    // Cooley-Tukey
+    for (uint32_t stride = 2; stride <= N; stride *= 2)
+    {
+        complex::complex_t<T> twiddle_factor = { cos((is_inverse ? 2.0 : -2.0) * 3.14 / stride), sin((is_inverse ? 2.0 : -2.0) * 3.14 / stride) };
```
devshgraphicsprogramming: use the proper full FP64 value of PI

include/nbl/builtin/hlsl/glsl_compat/subgroup_shuffle.hlsl
```diff
@@ -38,6 +38,18 @@ T subgroupShuffleDown(T value, uint delta)
 #endif
 }
 
+
+template<typename T>
+T subgroupShuffleXor(T value, uint32_t delta)
+{
+#ifdef NBL_GL_KHR_shader_subgroup_shuffle_relative
+    return spirv::groupShuffleXor<T>(3, value, delta);
```
devshgraphicsprogramming, line 46: @kpentaris do we assume shuffle relative always there?

kpentaris: It was not present in all devices if I remember correctly, which is why I added the #ifdef, but that might be the case for devices that are no longer supported by Nabla.

devshgraphicsprogramming: we require it now https://github.com/Devsh-Graphics-Programming/Nabla/blob/56d4efb3fbbfca936f6cf8936e53b31823d4ad1d/src/nbl/video/vulkan/profiles/NablaCore.json#L328

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,102 @@
...
+// Divide by scalar
+template<typename T>
+complex_t<T> div(const complex_t<T> c, T scalar)
+{
+    complex_t<T> result;
+    result.real = c.real / scalar;
+    result.imag = c.imag / scalar;
+
+    return result;
+}
```
devshgraphicsprogramming, line 93: @Fletterio according to @Przemog1 all of these can be real `operator`

its only `nbl/builtin/hlsl/binops.hlsl` that should receive some `nbl::hlsl::assign_add` functor structs with `void operator()(NBL_CONST_REF_ARG(T) other)` (void return cause there's no references in HLSL :( ) with specializations for `complex_t`

Note: @Przemog1 we want to do the same thing for `emualted_float64_t` in your PR

include/nbl/builtin/hlsl/radix_fft.hlsl
devshgraphicsprogramming: btwa nice piece of trivia @Fletterio , only GLSL defines fp64 transcendentals, HLSL doesn't (the sin,cos etc. functions always return float32, and if you try to use the GL.450.std extended instruction set to compute `cos(double)` this will end up emitting Float64 capability which is a definite DO NOT WANT)

So these definitely need to be tabulated as very precise constants and C-casted into the correct `nbl::hlsl::scalar_type_t<T>` so we don't accidentally emit Float64 capability

include/nbl/builtin/hlsl/glsl_compat/subgroup_shuffle.hlsl
devshgraphicsprogramming: remove the ifdef (if its still there after merging `master` to this branch and resolving conflicts)

devshgraphicsprogramming:
I see a certain ladder of features, and each should have their own branch (so I can merge in order without having all comments addressed):
1. having a `complex_t` struct to use
2. `subgroup::fft` utilities for DIT and DIF
3. `workgroup::fft` utilities for DIT and DIF
4. a way to do multiple input items per-workgroup, i.e. do a 2048 sized FFT with a 512 workgroup while having the option of only using enough shared memory for 512 FFT + auxilary accessor
5. optimizations for FFT real signals
6. resurrect the `nbl::ext` and 1 or both of the old FFT examples to test with

They can branch off each other.

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -7,96 +7,172 @@
 
 
 
-#ifdef __HLSL_VERSION
+#ifndef __HLSL_VERSION
```
devshgraphicsprogramming, line 10: imo no point guarding it

include/nbl/builtin/hlsl/radix_fft.hlsl
Fletterio: I'm not really following - we want to use the full PI value in f64, but not emit f64 capability?

include/nbl/builtin/hlsl/complex.hlsl
Fletterio: Marking this as resolved as this is now an operator

include/nbl/builtin/hlsl/radix_fft.hlsl
devshgraphicsprogramming: yes you want to cast the constant into a type you'd be using, see the Gauss Legendre / Quadrature header stuff that Erfan is using for example 62 CAD

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,191 @@
...
+#include <nbl/builtin/hlsl/binops.hlsl>
+
+namespace nbl
+{
+namespace hlsl
+{
+
+    template<typename Scalar>
+    struct complex_t
+    {
```
devshgraphicsprogramming, line 17: dont indent extra because of namespace

include/nbl/builtin/hlsl/binops.hlsl
```diff
@@ -0,0 +1,25 @@
...
+    { \
+        lhs = lhs OP rhs; \
+    } \
+}
+COMPOUND_ASSIGN(add,+);
+COMPOUND_ASSIGN(subtract,-);
+COMPOUND_ASSIGN(mul,*);
+COMPOUND_ASSIGN(div,/);
+
+#undef COMPOUND_ASSIGN
```
devshgraphicsprogramming, line 20: add them here https://github.com/Devsh-Graphics-Programming/Nabla/blob/66332959a4b5df61dc1e327dfa646fa91ea00de1/include/nbl/builtin/hlsl/functional.hlsl#L67

and make sure they have names `plus_assign`, `multiplies_assign` to be inline with the rest.

Also because its just 4 struct defs and not specialization, IMHO you shouldn't use macros because that just makes the code iffy to read. Or if you are to use a macro make sure it has the same form as the `ALIAS_STD` so you have to put th e `};` afterwards to be able to add extra fields (i.e. identity)

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,191 @@
...
+        static complex_t create(const Scalar real, const Scalar imag)
+        {
+                complex_t retVal = { real, imag };
+                return retVal;
+        }
+            
+        static complex_t create(const complex_t other)
+        {
+                return other;
+        }
```
devshgraphicsprogramming, line 31: no needed, regular assignment will do

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,191 @@
...
+            
+        // ----------------- Relational operators -----------------------------
+        bool operator==(const complex_t rhs)
+        {
+                return !(uint64_t(m_real) ^ uint64_t(rhs.m_real)) && !(uint64_t(m_imag) ^ uint64_t(rhs.m_imag));
+        }
+        bool operator!=(const complex_t rhs)
+        {
+                return uint64_t(m_real) ^ uint64_t(rhs.m_real) || uint64_t(m_imag) ^ uint64_t(rhs.m_imag);
```
devshgraphicsprogramming, line 120: don't cast to `uint64_t` :
1. you're not comparing bitpatterns, you're casting to uint64_T which truncatest the floats, etc.
2. `Scalar=float32_t` would have been 32bit
3. `!=` and `==` work on floats, so use them

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    template <typename Scalar>
+    using assign_add_complex = typename ComplexCompoundHelper<Scalar>::add;
+    template <typename Scalar>
+    using assign_subtract_complex = typename ComplexCompoundHelper<Scalar>::subtract;
+    template <typename Scalar>
+    using assign_mul_complex = typename ComplexCompoundHelper<Scalar>::mul;
+    template <typename Scalar>
+    using assign_div_complex = typename ComplexCompoundHelper<Scalar>::div;
```
devshgraphicsprogramming, line 141: I don't get this, we don't have them for regular `plus<T>`

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -0,0 +1,44 @@
...
+
+#include "nbl/builtin/hlsl/complex.hlsl"
+#include "nbl/builtin/hlsl/cpp_compat.hlsl"
+#include "nbl/builtin/hlsl/fft/common/twiddles.hlsl"
+
+namespace nbl 
+{
+namespace hlsl
+{
+namespace fft::common {
```
devshgraphicsprogramming, line 12: 1. HLSL doesn't support nested namespace syntax
2. use `detail` or `impl` for things you don't want to make public to others

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -0,0 +1,44 @@
...
+
+    template<typename Scalar, bool inverse> 
+    struct DIX {
+        //Assumes I have precomputed twiddles, so just pass the index. 
+        static void radix2Butterfly(NBL_CONST_REF_ARG(complex_t<Scalar>) twiddle, NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi){
```
devshgraphicsprogramming, line 17: drop `Butterfly` from the name, just `radix2` is fine

include/nbl/builtin/hlsl/fft/common/twiddles_impl.hlsl
```diff
@@ -0,0 +1,291 @@
+#ifndef float_t
+#error Define float_t before including 
+#endif
+#ifndef TYPED_NUMBER
+#error Define TYPED_NUMBER before including 
+#endif
```
devshgraphicsprogramming, line 6: but this is same as `float_t` isn't it?

include/nbl/builtin/hlsl/fft/common/twiddles_impl.hlsl
```diff
@@ -0,0 +1,291 @@
...
+#ifndef MID_TWIDDLE_BITS
+#error Define MID_TWIDDLE_BITS before including 
+#endif
+#ifndef HIGH_TWIDDLE_BITS
+#error Define HIGH_TWIDDLE_BITS before including 
+#endif
+
+#define NAMESPACE_NAME NBL_CONCATENATE(twiddles_, float_t) 
+
```
devshgraphicsprogramming, line 19: bury inside `nbl::hlsl::fft` as well

include/nbl/builtin/hlsl/fft/common/twiddles_impl.hlsl
```diff
@@ -0,0 +1,291 @@
...
+namespace NAMESPACE_NAME {
+    
+    // -------------------------------------------- 64th roots of unity --------------------------------------------------------
+
+    NBL_CONSTEXPR float_t low_real[1u << LOW_TWIDDLE_BITS] = {
+        TYPED_NUMBER(1),
```
devshgraphicsprogramming, line 25: I'm pretty sure that compiler will just choke with
```
too many elements in vector initialization (expected X element, have Y)
```
when `1<<LOW_TWIDDLE_BITS` is less than the length of your array, as this is what godbolt gives on
```
const static float arr[1] = {6.f,9.f}
```

include/nbl/builtin/hlsl/fft/common/twiddles.hlsl
```diff
@@ -0,0 +1,43 @@
...
+    #define TYPED_NUMBER(N) NBL_CONCATENATE(N, f) // to add f after floating point numbers and avoid casting warnings and emitting ShaderFloat64 Caps
+    #include <nbl/builtin/hlsl/fft/common/twiddles_impl.hlsl>
+    #undef TYPED_NUMBER
+    #undef float_t 
+
+    #define float_t float64_t
+    #define TYPED_NUMBER(N) N
+    #include <nbl/builtin/hlsl/fft/common/twiddles_impl.hlsl>
+    #undef TYPED_NUMBER
+    #undef float_t
```
devshgraphicsprogramming, line 37: don't bother precomputing twiddles, did this benchmark in shadertoy
```cpp
const float _cos[32] = float[32](
1.0,0.980785,0.92388,0.83147,0.707107,0.55557,0.382683,0.19509,-4.37114e-08,-0.19509,-0.382684,-0.55557,-0.707107,-0.83147,-0.92388,-0.980785,-1.0,-0.980785,-0.92388,-0.83147,-0.707107,-0.55557,-0.382683,-0.19509,1.19249e-08,0.19509,0.382684,0.55557,0.707107,0.83147,0.92388,0.980785
);

void mainImage( out vec4 fragColor, in vec2 fragCoord )
{
    // Normalized pixel coordinates (from 0 to 1)
    vec2 uv = fragCoord/iResolution.xy;

    // Time varying pixel color
    vec3 col = vec3(0.0);

    const int spins = 2000;
    for (int i=0; i<spins; i++)
    {
        int iarg = i;//+int(fragCoord.x)+int(fragCoord.y);
        float arg = float(iarg)/float(spins);
        float cs = cos(arg*6.2831853);
        col.rg += vec2(_cos[iarg]);
    }
    // Output to screen
    fragColor = vec4(col,1.0);
}
```
perf dropped 50%, it gets worse the bigger the array gets

include/nbl/builtin/hlsl/radix_fft.hlsl
```diff
@@ -0,0 +1,80 @@
...
+
+#include <nbl/builtin/hlsl/complex.hlsl>
+
+
+namespace nbl
+{
+namespace hlsl
+{
+
+const float64_t PI = 3.1415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679;
```
devshgraphicsprogramming, line 16: please make a `numbers.hlsl` a counterpart to https://en.cppreference.com/w/cpp/numeric/constants

include/nbl/builtin/hlsl/radix_fft.hlsl
```diff
@@ -0,0 +1,80 @@
...
+        for (uint32_t bit = 0; bit < log2N; ++bit)
+            j |= ((i >> bit) & 1) << (log2N - 1 - bit);
+
+        if (j > i)
+        {
+            complex_t<T> temp = data[i];
+            data[i] = data[j];
+            data[j] = temp;
+        }
+    }
```
devshgraphicsprogramming, line 37: don't do the bit reversal for me, its often not needed

include/nbl/builtin/hlsl/radix_fft.hlsl
```diff
@@ -0,0 +1,80 @@
...
+namespace hlsl
+{
+
+const float64_t PI = 3.1415926535897932384626433832795028841971693993751058209749445923078164062862089986280348253421170679;
+
+// Fast Fourier Transform
+template<typename T, uint32_t N, bool is_inverse>
+void fft_radix2(complex_t<T> data[N])
+{
+    const uint32_t log2N = uint32_t(log2(N));
```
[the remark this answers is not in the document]
devshgraphicsprogramming: is this supposed to be a workgroup FFT? a subgroup FFT or what?

include/nbl/builtin/hlsl/radix_fft.hlsl
```diff
@@ -0,0 +1,80 @@
...
+    {
+        complex_t<T> twiddle_factor;
+        if(is_inverse)
+        {
+            twiddle_factor = { cos(2.0 * PI / stride), sin(2.0 * PI / stride) };
+        }
+        else
+        {
+            twiddle_factor = { cos(-2.0 * PI / stride), sin(-2.0 * PI / stride) };
+        }
```
devshgraphicsprogramming, line 52: 1. the cosine is the same, because its an even function
2. sine is an odd function

TL;DR its enough for you to do
```cpp
const float arg = pi_v<T>*2.f;
twiddle_factor = {cos(arg),sin(arg)};
if (is_inverse)
    twiddle_factor.imag = -twiddle_factor.imag; // compiler should optimize to a bitflip
```

include/nbl/builtin/hlsl/radix_fft.hlsl
```diff
@@ -0,0 +1,80 @@
...
+            complex_t<T> temp = data[i];
+            data[i] = data[j];
+            data[j] = temp;
+        }
+    }
+
+    // Cooley-Tukey
+    assign_mul_complex<T> mul_assign;
+    assign_div_complex<T> div_assign;
+    for (uint32_t stride = 2; stride <= N; stride *= 2)
```
devshgraphicsprogramming, line 42: I'd do a loop from `stride=1` to `stride=N/2` instead of `2->N` because that's the actual stride between your butterfly elements, what you have right now is a width

include/nbl/builtin/hlsl/glsl_compat/subgroup_shuffle.hlsl
```diff
@@ -30,6 +30,14 @@ T subgroupShuffleDown(T value, uint32_t delta)
     return spirv::groupShuffleDown<T>(spv::ScopeSubgroup, value, delta);
 }
 
+
+template<typename T>
+T subgroupShuffleXor(T value, uint32_t delta)
```
devshgraphicsprogramming, line 35: does the spec call it a delta? I'd call it a mask

include/nbl/builtin/hlsl/radix_fft.hlsl
devshgraphicsprogramming: nvm about the tabulation... its not a perf win compared to just calling sin and cos

for now just assert that `T` for the FFT complex numbers is NOT `float64_t` because that opens a whole can of worms of std.glsl spir-v intrinsic header

include/nbl/builtin/hlsl/radix_fft.hlsl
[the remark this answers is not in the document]
Fletterio: This is all nahim's code, I didn't bother changing it yet

include/nbl/builtin/hlsl/binops.hlsl
Fletterio: To be able to define an identity for, say, complex numbers, I'd need the templated expressions `T(0)` or `T(1)` to make sense when replacing `T` by `complex_t<T>`. Is that doable?

include/nbl/builtin/hlsl/glsl_compat/subgroup_shuffle.hlsl
```diff
@@ -19,22 +19,21 @@ T subgroupShuffle(T value, uint32_t invocationId)
 }
 
 template<typename T>
-T subgroupShuffleUp(T value, uint32_t delta)
+T subgroupShuffleXor(T value, uint32_t delta)
 {
-    return spirv::groupShuffleUp<T>(spv::ScopeSubgroup, value, delta);
+    return spirv::groupShuffleXor<T>(spv::ScopeSubgroup, value, mask);
```
devshgraphicsprogramming, line 24: how does this even compile?, `mask` is not defined.?

include/nbl/builtin/hlsl/radix_fft.hlsl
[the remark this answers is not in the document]
devshgraphicsprogramming: ah ok!

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -12,178 +12,178 @@ namespace nbl
...
+};
+
+template <typename Scalar>
+using assign_add_complex = typename ComplexCompoundHelper<Scalar>::add;
+template <typename Scalar>
+using assign_subtract_complex = typename ComplexCompoundHelper<Scalar>::subtract;
+template <typename Scalar>
+using assign_mul_complex = typename ComplexCompoundHelper<Scalar>::mul;
+template <typename Scalar>
+using assign_div_complex = typename ComplexCompoundHelper<Scalar>::div;
```
devshgraphicsprogramming: if you need aliases, then don't have them hanging around globally imho, or we'll drown

include/nbl/builtin/hlsl/math/complex.hlsl
```diff
@@ -113,32 +108,54 @@ struct complex_t
...
-// ---------------------- Compound assignment functors ----------------------
-template <typename Scalar>
-struct ComplexCompoundHelper{              
-    using add = assign_add_t<complex_t<Scalar> >;
-    using subtract = assign_subtract_t<complex_t<Scalar> >;
-    using mul = assign_mul_t<complex_t<Scalar> >;
-    using div = assign_div_t<complex_t<Scalar> >;
+// ---------------------- Compound assign operators ------------------------
+// Specializations of the structs found in functional.hlsl
+// TODO: Figure out a way of making them have an identity (const static member struct doesn't compile on Godbolt)
```
devshgraphicsprogramming, line 121: oh yes, open a DXC issue about this!

devshgraphicsprogramming: aaah its a template, you're blocked by https://github.com/microsoft/hlsl-specs/issues/109

include/nbl/builtin/hlsl/math/complex.hlsl
```diff
@@ -113,32 +108,54 @@ struct complex_t
...
+
+template<typename Scalar> 
+struct plus_assign< complex_t<Scalar> > {
+    void operator()(NBL_REF_ARG(complex_t<Scalar>) lhs, NBL_CONST_REF_ARG(complex_t<Scalar>) rhs) {
+        lhs = lhs + rhs;                                                             
+    }                                          
+};
```
devshgraphicsprogramming, line 128: btw when it finally works, the `plus` would need a specialization instead of `plus_assign`.

`plus_assign` should just alias the identity of the `plus` hence no specialization needed

include/nbl/builtin/hlsl/math/complex.hlsl
```diff
@@ -113,32 +108,54 @@ struct complex_t
...
+template<typename Scalar> 
+struct multiplies_assign< complex_t<Scalar> > {
+    void operator()(NBL_REF_ARG(complex_t<Scalar>) lhs, NBL_CONST_REF_ARG(complex_t<Scalar>) rhs) {
+        lhs = lhs * rhs;                                                             
+    }
+
+    void operator()(NBL_REF_ARG(complex_t<Scalar>) lhs, NBL_CONST_REF_ARG(Scalar) rhs) {
+        lhs = lhs * rhs;                                                             
+    }                                          
+};
```
devshgraphicsprogramming, line 139: Ok I see you need to specialize because HLSL has no conversion operators that actually work 🤦 

However, you should specialize the non `_assign` and make the `_assign` use not the +-/* operators, but the non-assigns themselves

include/nbl/builtin/hlsl/functional.hlsl
```diff
@@ -64,6 +64,36 @@ ALIAS_STD(less_equal,<=) };
...
 #undef ALIAS_STD
 
+// ------------------------ Compound assignment operators ----------------------
+
+#define COMPOUND_ASSIGN(NAME,OP) template<typename T> struct NAME##_assign { \
+    using type_t = T; \
+    \
+    void operator()(NBL_REF_ARG(T) lhs, NBL_CONST_REF_ARG(T) rhs) \
+    { \
+        lhs = lhs OP rhs; \
```
devshgraphicsprogramming, line 74: don't take an OP, use the `NAME` functor from above, 

btw see if making the `rhs` template `U` on both `operator()` would break the rest of our shaders

include/nbl/builtin/hlsl/math/complex.hlsl
devshgraphicsprogramming: you might have to explicitly out-of-line initialize every specialization combo of `complex_t`

include/nbl/builtin/hlsl/math/numbers.hlsl
```diff
@@ -0,0 +1,36 @@
...
+    NBL_CONSTEXPR_STATIC_INLINE float_t inv_pi = float_t(0.3183098861837907);
+    NBL_CONSTEXPR_STATIC_INLINE float_t inv_sqrtpi = float_t(0.5641895835477563);
+    NBL_CONSTEXPR_STATIC_INLINE float_t ln2 = float_t(0.6931471805599453);
+    NBL_CONSTEXPR_STATIC_INLINE float_t ln10 = float_t(2.302585092994046);
+    NBL_CONSTEXPR_STATIC_INLINE float_t sqrt2 = float_t(1.4142135623730951);
+    NBL_CONSTEXPR_STATIC_INLINE float_t sqrt3 = float_t(1.7320508075688772);
+    NBL_CONSTEXPR_STATIC_INLINE float_t inv_sqrt3 = float_t(0.5773502691896257);
+    NBL_CONSTEXPR_STATIC_INLINE float_t egamma = float_t(0.5772156649015329);
+    NBL_CONSTEXPR_STATIC_INLINE float_t phi = float_t(1.618033988749895);               
+};
```
devshgraphicsprogramming, line 28: they're not in a struct in `std`, they're constexpr template variables so define them as
```cpp
namespace numbers
{

template<typename float_t>
NBL_CONSTEXPR float_t e = ...;

etc.
}
```

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -1,34 +1,46 @@
 #ifndef _NBL_BUILTIN_HLSL_FFT_COMMON_INCLUDED_
 #define _NBL_BUILTIN_HLSL_FFT_COMMON_INCLUDED_
 
-#include "nbl/builtin/hlsl/complex.hlsl"
+#include "nbl/builtin/hlsl/math/complex.hlsl"
```
devshgraphicsprogramming, line 4: if something is `<complex>` or `<tgmath>` in STL it shall be `nbl/builtin/hlsl/complex.hlsl` or `nbl/builtin/hlsl/tgmath.hlsl` respectively

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -1,34 +1,46 @@
...
 #include "nbl/builtin/hlsl/cpp_compat.hlsl"
-#include "nbl/builtin/hlsl/fft/common/twiddles.hlsl"
+#include "nbl/builtin/hlsl/math/numbers.hlsl"
```
devshgraphicsprogramming, line 6: same thing for numbers

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -1,34 +1,46 @@
...
+
+    // Computes the kth element in the group of N roots of unity
+    template<typename Scalar, bool inverse>
+    complex_t<Scalar> twiddle(uint32_t k, uint32_t N){
+        complex_t<Scalar> retVal;
+        retVal.real( cos(2.f * math::Numbers<Scalar>::pi * float(k) / N) );
+        if (! inverse)
+            retVal.imag( sin(2.f * math::Numbers<Scalar>::pi * float(k) / N) );
+        else
+            retVal.imag( sin(-2.f * math::Numbers<Scalar>::pi * float(k) / N) );
```
devshgraphicsprogramming, line 22: 1. don't add extra indent because of namespace
2. use the `{` on newline convention
3. hoist the computation of `2.f*math::Numbers<Scalar>::pi/Scalar(N)` into a variable on line 17
4. its not `float(k)` it should be `Scalar(k)` - style nitpick, nothing really will happen

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -0,0 +1,57 @@
...
+#include "nbl/builtin/hlsl/glsl_compat/subgroup_basic.hlsl"
+#include "nbl/builtin/hlsl/glsl_compat/subgroup_shuffle.hlsl"
+
+namespace nbl 
+{
+namespace hlsl
+{
+namespace subgroup
+{
+    // TODO: specialize subgorupShuffleXor to different Scalar types. As of right now this would only support 1-wide scalars (no vectors)
```
devshgraphicsprogramming, line 14: thats fine, lets do 1 channel at a time

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -0,0 +1,57 @@
...
+        lo.imag(shuffledLoHiPacked.y);
+        hi.real(shuffledLoHiPacked.z);
+        hi.imag(shuffledLoHiPacked.w);
+        // Get twiddle with k = subgroupID mod stride, N = 2 * stride
+        fft::DIX<Scalar, inverse>::radix2(fft::twiddle<Scalar, inverse>(glsl::gl_SubgroupID() & (stride - 1), stride << 1), lo, hi);   
+    }
+
+
+    template<typename Scalar, uint32_t SubgroupSize, bool inverse>
+    void FFT(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi) {
```
devshgraphicsprogramming, line 29: you need to document with comments what `lo` and `hi` are

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -0,0 +1,57 @@
...
+        const uint32_t doubleSubgroupSize = SubgroupSize << 1;
+        // special first iteration
+        if (! inverse)
+            fft::DIX<Scalar, inverse>::radix2(fft::twiddle<Scalar, inverse>(glsl::gl_SubgroupID(), doubleSubgroupSize), lo, hi);
```
devshgraphicsprogramming, line 33: using `DIX<Scalar,inverse>` when you're already inside an `if (inverse)` or `if (!inverse)` makes the code super unreadable, I'd use DIT or DIF directly

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -0,0 +1,57 @@
...
+            fft::DIX<Scalar, inverse>::radix2(fft::twiddle<Scalar, inverse>(glsl::gl_SubgroupID(), doubleSubgroupSize), lo, hi);                                                                                   
+        
+        // Decimation in Time
+        if (inverse)
+        for (uint32_t stride = 1; stride < SubgroupSize; stride <<= 1)
+            FFT_loop<Scalar, inverse>(stride, lo, hi);
+        // Decimation in Frequency
+        else
+        for (uint32_t stride = SubgroupSize >> 1; stride > 0; stride >>= 1)
+            FFT_loop<Scalar, inverse>(stride, lo, hi);
```
devshgraphicsprogramming, line 42: also substitute `inverse` if you already know the value

include/nbl/builtin/hlsl/functional.hlsl
Fletterio: It would break the cpp compat, since std::plus, minus etc only template on one parameter

https://github.com/Devsh-Graphics-Programming/Nabla/blob/65d5180d6ad584bd2abb9a087ad21cec18766f2d/include/nbl/builtin/hlsl/functional.hlsl#L17

include/nbl/builtin/hlsl/math/complex.hlsl
Fletterio: I specialize both for complex to support multiplication/division by scalars with less FMAs

Fletterio: Alternatively I can try making the non-assign have a templated operator() on a different type

include/nbl/builtin/hlsl/functional.hlsl
Fletterio: C++ has std::plus<void> with a templated operator(), this would be the closest thing I believe

include/nbl/builtin/hlsl/math/complex.hlsl
devshgraphicsprogramming: if you implement `X_assign` in terms of `X` you only need to specialize the `X`

include/nbl/builtin/hlsl/functional.hlsl
devshgraphicsprogramming: > Best I can do is something like `#define ARITHMETIC_OP(NAME,OP) template<typename LHSType, typename RHSType, typename ReturnType> struct NAME` and have the ones under `ALIAS_STD` be template specializations of this. Same goes for the assign ops
> 
> Reason for this is templating `operator()` seems to not be allowed, and the third template parameter `ReturnType` is needed because we don't have decltype

we do have decltype, instead of `auto` you can use `decltype(declval<RHS>() OP declval<LHS>())` instead of `auto`

Fletterio: Do I need to include anything to access the use of decltype?

devshgraphicsprogramming: type_traits.hlsl probably, its basically a macro, forwards to `__decltype()` which is the only Clang intrinsic that DXC hasn't killed

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,346 @@
...
+
+#define COMPLEX_ARITHMETIC_IDENTITIES(SCALAR) \
+template<> \
+const static complex_t< SCALAR > plus< complex_t< SCALAR > >::identity = { SCALAR (0), SCALAR (0)}; \
+template<> \
+const static complex_t< SCALAR > minus< complex_t< SCALAR > >::identity = { SCALAR (0),  SCALAR (0)}; \
+template<> \
+const static complex_t< SCALAR > multiplies< complex_t< SCALAR > >::identity = { SCALAR (1),  SCALAR (0)}; \
+template<> \
+const static complex_t< SCALAR > divides< complex_t< SCALAR > >::identity = { SCALAR (1),  SCALAR (0)};
```
devshgraphicsprogramming, line 191: you need to use `promote` from to get your 1 to expand to 1,1,... as neeeded

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,346 @@
...
+
+COMPLEX_ARITHMETIC_IDENTITIES(float32_t)  
+COMPLEX_ARITHMETIC_IDENTITIES(float64_t)
```
devshgraphicsprogramming, line 194: you also need `float32_t2`, 3, 4 etc and for `float64_t2`,...

include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -0,0 +1,346 @@
...
+const static complex_t< SCALAR > NAME##_assign< complex_t< SCALAR > >::identity = NAME < complex_t< SCALAR > >::identity;
+
+#define COMPLEX_COMPOUND_ASSIGN_IDENTITIES(SCALAR) \
+COMPLEX_COMPOUND_ASSIGN_IDENTITY(plus, SCALAR) \
+COMPLEX_COMPOUND_ASSIGN_IDENTITY(minus, SCALAR) \
+COMPLEX_COMPOUND_ASSIGN_IDENTITY(multiplies, SCALAR) \
+COMPLEX_COMPOUND_ASSIGN_IDENTITY(divides, SCALAR)
+
+COMPLEX_COMPOUND_ASSIGN_IDENTITIES(float32_t)
+COMPLEX_COMPOUND_ASSIGN_IDENTITIES(float64_t)
```
devshgraphicsprogramming, line 284: moar types!

include/nbl/builtin/hlsl/complex.hlsl
Fletterio: Where is `promote` declared?

PR #614, opened 2023-11-27
Nabla Shader Compile Tool [WIP]

AnastaZIuk:


tools/ndt/main.cpp
```diff
@@ -1 +1,52 @@
...
+    {
+        if (args[i] == "-no-nbl-builtins")
+            return true;
+    }
+    return false;
+}
+
+
+
+int main(int argc, char* argv[])
```
devshgraphicsprogramming, line 26: use ApplicationFramework

tools/ndt/main.cpp
```diff
@@ -3,15 +3,17 @@
...
 #include <cstdlib>
-#include <vector>
+
 
 
 bool no_nbl_builtins;
 
 
-bool noNblBuiltinsEnabled(const std::vector<std::string>& args)
+bool noNblBuiltinsEnabled(const core::vector<std::string>& args)
```
devshgraphicsprogramming, line 16: make it a method and you don't need to pass the `argv` around since its a member

nahiim: OK. I think I'll just remove it and write the loop directly.

devshgraphicsprogramming: anyway argument parsing is @Hazardu's job, you can hardcode all inputs and stuff meant to be gathered from cmdline

PR #621, opened 2023-11-30
Runtime injected code

Przemog1:


include/nbl/builtin/hlsl/subgroup/arithmetic_portability_impl.hlsl
```diff
@@ -22,17 +22,34 @@ namespace hlsl
...
+template<class Binop>
+struct exclusive_scan<Binop, true>;
+
+// portability
+template<class Binop>
+struct reduction<Binop, false>;
+template<class Binop>
+struct inclusive_scan<Binop, false>;
+template<class Binop>
+struct exclusive_scan<Binop, false>;
```
devshgraphicsprogramming, line 48: specializations don't need to be forward declared, this does nothing

PR #630, opened 2023-12-19
IES profile implementation

sunho:
## Description
Implements IES profile based light to ditt branch.

## Testing 
Basic testing done by attaching IES profile to lights in staircase scenes.

## TODO list:
- [x] Clean up git history
- [ ] We can cache some emitter instances


include/nbl/ext/MitsubaLoader/SContext.h
```diff
@@ -103,6 +103,18 @@ namespace MitsubaLoader
...
 
+		static std::string emissionProfileCacheKey(const CElementEmissionProfile& profile)
+		{
+			std::string key = profile.filename;
+			key += "?ies";
+			return key;
+		}
+
+		static auto emissionProfileSamplerParams(const CElementEmissionProfile& profile) {
+			asset::ISampler::SParams res = { asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETBC_INT_OPAQUE_BLACK, asset::ISampler::ETF_NEAREST, asset::ISampler::ETF_NEAREST, asset::ISampler::ESMM_NEAREST, 0u, false, asset::ECO_ALWAYS };
```
devshgraphicsprogramming, line 114: Linear interpolation please

src/nbl/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ option(_NBL_COMPILE_WITH_OPENEXR_LOADER_ "Compile with OpenEXR Loader" ON)
 option(_NBL_COMPILE_WITH_OPENEXR_WRITER_ "Compile with OpenEXR Writer" ON)
 option(_NBL_COMPILE_WITH_GLI_LOADER_ "Compile with GLI Loader" ON)
 option(_NBL_COMPILE_WITH_GLI_WRITER_ "Compile with GLI Writer" ON)
+option(_NBL_COMPILE_WITH_IES_LOADER_ "Compile with IES Loader" ON)
```
devshgraphicsprogramming, line 62: dont even make it an option

include/nbl/builtin/glsl/material_compiler/common_invariant_declarations.glsl
```diff
@@ -80,9 +80,19 @@ void nbl_glsl_MC_finalizeMicrofacet(inout nbl_glsl_MC_microfacet_t mf)
 
 #include <nbl/builtin/glsl/format/decode.glsl>
 
-struct nbl_glsl_MC_oriented_material_t
+#define NBL_GLSL_MC_INVALID_EMITTER_ID 0
```
devshgraphicsprogramming, line 83: use 0xfffffffffu as the invalid ID

include/nbl/asset/material_compiler/IR.h
```diff
@@ -424,11 +432,30 @@ class IR : public core::IReferenceCounted
...
+        CEmitterNode() : INode(ES_EMITTER) {}
+
+        struct EmissionProfile {
+            EmissionProfile() { }
+            operator bool() const {
+                return normalizeEnergy != 0.0f;
+            }
+            STextureSource texture;
+            color_t left;
+            color_t up;
```
devshgraphicsprogramming, line 446: eh this is weird as hell

devshgraphicsprogramming: please use the full translation/matrix node

include/nbl/builtin/glsl/material_compiler/common_invariant_declarations.glsl
```diff
@@ -80,9 +80,19 @@ void nbl_glsl_MC_finalizeMicrofacet(inout nbl_glsl_MC_microfacet_t mf)
...
+
+struct nbl_glsl_MC_emitter_t
 {
 	uvec2 emissive;
+	uvec2 emissionProfile;
+	float normalizeEnergy;
```
devshgraphicsprogramming, line 89: you don't need this, scaling should already be multiplied into `emissive` also you're messing up the struct alignment with padding

include/nbl/builtin/glsl/material_compiler/common.glsl
```diff
@@ -565,6 +565,46 @@ nbl_glsl_MC_eval_pdf_aov_t nbl_glsl_MC_instr_execute_BLEND(
...
+// 	return emissive;
+// 	// if (nbl_glsl_unpackWrapModes(orientedMaterial.texID) == uvec2(3,3)) {
+// 	// 	return nbl_glsl_decodeRGB19E7(orientedMaterial.emissive);
+// 	// }
+// 	// return nbl_glsl_vTextureGrad(orientedMaterial.texID.xy, vec2(0.5), mat2(1.0)).r * vec3(2.0);
+// }
+
+vec2 nbl_glsl_MC_convert_dir_to_ies_uv(vec3 dir) {
+	return vec2((atan(dir.x, dir.y) + nbl_glsl_PI) / (2.0*nbl_glsl_PI), acos(dir.z) / nbl_glsl_PI);
+}
```
devshgraphicsprogramming, line 589: why did you choose this mapping?

devshgraphicsprogramming: also this needs to go to some `nbl/builtin/glsl/ies` just how our envmap sampling and sphere sampling did

devshgraphicsprogramming: you know the GPU hates inverse trig already, use octahedral  instead plz

include/nbl/builtin/glsl/ext/MitsubaLoader/material_compiler_compatibility_impl.glsl
```diff
@@ -39,6 +39,11 @@ layout(set = 0, binding = 6, std430) restrict readonly buffer PREFETCH_INSTR_BUF
     nbl_glsl_MC_prefetch_instr_t data[];
 } prefetch_instr_buf;
 
+layout(set = 0, binding = 7, std430) restrict readonly buffer EMITTER_BUF
+{
+    nbl_glsl_MC_emitter_t data[];
+} emitter_buf;
```
devshgraphicsprogramming, line 45: I think a lot of stuff in Example 22 and the Mistuba loader expects the next "free slot" to be 7, can you show me where you solve that?

include/nbl/ext/MitsubaLoader/CElementEmissionProfile.h
```diff
@@ -0,0 +1,69 @@
...
+	bool addProperty(SNamedPropertyElement&& _property) override;
+	bool onEndTag(asset::IAssetLoader::IAssetLoaderOverride* _override, CMitsubaMetadata* globalMetadata) override {
+		return true;
+	}
+	bool processChildData(IElement* _child, const std::string& name) override;
+	IElement::Type getType() const override { return IElement::Type::EMISSION_PROFILE; }
+	std::string getLogName() const override { return "emissionprofile "; }
+
+	std::string filename;
+	CElementTransform transform;
```
devshgraphicsprogramming, line 61:
```xml
<shape type="sphere">
      <emitter type="area">
           <rgb name="radiance" value="1"/>
           <emissionprofile type="measured">
              <string name="filename" value="test.ies"/>
              <float name="normalizeEnergy" value="12.56636"/>
           </emissionprofile>
           <transform name="to_world">
              <lookat origin="1, 1, 1" target="1, 2, 1" up="0, 0, 1"/>
          </transform>
    </emitter>
</shape>
```
transform is already a member of `CElementEmitter` it has no business being here!

include/nbl/ext/MitsubaLoader/CElementEmissionProfile.h
```diff
@@ -0,0 +1,69 @@
...
+	float normalizeEnergy;
```
devshgraphicsprogramming, line 62: there are 3 modes you should support:
- no normalization (default)
- normalize energy
- normalize Max

you can use 0 to mean no normalization, positive values for energy and negative for max

include/nbl/ext/MitsubaLoader/CElementEmitter.h
```diff
@@ -251,6 +255,12 @@ class CElementEmitter : public IElement
 					animation = anim;
 					return true;
 					break;*/
+				case IElement::Type::EMISSION_PROFILE: {
+					if (type == Type::AREA) {
+						area.emissionProfile = static_cast<CElementEmissionProfile*>(_child);
+					}
+					return true;
```
devshgraphicsprogramming, line 262: we return `true` on success, and `false` on failure, this means that if you don't expect (i.e. not AREA, POINT or SPOT) to have an IES profile, you should return false... so move the `return true` into the if-clause

include/nbl/asset/material_compiler/IR.h
devshgraphicsprogramming: use 2x `float [3]` or some compat mini-`vec3` if we have it used somewhere else

include/nbl/asset/material_compiler/IR.h
```diff
@@ -424,11 +432,30 @@ class IR : public core::IReferenceCounted
...
+            float normalizeEnergy = 0.0f;
```
devshgraphicsprogramming, line 447: shouldn't be here, multiply it into the `intensity` of the parent node

include/nbl/ext/MitsubaLoader/CElementEmitter.h
devshgraphicsprogramming: Also above you have a `toWorld`  handling path, which didn't use to handle AREA, now you need to make it handle AREA since thats what orients the emission profile.

Right now you'll get a parsing error from 
```xml
<shape type="sphere">
      <emitter type="area">
           <rgb name="radiance" value="1"/>
           <emissionprofile type="measured">
              <string name="filename" value="test.ies"/>
              <float name="normalizeEnergy" value="12.56636"/>
           </emissionprofile>
           <transform name="toWorld">
              <lookat origin="1, 1, 1" target="1, 2, 1" up="0, 0, 1"/>
          </transform>
    </emitter>
</shape>
```

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,226 @@
...
+     if (error)
+         return false;
+     if (type_ <= 0 || type_ > 3) {
+         errorMsg = "unrecognized type";
+         return false;
+     }
+     CIESProfile::PhotometricType type =
+         static_cast<CIESProfile::PhotometricType>(type_);
+     assert(type == CIESProfile::PhotometricType::TYPE_C &&
+         "Only type C is supported for now");
```
devshgraphicsprogramming, line 80: log instead of assertingt, asserts only work in Debug and crash the app, return false instead

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,226 @@
...
+  CIESProfileMetadata(double maxIntensity)
+      : IAssetMetadata(), maxIntensity(maxIntensity) {}
+
+  _NBL_STATIC_INLINE_CONSTEXPR const char *LoaderName = "CIESProfileLoader";
+  const char *getLoaderName() const override { return LoaderName; }
+
+  double getMaxIntensity() const { return maxIntensity; }
+
+private:
+  double maxIntensity;
```
devshgraphicsprogramming, line 145: we also need the integral over the sphere, and the "last angle" (90 vs 180 tells me if the profile is full sphere or hemisphere, then we know whether to use octahedral or diamond mapping)

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,226 @@
...
+asset::SAssetBundle
+CIESProfileLoader::loadAsset(io::IReadFile* _file,
+    const asset::IAssetLoader::SAssetLoadParams& _params,
+    asset::IAssetLoader::IAssetLoaderOverride* _override,
+    uint32_t _hierarchyLevel) {
+    if (!_file)
+        return {};
+
+    core::vector<char> data(_file->getSize());
+    _file->read(data.data(), _file->getSize());
```
devshgraphicsprogramming, line 158: try to avoid this copy if the file is mappable

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    CIESProfileParser parser(data.data(), data.size());
+
+    CIESProfile profile;
+    if (!parser.parse(profile))
+        return {};
+
+    auto image = createTexture(profile, TEXTURE_WIDTH, TEXTURE_HEIGHT);
```
devshgraphicsprogramming, line 166: don't make it a fixed size, work out the smallest "gap" in the output mapping between source samples, work out the reciprocal and multiply by 2, that should be your output resolution.

devshgraphicsprogramming: and first and foremost, isotropic IES (rotationally invariant) should output a texture with a height of 1

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,226 @@
...
+    for (size_t i = 0; i < height; i++) {
+        for (size_t j = 0; j < width; j++) {
+            double I = profile.sample(i * vertAngleRate, j * horiAngleRate);
+            uint16_t value = static_cast<uint16_t>(
+                std::clamp(I * maxValueRecip * 65535.0, 0.0, 65535.0));
+            *reinterpret_cast<uint16_t*>(
+                bufferPtr + i * bufferRowLength * texelBytesz + j * texelBytesz) =
+                value;
+        }
+    }
```
devshgraphicsprogramming, line 218: octahedral/diamond mappings please

src/nbl/asset/interchange/CIESProfileLoader.h
```diff
@@ -0,0 +1,160 @@
...
+#include "nbl/asset/interchange/IAssetLoader.h"
+
+#include <sstream>
+
+namespace nbl {
+namespace asset {
+
+class CIESProfile {
+public:
+  enum PhotometricType : uint32_t {
```
devshgraphicsprogramming, line 25: use nested namespaces and extra indendation for `public/private` etc.

src/nbl/asset/interchange/CIESProfileLoader.h
```diff
@@ -0,0 +1,160 @@
...
+// For conditions of distribution and use, see copyright notice in nabla.h
+#ifndef __NBL_ASSET_C_IES_PROFILE_LOADER_H_INCLUDED__
+#define __NBL_ASSET_C_IES_PROFILE_LOADER_H_INCLUDED__
+
+#ifdef _NBL_COMPILE_WITH_IES_LOADER_
+
+#include "nabla.h"
+#include "nbl/asset/ICPUImage.h"
+#include "nbl/asset/ICPUShader.h"
+#include "nbl/video/IGPUShader.h"
```
devshgraphicsprogramming, line 12: `nabla.h` is for external projects, and `asset` namespace is not allowed to include `video`

src/nbl/asset/interchange/CIESProfileLoader.h
```diff
@@ -0,0 +1,160 @@
...
+    error = true;
+    if (!this->errorMsg)
+      this->errorMsg = errorMsg;
+    return -1.0;
+  }
+
+  bool error{false};
+  const char *errorMsg{nullptr};
+  std::stringstream ss;
+};
```
devshgraphicsprogramming, line 94: these are implementation details, hide them in the cpp file

src/nbl/asset/interchange/CIESProfileLoader.h
```diff
@@ -0,0 +1,160 @@
...
+      : IAssetMetadata(), maxIntensity(maxIntensity) {}
+
+  _NBL_STATIC_INLINE_CONSTEXPR const char *LoaderName = "CIESProfileLoader";
+  const char *getLoaderName() const override { return LoaderName; }
+
+  double getMaxIntensity() const { return maxIntensity; }
+
+private:
+  double maxIntensity;
+};
```
devshgraphicsprogramming, line 108: you have this class in the Cpp file as well!?

btw the Metadata classes are supposed to be in public headers, this means `include/nbl/asset/interchange` so that the users of the loader can cast to them and actually use them

src/nbl/asset/interchange/CIESProfileLoader.h
```diff
@@ -0,0 +1,160 @@
...
+  //! Loads an asset from an opened file, returns nullptr in case of failure.
+  asset::SAssetBundle
+      loadAsset(io::IReadFile* _file,
+          const asset::IAssetLoader::SAssetLoadParams& _params,
+          asset::IAssetLoader::IAssetLoaderOverride* _override = nullptr,
+          uint32_t _hierarchyLevel = 0u) override;
+
+private:
+    core::smart_refctd_ptr<asset::ICPUImage>
+        createTexture(const CIESProfile& profile, size_t width, size_t height);
```
devshgraphicsprogramming, line 152: IMHO you should load and return an ICPUImageView, since you know exactly:
- usage (sampled)
- final view format (R16_UNORM)
- the colorspace (identity)
- whether the sampler is meant to be 2D or 2D_ARRAY

which is not something we can say about the other "image loaders" like PNG, JPG, EXR

src/nbl/ext/MitsubaLoader/CElementEmissionProfile.cpp
```diff
@@ -0,0 +1,78 @@
...
+			return false;
+		transform = *tform;
+		return true;
+	}
+	break;
+	default:
+		break;
+	}
+	return false;
+}
```
devshgraphicsprogramming, line 74: transform is supposed to be in the parent node, so this one isn;t supposed to have children

include/nbl/builtin/glsl/ext/MitsubaLoader/material_compiler_compatibility_impl.glsl
devshgraphicsprogramming: ok I see its taken care of in Mitsuba Loadeer cpp

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -710,14 +717,28 @@ SContext::shape_ass_type CMitsubaLoader::loadBasicShape(SContext& ctx, uint32_t
 	{
 		auto bsdf = getBSDFtreeTraversal(ctx, shape->bsdf);
 		core::matrix3x4SIMD tform = core::concatenateBFollowedByA(relTform, shape->getAbsoluteTransform());
+		auto emitter = shape->obtainEmitter();
+		if (emitter.type == CElementEmitter::AREA) {
+			cacheEmissionProfile(ctx, emitter.area.emissionProfile);
+			auto* emissionNode = ctx.ir->allocNode<asset::material_compiler::IR::CEmissionNode>();
+			// TODO cache 
+			auto* emitterNode = ctx.frontend.createEmitterNode(ctx.ir.get(), &emitter, core::matrix4SIMD(tform));
```
devshgraphicsprogramming, line 725: ok this looks good

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -955,6 +976,49 @@ SContext::shape_ass_type CMitsubaLoader::loadBasicShape(SContext& ctx, uint32_t
...
+	{
+		ISampler::SParams samplerParams = ctx.emissionProfileSamplerParams(*profile);
+		const asset::IAsset::E_TYPE types[] = { asset::IAsset::ET_SAMPLER, asset::IAsset::ET_TERMINATING_ZERO };
+		const std::string samplerCacheKey = ctx.samplerCacheKey(samplerParams);
+		if (ctx.override_->findCachedAsset(samplerCacheKey, types, ctx.inner, 0u).getContents().empty())
+		{
+			SAssetBundle samplerBundle(nullptr, { core::make_smart_refctd_ptr<ICPUSampler>(samplerParams) });
+			ctx.override_->insertAssetIntoCache(std::move(samplerBundle), samplerCacheKey, ctx.inner, 0u);
+		}
+	}
```
devshgraphicsprogramming, line 1019: you won't need this if your IES Loader will already return a Image View, you'll only need to deduce a sampler to use from the metadata

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -55,6 +56,46 @@ auto CMitsubaMaterialCompilerFrontend::getTexture(const CElementTexture* _elemen
...
+}
+
+CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend::createEmitterNode(asset::material_compiler::IR* ir, const CElementEmitter* _emitter, core::matrix4SIMD transform) {
+    assert(_emitter->type == CElementEmitter::AREA);
+    auto* res = ir->allocNode<asset::material_compiler::IR::CEmitterNode>();
+    if (_emitter->area.emissionProfile) {
+        asset::material_compiler::IR::CEmitterNode::EmissionProfile profile;
+        asset::material_compiler::IR::INode::STextureSource tex;
+        std::tie(tex.image, tex.sampler, tex.scale) = getEmissionProfile(_emitter->area.emissionProfile);
+        assert(tex.image && "emission profile not loaded");
```
devshgraphicsprogramming, line 87: thou shalt not assert to handle failures, you will log and provide a default.

Also can you assert `tex.scale==1.f`

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -55,6 +56,46 @@ auto CMitsubaMaterialCompilerFrontend::getTexture(const CElementTexture* _elemen
...
+        profile.texture = tex;
+        profile.normalizeEnergy = _emitter->area.emissionProfile->normalizeEnergy;
+        auto inverseTransform = core::transpose(core::concatenateBFollowedByA(transform, _emitter->area.emissionProfile->transform.matrix));
+        profile.left = inverseTransform[0];
+        profile.up = inverseTransform[1];
```
devshgraphicsprogramming, line 92: hmm this is not robust, we need to handle either being LH/RH, non uniform scaling, etc.

Since you can't post-facto anisotropically scale a spherical distribution (maybe you can if you separate out the tensors, but lets not go down a rabbit hole), we require the 3x3 matrix to be orthonormal

I'd extract the Z+ (target axis) first, then build an orthogonal X+ around it and store that, with 1 bit denoting whether Y+ is supposed to be RH or LH relative.

you're missing the orthogonalization step making sure that Z+ and X+ are orthognal and unit length.


hmmm, I'm thinking, maybe we should just store the thing as a quaternion? its compressible into a single RGB21_UNORM because the length encodes the rotation around the axis, and you end up with 1 bit to denote if the Y+ is supposed to be LH or RH w.r.t Z+ and X+.

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -547,9 +589,6 @@ auto CMitsubaMaterialCompilerFrontend::compileToIRTree(asset::material_compiler:
         *dst = ir_node;
     }
 
-    ir->addRootNode(frontroot);
-    ir->addRootNode(backroot);
```
devshgraphicsprogramming, line 551: so who does this now?

include/nbl/builtin/glsl/material_compiler/common.glsl
```diff
@@ -565,6 +565,46 @@ nbl_glsl_MC_eval_pdf_aov_t nbl_glsl_MC_instr_execute_BLEND(
...
+
+vec3 nbl_glsl_MC_oriented_material_t_getEmissive(in nbl_glsl_MC_oriented_material_t orientedMaterial, vec3 dir)
+{
+	if (orientedMaterial.emitter_id == NBL_GLSL_MC_INVALID_EMITTER_ID) {
+		return vec3(0.0);
+	}
+	nbl_glsl_MC_emitter_t emitter = nbl_glsl_MC_fetchEmitterData(orientedMaterial.emitter_id - 1);
```
devshgraphicsprogramming, line 596: using fullbits pattern as  invalid saves you a -1 here

include/nbl/builtin/glsl/material_compiler/common.glsl
devshgraphicsprogramming: I'd take an extra 3 bit enum about the domain size
```cpp
MIRROR_180_BITS = 0b001, // Last Angle is 180, so map V slightly differently
MIRROR_90_BITS = 0b010, // Last Angle is 90, so map both U and V slightly differently
ISOTROPIC_BITS = 0b011, // texture to sample is Nx1, pretend v=middle always  
FULL_THETA_BIT = 0b100, // handle extended domain and rotate by 45 degrees for anisotropic
```

include/nbl/builtin/glsl/material_compiler/common.glsl
```diff
@@ -565,6 +565,46 @@ nbl_glsl_MC_eval_pdf_aov_t nbl_glsl_MC_instr_execute_BLEND(
...
+	vec3 emissive = nbl_glsl_decodeRGB19E7(emitter.emissive);
+#if _NBL_VT_FLOAT_VIEWS_COUNT
+	if (emitter.normalizeEnergy != 0.0f) {
+		vec3 left = vec3(emitter.orientation[0],emitter.orientation[1],emitter.orientation[2]);
+		vec3 up = vec3(emitter.orientation[3],emitter.orientation[4],emitter.orientation[5]);
+		vec3 view = cross(left, up);
+		return emitter.normalizeEnergy * emissive * nbl_glsl_vTextureGrad(emitter.emissionProfile, nbl_glsl_MC_convert_dir_to_ies_uv(normalize(mat3(left, up, view)*dir)), mat2(0.0)).r;
```
devshgraphicsprogramming, line 603: don't use normalizeEnergy at all, to check for null profile you can make a GLSL function in the VT headers to check orig-size = (0,0) on the VT-handle

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
```diff
@@ -22,15 +22,32 @@ namespace nbl::asset::material_compiler
...
 struct nbl_glsl_MC_oriented_material_t
 {
-    uvec2 emissive;
+	uint emitter_id; // 0 = invalid
     uint prefetch_offset;
     uint prefetch_count;
     uint instr_offset;
     uint rem_pdf_count;
     uint nprecomp_count;
     uint genchoice_count;
```
devshgraphicsprogramming, line 40: maybe pull in the GLSL structs (so the fields are in sync and don't repeat) and use inheritance to extend them with extra methods

include/nbl/ext/MitsubaLoader/CMitsubaLoader.h
```diff
@@ -73,8 +73,11 @@ class CMitsubaLoader : public asset::IRenderpassIndependentPipelineLoader
 		core::vector<SContext::shape_ass_type>	getMesh(SContext& ctx, uint32_t hierarchyLevel, CElementShape* shape);
 		core::vector<SContext::shape_ass_type>	loadShapeGroup(SContext& ctx, uint32_t hierarchyLevel, const CElementShape::ShapeGroup* shapegroup, const core::matrix3x4SIMD& relTform);
 		SContext::shape_ass_type				loadBasicShape(SContext& ctx, uint32_t hierarchyLevel, CElementShape* shape, const core::matrix3x4SIMD& relTform);
-		
+
 		void									cacheTexture(SContext& ctx, uint32_t hierarchyLevel, const CElementTexture* texture, const CMitsubaMaterialCompilerFrontend::E_IMAGE_VIEW_SEMANTIC semantic);
+		void									cacheBitmap(SContext& ctx, uint32_t hierarchyLevel, const CElementTexture::Bitmap* texture, const CMitsubaMaterialCompilerFrontend::E_IMAGE_VIEW_SEMANTIC semantic);
```
devshgraphicsprogramming, line 78: `cacheBitmap` looks new, what is it for?

src/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.cpp
```diff
@@ -1184,8 +1200,23 @@ auto CMaterialCompilerGLSLBackendCommon::compile(SContext* _ctx, IR* _ir, E_GENE
 	res.usedRegisterCount = 0u;
 	res.globalPrefetchRegCountFlags = 0u;
 
-	for (const IR::INode* root : _ir->roots)
-	{
+	core::unordered_map<IR::CEmitterNode*, uint32_t> emitterCache;
```
devshgraphicsprogramming, line 1203: you're hashing pointers to emitter nodes, not their actual values, I'd be surprised if this cache worked at all

src/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.cpp
```diff
@@ -1184,8 +1200,23 @@ auto CMaterialCompilerGLSLBackendCommon::compile(SContext* _ctx, IR* _ir, E_GENE
...
+	auto getOrCreateEmitter = [&](IR::CEmitterNode* node) -> uint32_t {
+		auto it = emitterCache.find(node);
+		if (it == emitterCache.end()) {
+			auto emitter = lowerEmitter(_ctx, node);
+			it = emitterCache.emplace(node, res.emitterData.size() + 1).first;
```
devshgraphicsprogramming, line 1208: use a different inbvalid pattern than 0, and you'll save a +1 here on encode

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
devshgraphicsprogramming: actually quat probably not the best choice since I'll want OBB MIS later, do what you want here, just make sure Z+ and X+ are orho and you can handle Y+ needing flip depending on LH vs RH (which you can infer from the determinant of `invereseTransform`

src/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.cpp
devshgraphicsprogramming: anyway since the emitter is supposed to be a root node with the whole BSDF as the child, can it be cached at all?

src/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.cpp
```diff
@@ -1303,29 +1334,48 @@ auto CMaterialCompilerGLSLBackendCommon::compile(SContext* _ctx, IR* _ir, E_GENE
...
+			auto* node = static_cast<const IR::CEmissionNode*>(root);
+			material = compileBSDFRootNode(node->children[0]);
+			material.emitter_id = getOrCreateEmitter(node->emitter);
+		}
+		else {
+			material = compileBSDFRootNode(root);
+			material.emitter_id = 0;
+		}
+
+		printf("asdfasf %d\n", material.emitter_id);
```
devshgraphicsprogramming, line 1377: we have a dedicated log for logging out the material compiler stuff, ex 22 produces a log.txt everytime with a summary of everything

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
sunho: We probably have to orthogonalize matrix as well if transform of emitter object does weird thing. I don't see orthogonalize function in core tho.

include/nbl/builtin/glsl/ies/functions.glsl
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2018-2024 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+#ifndef _NBL_GLSL_IES_FUNCTIONS_INCLUDED_
+#define _NBL_GLSL_IES_FUNCTIONS_INCLUDED_
+
+#include <nbl/builtin/glsl/math/constants.glsl>
+
+vec2 nbl_glsl_IES_convert_dir_to_uv(vec3 dir) {
```
devshgraphicsprogramming, line 9: write in the comments what I actually asked for as a TODO

include/nbl/ext/MitsubaLoader/SContext.h
```diff
@@ -110,8 +111,8 @@ namespace MitsubaLoader
 			return key;
 		}
 
-		static auto emissionProfileSamplerParams(const CElementEmissionProfile& profile) {
-			asset::ISampler::SParams res = { asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETBC_INT_OPAQUE_BLACK, asset::ISampler::ETF_NEAREST, asset::ISampler::ETF_NEAREST, asset::ISampler::ESMM_NEAREST, 0u, false, asset::ECO_ALWAYS };
+		static auto emissionProfileSamplerParams(const CElementEmissionProfile& profile, const asset::CIESProfileMetadata& meta) {
+			asset::ISampler::SParams res = { asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETC_CLAMP_TO_EDGE, asset::ISampler::ETBC_INT_OPAQUE_BLACK, asset::ISampler::ETF_LINEAR, asset::ISampler::ETF_LINEAR, asset::ISampler::ETF_LINEAR, 0u, false, asset::ECO_ALWAYS };
```
devshgraphicsprogramming, line 115: if you want the octahedral mapping to work right, you need REPEAT edge clamping on U and V

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
```diff
@@ -26,12 +26,14 @@ struct nbl_glsl_MC_emitter_t
 {
 	uvec2 emissive;
 	uvec2 emissionProfile;
-	float normalizeEnergy;
-	float orientation[6]; // not vector because of alignment, [0:3] left [3:6] up
+	float orientation[6]; // LSB of orientation[0] is 1 if right handed; [0:3] y axis [3:6] z axis
```
devshgraphicsprogramming, line 29: so [0,3] is y-axis, why did you call it "left" before?

include/nbl/builtin/glsl/material_compiler/common.glsl
```diff
@@ -584,23 +585,23 @@ nbl_glsl_MC_eval_pdf_aov_t nbl_glsl_MC_instr_execute_BLEND(
...
 #if _NBL_VT_FLOAT_VIEWS_COUNT
-	if (emitter.normalizeEnergy != 0.0f) {
-		vec3 left = vec3(emitter.orientation[0],emitter.orientation[1],emitter.orientation[2]);
-		vec3 up = vec3(emitter.orientation[3],emitter.orientation[4],emitter.orientation[5]);
-		vec3 view = cross(left, up);
-		return emitter.normalizeEnergy * emissive * nbl_glsl_vTextureGrad(emitter.emissionProfile, nbl_glsl_MC_convert_dir_to_ies_uv(normalize(mat3(left, up, view)*dir)), mat2(0.0)).r;
+	if (nbl_glsl_isValidVirtualTex(emitter.emissionProfile)) {
+		vec3 up = vec3(emitter.orientation[0],emitter.orientation[1],emitter.orientation[2]);
+		vec3 view = vec3(emitter.orientation[3],emitter.orientation[4],emitter.orientation[5]);
+		vec3 right = cross(up, view);
```
devshgraphicsprogramming, line 600: getting a headache here, you called [0,3) "left", now in the comments its the "y-axis" but you call it "up" here
I can kinda  understand if `view` means `target` then Z+ kinda makes sense to be called that, but you comment called it "up"

Now I can see that if they are now y+ and Z+, then cross producting z+ and y+ (in that order) makes the result x-, and are you sure you want that?

src/nbl/asset/interchange/CIESProfileLoader.h
devshgraphicsprogramming: not really

src/nbl/asset/interchange/CIESProfileLoader.cpp
devshgraphicsprogramming: > and first and foremost, isotropic IES (rotationally invariant - last angle 0) should output a texture with a height of 1

ok I can let off on that

include/nbl/builtin/glsl/ies/functions.glsl
devshgraphicsprogramming: basically this as a comment https://github.com/Devsh-Graphics-Programming/Nabla/pull/630#discussion_r1431395101

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -57,42 +57,54 @@ auto CMitsubaMaterialCompilerFrontend::getTexture(const CElementTexture* _elemen
...
+
+        auto inverseTransform = core::transpose(core::concatenateBFollowedByA(transform, _emitter->transform.matrix));
+        profile.right_hand = core::determinant(inverseTransform) >= 0.0f;
+        profile.up = core::normalize(inverseTransform[1]);
+        profile.view = core::normalize(inverseTransform[2]);
+
         res->emissionProfile = profile;
+        float normalizeEnergy = _emitter->area.emissionProfile->normalizeEnergy;
+        if (normalizeEnergy < 0.0f) {
+            res->intensity *= meta->getMaxIntensity();
```
devshgraphicsprogramming, line 102: normalization implies division though?

devshgraphicsprogramming:
btw Example 22 Readme.md needs touching to show an example of how to use the IES

src/nbl/asset/interchange/CIESProfileLoader.cpp
devshgraphicsprogramming: I still want the `lastAngle` info, but as a boolean `bool fullSphere; // or only a hemisphere?`

src/nbl/asset/interchange/CIESProfileLoader.cpp
devshgraphicsprogramming: no, check if `IFileBase::getMappedPointer()` returns null or not

include/nbl/builtin/glsl/material_compiler/common.glsl
```diff
@@ -584,23 +585,23 @@ nbl_glsl_MC_eval_pdf_aov_t nbl_glsl_MC_instr_execute_BLEND(
...
+		if ((floatBitsToInt(emitter.orientation[0])&1u) != 1u) {
+			right *= -1;
+		}
+		return emissive * nbl_glsl_vTextureGrad(emitter.emissionProfile, nbl_glsl_IES_convert_dir_to_uv(normalize(mat3(right, up, view)*dir)), mat2(0.0)).r;
```
devshgraphicsprogramming, line 604: Pro-Tip, if the `dir` is already normalized going into the function (it most likely is) and the matrix is orthonormal (it really should be), there's no need to call `normalize` here

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,352 @@
...
+        return {};
+
+    core::vector<char> data(_file->getSize());
+    _file->read(data.data(), _file->getSize());
+
+    CIESProfileParser parser(data.data(), data.size());
+
+    CIESProfile profile;
+    if (!parser.parse(profile))
+        return {};
```
devshgraphicsprogramming, line 260: you set the errorMsg on the parser, but you never actually print it, XD

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,352 @@
...
+    const char* errorMsg{ nullptr };
+    std::stringstream ss;
+};
+
+double CIESProfile::sample(double vAngle, double hAngle) const {
+    vAngle = fmod(vAngle, vAngles.back()); // warp angle
+    hAngle = hAngles.back() == 0.0
+                 ? 0.0
+                 : fmod(hAngle,
+                        hAngles.back()); // when last horizontal angle is zero
```
devshgraphicsprogramming, line 94: `fmod` gives you weird warping for vertical angles, I'd rather clamp when the range of the values isn't 0,2PI

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,352 @@
...
+    double dTheta = core::PI<double>() / numVSamples;
+    double dPhi = 2 * core::PI<double>() / numHSamples;
+    double dThetaInAngle = MAX_VANGLE / numVSamples;
+    double dPhiInAngle = MAX_HANGLE / numHSamples;
+    double res = 0;
+    for (size_t i = 0; i < numVSamples; i++) {
+        for (size_t j = 0; j < numHSamples; j++) {
+            res += dPhi * dTheta * sin(dTheta * i) * sample(i* dThetaInAngle,j*dPhiInAngle);
+        }
+    }
```
devshgraphicsprogramming, line 134: I'd compute it while parsing it in, work out the approximate solid angle of the sample from the spacing between itself and the ones befor/after to form a spherical square around it, then work out solid angle of the square (you've already implemented functions for it in GLSL looong time ago), multiply the sample value by that solid angle

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -980,38 +968,31 @@ void CMitsubaLoader::cacheEmissionProfile(SContext& ctx, const CElementEmissionP
...
-		asset::SAssetBundle viewBundle(nullptr, { ICPUImageView::create(std::move(viewParams)) });
-		ctx.override_->insertAssetIntoCache(std::move(viewBundle), cacheKey, ctx.inner, 0u);
-	}
 	
-	{
-		ISampler::SParams samplerParams = ctx.emissionProfileSamplerParams(*profile);
-		const asset::IAsset::E_TYPE types[] = { asset::IAsset::ET_SAMPLER, asset::IAsset::ET_TERMINATING_ZERO };
+		auto meta = core::make_smart_refctd_ptr<asset::CIESProfileMetadata>(*static_cast<const asset::CIESProfileMetadata*>(assetLoaded.getMetadata()));
+		asset::SAssetBundle viewBundle(std::move(meta), {assetLoaded.getContents().begin()[0]});
+		ctx.override_->insertAssetIntoCache(std::move(viewBundle), cacheKey, ctx.inner, 0u);
```
devshgraphicsprogramming, line 990: this makes no sense, you've just asked for an asset to be loaded using `interm_getAssetInHierarchy` its already inserted into the cache, the whole bundle!

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -980,38 +968,31 @@ void CMitsubaLoader::cacheEmissionProfile(SContext& ctx, const CElementEmissionP
...
+
+		ISampler::SParams samplerParams = ctx.emissionProfileSamplerParams(*profile, reinterpret_cast<const asset::CIESProfileMetadata&>(*assetLoaded.getMetadata()));
+		const asset::IAsset::E_TYPE types2[] = { asset::IAsset::ET_SAMPLER, asset::IAsset::ET_TERMINATING_ZERO };
 		const std::string samplerCacheKey = ctx.samplerCacheKey(samplerParams);
-		if (ctx.override_->findCachedAsset(samplerCacheKey, types, ctx.inner, 0u).getContents().empty())
+		if (ctx.override_->findCachedAsset(samplerCacheKey, types2, ctx.inner, 0u).getContents().empty())
 		{
 			SAssetBundle samplerBundle(nullptr, { core::make_smart_refctd_ptr<ICPUSampler>(samplerParams) });
 			ctx.override_->insertAssetIntoCache(std::move(samplerBundle), samplerCacheKey, ctx.inner, 0u);
```
devshgraphicsprogramming, line 998: this should be in the external scope, not depending on whether you find the profile in the cache the first time round

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -980,38 +968,31 @@ void CMitsubaLoader::cacheEmissionProfile(SContext& ctx, const CElementEmissionP
...
 	if (ctx.override_->findCachedAsset(cacheKey, types, ctx.inner, 0u).getContents().empty())
 	{
 		auto assetLoaded = interm_getAssetInHierarchy(m_assetMgr, profile->filename, ctx.inner.params, 0u, ctx.override_);
 
-		ICPUImageView::SCreationParams viewParams = {};
+	
 		auto contentRange = assetLoaded.getContents();
 		if (contentRange.empty())
 		{
 			os::Printer::log("[ERROR] Could Not Find Texture: " + cacheKey, ELL_ERROR);
```
devshgraphicsprogramming, line 984: not texture, Emission Profile, more useful message for me

include/nbl/ext/MitsubaLoader/SContext.h
```diff
@@ -110,8 +111,8 @@ namespace MitsubaLoader
 			return key;
 		}
```
devshgraphicsprogramming, line 112: this function kinda makes no sense to exist

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
devshgraphicsprogramming: ok so you're doing all these gymnastics to "double cache" `something.ies` as `something.ies?ies`

just stop it https://github.com/Devsh-Graphics-Programming/Nabla/pull/630/files/faf4f805df55fe3af3e1a80188494c324e015919..cb61c15d7f6695c3e22f3270d9173d72f401326e#r1456676288

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -980,38 +968,31 @@ void CMitsubaLoader::cacheEmissionProfile(SContext& ctx, const CElementEmissionP
...
 		return;
-	
+
 	const auto cacheKey = ctx.emissionProfileCacheKey(*profile);
 
 	const asset::IAsset::E_TYPE types[]{ asset::IAsset::ET_IMAGE_VIEW, asset::IAsset::ET_TERMINATING_ZERO };
+
 	if (ctx.override_->findCachedAsset(cacheKey, types, ctx.inner, 0u).getContents().empty())
 	{
 		auto assetLoaded = interm_getAssetInHierarchy(m_assetMgr, profile->filename, ctx.inner.params, 0u, ctx.override_);
```
devshgraphicsprogramming, line 978: due to not needing to cache under a special key: https://github.com/Devsh-Graphics-Programming/Nabla/pull/630/files/faf4f805df55fe3af3e1a80188494c324e015919..cb61c15d7f6695c3e22f3270d9173d72f401326e#r1456673824

just use the `interm_getAssetInHierarchy` with the filename, only do special caching for the sampler cause you create that on-demand

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -57,42 +57,54 @@ auto CMitsubaMaterialCompilerFrontend::getTexture(const CElementTexture* _elemen
...
-        assert(tex.image && "emission profile not loaded");
-        profile.texture = tex;
-        profile.normalizeEnergy = _emitter->area.emissionProfile->normalizeEnergy;
-        auto inverseTransform = core::transpose(core::concatenateBFollowedByA(transform, _emitter->area.emissionProfile->transform.matrix));
-        profile.left = inverseTransform[0];
-        profile.up = inverseTransform[1];
+        auto [image, sampler, meta] = getEmissionProfile(_emitter->area.emissionProfile);
+        if (!image) {
+            os::Printer::log("ERROR: Emission profile not loaded", ELL_ERROR);
+            return nullptr;
```
devshgraphicsprogramming, line 90: that doesn't provide a default, it actually murders my emitter completely, please fallback to produce the same result as if `_emitter->area.emissionProfile` was null

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
devshgraphicsprogramming: > We probably have to orthogonalize matrix as well if transform of emitter object does weird thing. I don't see orthogonalize function in core tho.

there isn't one, I'd check that the `inverseTransform` determinant is sufficiently far away from 0, else turn off the emission profile and log a warning/error

then use the scale extraction method of `core::matrix3x4SIMD`, if the scale isn't sufficiently uniform (don't do exact comparisons please cause fp32) for each axis, output a warning and ortho-normalize everything around the `view` (the straight up angle in the IES) and then the `hAngle=0` direction

src/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.cpp
```diff
@@ -1363,18 +1364,19 @@ auto CMaterialCompilerGLSLBackendCommon::compile(SContext* _ctx, IR* _ir, E_GENE
 
 	for (const IR::INode* root : _ir->roots)
 	{
+		if (root->symbol != IR::INode::ES_ROOT) {
+			os::Printer::log("Material compiler: Non root node added as root; ignoring it", ELL_WARNING);
```
devshgraphicsprogramming, line 1368: technically its an error

src/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.cpp
devshgraphicsprogramming: please print to my debug log

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
sunho: it was x axis and y axis before but I changed it cause you asked one of them to be z.

include/nbl/builtin/glsl/material_compiler/common.glsl
sunho: yeah must be in reverse order.

include/nbl/builtin/glsl/material_compiler/common_invariant_declarations.glsl
```diff
@@ -80,9 +80,18 @@ void nbl_glsl_MC_finalizeMicrofacet(inout nbl_glsl_MC_microfacet_t mf)
...
 #include <nbl/builtin/glsl/format/decode.glsl>
 
-struct nbl_glsl_MC_oriented_material_t
+#define NBL_GLSL_MC_INVALID_EMITTER_ID 0xffffffffu
+
+struct nbl_glsl_MC_emitter_t
 {
 	uvec2 emissive;
+	uvec2 emissionProfile;
+	float orientation[6]; // [0:3] left vector [3:6] up
```
sunho, line 89: this is outdated

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
sunho: this path wouldnt be run if emisison profile is null right?

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
sunho: wasn't that what you wanted?

sunho: This is the "max mode" implementation. Normalization by energy is done in the next line.

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
sunho: oh nvm I misread your comment. I can just make this warning and keep going.

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
sunho: I tried to use interm_getAssetInHierarchy inside the IR frontend but I couldn't access m_asserManager from there. Was that intended design or can interm_getAssetInHierarchy be accessed other way?

src/nbl/asset/interchange/CIESProfileLoader.cpp
sunho: vAngles.back() can be 90 degrees or 180 degrees. Warping is explicitly needed here. I can clamp vAngle to 0,2pi though.

devshgraphicsprogramming: > vAngles.back() can be 90 degrees or 180 degrees. Warping is explicitly needed here. I can clamp vAngle to 0,2pi though.

If you're over 180 you're wrapping around over the south pole of the sphere, and then you end up flipping your horizontal angle.

Also for 90 degs, fmod will make you wrap around to a higher angle so the value at 0 will repeat just past, 90.

You probably want MIRROR wrapping here, not REPEAT

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
devshgraphicsprogramming: `interm_getAssetInHierarchy` is only accesible and usable by other IAssetLoaders (friendship)

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
devshgraphicsprogramming: yep, and uniform profile (no profile) but still having an emitter

include/nbl/asset/material_compiler/CMaterialCompilerGLSLBackendCommon.h
devshgraphicsprogramming: ok. I remember, cause isotropic gives no fs about X and Y, only Z

src/nbl/asset/interchange/CIESProfileLoader.cpp
sunho: Maybe it's not available in ditt branch?

devshgraphicsprogramming: ah yeah its not! 

Forgot how old it was!

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -0,0 +1,352 @@
...
+    double vertInv = 1.0 / height;
+    double horiInv = 1.0 / width;
+    char* bufferPtr = reinterpret_cast<char*>(buffer->getPointer());
+    for (size_t i = 0; i < height; i++) {
+        for (size_t j = 0; j < width; j++) {
+            auto dir = octahdronUVToDir(j * vertInv, i * horiInv);
+            auto [theta, phi] = sphericalDirToAngles(dir);
+            double I = profile.sample(core::degrees(theta), core::degrees(phi));
+            uint16_t value = static_cast<uint16_t>(
+                std::clamp(I * maxValueRecip * 65535.0, 0.0, 65535.0));
```
devshgraphicsprogramming, line 339: remember to add +0.5 so that [0,0.5) ->0 and [0.5,1.5) -> 1

very important so you can actually have a Max Value thats reachable

I just fixed a bug resulting from such quanitzation
https://github.com/Devsh-Graphics-Programming/Nabla/commit/9804926db0c7809958c84496be3137848a3642a0

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
devshgraphicsprogramming: ok I've had to reheat this a bit in my brain cause I lost the thread a bit.

> there are 3 modes you should support:
> 
>     * no normalization (default)
> 
>     * normalize energy
> 
>     * normalize Max
> 
> 
> you can use 0 to mean no normalization, positive values for energy and any negative for max

but you load the IES as a UNORM which needs to be multiplied with the max value so that you recover the original values.

So:
no normalization [zero] => `intensity * meta->getMaxIntensity()`
normalization by Max [negative] => `intensity` (leave alone, don't do anything
normalization by Energy => `intensity * meta->getMaxIntensity()/meta->getIntegral()`

src/nbl/asset/interchange/CIESProfileLoader.cpp
devshgraphicsprogramming: fixed in #660 it was actually CLAMP_TO_BORDER with border being black thats correct

PR #632, opened 2023-12-21
Add concepts.hlsl to cpp_compat.

theoreticalphysicsftw:
## Description
Add a basic `concepts.hlsl` header to work with C++20 concepts when the hlsl builtins are being included in C++20 mode.

## Testing 
MSVC++ compilation

## TODO list:
More testing with code using it.

PR #633, opened 2023-12-21
Refactor common BxDF header.

theoreticalphysicsftw:
## Description
Refactor the common BxDF structs. This depends on #631 and #632 .

## Testing 
Just raw compilation. This will be properly tested when I'm done with the BxDF tests.

## TODO list:
Port the rest of the headers and test how this works out with them.


devshgraphicsprogramming:
#811 does this

PR #648, opened 2024-01-18
Nabla Shader Compiler Tool

AnastaZIuk:


src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -192,12 +265,22 @@ std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADE
...
+
+
     return resolvedString;
 }
 
+std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADER_STAGE& stage, const SPreprocessorOptions& preprocessOptions) const
+{
+    std::vector<std::string> extra_dxc_compile_flags = {};
+    return preprocessShader(std::move(code), stage, extra_dxc_compile_flags, preprocessOptions);
+}
```
devshgraphicsprogramming, line 283: why notprovide a deault arg for the extra compile flags and have it last, then no extra wrapper fun neeeded.

src/nbl/asset/utils/waveContext.h
```diff
@@ -64,6 +68,7 @@ struct preprocessing_hooks final : public boost::wave::context_policies::default
         typename ContextT::token_type const& act_token
     )
     {
+        hash_token_occurences++;
```
devshgraphicsprogramming, line 71: cool but you also need to count them on includes, ifdefs, etc.

src/nbl/asset/utils/waveContext.h
```diff
@@ -97,6 +102,34 @@ struct preprocessing_hooks final : public boost::wave::context_policies::default
             m_pragmaStage = found->second;
             return true;
         }
+        
+        if (strcmp(optionStr, "dxc_compile_flags") == 0 && hash_token_occurences == 1) {
```
devshgraphicsprogramming, line 106: you need to spam me with an ERROR or WARNING if you encounter `dxc_compile_flags` and the hast token occurence counter isn't 1,  instead of silently skipping

src/nbl/asset/utils/waveContext.h
```diff
@@ -454,6 +490,9 @@ template<> inline bool boost::wave::impl::pp_iterator_functor<nbl::wave::context
         else
             result = includeFinder->getIncludeRelative(ctx.get_current_directory(),file_path);
     }
+    else {
+        ctx.get_hooks().m_logger.log("Include finder not assigned, preprocessor will not include file " + file_path, nbl::system::ILogger::ELL_ERROR);
```
devshgraphicsprogramming, line 494: thats a straight up error, you need to throw and return false

include/nbl/asset/utils/CHLSLCompiler.h
```diff
@@ -47,18 +51,30 @@ class NBL_API2 CHLSLCompiler final : public IShaderCompiler
...
+			L"-spirv",
+			L"-Zpr",
+			L"-enable-16bit-types",
+			L"-fvk-use-scalar-layout",
+			L"-Wno-c++11-extensions",
+			L"-Wno-c++1z-extensions",
+			L"-Wno-gnu-static-float-init",
+			L"-fspv-target-env=vulkan1.3"
+		};
+		constexpr static inline uint32_t RequiredArgumentCount = 8;
```
devshgraphicsprogramming, line 67: sizeof() on sized arrays exists and returns their length

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,19 +71,71 @@ CHLSLCompiler::~CHLSLCompiler()
...
+static void try_upgrade_hlsl_version(std::vector<std::wstring>& arguments) 
 {
-    ComPtr<IDxcBlobEncoding> errorMessages;
-    ComPtr<IDxcBlob> objectBlob;
-    ComPtr<IDxcResult> compileResult;
+    auto foundShaderStageArgument = std::find(arguments.begin(), arguments.end(), L"-HV");
+    if (foundShaderStageArgument != arguments.end()) {
+        auto foundShaderStageArgumentValueIdx = foundShaderStageArgument - arguments.begin() + 1;
+        std::wstring version = arguments[foundShaderStageArgumentValueIdx];
+        if (version.length() >= 4 && (version[2] < '2' || (version[2] == '2' && version[3] <= '0')))
```
devshgraphicsprogramming, line 80: use `atoi` or some other string2int function then compare against being less than `2021`

(also check for last [back()] char being a letter instead of a number, then ignore the check)

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,19 +71,71 @@ CHLSLCompiler::~CHLSLCompiler()
...
-    std::string GetErrorMessagesString()
+
+static void try_upgrade_shader_stage(std::vector<std::wstring> &arguments) {
+    auto foundShaderStageArgument = std::find(arguments.begin(), arguments.end(), L"-T");
+    if (foundShaderStageArgument != arguments.end()) {
+        auto foundShaderStageArgumentValueIdx = foundShaderStageArgument - arguments.begin() + 1;
+        std::wstring targetProfile = arguments[foundShaderStageArgumentValueIdx];
+        if (targetProfile.length() >= 6) {
+            int argument_version = (targetProfile[3] - '0') * 10 + (targetProfile[5] - '0');
+            if (argument_version < 67)
```
devshgraphicsprogramming, line 93: don't check length, find the SECOND `_` from the back, take the suffix, replace the last `_` with a `.`, then parse as two ints (version), finally do your upgrades and re-assembly

the way you wrote it now will break if we ever get SM 6.10 (they're shipping 6.8 now)

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,19 +71,71 @@ CHLSLCompiler::~CHLSLCompiler()
...
+    for (int i = 0; i < argc; i++)
     {
-        return std::string(reinterpret_cast<char*>(errorMessages->GetBufferPointer()), errorMessages->GetBufferSize());
+        for (int j = 0; j < CHLSLCompiler::RequiredArgumentCount; j++)
+        {
+            if (arguments[i] == CHLSLCompiler::RequiredArguments[j]) {
+                found_arg_flags[j] = true;
+                break;
+            }
+        }
```
devshgraphicsprogramming, line 114: use `std::find_if` or declare an `static map<std::wstring_view>` initialized from the constexpr in the function

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -208,9 +280,9 @@ core::smart_refctd_ptr<ICPUShader> CHLSLCompiler::compileToSPIRV(const char* cod
         hlslOptions.preprocessorOptions.logger.log("code is nullptr", system::ILogger::ELL_ERROR);
         return nullptr;
     }
-
+    std::vector<std::string> dxc_compile_flags = {};
```
devshgraphicsprogramming, line 283: you could just initialize them to `hlslOptions.dxcFlags` thing or whatever that span member is called

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -255,49 +327,59 @@ core::smart_refctd_ptr<ICPUShader> CHLSLCompiler::compileToSPIRV(const char* cod
...
-
-    std::vector<LPCWSTR> arguments = {
-        L"-spirv",
+    
+    std::vector<std::wstring> arguments = {};
+    if (dxc_compile_flags.size()) { // #pragma dxc_compile_flags takes priority
+        populate_arguments_with_type_conversion(arguments, dxc_compile_flags, hlslOptions.preprocessorOptions.logger);
+    }
+    else if (hlslOptions.dxcOptions.size()) { // second in order of priority is command line arguments
+        populate_arguments_with_type_conversion(arguments, hlslOptions.dxcOptions, hlslOptions.preprocessorOptions.logger);
```
devshgraphicsprogramming, line 336: you could unify both branches if you only initialized `dxc_compile_flags` to `hlslOptions.dxcOptions`

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,19 +71,71 @@ CHLSLCompiler::~CHLSLCompiler()
...
+        std::wstring targetProfile = arguments[foundShaderStageArgumentValueIdx];
+        if (targetProfile.length() >= 6) {
+            int argument_version = (targetProfile[3] - '0') * 10 + (targetProfile[5] - '0');
+            if (argument_version < 67)
+            {
+                targetProfile.replace(3, 3, L"6_7");
+                arguments[foundShaderStageArgumentValueIdx] = targetProfile;
+            }
+        }
+    }
```
devshgraphicsprogramming, line 99: you also need to handle the case when there's no `-T` arg at all

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,19 +71,71 @@ CHLSLCompiler::~CHLSLCompiler()
...
-    ComPtr<IDxcBlobEncoding> errorMessages;
-    ComPtr<IDxcBlob> objectBlob;
-    ComPtr<IDxcResult> compileResult;
+    auto foundShaderStageArgument = std::find(arguments.begin(), arguments.end(), L"-HV");
+    if (foundShaderStageArgument != arguments.end()) {
+        auto foundShaderStageArgumentValueIdx = foundShaderStageArgument - arguments.begin() + 1;
+        std::wstring version = arguments[foundShaderStageArgumentValueIdx];
+        if (version.length() >= 4 && (version[2] < '2' || (version[2] == '2' && version[3] <= '0'))) 
+            arguments[foundShaderStageArgumentValueIdx] = L"2021";
+    }
```
devshgraphicsprogramming, line 82: you also need to handle the case when there's no `-HV` arg at all

tools/nsc/main.cpp
```diff
@@ -0,0 +1,155 @@
...
+	bool keepRunning() override { return false; }
+
+
+private:
+
+	core::smart_refctd_ptr<ICPUShader> compile_shader(std::string& shader_code, std::string_view sourceIdentifier) {
+		constexpr uint32_t WorkgroupSize = 256;
+		constexpr uint32_t WorkgroupCount = 2048;
+		const string WorkgroupSizeAsStr = std::to_string(WorkgroupSize);
+		const IShaderCompiler::SPreprocessorOptions::SMacroDefinition WorkgroupSizeDefine = { "WORKGROUP_SIZE",WorkgroupSizeAsStr };
```
devshgraphicsprogramming, line 117: remove this

tools/nsc/main.cpp
```diff
@@ -0,0 +1,155 @@
...
+		smart_refctd_ptr<CHLSLCompiler> hlslcompiler = make_smart_refctd_ptr<CHLSLCompiler>(smart_refctd_ptr(m_system));
+
+		CHLSLCompiler::SOptions options = {};
+		// want as much debug as possible
+		options.stage = asset::IShader::E_SHADER_STAGE::ESS_UNKNOWN;
+		options.debugInfoFlags = IShaderCompiler::E_DEBUG_INFO_FLAGS::EDIF_LINE_BIT;
+		// this lets you source-level debug/step shaders in renderdoc
+		//if (physDev->getLimits().shaderNonSemanticInfo)
+		//options.debugInfoFlags |= IShaderCompiler::E_DEBUG_INFO_FLAGS::EDIF_NON_SEMANTIC_BIT;
+		// if you don't set the logger and source identifier you'll have no meaningful errors
```
devshgraphicsprogramming, line 128: dont set those

tools/nsc/main.cpp
```diff
@@ -0,0 +1,155 @@
...
+		options.preprocessorOptions.sourceIdentifier = sourceIdentifier;
+		options.preprocessorOptions.logger = m_logger.get();
+		options.preprocessorOptions.extraDefines = { &WorkgroupSizeDefine,&WorkgroupSizeDefine + 1 };
```
devshgraphicsprogramming, line 131: empty

tools/nsc/main.cpp
```diff
@@ -0,0 +1,155 @@
...
+		constexpr uint32_t WorkgroupSize = 256;
+		constexpr uint32_t WorkgroupCount = 2048;
+		const string WorkgroupSizeAsStr = std::to_string(WorkgroupSize);
+		const IShaderCompiler::SPreprocessorOptions::SMacroDefinition WorkgroupSizeDefine = { "WORKGROUP_SIZE",WorkgroupSizeAsStr };
+
+		smart_refctd_ptr<CHLSLCompiler> hlslcompiler = make_smart_refctd_ptr<CHLSLCompiler>(smart_refctd_ptr(m_system));
+
+		CHLSLCompiler::SOptions options = {};
+		// want as much debug as possible
+		options.stage = asset::IShader::E_SHADER_STAGE::ESS_UNKNOWN;
```
devshgraphicsprogramming, line 123: use the asset loader to load your input file as ICPUShader, it might be able to deduce the stage for you
(in return you have to check that the content type is  HLSL and not something else)

src/nbl/asset/utils/waveContext.h
Hazardu: In boost wave `found_directive` hook triggers on every preprocessor directive, so i will count it there instead
```
template <typename ContextT, typename TokenT>
bool found_directive(ContextT const& ctx, TokenT const& directive)
{
    hash_token_occurences++;
    return false;
}
```

src/nbl/asset/utils/CHLSLCompiler.cpp
Hazardu: I was told to have any arguments read from `#pragma` override any arguments passed by command line, thus it has to be a new vector, that replaces hlslOptions.dxcOptions.

tools/nsc/main.cpp
Hazardu: I remember that the code that emits `#pragma wave dxc_compile_flags` is conditional, it only adds the pragma to compiled code if this debug flag `IShaderCompiler::E_DEBUG_INFO_FLAGS::EDIF_LINE_BIT` or `EDIF_NON_SEMANTIC_BIT` or `EDIF_SOURCE_BIT` is on 

How should we go about this? Do you want the pragma to always be emitted, or only when user gives the corresponding `-fspv-debug=file|source|line` arg in command line when running NSC tool?

src/nbl/asset/utils/waveContext.h
```diff
@@ -103,24 +109,27 @@ struct preprocessing_hooks final : public boost::wave::context_policies::default
             return true;
         }
         
-        if (strcmp(optionStr, "dxc_compile_flags") == 0 && hash_token_occurences == 1) {
+        if (strcmp(optionStr, "dxc_compile_flags") ) {
+            if (0 && hash_token_occurences != 1) {
```
devshgraphicsprogramming, line 113: some weird typo here, you go into the if block if your optionStr DOES NOT equal `dxc_compile_flags`

CMakeLists.txt
```diff
@@ -352,10 +352,8 @@ add_subdirectory(${THIRD_PARTY_SOURCE_DIR})
 
 add_subdirectory(src/nbl)
 add_subdirectory("${NBL_PYTHON_MODULE_ROOT_PATH}" tests) # Python Framework
-
-if(NBL_BUILD_EXAMPLES)
-	add_subdirectory(examples_tests)
-endif()
+add_subdirectory(examples_tests)
```
devshgraphicsprogramming, line 355: this needs a revert!

cmake/common.cmake
```diff
@@ -56,18 +56,28 @@ macro(nbl_create_executable_project _EXTRA_SOURCES _EXTRA_OPTIONS _EXTRA_INCLUDE
...
+			endif()
+		
+			add_dependencies(${EXECUTABLE_NAME} ${NBL_EXECUTABLE_COMMON_API_TARGET})
+			target_link_libraries(${EXECUTABLE_NAME} PUBLIC ${NBL_EXECUTABLE_COMMON_API_TARGET})
+			target_precompile_headers("${EXECUTABLE_NAME}" REUSE_FROM "${NBL_EXECUTABLE_COMMON_API_TARGET}")
 		endif()
-	
-		add_dependencies(${EXECUTABLE_NAME} ${NBL_EXECUTABLE_COMMON_API_TARGET})
-		target_link_libraries(${EXECUTABLE_NAME} PUBLIC ${NBL_EXECUTABLE_COMMON_API_TARGET})
-		target_precompile_headers("${EXECUTABLE_NAME}" REUSE_FROM "${NBL_EXECUTABLE_COMMON_API_TARGET}")
```
devshgraphicsprogramming, line 70: @AnastaZIuk review this and okay it

tools/nsc/main.cpp
devshgraphicsprogramming: still outstanding

tools/nsc/main.cpp
devshgraphicsprogramming: you can always emit it, its just that DXC will "loose" the original source and comments AFTER wave is done preprocessing without the `-fspv-debug` which is fine... don't make emission of `#pragma wave dxc_compile_flags` conditional, always do it

tools/nsc/main.cpp
devshgraphicsprogramming: still outstanding

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -303,50 +301,32 @@ std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADE
...
-    {
-        // Shader Stage was overriden using #pragma
-        // #pragma wave shaderStage overrides all other definitions of shader stage, such as
-        // -T argument (passed here as options.stage)
-        // shader stage inferred from file extension in asset namespace
-        stage = stageOverrideFromPragma;
-    }
    
     std::vector<std::wstring> arguments = {};
     if (dxc_compile_flags.size() || hlslOptions.dxcOptions.size()) { // #pragma dxc_compile_flags takes priority
```
devshgraphicsprogramming, line 336: what the actual F? have you looked at your IF condition!?

tools/nsc/main.cpp
```diff
@@ -110,38 +105,54 @@ class ShaderCompiler final : public system::IApplicationFramework
...
-	std::string open_shader_file(std::string& filepath) {
-		std::ifstream stream(filepath);
-		std::string str((std::istreambuf_iterator<char>(stream)),
-			std::istreambuf_iterator<char>());
-		return str;
+	const ICPUShader* open_shader_file(std::string& filepath) {
+
+		m_assetMgr = make_smart_refctd_ptr<asset::IAssetManager>(smart_refctd_ptr(m_system));
+		auto resourceArchive = make_smart_refctd_ptr<system::CMountDirectoryArchive>(localInputCWD, smart_refctd_ptr(m_logger), m_system.get());
+		m_system->mount(std::move(resourceArchive));
```
devshgraphicsprogramming: should have just set `lp.workingDirectory` to CWD instead of mounting a directory XD

tools/nsc/main.cpp
```diff
@@ -0,0 +1,154 @@
...
+			return false;
+		}
+
+		m_arguments = std::vector<std::string>(argv.begin() + 2, argv.end()); // turn argv into vector for convenience
+		std::string file_to_compile = argv[1];
+
+		if (!m_system->exists(file_to_compile, IFileBase::ECF_READ)) {
+			m_logger->log("Incorrect arguments. Expecting second argument to be filename of the shader intended to compile.", ILogger::ELL_ERROR);
+			return false;
+		}
```
devshgraphicsprogramming, line 45: I think DXC takes the inputs from the back, not the front

```
dxc.exe [options] <inputs>
```

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,105 @@ CHLSLCompiler::~CHLSLCompiler()
...
+				}
+				s.replace(0, 2, stageStr);
+			}
+
+			std::vector<std::wstring::iterator> underscorePositions = {};
+			auto it = std::find(s.begin(), s.end(), '_');
+            while (it != s.end()) {
+				underscorePositions.push_back(it);
+                it = std::find(it + 1, s.end(), '_');
+            }
```
devshgraphicsprogramming, line 138: formatting

src/nbl/asset/utils/CHLSLCompiler.cpp
devshgraphicsprogramming: still outstanding!

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,105 @@ CHLSLCompiler::~CHLSLCompiler()
...
+
+static void try_upgrade_shader_stage(std::vector<std::wstring>& arguments, asset::IShader::E_SHADER_STAGE shaderStageOverrideFromPragma, system::logger_opt_ptr& logger) {
+	
+	constexpr int MajorReqVersion = 6,
+		MinorReqVersion = 7;
+
+	auto foundShaderStageArgument = std::find(arguments.begin(), arguments.end(), L"-T");
+	if (foundShaderStageArgument != arguments.end() && foundShaderStageArgument + 1 != arguments.end()) {
+		auto foundShaderStageArgumentValueIdx = foundShaderStageArgument - arguments.begin() + 1;
+		std::wstring s = arguments[foundShaderStageArgumentValueIdx];
```
devshgraphicsprogramming, line 120: you know you could just write `s = *(foundShaderStageArgument+1);`

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,105 @@ CHLSLCompiler::~CHLSLCompiler()
...
+		if (s.length() >= 6) {
```
devshgraphicsprogramming, line 121: where's the else case to replace it wholesale upon failure?

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,105 @@ CHLSLCompiler::~CHLSLCompiler()
...
+			if (shaderStageOverrideFromPragma != IShader::ESS_UNKNOWN)
+			{
+				// replace first 2 characters if shaderStageOverrideFromPragma != Unknown
+				auto stageStr = ShaderStageToString(shaderStageOverrideFromPragma);
+				if (!stageStr) {
+					logger.log("invalid shader stage %i", system::ILogger::ELL_ERROR, shaderStageOverrideFromPragma);
+					return;
+				}
+				s.replace(0, 2, stageStr);
```
devshgraphicsprogramming, line 130: replace everything up to the first `_`

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,105 @@ CHLSLCompiler::~CHLSLCompiler()
...
+            }
+			std::wstring majorVersionString, minorVersionString;
+			int size = underscorePositions.size();
+			auto secondLastUnderscore = underscorePositions[size - 2];
+			auto lastUnderscore = underscorePositions[size - 1];
+			majorVersionString = std::wstring(secondLastUnderscore + 1, lastUnderscore);
+			minorVersionString = std::wstring(lastUnderscore + 1, s.end());
+
+			int major = std::stoi(majorVersionString);
+			int minor = std::stoi(minorVersionString);
```
devshgraphicsprogramming, line 147: `stoi` throws `std::invalid_argument` etc. , so these need to be in a try-catch block and the `major` and `minor` need initialization to som einvalid balue

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -255,49 +327,59 @@ core::smart_refctd_ptr<ICPUShader> CHLSLCompiler::compileToSPIRV(const char* cod
...
+            arguments.push_back(L"-fspv-debug=file");
+        if (hlslOptions.debugInfoFlags.hasFlags(E_DEBUG_INFO_FLAGS::EDIF_SOURCE_BIT))
+            arguments.push_back(L"-fspv-debug=source");
+        if (hlslOptions.debugInfoFlags.hasFlags(E_DEBUG_INFO_FLAGS::EDIF_LINE_BIT))
+            arguments.push_back(L"-fspv-debug=line");
+        if (hlslOptions.debugInfoFlags.hasFlags(E_DEBUG_INFO_FLAGS::EDIF_TOOL_BIT))
+            arguments.push_back(L"-fspv-debug=tool");
+        if (hlslOptions.debugInfoFlags.hasFlags(E_DEBUG_INFO_FLAGS::EDIF_NON_SEMANTIC_BIT))
+            arguments.push_back(L"-fspv-debug=vulkan-with-source");
     }
```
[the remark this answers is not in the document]
devshgraphicsprogramming: still outstanding

tools/nsc/main.cpp
devshgraphicsprogramming: I think I did that already, check and if true resolve

src/nbl/asset/utils/CHLSLCompiler.cpp
devshgraphicsprogramming: Still outstanding

src/nbl/asset/utils/CHLSLCompiler.cpp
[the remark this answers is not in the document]
devshgraphicsprogramming: still outstanding

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,123 @@ CHLSLCompiler::~CHLSLCompiler()
...
+                logger.log("Parsing shader version failed, invalid argument exception: %s", system::ILogger::ELL_ERROR, e.what());
+            }
+            catch (const std::out_of_range& e) {
+                logger.log("Parsing shader version failed, out of range exception: %s", system::ILogger::ELL_ERROR, e.what());
+            }
+        }
+        else {
+            logger.log("invalid shader stage '%s' argument, expecting a string of length >= 6 ", system::ILogger::ELL_ERROR, s);
+        }
+	}
```
devshgraphicsprogramming, line 171: last else case missing, also you need to HANDLE the errors, i.e. provide a default SM version and stage, cause you do have a `shaderStageOverrideFromPragma` after all!

src/nbl/asset/utils/CHLSLCompiler.cpp
devshgraphicsprogramming: see https://github.com/Devsh-Graphics-Programming/Nabla/pull/648/files#r1497438169

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,123 @@ CHLSLCompiler::~CHLSLCompiler()
...
+	if (foundShaderStageArgument != arguments.end() && foundShaderStageArgument + 1 != arguments.end()) {
+		auto foundShaderStageArgumentValueIdx = foundShaderStageArgument - arguments.begin() + 1;
+		std::wstring s = arguments[foundShaderStageArgumentValueIdx];
+        if (s.length() >= 6) {
+            std::vector<std::wstring::iterator> underscorePositions = {};
+            auto it = std::find(s.begin(), s.end(), '_');
+            while (it != s.end()) {
+                underscorePositions.push_back(it);
+                it = std::find(it + 1, s.end(), '_');
+            }
```
devshgraphicsprogramming, line 127: magic: https://en.cppreference.com/w/cpp/string/basic_string/rfind

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,123 @@ CHLSLCompiler::~CHLSLCompiler()
...
+                    logger.log("invalid shader stage %i", system::ILogger::ELL_ERROR, shaderStageOverrideFromPragma);
+                    return;
+                }
+                if (underscorePositions.size() == 0)
+                {
+                    logger.log("incorrect format for shader stage parameter, expecting shader stage, underscore, then version", system::ILogger::ELL_ERROR, shaderStageOverrideFromPragma);
+                    return;
+                }
+                int len = underscorePositions[0] - s.begin();
+                s.replace(0, len, stageStr);
```
devshgraphicsprogramming, line 142: do you even understand what you're doing here?

you could have just deduced the SM version first! then fixed up the stage, you wrote complete sphagetti code

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,123 @@ CHLSLCompiler::~CHLSLCompiler()
...
+                    logger.log("Upgrading shader stage version number to %i %i", system::ILogger::ELL_DEBUG, MajorReqVersion, MinorReqVersion);
+                    arguments[foundShaderStageArgumentValueIdx] = std::wstring(s.begin(), secondLastUnderscore + 1) + std::to_wstring(MajorReqVersion) + L"_" + std::to_wstring(MinorReqVersion);
+                }
+            }
+            catch (const std::invalid_argument& e) {
+                logger.log("Parsing shader version failed, invalid argument exception: %s", system::ILogger::ELL_ERROR, e.what());
+            }
+            catch (const std::out_of_range& e) {
+                logger.log("Parsing shader version failed, out of range exception: %s", system::ILogger::ELL_ERROR, e.what());
+            }
```
devshgraphicsprogramming, line 166: youre supposed to recover, not just log errors.

Also catch separately for major and minor

src/nbl/asset/utils/CHLSLCompiler.cpp
Hazardu: Thanks

src/nbl/asset/utils/CHLSLCompiler.cpp
devshgraphicsprogramming: ok, still recovery needed

src/nbl/asset/utils/CHLSLCompiler.cpp
devshgraphicsprogramming: for a `-T stage_x_y`

malformed `stage` (length or not found in your array of stages), you can fallback to `shaderStageOverrideFromPragma` if its valid, if not you have a true error
malformed `_x_y`, you can fallback to the minimum required version

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,149 @@ CHLSLCompiler::~CHLSLCompiler()
...
+            
+            // In case of an exception 
+            arguments[foundShaderStageArgumentValueIdx] = stageStr + L"_" + std::to_wstring(MajorReqVersion) + L"_" + std::to_wstring(MinorReqVersion);
+            return;
+        }
+        else {
+            logger.log("invalid shader stage '%s' argument, expecting a string of length >= 6 ", system::ILogger::ELL_ERROR, s);
+        }
+	}
```
devshgraphicsprogramming, line 188: First thing: fix the formatiing/indentation !

Second thign: start using this pattern
```cpp
if (cond0)
{
   if (cond1)
   {
      success Case
   }
}

Failure Case
```
or this one
```cpp
Set Failure Case
if (cond0)
{
   if (cond1)
   {
      Overwrite Failure with Success Case
   }
}
```

Right now this code is unreadable mess, and if you hit the else case of `s.length()<6` you'll only end up with an error being logged, not a reasonable fallback.

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,149 @@ CHLSLCompiler::~CHLSLCompiler()
...
+            while (it != s.end()) {
+                underscorePositions.push_back(it);
+                it = std::find(it + 1, s.end(), '_');
+            }
+
+            // Bad input check
+            if (underscorePositions.size() < 2)
+            {
+                logger.log("Incorrect -T argument value.\nExpecting string with at least 2 '_' delimiters: between shader stage, version major and version minor.",
+                    system::ILogger::ELL_ERROR);
```
devshgraphicsprogramming, line 134: IMHO you're looking for exactly 2

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,149 @@ CHLSLCompiler::~CHLSLCompiler()
...
 {
-    ComPtr<IDxcBlobEncoding> errorMessages;
-    ComPtr<IDxcBlob> objectBlob;
-    ComPtr<IDxcResult> compileResult;
+    auto stageArgumentPos = std::find(arguments.begin(), arguments.end(), L"-HV");
+    if (stageArgumentPos != arguments.end()) {
+        auto index = stageArgumentPos - arguments.begin() + 1; // -HV XXXXX, get index of second
+        std::wstring version = arguments[index];
+        if (!isalpha(version.back()) && version.length() >= 4 && std::stoi(version) < 2021)
+            arguments[index] = L"2021";
```
devshgraphicsprogramming, line 102: this is weird, means you'll only rest to 2021 if all the above hold, should be
```cpp
auto& version = argumetns[index];
try
{
  // remove x and replace with 1
  if (isalpha(version.back()))
     version.back() = L'1';  
  if (version.length()!=4)
     throw std::invalid_argument();
  if (std::stoi(version)<2021)
     throw std::invalid_argument();
}
catch ()
{
   version = L"2021";
}
```

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,149 @@ CHLSLCompiler::~CHLSLCompiler()
...
 
 
-struct DxcCompilationResult
+static void try_upgrade_hlsl_version(std::vector<std::wstring>& arguments, system::logger_opt_ptr& logger)
 {
-    ComPtr<IDxcBlobEncoding> errorMessages;
-    ComPtr<IDxcBlob> objectBlob;
-    ComPtr<IDxcResult> compileResult;
+    auto stageArgumentPos = std::find(arguments.begin(), arguments.end(), L"-HV");
+    if (stageArgumentPos != arguments.end()) {
```
devshgraphicsprogramming, line 98: you should be checking that `stageArguementPos+1<arguments.end()` since you intend to access the next thign

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,149 @@ CHLSLCompiler::~CHLSLCompiler()
...
+            stageStr = std::wstring(s.begin(), underscorePositions[0]);
+            if (shaderStageOverrideFromPragma != IShader::ESS_UNKNOWN) // replace first 2 characters if shaderStageOverrideFromPragma != Unknown
+            {
+                if (!overrideStageStr) {
+                    logger.log("Invalid shader stage with int value '%i'.\nThis value does not have a known string representation.",
+                        system::ILogger::ELL_ERROR, shaderStageOverrideFromPragma);
+                    return;
+                }
+                stageStr = std::wstring(overrideStageStr);
+            }
```
devshgraphicsprogramming, line 148: why do you do the check here?

I mean you initialize the "no -T found` from this as well, so you SHOULD do this check at the top!

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -207,92 +357,48 @@ std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADE
...
-        case asset::IShader::ESS_MESH:
-            targetProfile.replace(0, 2, L"ms");
-            break;
-        default:
-            hlslOptions.preprocessorOptions.logger.log("invalid shader stage %i", system::ILogger::ELL_ERROR, stage);
+        // Set profile two letter prefix based on stage
+        auto stageStr = ShaderStageToString(stage);
+        if (!stageStr) {
+            logger.log("invalid shader stage %i", system::ILogger::ELL_ERROR, stage);
             return nullptr;
```
devshgraphicsprogramming, line 392: why have you got some duplicate logic, you're calling `try_upgrade_shader_stage` after this anyway!


Just don't include the `-T` parameter and test the fallbacks being correct in `try_upgrade_shader_stage` and `try_upgrade_hlsl_version`

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -317,17 +423,25 @@ core::smart_refctd_ptr<ICPUShader> CHLSLCompiler::compileToSPIRV(const char* cod
             arguments.push_back(L"-fspv-debug=vulkan-with-source");
```
devshgraphicsprogramming, line 423: the debug flags should always be added whenever `dxc_compile_flags.empty`

devshgraphicsprogramming: this means they should be added whenever `dxcOptions` doesn't have them!

src/nbl/asset/utils/CHLSLCompiler.cpp
devshgraphicsprogramming: it doesn't matter if you log an ERROR and don't stop the compilation beceause the override wasn't needed and the other value was ok, but the else case for find `-T` needs the check

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,158 @@ CHLSLCompiler::~CHLSLCompiler()
...
     {
-        return std::string(reinterpret_cast<char*>(errorMessages->GetBufferPointer()), errorMessages->GetBufferSize());
+        logger.log("Invalid shader stage with int value '%i'.\nThis value does not have a known string representation.",
+            system::ILogger::ELL_ERROR, shaderStageOverrideFromPragma);
+        return;
     }
-};
+    bool setDefaultValue = true;
+    auto foundShaderStageArgument = std::find(arguments.begin(), arguments.end(), L"-T");
+    if (foundShaderStageArgument != arguments.end() && foundShaderStageArgument +1 != arguments.end()) {
```
devshgraphicsprogramming, line 137: if I only have `-T`, then we wont enter but `setDefaultValue==true`, nitpick.

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -61,29 +92,158 @@ CHLSLCompiler::~CHLSLCompiler()
...
+            else
+            {
+                logger.log("Incorrect -T argument value.\nExpecting string with exactly 2 '_' delimiters: between shader stage, version major and version minor.",
+                    system::ILogger::ELL_ERROR);
+            }
+        }
+        else 
+        {
+            logger.log("invalid shader stage '%s' argument, expecting a string of length >= 6 ", system::ILogger::ELL_ERROR, s);
+        }
```
devshgraphicsprogramming, line 194: but you don't recover, you don't set the `stageStr` to anything before you start testing for `s.length()>=6`

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -207,127 +366,88 @@ std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADE
...
-    // TODO: Figure out a way to get the shader model version automatically
-    // 
-    // We can't get it from the DXC library itself, as the different versions and the parsing
-    // use a weird lexer based system that resolves to a hash, and all of that is in a scoped variable
-    // (lib/DXIL/DxilShaderModel.cpp:83)
-    // 
-    // Another option is trying to fetch it from the commandline tool, either from parsing the help message
-    // or from brute forcing every -T option until one isn't accepted
-    //
     std::wstring targetProfile(SHADER_MODEL_PROFILE);
```
devshgraphicsprogramming, line 386: unused variable

PR #649, opened 2024-01-25
Work on property pool HLSL impl

deprilula28:
## Description

Implementing `CPropertyPoolHandler` and `CPropertyPool` in HLSL, using direct buffer address instead of allocating descriptors sets for buffers.
Notes about impl:
```
-> Currently uses descritor pools (needs to allocate every time)
    -> Use BDA and root constants with the addresses instead
-> Device capabilities traits 
    -> Example version: https://github.com/Devsh-Graphics-Programming/Nabla/blob/vulkan_1_3/include/nbl/builtin/hlsl/device_capabilities_traits.hlsl
    -> maxOptimallyResidentWorkgroupInvocations
    -> Can use nbl::hlsl::jit::device_capabilities struct with JIT generated "constexpr" variables for maximally optimal workgroup invocations
    https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/blob/master/23_ArithmeticUnitTest/app_resources/shaderCommon.hlsl#L9
    https://github.com/microsoft/DirectXShaderCompiler/issues/6144


=== tasks ===

-> Port https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/glsl/property_pool/copy.comp to HLSL
    -> Persitently resident threads, scrolling over, maximally using the GPU workgroup size
    -> Dispatch: 2D (x: DWORD in the property ID, y: property ID)
        -> property id: which buffer youre touching/analogous to draw id
            -> indexes into transferData
            -> new version: use null pointer as invalid pointer
    -> transferData: List of copy "commands"
        -> new version: Replaced by push constant with BDA address
    -> addresses: "Index buffer"
        -> invalid pointer: IOTA (analogous to not using an index buffer, use iteration index as the fetching index)
    -> Use shorts (uint16) instead of DWORDs (uint32)
        -> Transfer data struct uses bytes for future proofing
    -> Specialize on:
        -> Whether or not source is a fill
        -> Type of index (uint8, uint16, uint32, uint64)
        -> Src index is IOTA
        -> Dst index is IOTA
    -> Keep optimization for modulos (line 38 & 52)

-> CPU Code
    -> CPropertyPoolHandler
        -> Nuke m_maxPropertiesPerPass, getMaxScratchSize (not relevant with BDA version)
    -> TransferRequest on CPU keeps reference to the buffer and places it in the command buffer for lifetime tracking
        -> Have a custom command that just keeps track of a **variable number** of reference counted objects for preserving lifetimes (LinkedPreservedLifetimes?)
            -> Take a span of IGPUReferenceCounted
            -> Example: https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/src/nbl/video/IGPUCommandBuffer.cpp#L104C54-L104C54
            -> For variable amount of stuff: https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/src/nbl/video/IGPUCommandBuffer.cpp#L403C90-L403C90
            -> Signature example: `IGPUCommandBuffer::preserveLifetime(std::span<const core::IReferenceCounted>)`
    -> New transfer property signature
        -> make pipeline barriers more robust (or require everything to be done properly outside the function)
        -> First parameter: SIntendedSubmitInfo (IUtilities-independent submit info struct thing for handling overflows)
            -> Source for it: https://github.com/Devsh-Graphics-Programming/Nabla/blob/vulkan_1_3/include/nbl/video/utilities/SIntendedSubmitInfo.h
            -> Move IUtilities::autoSubmit and IUtilities::autoSubmitAndBlock to SIntendedSubmitInfo as static method (no more relation to IUtilities)
        -> Second parameter: struct with parameters
            `const asset::SBufferBinding<video::IGPUBuffer>& scratch, system::logger_opt_ptr logger, const size_t baseDWORD=0ull, const size_t endDWORD=~0u`
            -> Additional parameters that are optional including additional pipeline barrier values
            bitfield/boolean [pre|post]ScratchBarrier = true
    -> lets keep MaxPropertiesPerDispatch and have it equal to 64kb/sizeof(nbl::hlsl::property_pools::transferTrequest)
        -> instead of copy lambda logic at https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/src/nbl/video/utilities/CPropertyPoolHandler.cpp#L172, fail if over MaxPropertiesPerDispatch
    -> leave upstreaming thing & contiguous buffers for later (#ifdef 0 it out)
        -> transferProperties with upstreaming & freeProperties
    -> IPropertyPool
        -> allocateProperties: use span instead of begin & end
            -> (behaviour) 
                -> goes through indices to find empty ones and allocate them
                -> if it's contiguous: add mapping from index to addr and addr to index
        -> nuke descriptor set stuff (line 198 -> 211)
        -> validateBlocks: change offset check (https://github.com/Devsh-Graphics-Programming/Nabla/blob/64cbb652e39acf0239a61bcee7fc26d70ab8d089/src/nbl/video/utilities/IPropertyPool.cpp#L38) to BDA
            -> check usages & non null address
    -> CPropertyPool: don't change anything, just make sure identation is right

    -> MegaDescriptorSet (Descriptor set sub-allocate)
        -> Have a multi-timeline event functor with IFuture await
        ```cpp
            MultiTimelineEventHandlerST<DeferredFreeFunctor> deferredFrees;
            deferredFrees.latch(futureWait,std::move(functor));
        ```
            -> Also have it on IPropertyPool
            -> Solve synchronization issues

    -> create example testing downloads, uploads of properties
        -> with IB, without IB, fills, etc etc
        -> use regular buffer for everything
        -> later test the streaming buffers (ifdef them back in)
```

## Testing 
<!-- Explain how this change was tested. -->

## TODO list:
<!-- A list of things that have to be finished before this PR can be merged -->

- [ ] Verify why things aren't being written accurately
- [ ] Implement address buffer handling
- [ ] Baseline test
- [ ] Test with IOTA
- [ ] Test with fill buffers
- [ ] Test with different element sizes
- [ ] Test with different element counts 
- [ ] Test with different transfer amounts

<!--
By creating this pull request into Nabla, you agree to release all your past (even from previous commits) and present contributions in the Nabla repository under the Apache 2.0 license. If you're not the sole contributor, ensure that all contributors have signed the CLA agreeing to this.
-->


include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -0,0 +1,148 @@
...
+    vk::RawBufferStore<uint>(globals.transferCommandsAddress + 40 * 5, vk::RawBufferLoad<uint>(transferRequest.srcAddr + sizeof(uint16_t) * 3));
+    //if (fill) { TransferLoopPermutationFill<true> loop; loop.copyLoop(invocationIndex, propertyId, transferRequest, dispatchSize); }
+    //else { TransferLoopPermutationFill<false> loop; loop.copyLoop(invocationIndex, propertyId, transferRequest, dispatchSize); }
+}
+
+}
+}
+}
+
+[numthreads(1,1,1)]
```
devshgraphicsprogramming, line 143: either:
- hardcode to `(256,1,1)`
- require somone else to make this entrypoint and provide the workgroup size by themself

I lean towards option 2 here

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -0,0 +1,148 @@
...
+    transferRequest.dstIndexAddr = vk::RawBufferLoad<uint64_t>(globals.transferCommandsAddress + sizeof(uint64_t) * 3);
+    // Remaining elements are part of the same bitfield
+    // TODO: Do this only using raw buffer load?
+    uint64_t bitfieldType = vk::RawBufferLoad<uint64_t>(globals.transferCommandsAddress + sizeof(uint64_t) * 4);
+    transferRequest.elementCount32 = uint32_t(bitfieldType);
+    transferRequest.elementCountExtra = uint32_t(bitfieldType);
+    transferRequest.propertySize = uint32_t(bitfieldType >> 3);
+    transferRequest.fill = uint32_t(bitfieldType >> (3 + 24));
+    transferRequest.srcIndexSizeLog2 = uint32_t(bitfieldType >> (3 + 24 + 1));
+    transferRequest.dstIndexSizeLog2 = uint32_t(bitfieldType >> (3 + 24 + 1 + 2));
```
devshgraphicsprogramming, line 127: make a factory function (`static TransferRequest TransferRequest::construct(const uint64_t)`) which does this instead of keeping it around here

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -0,0 +1,148 @@
...
+#include "nbl/builtin/hlsl/property_pool/transfer.hlsl"
+
+namespace nbl
+{
+namespace hlsl
+{
+namespace property_pools
+{
+
+[[vk::push_constant]] GlobalPushContants globals;
```
devshgraphicsprogramming, line 12: this is nasty, pass the struct around as a `NBL_CONST_REF` function parameter

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -0,0 +1,148 @@
...
+    // TODO: Do this only using raw buffer load?
+    uint64_t bitfieldType = vk::RawBufferLoad<uint64_t>(globals.transferCommandsAddress + sizeof(uint64_t) * 4);
+    transferRequest.elementCount32 = uint32_t(bitfieldType);
+    transferRequest.elementCountExtra = uint32_t(bitfieldType);
+    transferRequest.propertySize = uint32_t(bitfieldType >> 3);
+    transferRequest.fill = uint32_t(bitfieldType >> (3 + 24));
+    transferRequest.srcIndexSizeLog2 = uint32_t(bitfieldType >> (3 + 24 + 1));
+    transferRequest.dstIndexSizeLog2 = uint32_t(bitfieldType >> (3 + 24 + 1 + 2));
+
+    const uint dispatchSize = nbl::hlsl::device_capabilities_traits<device_capabilities>::maxOptimallyResidentWorkgroupInvocations;
```
devshgraphicsprogramming, line 129: its slightly overkill to bring out the JIT device traits, I'd rather take the dispatch size as a template parameter and do the JIT device traints outside

include/nbl/builtin/hlsl/property_pool/transfer.hlsl
```diff
@@ -0,0 +1,57 @@
...
+
+struct TransferRequest
+{
+    // This represents a transfer command/request
+    uint64_t srcAddr;
+    uint64_t dstAddr;
+    uint64_t srcIndexAddr; // IOTA default
+    uint64_t dstIndexAddr; // IOTA default
+    // TODO: go back to this ideal layout when things work
+    // (Getting a fatal error from DXC when using 64-bit bitfields:)
```
devshgraphicsprogramming, line 21: open an issue

include/nbl/builtin/hlsl/property_pool/transfer.hlsl
```diff
@@ -0,0 +1,57 @@
...
+    //uint64_t dstIndexSizeLog2 : 2;
+    uint32_t elementCount32; // 32 first bits
+    uint32_t elementCountExtra : 3; // 3 last bits
+    uint32_t propertySize : 24;
+    uint32_t fill: 1;
+    uint32_t srcIndexSizeLog2 : 2;
+    uint32_t dstIndexSizeLog2 : 2;
+};
+
+struct GlobalPushContants
```
devshgraphicsprogramming, line 39: maybe call it TransfersInfo or Parameters instead of Push Constants

include/nbl/builtin/hlsl/property_pool/transfer.hlsl
```diff
@@ -0,0 +1,57 @@
...
+    // BDA address (GPU pointer) into the transfer commands buffer
+    uint64_t transferCommandsAddress;
+    // Define the range of invocations (X axis) that will be transfered over in this dispatch
+    // May be sectioned off in the case of overflow or any other situation that doesn't allow
+    // for a full transfer
+    uint64_t beginOffset;
+    uint64_t endOffset;
+};
+
+NBL_CONSTEXPR uint32_t MaxPropertiesPerDispatch = 128;
```
devshgraphicsprogramming, line 50: is there any reason to keep this around anymore?

include/nbl/builtin/hlsl/property_pool/transfer.hlsl
```diff
@@ -0,0 +1,57 @@
...
+
+struct GlobalPushContants 
+{
+    // BDA address (GPU pointer) into the transfer commands buffer
+    uint64_t transferCommandsAddress;
+    // Define the range of invocations (X axis) that will be transfered over in this dispatch
+    // May be sectioned off in the case of overflow or any other situation that doesn't allow
+    // for a full transfer
+    uint64_t beginOffset;
+    uint64_t endOffset;
```
devshgraphicsprogramming, line 47: would be useful to make it clear we're counting in DWORDs or shorts (if you want to do 16bit transfer atoms instead)

include/nbl/video/utilities/IPropertyPool.h
```diff
@@ -11,6 +11,11 @@
 #include "nbl/video/ILogicalDevice.h"
 #include "nbl/video/IGPUDescriptorSetLayout.h"
 
+#include "glm/glm/glm.hpp"
+#include <nbl/builtin/hlsl/cpp_compat/matrix.hlsl>
+#include <nbl/builtin/hlsl/cpp_compat/vector.hlsl>
```
devshgraphicsprogramming, line 16: should be neough to just include the `cpp_compat.hlsl`

include/nbl/video/utilities/CPropertyPoolHandler.h
```diff
@@ -12,18 +12,14 @@
 #include "nbl/video/utilities/IDescriptorSetCache.h"
 #include "nbl/video/utilities/IPropertyPool.h"
 
+#include "glm/glm/glm.hpp"
+#include <nbl/builtin/hlsl/cpp_compat/matrix.hlsl>
+#include <nbl/builtin/hlsl/cpp_compat/vector.hlsl>
+#include "nbl/builtin/hlsl/property_pool/transfer.hlsl"
```
devshgraphicsprogramming, line 18: same as https://github.com/Devsh-Graphics-Programming/Nabla/pull/649/files#r1488279071

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -0,0 +1,148 @@
...
+template<bool Fill>
+struct TransferLoopPermutationFill
+{
+    void copyLoop(uint baseInvocationIndex, uint propertyId, TransferRequest transferRequest, uint dispatchSize)
+    {
+        bool srcIota = transferRequest.srcIndexAddr == 0;
+        if (srcIota) { TransferLoopPermutationSrcIota<Fill, true> loop; loop.copyLoop(baseInvocationIndex, propertyId, transferRequest, dispatchSize); }
+        else { TransferLoopPermutationSrcIota<Fill, false> loop; loop.copyLoop(baseInvocationIndex, propertyId, transferRequest, dispatchSize); }
+    }
+};
```
devshgraphicsprogramming, line 104: only use such struct functor trampolines if you need partial specialization.

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -0,0 +1,148 @@
...
+    {
+        bool dstIota = transferRequest.dstIndexAddr == 0;
+        if (dstIota) { TransferLoopPermutationDstIota<Fill, SrcIndexIota, true> loop; loop.copyLoop(baseInvocationIndex, propertyId, transferRequest, dispatchSize); }
+        else { TransferLoopPermutationDstIota<Fill, SrcIndexIota, false> loop; loop.copyLoop(baseInvocationIndex, propertyId, transferRequest, dispatchSize); }
+    }
+};
+
+template<bool Fill>
+struct TransferLoopPermutationFill
+{
```
devshgraphicsprogramming, line 97: only use structs instead of templated functions when you need partial specialization

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,119 @@
...
+  public:
+        using AddressAllocator = AddrAllocator;
+        using ReservedAllocator = ReservAllocator;
+        using size_type = typename AddressAllocator::size_type;
+        using value_type = typename AddressAllocator::size_type;
+        static constexpr value_type invalid_value = AddressAllocator::invalid_address;
+
+        // constructors
+        template<typename... Args>
+        inline SubAllocatedDescriptorSet(const std::span<const video::IGPUDescriptorSetLayout::SBinding> bindings,
```
devshgraphicsprogramming, line 29: why didn't you take a layout instead of a span of bindings?

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,119 @@
...
+                }
+				m_allocatableRanges.push_back(range);
+            }
+
+			m_addressAllocator = AddrAllocator(
+				_reservedAllocator.allocate(AddressAllocator::reserved_size(maxAllocatableAlignment, static_cast<size_type>(allocatableDescriptors), args...), _NBL_SIMD_ALIGNMENT),
+				static_cast<size_type>(0), 0u, maxAllocatableAlignment, static_cast<size_type>(allocatableDescriptors), std::forward<Args>(args)...
+			);
+			m_reservedAllocator = ReservedAllocator(std::move(_reservedAllocator));
+            m_reservedSize = allocatableDescriptors;
```
devshgraphicsprogramming, line 55: nooo, you need to prime it with `AddressAllocator::reserved_size` ou allocated the reserved space with

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,119 @@
...
+        {
+            auto ptr = reinterpret_cast<const uint8_t*>(core::address_allocator_traits<AddressAllocator>::getReservedSpacePtr(m_addressAllocator));
+            m_reservedAllocator.deallocate(const_cast<uint8_t*>(ptr),m_reservedSize);
+        }
+
+        // anyone gonna use it?
+        inline const AddressAllocator& getAddressAllocator() const {return m_addressAllocator;}
+
+        //
+        inline ReservedAllocator& getReservedAllocator() {return m_reservedAllocator;}
```
devshgraphicsprogramming, line 74: keep it a private implementation detail

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,119 @@
...
+        //! Warning `outAddresses` needs to be primed with `invalid_value` values, otherwise no allocation happens for elements not equal to `invalid_value`
+        template<typename... Args>
+        inline void multi_allocate(uint32_t count, value_type* outAddresses, const size_type* sizes, const Args&... args)
+        {
+            core::address_allocator_traits<AddressAllocator>::multi_alloc_addr(m_addressAllocator,count,outAddresses,sizes,1,args...);
+        }
+        inline void multi_deallocate(uint32_t count, const size_type* addr, const size_type* sizes)
+        {
+            core::address_allocator_traits<AddressAllocator>::multi_free_addr(m_addressAllocator,count,addr,sizes);
+        }
```
devshgraphicsprogramming, line 87: the allocations from different bindings should not be interchangeable, so this method should take the binding index as the first parameter (before the count)

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    allocatableDescriptors += binding.count;
+                }
+				m_allocatableRanges.push_back(range);
+            }
+
+			m_addressAllocator = AddrAllocator(
+				_reservedAllocator.allocate(AddressAllocator::reserved_size(maxAllocatableAlignment, static_cast<size_type>(allocatableDescriptors), args...), _NBL_SIMD_ALIGNMENT),
+				static_cast<size_type>(0), 0u, maxAllocatableAlignment, static_cast<size_type>(allocatableDescriptors), std::forward<Args>(args)...
+			);
```
devshgraphicsprogramming, line 53: https://github.com/Devsh-Graphics-Programming/Nabla/pull/649/files#r1488635235

I'd have one address allocator per `AFTER_BIND & UPDATE_WHILE_PENDING & PARTIALLY_BOUND` binding

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
deprilula28: Layout doesn't have the list of bindings

include/nbl/video/utilities/CPropertyPoolHandler.h
```diff
@@ -12,18 +12,14 @@
 #include "nbl/video/utilities/IDescriptorSetCache.h"
 #include "nbl/video/utilities/IPropertyPool.h"
 
+#include "glm/glm/glm.hpp"
+#include <nbl/builtin/hlsl/cpp_compat/matrix.hlsl>
+#include <nbl/builtin/hlsl/cpp_compat/vector.hlsl>
```
devshgraphicsprogramming, line 17: just `#include "nbl/builtin/hlsl/cpp_compat.hlsl"`

include/nbl/video/utilities/IPropertyPool.h
devshgraphicsprogramming, line 16: just `#include "nbl/builtin/hlsl/cpp_compat.hlsl"`

include/nbl/video/utilities/IPropertyPool.h
```diff
@@ -21,8 +26,7 @@ class NBL_API2 IPropertyPool : public core::IReferenceCounted
 	public:
 		using PropertyAddressAllocator = core::PoolAddressAllocatorST<uint32_t>;
 
-        static inline constexpr auto invalid = PropertyAddressAllocator::invalid_address;
-
+        static inline constexpr uint64_t invalid = 0;
```
devshgraphicsprogramming, line 29: when did we agree to change the `invalid` from the AddressAllocator invalid (0xdeadbeefu) to 0?

include/nbl/video/utilities/IPropertyPool.h
```diff
@@ -34,19 +38,19 @@ class NBL_API2 IPropertyPool : public core::IReferenceCounted
...
         //
-        inline uint32_t getFree() const
+        inline uint64_t getFree() const
         {
             return indexAllocator.get_free_size();
         }
 
         //
-        inline uint32_t getCapacity() const
+        inline uint64_t getCapacity() const
```
devshgraphicsprogramming, line 53: no point changing them to `uint64_t` if your address allocator is still  `uint32_t`

devshgraphicsprogramming: also this is a poster child fro `using` aliasing the `PropertyAddressAllocator::value_type` and using that throughout

src/nbl/video/utilities/CPropertyPoolHandler.cpp
```diff
@@ -5,11 +5,36 @@
...
+			asset::IAssetLoader::SAssetLoadParams params = {};
+			auto assetBundle = assetManager->getAsset(path, params);
+			auto assets = assetBundle.getContents();
+			assert(!assets.empty());
+
+			auto cpuShader = asset::IAsset::castDown<asset::ICPUShader>(assets[0]);
+			auto shader = m_device->createShader(cpuShader.get());
+			return shader;
+		};
+	auto shader = loadShader("../../../include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl");
```
devshgraphicsprogramming, line 26: I'd rather use `system->openFile` and use the `nbl/builtin/hlsl/property_pool/copy.comp.hlsl` path directly toi opena  memory mapped file and create `IGPUShader` from it directly

devshgraphicsprogramming: look at the old `loadBuiltinData` code

src/nbl/video/utilities/CPropertyPoolHandler.cpp
```diff
@@ -59,15 +84,89 @@ CPropertyPoolHandler::CPropertyPoolHandler(core::smart_refctd_ptr<ILogicalDevice
...
+		uint32_t requestsThisPass = core::min<uint32_t>(std::distance(transferPassRequests, requestsEnd), MaxPropertiesPerDispatch);
+		uint64_t maxElements = 0;
+		for (uint32_t i = 0; i < requestsThisPass; i ++)
+		{
+			auto& transferRequest = transferRequestsData[i];
+			auto srcRequest = transferPassRequests + i;
+			transferRequest.srcAddr = srcRequest->memblock.buffer.get()->getDeviceAddress() + srcRequest->memblock.offset;
+			transferRequest.dstAddr = srcRequest->buffer.buffer.get()->getDeviceAddress() + srcRequest->buffer.offset;
+			transferRequest.srcIndexAddr = srcRequest->srcAddressesOffset ? addressBufferDeviceAddr + srcRequest->srcAddressesOffset : 0;
+			transferRequest.dstIndexAddr = srcRequest->dstAddressesOffset ? addressBufferDeviceAddr + srcRequest->dstAddressesOffset : 0;
```
devshgraphicsprogramming, line 134: I think I told you explicitly to get rid of the `addresses` buffer, and just have each transferRequest supply the correctly offsetted BDA or SBufferBinding instead

devshgraphicsprogramming: given how you're doing `srcAddr` and `dstAddr`, the `SBufferBinding` route is preferrable

src/nbl/video/utilities/CPropertyPoolHandler.cpp
```diff
@@ -59,15 +84,89 @@ CPropertyPoolHandler::CPropertyPoolHandler(core::smart_refctd_ptr<ILogicalDevice
...
+			transferRequest.elementCount32 = uint32_t(srcRequest->elementCount & (uint64_t(1) << 32) - 1);
+			transferRequest.elementCountExtra = uint32_t(srcRequest->elementCount >> 32);
+			transferRequest.propertySize = srcRequest->elementSize;
+			transferRequest.fill = 0; // TODO
+			transferRequest.srcIndexSizeLog2 = 1u; // TODO
+			transferRequest.dstIndexSizeLog2 = 1u; // TODO
+			assert(getAlignment(transferRequest.srcAddr) == 0);
+			assert(getAlignment(transferRequest.dstAddr) == 0);
```
devshgraphicsprogramming, line 142: I'd rather error and get a nice log message than crash

src/nbl/video/utilities/CPropertyPoolHandler.cpp
```diff
@@ -59,15 +84,89 @@ CPropertyPoolHandler::CPropertyPoolHandler(core::smart_refctd_ptr<ILogicalDevice
...
+		logger.log("CPropertyPoolHandler: Need a valid scratch buffer which can have updates staged from the commandbuffer!",system::ILogger::ELL_ERROR);
+		return false;
+	}
+	// TODO: validate usage flags
+	uint32_t maxScratchSize = MaxPropertiesPerDispatch * sizeof(nbl::hlsl::property_pools::TransferRequest);
+	if (scratch.offset + maxScratchSize > scratch.buffer->getSize())
+	    logger.log("CPropertyPoolHandler: The scratch buffer binding provided might not be big enough in the worst case! (Scratch buffer size: %i Max scratch size: %i)",
+			system::ILogger::ELL_WARNING,
+			scratch.buffer->getSize() - scratch.offset,
+			maxScratchSize);
```
devshgraphicsprogramming, line 112: I'd validate better, also right now your code will choke/crash on a scratch buffer thats too small, you're still using `MaxPropertiesPerDispatch` to figure out the `numberOfPasses`

src/nbl/video/utilities/CPropertyPoolHandler.cpp
```diff
@@ -59,15 +84,89 @@ CPropertyPoolHandler::CPropertyPoolHandler(core::smart_refctd_ptr<ILogicalDevice
...
+		}
+		cmdbuf->updateBuffer({ scratch.offset,sizeof(TransferRequest) * requestsThisPass, core::smart_refctd_ptr(scratch.buffer) }, transferRequestsData);
+		// TODO: pipeline barrier
+		cmdbuf->bindComputePipeline(m_pipeline.get());
+		
+		nbl::hlsl::property_pools::GlobalPushContants pushConstants;
+		{
+			pushConstants.beginOffset = baseDWORD;
+			pushConstants.endOffset = endDWORD;
+			pushConstants.transferCommandsAddress = scratchBufferDeviceAddr;
```
devshgraphicsprogramming, line 154: you need to properly offset by `transferPassRequestsIndex` or something related

src/nbl/video/utilities/CPropertyPoolHandler.cpp
```diff
@@ -59,15 +84,89 @@ CPropertyPoolHandler::CPropertyPoolHandler(core::smart_refctd_ptr<ILogicalDevice
...
+			maxElements = core::max<uint64_t>(maxElements, srcRequest->elementCount);
+		}
+		cmdbuf->updateBuffer({ scratch.offset,sizeof(TransferRequest) * requestsThisPass, core::smart_refctd_ptr(scratch.buffer) }, transferRequestsData);
+		// TODO: pipeline barrier
+		cmdbuf->bindComputePipeline(m_pipeline.get());
+		
+		nbl::hlsl::property_pools::GlobalPushContants pushConstants;
+		{
+			pushConstants.beginOffset = baseDWORD;
+			pushConstants.endOffset = endDWORD;
```
devshgraphicsprogramming, line 153: this is why naming your variables matters, Offset == ByteOffset, but we're clearly counting in DWORDs?

src/nbl/video/utilities/CPropertyPoolHandler.cpp
```diff
@@ -59,15 +84,89 @@ CPropertyPoolHandler::CPropertyPoolHandler(core::smart_refctd_ptr<ILogicalDevice
...
+		}
+		assert(getAlignment(scratchBufferDeviceAddr) == 0);
+		assert(getAlignment(sizeof(TransferRequest)) == 0);
+		cmdbuf->pushConstants(m_pipeline->getLayout(), asset::IShader::ESS_COMPUTE, 0u, sizeof(pushConstants), &pushConstants);
+
+		// dispatch
+		{
+			const auto& limits = m_device->getPhysicalDevice()->getLimits();
+			const auto invocationCoarseness = limits.maxOptimallyResidentWorkgroupInvocations * requestsThisPass;
+			cmdbuf->dispatch(limits.computeOptimalPersistentWorkgroupDispatchSize(maxElements,invocationCoarseness), requestsThisPass, 1u);
```
devshgraphicsprogramming, line 164: you're probably getting the wrong thing here, you want to round-up-divide the `maxElements` by `requestsThisPass` before shoving it into `computeOptimalPersistentWorkgroupDispatchSize`

NOTE THAT `computeOptimalPersistentWorkgroupDispatchSize` SECOND ARGUMENT IS WORKGROUP SIZE! NOT DISPATCH SIZE

src/nbl/video/utilities/IPropertyPool.cpp
```diff
@@ -20,7 +20,7 @@ IPropertyPool::IPropertyPool(uint32_t capacity, void* reserved, bool contiguous)
 {
     if (contiguous)
     {
-        m_indexToAddr = reinterpret_cast<uint32_t*>(reinterpret_cast<uint8_t*>(reserved)+getReservedSize(capacity));
+        m_indexToAddr = reinterpret_cast<uint64_t*>(reinterpret_cast<uint8_t*>(reserved)+getReservedSize(capacity));
```
devshgraphicsprogramming, line 23: you can't touch this up without making the reserved mem bigger!

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: Last I checked `IGPUDescriptorSetLayout` inherits from `asset::IDescriptorSetLayout` which has the following public methods:
- `uint32_t getTotalBindingCount()`
- `uint32_t getTotalDescriptorCount(const IDescriptor::E_TYPE type)`
- `const CBindingRedirect& getDescriptorRedirect(const IDescriptor::E_TYPE type) const`

What do you need in `CBindingRedirect`  ?  a getter for `CBindingRedirect::m_createFlags` ?

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
deprilula28: I'm not sure what you mean

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
deprilula28: Ah ok I think I missed CBindingRedirect

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
devshgraphicsprogramming: The struct functor only makes sense if:
- you do a partial specialization, i.e. `template<typename ArbitraryType> struct MyFunctor<true,ArbitraryType>` because functions can be only fully specialized
- you need to pass the functor as a template arg/lambda because HLSL202x doesn't allow function pointers/references  or you want a stateful functor
```cpp
template<typename Accessor, typename Compare>
uint32_t find_first(inout Accessor accessor, const Compare comparator);
```

if neither of the above applies, just use a templated function

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -111,36 +113,49 @@ void main(uint32_t3 dispatchId)
...
+    else { TransferLoopPermutationFill<false> loop; loop.copyLoop(invocationIndex, propertyId, transferRequest, dispatchSize); }
 }
 
 }
 }
 }
 
-[numthreads(1,1,1)]
+// TODO: instead use some sort of replace function for getting optimal size?
+[numthreads(512,1,1)]
```
devshgraphicsprogramming, line 158: I already wrote on discord to codegen a 5 line compute shader with `main` in the C++ and leave `copy.hlsl` as stage agnostic

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -111,36 +113,49 @@ void main(uint32_t3 dispatchId)
...
+    uint64_t transferCmdAddr = globals.transferCommandsAddress + sizeof(TransferRequest) * propertyId;
     TransferRequest transferRequest;
-    transferRequest.srcAddr = vk::RawBufferLoad<uint>(globals.transferCommandsAddress) | vk::RawBufferLoad<uint>(globals.transferCommandsAddress + sizeof(uint)) << 32;
-    transferRequest.dstAddr = vk::RawBufferLoad<uint64_t>(globals.transferCommandsAddress + sizeof(uint64_t));
-    transferRequest.srcIndexAddr = vk::RawBufferLoad<uint64_t>(globals.transferCommandsAddress + sizeof(uint64_t) * 2);
-    transferRequest.dstIndexAddr = vk::RawBufferLoad<uint64_t>(globals.transferCommandsAddress + sizeof(uint64_t) * 3);
+    transferRequest.srcAddr = vk::RawBufferLoad<uint64_t>(transferCmdAddr,8);
+    transferRequest.dstAddr = vk::RawBufferLoad<uint64_t>(transferCmdAddr + sizeof(uint64_t),8);
+    transferRequest.srcIndexAddr = vk::RawBufferLoad<uint64_t>(transferCmdAddr + sizeof(uint64_t) * 2,8);
+    transferRequest.dstIndexAddr = vk::RawBufferLoad<uint64_t>(transferCmdAddr + sizeof(uint64_t) * 3,8);
```
devshgraphicsprogramming, line 121: make a wrapper for `vk::RawBufferLoad` and `Store` in `nbl::hlsl::legacy` which uses our type traits to default the alignment

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -19,18 +19,20 @@ struct TransferLoop
...
+        const uint64_t srcAddressBufferOffset = SrcIndexIota ? srcOffset : vk::RawBufferLoad<uint32_t>(transferRequest.srcIndexAddr + srcOffset * sizeof(uint32_t));
+        const uint64_t dstAddressBufferOffset = DstIndexIota ? dstOffset : vk::RawBufferLoad<uint32_t>(transferRequest.dstIndexAddr + dstOffset * sizeof(uint32_t));
 
         const uint64_t srcAddressMapped = transferRequest.srcAddr + srcAddressBufferOffset * srcIndexSize; 
         const uint64_t dstAddressMapped = transferRequest.dstAddr + dstAddressBufferOffset * dstIndexSize; 
 
+        //vk::RawBufferStore<uint64_t>(transferRequest.dstAddr + invocationIndex * sizeof(uint64_t) * 2, srcAddressMapped,8);
+        //vk::RawBufferStore<uint64_t>(transferRequest.dstAddr + invocationIndex * sizeof(uint64_t) * 2 + sizeof(uint64_t), dstAddressMapped,8);
         if (SrcIndexSizeLog2 == 0) {} // we can't write individual bytes
         else if (SrcIndexSizeLog2 == 1) vk::RawBufferStore<uint16_t>(dstAddressMapped, vk::RawBufferLoad<uint16_t>(srcAddressMapped));
```
devshgraphicsprogramming, line 37: you've misundestood, we always store in atoms of `uint32_t` (you can upgrade to `uint16_t` if you want), its only the index that changes size

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -19,18 +19,20 @@ struct TransferLoop
...
         
-        const uint64_t srcIndexAddress = Fill ? transferRequest.srcIndexAddr + srcOffset : transferRequest.srcIndexAddr;
-        const uint64_t dstIndexAddress = Fill ? transferRequest.dstIndexAddr + dstOffset : transferRequest.dstIndexAddr;
-
-        const uint64_t srcAddressBufferOffset = SrcIndexIota ? srcIndexAddress : vk::RawBufferLoad<uint32_t>(srcIndexAddress);
-        const uint64_t dstAddressBufferOffset = DstIndexIota ? dstIndexAddress : vk::RawBufferLoad<uint32_t>(dstIndexAddress);
+        // IOTA: Use the index as the fetching offset
+        // Non IOTA: Read the address buffer ("index buffer") to select fetching offset
+        const uint64_t srcAddressBufferOffset = SrcIndexIota ? srcOffset : vk::RawBufferLoad<uint32_t>(transferRequest.srcIndexAddr + srcOffset * sizeof(uint32_t));
+        const uint64_t dstAddressBufferOffset = DstIndexIota ? dstOffset : vk::RawBufferLoad<uint32_t>(transferRequest.dstIndexAddr + dstOffset * sizeof(uint32_t));
```
devshgraphicsprogramming, line 29: its here where you should specialize for `IndexSizeLog2` !

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
deprilula28: ah ok, I think your original comment was the wrong way around

include/nbl/builtin/hlsl/property_pool/transfer.hlsl
```diff
@@ -49,6 +52,9 @@ struct GlobalPushContants
 
 NBL_CONSTEXPR uint32_t MaxPropertiesPerDispatch = 128;
 
+// TODO: instead use some sort of replace function for getting optimal size?
+NBL_CONSTEXPR uint32_t OptimalDispatchSize = 256;
```
devshgraphicsprogramming, line 56: you can use the device JIT to query the max compute dispatch size, I'd round it down to nearest PoT though, so the divisions aren't expensive

include/nbl/builtin/hlsl/property_pool/transfer.hlsl
devshgraphicsprogramming: in the DXC repo

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
devshgraphicsprogramming: > ah ok, I think your original comment was the wrong way around

"only use structs instead of templated functions when you need partial specialization"
1. you don't need partial specialization
2. you're using structs
3. but you only use structs when you have partial specialization
4. dont use structs here.

include/nbl/builtin/hlsl/property_pool/transfer.hlsl
deprilula28: https://github.com/microsoft/DirectXShaderCompiler/issues/6327

include/nbl/builtin/hlsl/property_pool/copy.comp.hlsl
```diff
@@ -62,58 +62,53 @@ struct TransferLoop
...
-    const uint propertyId = dispatchId.y;
-    const uint invocationIndex = dispatchId.x;
-
-    // Loading transfer request from the pointer (can't use struct
-    // with BDA on HLSL SPIRV)
-    uint64_t transferCmdAddr = globals.transferCommandsAddress + sizeof(TransferRequest) * propertyId;
+// Loading transfer request from the pointer (can't use struct
+// with BDA on HLSL SPIRV)
+static TransferRequest TransferRequest::newFromAddress(const uint64_t transferCmdAddr)
+{
```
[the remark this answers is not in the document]
deprilula28: The struct is shared with c++ code, so i wouldn't be able to use vk::rawbufferread; I could take the 64 bit value though

PR #650, opened 2024-01-26
Bitscan fix

kpentaris:
## Description
Subgroups invocations greater than 32 were using inappropriate
subgroupLtLe mask (using the first when they should be using
the one matching their subgroup invocation DWORD)

## Testing 
To properly test a gpu with max subgroup size > 32 is required (e.g. AMD)

## TODO list:
N/A


kpentaris:
ffs not sure why these other commits came about -_-

PR #656, opened 2024-02-19
Added bit_field intrisic

Eichenherz:
## Description
Added BitField(Un)SignedExtract intrinsic.


include/nbl/builtin/hlsl/spirv_intrinsics/core.hlsl
```diff
@@ -129,6 +129,15 @@ void memoryBarrier(uint32_t memoryScope, uint32_t memorySemantics);
...
 [[vk::ext_instruction(spv::OpBitcast)]]
 T bitcast(U);
+
+template<typename Unsigned>
+[[vk::ext_instruction( spv::OpBitFieldUExtract )]]
+Unsigned bitfieldExtractUnsigned( Unsigned val, uint32_t offsetBits, uint32_t numBits );
+
+template<typename Signed>
+[[vk::ext_instruction( spv::OpBitFieldSExtract )]]
+Signed bitfieldExtractSigned( Signed val, uint32_t offsetBits, uint32_t numBits );
```
devshgraphicsprogramming, line 139: keep the naming 1:1 with SPIR-V but without `Op` prefix, so `bitFieldUExtract` and `bitFieldSExtract`

include/nbl/builtin/hlsl/bit.hlsl
```diff
@@ -102,6 +108,13 @@ uint16_t countl_zero(T n)
     return impl::clz<sizeof(T)*8>(n);
 }
 
+// TODO: templated "switch" depending on signedness  
+template<typename Unsigned>
+Unsigned bitfield_extract_unsigned( Unsigned val, uint32_t offsetBits, uint32_t numBits )
+{
+    return spirv::bitfieldExtractUnsigned<Unsigned>( val, offsetBits, numBits );
+}
+
```
devshgraphicsprogramming, line 117: move this to the `glsl` namespace header and conform to
https://registry.khronos.org/OpenGL-Refpages/gl4/html/bitfieldExtract.xhtml

devshgraphicsprogramming: btw if you want to "template switch" use a struct  and partial specialization as the implementation behind the scenes

Eichenherz: how do I use bitfieldExtract in hlsl/cpp "common" code ? glsl:: does not implement any cpp functions. Is it ok to do this or do I just move the hlsl that uses it out of the "common" scope ?

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -56,6 +56,18 @@ T atomicCompSwap(NBL_REF_ARG(T) ptr, T comparator, T value)
...
 
+template<typename T>
+T bitfieldExtract( T val, uint32_t offsetBits, uint32_t numBits )
+{
+    static_assert( is_integral<T>::value );
+    if( is_unsigned<T>::value )
+    {
+        return spirv::bitFieldUExtract<T>( val, offsetBits, numBits );
+    }
+    //if( is_signed<T>::value )
```
devshgraphicsprogramming, line 67: else ?

include/nbl/builtin/hlsl/glsl_compat/bit.hlsl
```diff
@@ -0,0 +1,46 @@
...
+        {
+            static_assert( false, "T must be an integral type" );
+        }
+        if( isSigned )
+        {
+            return spirv::bitFieldSExtract<T>( val, offsetBits, numBits );
+        }
+        return spirv::bitFieldUExtract<T>( val, offsetBits, numBits );
+    }
+};
```
devshgraphicsprogramming, line 36: thats not how partial specialization works

devshgraphicsprogramming: also the impl goes into `glsl_compat/core.hlsl`

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -83,6 +83,50 @@ void tess_ctrl_barrier() {
...
+        return spirv::bitFieldUExtract<T>( val, offsetBits, numBits );
+    } 
+};
+
+}
+
+template<typename T>
+T bitfieldExtract( T val, uint32_t offsetBits, uint32_t numBits )
+{
+    return impl::bitfieldExtract<T, is_signed<T>::value, is_integral<T>::value>( val, offsetBits, numBits );
```
devshgraphicsprogramming, line 127: did you actually test your code though, cause I can see with my very own eyes this is invalid syntax.

include/nbl/builtin/hlsl/bit.hlsl
devshgraphicsprogramming: create an alias a for the GLM function that does in the `ifndef __HLSL_VERSION` `else` case

include/nbl/builtin/hlsl/bit.hlsl
```diff
@@ -3,6 +3,7 @@
 
 
 #include <nbl/builtin/hlsl/cpp_compat.hlsl>
+#include <nbl/builtin/hlsl/type_traits.hlsl>
```
devshgraphicsprogramming, line 6: you're not using it here so don't add it

include/nbl/builtin/hlsl/bit.hlsl
devshgraphicsprogramming: ok just provide the C++ alias, and we can merge

Issue #352, opened 2022-05-18
[Proposal] NBL_ASSERT Macros and Cleaning up usages for verification and validation


Erfan-Ahmadi:
## Description
Currently in implementation of nbl functions we use variety of ways to validate inputs and calculations using asserts, ifs and debug breaks.

Problems with the current state of things invlove:
- Most of the time we debug in RelWithDebug and we might not get important asserts and delay fixing an important bug
- The code get's longer, dirty and a less unreadable with a lot of `assert(false)`s 
- There is no concept of a ASSERT Level based on build configuration since we use the `assert` from c
- Sometimes we want to get the ASSERT but be able to continue and not terminate the application (Based on the configuration)
- If someone is not in VisualStudio and not debugging they won't be able to see where the application failed or hit an assert (file, function, line and all)

Of course we can do all above using DEBUG_BREAK, ASSERT, ILogger. but it would require longer and redundant code

Here is some patterns from around the engine:
```c
if (!condition)
{
logger.log("some error happened");
assert(false);
}
```
```c
if (!condition)
{
assert(false);
return false;
}
```

My solution described below fixes the mentioned problems. and in addition brings some beauty to how we get the errors and asserts.

## Solution proposal

We define and `NBL_ASSERT_LEVEL` which can take values 0, 1, 2
`NBL_ASSERT_LEVEL == 2` -> Strictest -> is for DEBUG
`NBL_ASSERT_LEVEL == 1` -> somewhat strict -> is for not DEBUG
`NBL_ASSERT_LEVEL == 0` -> Doesn't do anything on Asserts -> can be overriden by cmake (by defining it first)

Here we define 3 new macros:
```
NBL_ASSERT 
NBL_ASSERT_STRONG
NBL_ASSERT_IF
```

** I'll define what `NblAssert()` function is below

`NBL_ASSERT`: does nothing in level 0 and 1, but does `NblAssert()` for level 2
`NBL_ASSERT_STRONG`: does nothing in level 0, but does `NblAssert()` for level 1 and 2
`NBL_ASSERT_IF`: returns `(condition)` at level 0 and 1, but does `NblAssert()` for 2
this is a special macro for us to be able to use an assert and check for an if at the same time.
```c
if (NBL_ASSERT_IF(VK_RET==VK_SUCCESS))
 return CVulkanSomething;
```

## Assertion Process
`NblAssert()` is our internal static function used for assertions and in the macros above

```c
bool NBL_API
[...]

PR #657, opened 2024-02-19
Add sub-allocated descriptor sets

deprilula28:
## Description
Sister PR: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/95

Adds a sub-allocated descriptor set, which takes a binding that has the required flags and allows for allocating within its range of descriptors.

## Testing 
- Created an example in the sister PR to test its functionality.

## TODO list:

- [x] PR review from old PR: https://github.com/Devsh-Graphics-Programming/Nabla/pull/649#discussion_r1488629923


include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+		return *m_allocatableRanges[binding].addressAllocator; 
+	}
+
+	// main methods
+
+	//! Warning `outAddresses` needs to be primed with `invalid_value` values, otherwise no allocation happens for elements not equal to `invalid_value`
+	template<typename... Args>
+	inline void multi_allocate(uint32_t binding, uint32_t count, value_type* outAddresses, const size_type* sizes, const Args&... args)
+	{
+		core::address_allocator_traits<AddressAllocator>::multi_alloc_addr(getBindingAllocator(binding), count, outAddresses, sizes, 1, args...);
```
devshgraphicsprogramming, line 108: maybe don't use traits, so you can skip the `sizes` completely and just have a `(const uint32_t binding, const uint32_t count, value_type* outAddresses)` signature and no templating

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+	}
+	inline void multi_deallocate(uint32_t binding, uint32_t count, const size_type* addr, const size_type* sizes)
+	{
+		auto& range = m_allocatableRanges[binding];
+		assert(range.reservedSize); // Check if this binding has an allocator
+		core::address_allocator_traits<AddressAllocator>::multi_free_addr(getBindingAllocator(binding), count, addr, sizes);
+	}
```
devshgraphicsprogramming, line 115: you need a functor like this https://github.com/Devsh-Graphics-Programming/Nabla/blob/bd8b9dcd2e1906dff4a6863f88e0874025a33a79/include/nbl/video/alloc/CAsyncSingleBufferSubAllocator.h#L36

which in the `operator()` will wipe/write/update the descriptor bindings that have been freed, and then run the `free_addr` on the bindings

you should only have one dynamic array, which is just to store the list of bindings & indices you want to free

P.S. I recommend adding some handling of descriptor set writes with `nullptr` descriptors, because if you don't do this, the IReferenceCounted in the binding will never get dropped.

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+
+namespace nbl::video
+{
+
+class SubAllocatedDescriptorSet : public core::IReferenceCounted 
+{
+public:
+	// address allocator gives offsets
+	// reserved allocator allocates memory to keep the address allocator state inside
+	using AddressAllocator = core::GeneralpurposeAddressAllocator<uint32_t>;
```
devshgraphicsprogramming, line 20: POOL ADDRESS ALLOCATOR

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+	using ReservedAllocator = core::allocator<uint8_t>;
+	using size_type = typename AddressAllocator::size_type;
+	using value_type = typename AddressAllocator::size_type;
+	static constexpr value_type invalid_value = AddressAllocator::invalid_address;
+
+protected:
+	struct SubAllocDescriptorSetRange {
+		std::shared_ptr<AddressAllocator> addressAllocator;
+		std::shared_ptr<ReservedAllocator> reservedAllocator;
```
devshgraphicsprogramming, line 29: unique_ptr can work just as well

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+		size_t reservedSize;
+	};
+	std::vector<SubAllocDescriptorSetRange> m_allocatableRanges = {};
```
devshgraphicsprogramming, line 32: make it a `map<binding,SEtRange>`

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+				auto binding = redirect.getBinding(i);
+				auto storageIndex = redirect.findBindingStorageIndex(binding);
+
+				auto count = redirect.getCount(storageIndex);
+				auto flags = redirect.getCreateFlags(storageIndex);
+
+				for (uint32_t j = m_allocatableRanges.size(); j < binding.data; j++)
+				{
+					m_allocatableRanges.push_back({});
+				}
```
devshgraphicsprogramming, line 56: just use a map, dont do this

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: not an `unordered_map` the expectation is very few bindings

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+	static constexpr value_type invalid_value = AddressAllocator::invalid_address;
+
+protected:
+	struct SubAllocDescriptorSetRange {
+		std::shared_ptr<AddressAllocator> addressAllocator;
+		std::shared_ptr<ReservedAllocator> reservedAllocator;
+		size_t reservedSize;
+	};
+	std::vector<SubAllocDescriptorSetRange> m_allocatableRanges = {};
+
```
devshgraphicsprogramming, line 33: add a recursive mutex under NBL_DEBUG just like https://github.com/Devsh-Graphics-Programming/Nabla/blob/bd8b9dcd2e1906dff4a6863f88e0874025a33a79/include/nbl/video/alloc/CAsyncSingleBufferSubAllocator.h#L26

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+	~SubAllocatedDescriptorSet()
+	{
+		for (uint32_t i = 0; i < m_allocatableRanges.size(); i++)
+		{
+			auto& range = m_allocatableRanges[i];
+			if (range.reservedSize == 0)
+				continue;
+
+			auto ptr = reinterpret_cast<const uint8_t*>(core::address_allocator_traits<AddressAllocator>::getReservedSpacePtr(*range.addressAllocator));
+			range.reservedAllocator->deallocate(const_cast<uint8_t*>(ptr), range.reservedSize);
```
devshgraphicsprogramming, line 86: you probably want to force the `addressAllocator` destructor to run first before pulling the reserved memory rug from under its feet

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+
+	// amount of bindings in the descriptor set layout used
+	uint32_t getLayoutBindingCount() { return m_allocatableRanges.size(); }
```
devshgraphicsprogramming, line 91: probably not needed

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: I'd probably replace all usages of `Address` with `BindingIndex` for clarity

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: btw your `DeferredFreeFunctor::operator()` needs to come in two overloads:
- `void` that just does the deallocation (note that the destructor DOES NOT call `operator()`)
- `bool` that does same as `void` but also takes a count of allocations we want to free up as a reference, you return true if you freed more than was there or 0, and false if there's some left (don't forget to update the counter)

The reason for this is because of the `poll` and `wait` functors of the event handler, you don't want to poll all events or wait till timeout just because some event on the queue wasn't ready. You want to bail as soon as you've made "enough space"

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: you want to slap a `MultiTimelineEventHandlerST<DeferredFreeFunctor>` per binding here as well

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -91,27 +85,42 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
+	{ 
+		return m_allocatableRanges.find(binding) != m_allocatableRanges.end(); 
+	}
 
 	AddressAllocator& getBindingAllocator(uint32_t binding) 
 	{ 
-		assert(isBindingAllocatable(binding)); // Check if this binding has an allocator
-		return *m_allocatableRanges[binding].addressAllocator; 
+		auto range = m_allocatableRanges.find(binding);
+		assert(range != m_allocatableRanges.end());// Check if this binding has an allocator
```
devshgraphicsprogramming, line 96: you might want to change the signature of the method to return a poitner instead of a reference so you can gracefully handle failures

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -29,13 +29,19 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
 	};
-	std::unordered_map<uint32_t, SubAllocDescriptorSetRange> m_allocatableRanges = {};
+	std::map<uint32_t, SubAllocDescriptorSetRange> m_allocatableRanges = {};
 
+	#ifdef _NBL_DEBUG
+	std::recursive_mutex stAccessVerfier;
+	#endif // _NBL_DEBUG
+
+	constexpr static inline uint32_t MaxDescriptorSetAllocationAlignment = 64u*1024u; // if you need larger alignments then you're not right in the head
+	constexpr static inline uint32_t MinDescriptorSetAllocationSize = 1u;
```
devshgraphicsprogramming, line 39: wuut, you don't need this though....? just use 1 for both, Pool Address allocator really doesn't care

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -75,51 +82,55 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
 			auto& range = m_allocatableRanges[i];
 			if (range.reservedSize == 0)
 				continue;
-
 			auto ptr = reinterpret_cast<const uint8_t*>(core::address_allocator_traits<AddressAllocator>::getReservedSpacePtr(*range.addressAllocator));
+			range.addressAllocator->~PoolAddressAllocator();
```
devshgraphicsprogramming, line 86: this is not how a shared or unique pointer works, you'll crash

just assign nullptr and it will kick in

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -23,26 +23,66 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
+	public:
+		inline DeferredFreeFunctor(SubAllocatedDescriptorSet* composed, uint32_t binding, size_type count, value_type* addresses)
+			: m_addresses(addresses, addresses + count), m_binding(binding), m_composed(composed)
+		{
+		}
+
+		// Just does the de-allocation
+		inline void operator()()
+		{
+			// isn't assert already debug-only?
```
devshgraphicsprogramming, line 37: yes, but Erfan wanted to change our debugbreaks and asserts a bit #352 so doing them like that for extra clarity

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -23,26 +23,66 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
+			#ifdef _NBL_DEBUG
+			assert(m_composed);
+			#endif // _NBL_DEBUG
+			m_composed->multi_deallocate(m_binding, m_addresses.size(), &m_addresses[0]);
```
devshgraphicsprogramming, line 41: style nitpick, use `xxx.data()` instead of `&xxx[0]`

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -23,26 +23,66 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
+		}
+
+		// Takes count of allocations we want to free up as reference, true is returned if
+		// the amount of allocations freed was >= allocationsToFreeUp
+		// False is returned if there are more allocations to free up
+		inline bool operator()(size_type allocationsToFreeUp)
```
devshgraphicsprogramming, line 47: you didn't take it as a reference though!

devshgraphicsprogramming: it needs to be updated because its passed between all `f.operator()`  of all `f` in the queue

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -23,26 +23,66 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
+			auto totalFreed = m_addresses.size() - prevCount;
+
+			// This does the same logic as bool operator()(size_type&) on 
+			// CAsyncSingleBufferSubAllocator
+			return totalFreed >= allocationsToFreeUp;
+		}
+	protected:
+		SubAllocatedDescriptorSet* m_composed;
+		uint32_t m_binding;
+		std::vector<value_type> m_addresses;
```
devshgraphicsprogramming, line 60: order your members from fattest alignment to smallest

I'd use `smart_refctd_dynamic_array`, because its faster to swap/move pointers than vectors (even if you use vector std::move, the state is larger)

P.S. you need to write the move operator for this functor.

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -23,26 +23,66 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
-	constexpr static inline uint32_t MaxDescriptorSetAllocationAlignment = 64u*1024u; // if you need larger alignments then you're not right in the head
+	constexpr static inline uint32_t MaxDescriptorSetAllocationAlignment = 1u; 
 	constexpr static inline uint32_t MinDescriptorSetAllocationSize = 1u;
 
 public:
+
 	// constructors
 	template<typename... Args>
-	inline SubAllocatedDescriptorSet(video::IGPUDescriptorSetLayout* layout)
+	inline SubAllocatedDescriptorSet(video::IGPUDescriptorSet* descriptorSet)
```
devshgraphicsprogramming, line 83: always take single pointers you want to take shared ownership of as `core::smart_refctd_ptr<T>&&` into functions like this

Also remove the `template<typename... Args>`

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -23,26 +23,66 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
+		uint32_t m_binding;
+		std::vector<value_type> m_addresses;
+	};
 protected:
 	struct SubAllocDescriptorSetRange {
 		std::shared_ptr<AddressAllocator> addressAllocator;
 		std::shared_ptr<ReservedAllocator> reservedAllocator;
 		size_t reservedSize;
 	};
+	MultiTimelineEventHandlerST<DeferredFreeFunctor> eventHandler;
```
devshgraphicsprogramming, line 68: The reason why I wrote https://github.com/Devsh-Graphics-Programming/Nabla/pull/657#discussion_r1496617502

Was so that the timelines aren't mixed, basically if you have bindings 0,1,2,3 and binding 1 has 400 deferred free events in the farther past, you don't need to poll/try-free them when you ask for a an allocation from binding 3

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -100,39 +141,67 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
 		}
 	}
+	//!
+	inline void multi_deallocate(const ISemaphore::SWaitInfo& futureWait, DeferredFreeFunctor&& functor) noexcept
+	{
+		auto debugGuard = stAccessVerifyDebugGuard();
+		eventHandler.latch(futureWait,std::move(functor));
+	}
+	// TODO: improve signature of this function in the future
+	template<typename T=core::IReferenceCounted>
```
devshgraphicsprogramming, line 191: doesn't need to be templated, you don't take any extras to kepp alive

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -100,39 +141,67 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
+		if (futureWait.semaphore)
+			multi_deallocate(futureWait, DeferredFreeFunctor(&this, binding, count, addr));
+		else
+			multi_deallocate(binding, count, addr);
+	}
+	//! Returns free events still outstanding
+	inline uint32_t cull_frees() noexcept
+	{
+		auto debugGuard = stAccessVerifyDebugGuard();
+		return eventHandler.poll().eventsLeft;
```
devshgraphicsprogramming, line 203: when doing one handler per binding, go across all and add them up

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -100,39 +141,67 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
 
 		auto allocator = getBindingAllocator(binding);
-		for (uint32_t i=0; i<count; i++)
+		for (size_type i=0; i<count; i++)
 		{
 			if (outAddresses[i]!=AddressAllocator::invalid_address)
 				continue;
 
 			outAddresses[i] = allocator->alloc_addr(1,1);
+			// TODO: should also write something to the descriptor set (or probably leave that to the caller?)
```
devshgraphicsprogramming, line 167: should go outside the loop and happen all at once in bulk.

reserve a vector of `SDescriptorInfo` for one descriptor write and fire it off

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -100,39 +141,67 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
 
 		auto allocator = getBindingAllocator(binding);
-		for (uint32_t i=0; i<count; i++)
+		for (size_type i=0; i<count; i++)
 		{
 			if (addr[i]==AddressAllocator::invalid_address)
 				continue;
 
 			allocator->free_addr(addr[i],1);
+			// TODO: should also write something to the descriptor sets
```
devshgraphicsprogramming, line 181: as we discussed we need a new method for the class that handles descriptor set write/copy, to nullify descriptor set tracking

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,120 @@
...
+	// whether that binding index can be sub-allocated
+	bool isBindingAllocatable(uint32_t binding) { return m_allocatableRanges[binding].reservedSize != 0; }
+
+	AddressAllocator& getBindingAllocator(uint32_t binding) 
+	{ 
+		assert(isBindingAllocatable(binding)); // Check if this binding has an allocator
+		return *m_allocatableRanges[binding].addressAllocator; 
+	}
+
+	// main methods
```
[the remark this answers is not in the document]
devshgraphicsprogramming: outstanding:
- `getDescriptorSet()`
- `multi_allocate` with a timeout point, and handling timeouts on them in general
- implementation details of allocating and freeing stuff

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: ok in-case I wasn't clear, yes... leave that to the caller

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: basically if you have the `nullDescriptor` device feature enabled, you would indeed write to the DS, else you'd just drop the refcounted references

this needs to be done as a `IGPUDescriptorSet::nullify(const uint32_t binding, std::span<uint32_t> indices)` function + a virtual nullify_impl

include/nbl/video/IGPUDescriptorSet.h
```diff
@@ -45,6 +45,14 @@ class IGPUDescriptorSet : public asset::IDescriptorSet<const IGPUDescriptorSetLa
             uint32_t count;
         };
 
+        struct SDropDescriptorSet
+        {
+            IGPUDescriptorSet* dstSet;
```
devshgraphicsprogramming, line 50: you already know the `dstSet` you'll drop from by calling `this->dropDescriptors`, its the `this`

devshgraphicsprogramming: also call it nullify, cause we not only drop, we nullify as well if we can, they're not supposed to be separate ops

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -180,7 +180,13 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
 	video::IGPUDescriptorSet* getDescriptorSet() { return m_descriptorSet.get(); }
 
 	//! Warning `outAddresses` needs to be primed with `invalid_value` values, otherwise no allocation happens for elements not equal to `invalid_value`
-	inline size_type try_multi_allocate(uint32_t binding, size_type count, video::IGPUDescriptorSet::SDescriptorInfo* descriptors, value_type* outAddresses)
+	inline size_type try_multi_allocate(
+		uint32_t binding, 
+		size_type count, 
+		video::IGPUDescriptorSet::SDescriptorInfo* descriptors, 
+		video::IGPUDescriptorSet::SWriteDescriptorSet* outDescriptorWrites,
```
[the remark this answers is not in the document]
devshgraphicsprogramming: its a conflation of responsibility to have the allocator set these up

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -217,17 +223,21 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
 				// can we change it?
 				write.info = &descriptor;
 			}
-			infos.push_back(descriptor);
-			writes.push_back(write);
+			outDescriptorWrites[i] = write;
```
devshgraphicsprogramming, line 228: why not call the time out version with a timepoint of `now()`

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -248,22 +258,34 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
 		{
 			eventHandler.wait(maxWaitPoint, unallocatedSize);
 
-			unallocatedSize = try_multi_allocate(binding, unallocatedSize, &descriptors[count - unallocatedSize], &outAddresses[count - unallocatedSize]);
+			unallocatedSize = try_multi_allocate(
+				binding, 
+				unallocatedSize, 
+				&descriptors[count - unallocatedSize], 
+				&outDescriptorWrites[count - unallocatedSize], 
+				&outAddresses[count - unallocatedSize]
```
devshgraphicsprogramming, line 268: another reason not to take a list of descriptors, writes and addresses to allocate

gaps might be in weird places, you're  sure to write them out wrong

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: ah ok there's an oberload of `multi_allocate` that does this, then the `try_multi_allocate` should be `protected`

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -285,6 +307,12 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
 			//
 			// this needs to be done as a IGPUDescriptorSet::nullify(const uint32_t binding, 
 			// std::span<uint32_t> indices) function + a virtual nullify_impl
+			video::IGPUDescriptorSet::SDropDescriptorSet dropDescriptorSet;
+			dropDescriptorSet.dstSet = m_descriptorSet.get();
+			dropDescriptorSet.binding = binding;
+			dropDescriptorSet.arrayElement = i;
+			dropDescriptorSet.count = 1;
+			m_logicalDevice->nullifyDescriptors({ &dropDescriptorSet, 1 });
```
devshgraphicsprogramming: there's a way to make this slightly more efficient, which is to do what I tell you NOT to do in the `multi_allocate` which is to take the `SDropDescriptorSet` (which should be called `SNullifyDescriptors`) pointer from outside, and just write out the dops here and submit the `nullify` outside

This would allow you to do just one call to `nullifyDescriptors` in `cull_frees` and `try_multi_allocate`

You'd obviously need an override that calls it automagically if no arguments supplied

include/nbl/video/IGPUDescriptorSet.h
devshgraphicsprogramming: hmm we wouldn't need the fast path if https://github.com/Devsh-Graphics-Programming/Nabla/pull/657/files/d608e780494b0a5bde138c9ee679269f3d2b2c59..894a47cd862f678c92991a1ef1861920b4d91f6d#r1512336030 was done

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -124,6 +124,29 @@ void IGPUDescriptorSet::processWrite(const IGPUDescriptorSet::SWriteDescriptorSe
...
+        if (dstDescriptors)
+            for (uint32_t i = 0; i < drop.count; i++)
+                dstDescriptors[drop.arrayElement + i] = nullptr;
+
+        if (dstSamplers)
+            for (uint32_t i = 0; i < drop.count; i++)
+                dstSamplers[drop.arrayElement + i] = nullptr;
+    }
+
+    incrementVersion();
```
devshgraphicsprogramming, line 147: only increment the version in the `_impl`, reason is that we only increment the version to detect UPDATE-AFTER-BIND and automagically invalidate descriptor sets. If there's no `nullDescriptor` feature (you can check outside of `CVulkanLogicalDevice` and gate the call to `_impl` with that), there will be no write to the underlying desc set.

Leave a comment about this.

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -124,6 +124,29 @@ void IGPUDescriptorSet::processWrite(const IGPUDescriptorSet::SWriteDescriptorSe
...
+void IGPUDescriptorSet::dropDescriptors(const IGPUDescriptorSet::SDropDescriptorSet& drop)
+{
+    assert(drop.dstSet == this);
+
+    for (uint32_t t = 0; t < static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_COUNT); ++t)
+    {
+        const auto type = static_cast<asset::IDescriptor::E_TYPE>(t);
+
+        auto* dstDescriptors = drop.dstSet->getDescriptors(type, drop.binding);
+        auto* dstSamplers = drop.dstSet->getMutableSamplers(drop.binding);
```
devshgraphicsprogramming, line 136: have the same asserts and checks as the `processWrite` function above

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -420,6 +420,14 @@ bool ILogicalDevice::updateDescriptorSets(const std::span<const IGPUDescriptorSe
     return true;
 }
 
+void ILogicalDevice::nullifyDescriptors(const std::span<const IGPUDescriptorSet::SDropDescriptorSet> dropDescriptors)
+{
+    for (const auto& drop : dropDescriptors)
```
devshgraphicsprogramming, line 449: you need the same validation that `updateDescriptorSets` has for the writes, insofar as its applicable

Function should also return a bool of whether we've had success or not

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: nevermind, I see the overload has a default time point for the wait. 

Discard everything I said, resolved.

include/nbl/video/IGPUDescriptorSet.h
devshgraphicsprogramming: Complete suggestion here: https://github.com/Devsh-Graphics-Programming/Nabla/pull/662

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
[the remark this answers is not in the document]
devshgraphicsprogramming: done in https://github.com/Devsh-Graphics-Programming/Nabla/pull/662

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: done in https://github.com/Devsh-Graphics-Programming/Nabla/pull/662

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -47,6 +51,28 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
 			m_composed->m_totalDeferredFrees -= getWorstCaseCount();
 		}
 
+		DeferredFreeFunctor(const DeferredFreeFunctor& other) = delete;
+		DeferredFreeFunctor& operator=(const DeferredFreeFunctor& other) = delete;
+		inline DeferredFreeFunctor& operator=(DeferredFreeFunctor&& other)
+		{
+			m_composed = other.m_composed;
+			m_addresses = other.m_addresses;
+			m_binding = other.m_binding;
```
devshgraphicsprogramming, line 60: you need to add the nullification of `other` here, don't ask why, its just weird that way

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -47,6 +51,28 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
...
+		// This is needed for the destructor of TimelineEventHandlerST
+		// Don't call this directly
+		// TODO: Find a workaround for this
+		inline void operator()()
+		{
+			core::vector<IGPUDescriptorSet::SDropDescriptorSet> nulls(m_addresses->size());
+			auto ptr = nulls.data();
+			operator()(ptr);
+			auto size = ptr - nulls.data();
+			m_composed->m_logicalDevice->nullifyDescriptors({nulls.data(),size_type(size)});
```
devshgraphicsprogramming, line 73: let it be, but assert(false) at the top, then you'll know you were supposed to drain the timeline first

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -0,0 +1,366 @@
...
+			reservedSize = other.reservedSize;
+			descriptorType = other.descriptorType;
+
+			// Nullify other
+			other.addressAllocator = nullptr;
+			other.reservedAllocator = nullptr;
+			other.reservedSize = 0u;
+			other.descriptorType = asset::IDescriptor::E_TYPE::ET_COUNT;
+			return *this;
+		}
```
devshgraphicsprogramming: as a matter of good practice you should `=delete` the `const&` assignment and constructor (and really any ctor/assignment you're not using)

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -124,6 +124,35 @@ void IGPUDescriptorSet::processWrite(const IGPUDescriptorSet::SWriteDescriptorSe
...
+		assert(dstDescriptors);
+		assert(dstSamplers);
```
devshgraphicsprogramming: see that we only require this if 
```cpp
if (type==asset::IDescriptor::E_TYPE::ET_COMBINED_IMAGE_SAMPLER && write.info->info.image.sampler)
```
in  the above functions!

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -124,6 +124,35 @@ void IGPUDescriptorSet::processWrite(const IGPUDescriptorSet::SWriteDescriptorSe
...
+
+        if (dstDescriptors)
```
devshgraphicsprogramming: if you assert that `dstDescriptors` are not null, then you shouldn't if them

src/nbl/video/IGPUDescriptorSet.cpp
devshgraphicsprogramming: I'm dumb for reviewing old code and not refreshing my browser tab

src/nbl/video/IGPUDescriptorSet.cpp
devshgraphicsprogramming: I'm dumb for reviewing old code and not refreshing my browser tab

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -124,6 +124,33 @@ void IGPUDescriptorSet::processWrite(const IGPUDescriptorSet::SWriteDescriptorSe
...
+    assert(drop.dstSet == this);
+
+	auto* dstDescriptors = drop.dstSet->getDescriptors(type, drop.binding);
+	auto* dstSamplers = drop.dstSet->getMutableSamplers(drop.binding);
+	assert(dstDescriptors);
+    // Samplers are only used in the combined image sampler descriptors
+    if (type == asset::IDescriptor::E_TYPE::ET_COMBINED_IMAGE_SAMPLER)
+	    assert(dstSamplers); 
+
+	if (dstDescriptors)
```
devshgraphicsprogramming, line 138: can;t have it both ways, either assert and don't if `dstDescriptors`, or if and don't assert

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -727,6 +727,76 @@ void CVulkanLogicalDevice::updateDescriptorSets_impl(const SUpdateDescriptorSets
     m_devf.vk.vkUpdateDescriptorSets(m_vkdev,vk_writeDescriptorSets.size(),vk_writeDescriptorSets.data(),vk_copyDescriptorSets.size(),vk_copyDescriptorSets.data());
 }
 
+void CVulkanLogicalDevice::nullifyDescriptors_impl(const std::span<const IGPUDescriptorSet::SDropDescriptorSet> drops, asset::IDescriptor::E_TYPE descriptorType)
+{
+    if (getEnabledFeatures().nullDescriptor)
+    {
```
devshgraphicsprogramming, line 733: my preferred way of bailing is returning at the start, instead of nesting oneself in extra `{}`, style nitpick

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -727,6 +727,76 @@ void CVulkanLogicalDevice::updateDescriptorSets_impl(const SUpdateDescriptorSets
...
+			case asset::IDescriptor::EC_IMAGE:
+                descriptorSize = sizeof(VkDescriptorImageInfo);
+                break;
+			case asset::IDescriptor::EC_BUFFER_VIEW:
+                descriptorSize = sizeof(VkBufferView);
+                break;
+			case asset::IDescriptor::EC_ACCELERATION_STRUCTURE:
+                descriptorSize = sizeof(VkAccelerationStructureKHR);
+                break;
+        }
```
devshgraphicsprogramming, line 758: what on earth is going on with your tab vs spaces?

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -727,6 +727,76 @@ void CVulkanLogicalDevice::updateDescriptorSets_impl(const SUpdateDescriptorSets
     m_devf.vk.vkUpdateDescriptorSets(m_vkdev,vk_writeDescriptorSets.size(),vk_writeDescriptorSets.data(),vk_copyDescriptorSets.size(),vk_copyDescriptorSets.data());
 }
 
+void CVulkanLogicalDevice::nullifyDescriptors_impl(const std::span<const IGPUDescriptorSet::SDropDescriptorSet> drops, asset::IDescriptor::E_TYPE descriptorType)
+{
+    if (getEnabledFeatures().nullDescriptor)
+    {
+		core::vector<VkWriteDescriptorSet> vk_writeDescriptorSets(drops.size(),{VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET,nullptr});
+		core::vector<VkWriteDescriptorSetAccelerationStructureKHR> vk_writeDescriptorSetAS(69u,{VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET_ACCELERATION_STRUCTURE_KHR,nullptr});
```
devshgraphicsprogramming, line 735: nice literal constant.

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -727,6 +727,76 @@ void CVulkanLogicalDevice::updateDescriptorSets_impl(const SUpdateDescriptorSets
     m_devf.vk.vkUpdateDescriptorSets(m_vkdev,vk_writeDescriptorSets.size(),vk_writeDescriptorSets.data(),vk_copyDescriptorSets.size(),vk_copyDescriptorSets.data());
 }
 
+void CVulkanLogicalDevice::nullifyDescriptors_impl(const std::span<const IGPUDescriptorSet::SDropDescriptorSet> drops, asset::IDescriptor::E_TYPE descriptorType)
```
devshgraphicsprogramming, line 730: thats some broken design, taking a span of drops and a type!?

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -727,6 +727,76 @@ void CVulkanLogicalDevice::updateDescriptorSets_impl(const SUpdateDescriptorSets
...
+                descriptorSize = sizeof(VkDescriptorBufferInfo);
+                break;
+			case asset::IDescriptor::EC_IMAGE:
+                descriptorSize = sizeof(VkDescriptorImageInfo);
+                break;
+			case asset::IDescriptor::EC_BUFFER_VIEW:
+                descriptorSize = sizeof(VkBufferView);
+                break;
+			case asset::IDescriptor::EC_ACCELERATION_STRUCTURE:
+                descriptorSize = sizeof(VkAccelerationStructureKHR);
```
devshgraphicsprogramming, line 756: btw if you look athe the  validation params for copies/updates they gather these counts in advance during validation

src/nbl/video/CVulkanLogicalDevice.cpp
deprilula28: I just noticed that, I think I copied it from the update descriptors one

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -444,6 +444,21 @@ bool ILogicalDevice::updateDescriptorSets(const std::span<const IGPUDescriptorSe
...
 
+bool ILogicalDevice::nullifyDescriptors(const std::span<const IGPUDescriptorSet::SDropDescriptorSet> dropDescriptors, asset::IDescriptor::E_TYPE descriptorType)
+{
+    for (const auto& drop : dropDescriptors)
+    {
+        auto ds = drop.dstSet;
+        if (!ds || !ds->wasCreatedBy(this))
+            return false;
+
+        ds->dropDescriptors(drop, descriptorType);
```
devshgraphicsprogramming, line 455: do all validation first, then drop them, right now if you have a validation error midway you'll end up dropping some and not others

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -444,6 +444,21 @@ bool ILogicalDevice::updateDescriptorSets(const std::span<const IGPUDescriptorSe
     return true;
 }
 
+bool ILogicalDevice::nullifyDescriptors(const std::span<const IGPUDescriptorSet::SDropDescriptorSet> dropDescriptors, asset::IDescriptor::E_TYPE descriptorType)
+{
+    for (const auto& drop : dropDescriptors)
+    {
+        auto ds = drop.dstSet;
+        if (!ds || !ds->wasCreatedBy(this))
+            return false;
```
devshgraphicsprogramming, line 453: you need to validate other things, see the `updateDescriptorSets`

devshgraphicsprogramming: like the binding existing, etc.

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -124,6 +124,33 @@ void IGPUDescriptorSet::processWrite(const IGPUDescriptorSet::SWriteDescriptorSe
...
+void IGPUDescriptorSet::dropDescriptors(const IGPUDescriptorSet::SDropDescriptorSet& drop, const asset::IDescriptor::E_TYPE type)
+{
+    assert(drop.dstSet == this);
+
+	auto* dstDescriptors = drop.dstSet->getDescriptors(type, drop.binding);
+	auto* dstSamplers = drop.dstSet->getMutableSamplers(drop.binding);
+	assert(dstDescriptors);
+    // Samplers are only used in the combined image sampler descriptors
+    if (type == asset::IDescriptor::E_TYPE::ET_COMBINED_IMAGE_SAMPLER)
+	    assert(dstSamplers);
```
devshgraphicsprogramming, line 136: assert may compile into nothingness, this leaves you with a kind-of dangling scope, use `{}` around the assert

include/nbl/video/ILogicalDevice.h
```diff
@@ -632,6 +632,9 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
             return updateDescriptorSets({pDescriptorWrites,descriptorWriteCount},{pDescriptorCopies,descriptorCopyCount});
         }
 
+        // should this be joined together with the existing updateDescriptorSets?
+        bool nullifyDescriptors(const std::span<const IGPUDescriptorSet::SDropDescriptorSet> dropDescriptors, asset::IDescriptor::E_TYPE descriptorType);
```
devshgraphicsprogramming, line 636: no don't join, cause `updateDescriptorSets` can't write null descs and takes a different parameter struct

Making this take a `descriptorType` while allowing for a full `SDropDescriptorSet` is some seriously broken design

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -635,12 +635,15 @@ core::smart_refctd_ptr<IDescriptorPool> CVulkanLogicalDevice::createDescriptorPo
...
+// a lot of empirical research went into defining this constant
+constexpr uint32_t MaxDescriptorSetAsWrites = 69u;
+
 void CVulkanLogicalDevice::updateDescriptorSets_impl(const SUpdateDescriptorSetsParams& params)
 {
     // Each pNext member of any structure (including this one) in the pNext chain must be either NULL or a pointer to a valid instance of
     // VkWriteDescriptorSetAccelerationStructureKHR, VkWriteDescriptorSetAccelerationStructureNV, or VkWriteDescriptorSetInlineUniformBlockEXT
     core::vector<VkWriteDescriptorSet> vk_writeDescriptorSets(params.writes.size(),{VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET,nullptr});
-    core::vector<VkWriteDescriptorSetAccelerationStructureKHR> vk_writeDescriptorSetAS(69u,{VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET_ACCELERATION_STRUCTURE_KHR,nullptr});
+    core::vector<VkWriteDescriptorSetAccelerationStructureKHR> vk_writeDescriptorSetAS(MaxDescriptorSetAsWrites,{VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET_ACCELERATION_STRUCTURE_KHR,nullptr});
```
devshgraphicsprogramming, line 646: wouldn't it just make sense to add a `asCount` to `SUpdateDescriptorSetsParams` ?

devshgraphicsprogramming: there's a thing already there.... `accelerationStructureCount`

thats the the total count of `infos` (count of items in an array binding), we'd just have to count the writes themselves as well

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -727,40 +730,41 @@ void CVulkanLogicalDevice::updateDescriptorSets_impl(const SUpdateDescriptorSets
...
-        case asset::IDescriptor::EC_ACCELERATION_STRUCTURE:
+		case asset::IDescriptor::EC_ACCELERATION_STRUCTURE:
 			descriptorSize = sizeof(VkAccelerationStructureKHR);
 			break;
+		}
+		maxSize = core::max(maxSize, write.count * descriptorSize);
 	}
 
-	core::vector<uint8_t> nullDescriptors(maxSize * descriptorSize, 0u);
+	core::vector<uint8_t> nullDescriptors(maxSize, 0u);
```
devshgraphicsprogramming, line 767: count all the reservations you need outside, during validation

include/nbl/video/alloc/SubAllocatedDescriptorSet.h
devshgraphicsprogramming: please don't mark things as resolved yourself, makes me loose track, adn you still don't set `other.m_ = nullptr`

PR #658, opened 2024-02-20
Task central limit box blur

Eichenherz:
## Description
HLSL implementation of the central limit box blur.


include/nbl/builtin/hlsl/cpp_compat/intrinsics.h
```diff
@@ -71,12 +71,18 @@ inline matrix<T,M,N> transpose(const matrix<T,N,M>& m)
     return glm::transpose(reinterpret_cast<typename matrix<T,N,M>::Base const&>(m));
 }
 
+namespace glsl // TODO ?
+{
```
devshgraphicsprogramming, line 75: in this case the alias goes into the #else case in this file
https://github.com/Devsh-Graphics-Programming/Nabla/blob/7d90baa1d8009f5c0040b74deb0d4b8f5e3f9584/include/nbl/builtin/hlsl/glsl_compat/core.hlsl#L130

.gitmodules
```diff
@@ -73,7 +73,7 @@
 	branch = main
 [submodule "examples_tests"]
 	path = examples_tests
-	url = git@github.com:Devsh-Graphics-Programming/Nabla-Examples-and-Tests.git
+	url = git@github.com:Eichenherz/Nabla-Examples-and-Tests.git
```
devshgraphicsprogramming, line 76: reminder to undo before we merge

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -124,8 +124,15 @@ struct bitfieldExtract<T, false, true>
...
-        ( val, offsetBits, numBits );
+    return impl::bitfieldExtract<T, is_signed<T>::value, is_integral<T>::value>::__call( val, offsetBits, numBits );
+}
+
+#else
+
+template<typename T>
+inline T bitfieldExtract( T val, uint32_t offsetBits, uint32_t numBits )
+{
+    return glm::bitfieldExtract( val, int32_t( offsetBits ), int32_t( numBits ) );
```
devshgraphicsprogramming, line 135: if GLM's version is templated, please provide the `<T>` explicitly

include/nbl/builtin/hlsl/blur/common.hlsl
```diff
@@ -0,0 +1,87 @@
...
+
+enum BorderColor : uint32_t
+{
+	BORDER_COLOR_FLOAT_TRANSPARENT_BLACK,
+	BORDER_COLOR_INT_TRANSPARENT_BLACK,
+	BORDER_COLOR_FLOAT_OPAQUE_BLACK,
+	BORDER_COLOR_INT_OPAQUE_BLACK,
+	BORDER_COLOR_FLOAT_OPAQUE_WHITE,
+	BORDER_COLOR_INT_OPAQUE_WHITE
+};
```
devshgraphicsprogramming, line 30: maybe lets move the same enums defined in `nbl::asset` to a common HLSL header and alias the old enums in ISampler to those?

include/nbl/builtin/hlsl/blur/common.hlsl
```diff
@@ -0,0 +1,87 @@
+#pragma once
```
devshgraphicsprogramming, line 1: no point, you already have a header guard

include/nbl/builtin/hlsl/blur/common.hlsl
```diff
@@ -0,0 +1,87 @@
...
+
+#ifndef _NBL_BOX_BLUR_COMMONS_
+#define _NBL_BOX_BLUR_COMMONS_
```
devshgraphicsprogramming, line 4: you should have the full namespace in the header define, see other HLSL header files

include/nbl/builtin/hlsl/blur/common.hlsl
```diff
@@ -0,0 +1,87 @@
...
+
+#include <nbl/builtin/hlsl/cpp_compat.hlsl>
+#include <nbl/builtin/hlsl/cpp_compat/intrinsics.h>
+#include <nbl/builtin/hlsl/glsl_compat/core.hlsl>
+
+//_NBL_STATIC_INLINE_CONSTEXPR uint32_t WORKGROUP_SIZE = 256u;
+//_NBL_STATIC_INLINE_CONSTEXPR uint32_t PASSES_PER_AXIS = 3u;
+//_NBL_STATIC_INLINE_CONSTEXPR uint32_t AXIS_DIM = 3u; // HUH ?
```
devshgraphicsprogramming, line 12: no longer needed.

But `AXIS_DIM` means if you have a 1D, 2D or 3D target

include/nbl/builtin/hlsl/blur/common.hlsl
devshgraphicsprogramming: actually you're missing a namespace for the whole thing, I guess it should be `nbl::hlsl::central_limit_blur`

the folder should be `central_limit_blur` instead of just `blur`

include/nbl/builtin/hlsl/blur/common.hlsl
```diff
@@ -0,0 +1,87 @@
...
+
+    uint32_t getWrapMode()
+    {
+        return nbl::hlsl::glsl::bitfieldExtract( inputDimensions.w, 25, 2 );
+    }
+
+    uint32_t getBorderColorType()
+    {
+        return nbl::hlsl::glsl::bitfieldExtract( inputDimensions.w, 22, 3 );
+    }
```
devshgraphicsprogramming, line 55: please don't use the old GLSL packing, HLSL has bitfields now

Also we can use `uint16_t` and scalar everywhere, so:
- have the input dimensions as a `uint16_t3`
- then the rest except the radius, packed into a sinlge `uint16_t`

Can you remind me why `chosenAxis` wastes 2 full 32bit uints?

include/nbl/builtin/hlsl/blur/common.hlsl
```diff
@@ -0,0 +1,87 @@
...
+		}
+		return borderColor;
+	}
+};
+
+//struct BufferAccessor;
+/*
+{
+	nbl::hlsl::float32_t getPaddedData( const nbl::hlsl::uint32_t3 coordinate, const uint32_t channel );
+	void setData( const nbl::hlsl::uint32_t3 coordinate, const uint32_t channel, const nbl::hlsl::float32_t val );
```
devshgraphicsprogramming, line 84: you can actually make the coordinates single dimensional, because every axis-pass can declare a different accessor

before we didn't have Static Polymorphism, hence the argument spam

include/nbl/builtin/hlsl/blur/box_blur.hlsl
```diff
@@ -0,0 +1,318 @@
...
+
+#include <nbl/builtin/hlsl/blur/common.hlsl>
+
+#include <nbl/builtin/hlsl/glsl_compat/core.hlsl>
+#include <nbl/builtin/hlsl/workgroup/arithmetic.hlsl>
+#include <nbl/builtin/hlsl/workgroup/scratch_size.hlsl>
+
+
+// https://github.com/microsoft/DirectXShaderCompiler/issues/6144
+static nbl::hlsl::uint32_t3 /*nbl::hlsl::glsl::*/gl_WorkGroupSize() { return uint32_t3( WORKGROUP_SIZE, 1, 1 ); }
```
devshgraphicsprogramming, line 11: you can forward declare this function and just let the user define it, instead of having a `WORKGROUP_SIZE` define we need to populate later

include/nbl/builtin/hlsl/blur/box_blur.hlsl
```diff
@@ -0,0 +1,318 @@
+#pragma once
```
devshgraphicsprogramming, line 1: header guard with defines instead, also namespace the whole thing

include/nbl/builtin/hlsl/blur/box_blur.hlsl
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	void workgroupExecutionAndMemoryBarrier()
+	{
+		nbl::hlsl::glsl::barrier();
+	}
+};
+
+ScratchProxy<float32_t, 0> prefixSumsAccessor; 
+ScratchProxy<float32_t, arithmeticSz> broadcastAccessor;
```
devshgraphicsprogramming, line 41: this whole thing belongs in the example/user-space code

include/nbl/builtin/hlsl/blur/box_blur.hlsl
```diff
@@ -0,0 +1,318 @@
...
+#include <nbl/builtin/hlsl/glsl_compat/core.hlsl>
+#include <nbl/builtin/hlsl/workgroup/arithmetic.hlsl>
+#include <nbl/builtin/hlsl/workgroup/scratch_size.hlsl>
+
+
+// https://github.com/microsoft/DirectXShaderCompiler/issues/6144
+static nbl::hlsl::uint32_t3 /*nbl::hlsl::glsl::*/gl_WorkGroupSize() { return uint32_t3( WORKGROUP_SIZE, 1, 1 ); }
+
+static const uint32_t ITEMS_PER_THREAD = ( AXIS_DIM + WORKGROUP_SIZE - 1 ) / WORKGROUP_SIZE;
+static const uint32_t ITEMS_PER_WG = ITEMS_PER_THREAD * WORKGROUP_SIZE;
```
devshgraphicsprogramming, line 14: you're computing it wrong, the arithmetic will always prefix sum `WORKGROUP_SIZE` elements, because you'll join up the prefix sums yourself

Now you're making us use more shared memory than is necessary.

devshgraphicsprogramming: Also I'm not sure you need `ITEMS_PER_THREAD` anywhere.

include/nbl/builtin/hlsl/blur/box_blur.hlsl
```diff
@@ -0,0 +1,318 @@
...
+( IMPL_LOCAL_SPILLAGE > ITEMS_PER_THREAD ) ? ITEMS_PER_THREAD :
+	( ( IMPL_LOCAL_SPILLAGE < ITEMS_PER_THREAD ) ? SPILLAGE_LOWER_BOUND : IMPL_LOCAL_SPILLAGE );
+
+
+uint32_t DataIndex( uint32_t i ) // ?
+{
+	return ( i * gl_WorkGroupSize().x ) + nbl::hlsl::glsl::gl_LocalInvocationIndex();
+}
+
+//TODO: glsl's mod(x, y) = x - y * floor(x/y). Would prolly need to implement a glsl mod
```
devshgraphicsprogramming, line 101: create a SPV intrinsic for it, you will have to include `unified/glsl.std450.hpp` or something like that (and add it to Nabla builtins - see how the core spirv header was added in CMake)

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+
+#include <nbl/builtin/hlsl/central_limit_blur/common.hlsl>
+
+#include <nbl/builtin/hlsl/glsl_compat/core.hlsl>
+#include <nbl/builtin/hlsl/workgroup/arithmetic.hlsl>
+#include <nbl/builtin/hlsl/workgroup/scratch_size.hlsl>
+
+
+// https://github.com/microsoft/DirectXShaderCompiler/issues/6144
+static nbl::hlsl::uint32_t3 /*nbl::hlsl::glsl::*/gl_WorkGroupSize() { return uint32_t3( WORKGROUP_SIZE, 1, 1 ); }
```
devshgraphicsprogramming, line 12: we even fwd declare this function in some header I think, maybe the `glsl_compat` stuff?

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+	( ( IMPL_LOCAL_SPILLAGE < WORKGROUP_SIZE ) ? SPILLAGE_LOWER_BOUND : IMPL_LOCAL_SPILLAGE );
+
+
+uint32_t DataIndex( uint32_t i ) // ?
+{
+	return ( i * gl_WorkGroupSize().x ) + nbl::hlsl::glsl::gl_LocalInvocationIndex();
+}
+
+//TODO: glsl's mod(x, y) = x - y * floor(x/y). Would prolly need to implement a glsl mod
+void BoxBlur(
```
devshgraphicsprogramming, line 90: rollup every free floating function and constant into a member in
```cpp
template<uint16_t WorkgroupSize, typename device_capabilities_t=void>
struct Blur
```

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+// This:
+// `#define LOCAL_SPILLAGE ITEMS_PER_THREAD`
+// seems to be a safe lower bound, however, it must be inefficient.
+static const uint32_t SPILLAGE_LOWER_BOUND = ( WORKGROUP_SIZE / 4 );
+
+static const uint32_t IMPL_LOCAL_SPILLAGE = ( ( arithmeticSz - 1 ) / WORKGROUP_SIZE + 1 );
+
+static const uint32_t LOCAL_SPILLAGE =
+( IMPL_LOCAL_SPILLAGE > WORKGROUP_SIZE ) ? WORKGROUP_SIZE :
+	( ( IMPL_LOCAL_SPILLAGE < WORKGROUP_SIZE ) ? SPILLAGE_LOWER_BOUND : IMPL_LOCAL_SPILLAGE );
```
devshgraphicsprogramming, line 81:
```cpp
NBL_CONSTEXPR_STATIC_INLINE bool needToSpill = MaxItemCount>WorkgroupSize;
NBL_CONSTEXPR_STATIC_INLINE uint16_t spillCount = (MaxItemCount-1)/WorkgroupSize+1;
```

no clue if we have an `nbl::hlsl::mpl` version of a divide-and-round/divide-and-ceil

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+
+
+uint32_t DataIndex( uint32_t i ) // ?
+{
+	return ( i * gl_WorkGroupSize().x ) + nbl::hlsl::glsl::gl_LocalInvocationIndex();
```
devshgraphicsprogramming, line 86: no, you're not guaranteed that LocalInvocationsIndex maps to subgroups properly, do what @kpentaris does and use the custom mapping ex 23 / workgroup prefix sum uses

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+	}
+};
+
+prefix_sum_t workgroupPrefixSum;
+
+float32_t workgroupBroadcast( NBL_CONST_REF_ARG( float32_t ) value, uint32_t index )
+{
+	using accessor_t = ScratchProxy<float32_t, arithmeticSz>;
+	return nbl::hlsl::workgroup::Broadcast<float32_t, accessor_t>( value, broadcastAccessor, index );
+}
```
devshgraphicsprogramming, line 68: no point wrapping this up IMHO

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+	type_t operator()( NBL_CONST_REF_ARG( type_t ) value )
+	{
+		type_t retval = nbl::hlsl::workgroup::inclusive_scan<bin_op_t, arithmeticSz>::template __call<ScratchProxy<type_t, 0> >( value, prefixSumsAccessor );
+		// we barrier before because we alias the accessors for Binop
+		prefixSumsAccessor.workgroupExecutionAndMemoryBarrier();
+		return retval;
+	}
+};
+
+prefix_sum_t workgroupPrefixSum;
```
devshgraphicsprogramming, line 62: don't leave globals like this laying around in headers, also the acessor needs to be configurable, so there's literally nothing you do here, except choose your bin_op and type_t 

But then this wrapper needs to move into a general struct `Blur` that encompasses everything

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
devshgraphicsprogramming: `workgroup::SubgroupContiguousIndex()`

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+struct prefix_sum_t
+{
+	using bin_op_t = nbl::hlsl::plus<float32_t>;
+	using type_t = bin_op_t::type_t;
+
+	type_t operator()( NBL_CONST_REF_ARG( type_t ) value )
+	{
+		type_t retval = nbl::hlsl::workgroup::inclusive_scan<bin_op_t, arithmeticSz>::template __call<ScratchProxy<type_t, 0> >( value, prefixSumsAccessor );
+		// we barrier before because we alias the accessors for Binop
+		prefixSumsAccessor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 57: you  don't alias the accessors though!

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+	float32_t blurred = textureAccessor.get( DataIndex( i ), ch );
+
+	for( uint32_t pass = 0u; pass < PASSES_PER_AXIS; ++pass )
+	{
+		float32_t previousBlockSum = 0.f;
+
+		float32_t spill[ LOCAL_SPILLAGE ];
+		for( uint32_t i = 0u; i < LOCAL_SPILLAGE; ++i )
+		{
+			float32_t scanResult = workgroupPrefixSum( blurred ) + previousBlockSum;
```
devshgraphicsprogramming, line 106: there's a more clever/efficient way to do this
```cpp
if (workgroup::SubgroupContiguousIndex()==0)
   blurred += previousBlockSum;
scanResult = workgroupPrefixSum(blurred);
```

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+			spill[ i ] = scanResult;
```
devshgraphicsprogramming, line 107: I'd unroll the loop for the `i=0` case, and start from `i=1` so you can skip an add.

Then you can move the broadcast to the start of the loop and read from `spill[i-1]`

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+			previousBlockSum = workgroupBroadcast( spill[ i ], gl_WorkGroupSize().x - 1u );
```
devshgraphicsprogramming, line 108: there's no need to broadcast to everyone, its enough to write out the value to memory for invocation `workgroup::SubgroupContiguousIndex()==WorkgroupSize-1` then only read it by `workgroup::SubgroupContiguousIndex()==0` (see https://github.com/Devsh-Graphics-Programming/Nabla/pull/658/files#r151329896)

You'd still need a barrier between the get and the set

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+		{
+			uint32_t idx = DataIndex( i ); // ?
+			prefixSumsAccessor.set( idx, spill[ i ] );
+		}
+		prefixSumsAccessor.workgroupExecutionAndMemoryBarrier();
+
+		uint32_t idx = DataIndex( i );
+
+		if( idx < AXIS_DIM )
+		{
```
devshgraphicsprogramming, line 127: what is this if statement even?

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+			const float32_t radius = inRadius * AXIS_DIM;
```
devshgraphicsprogramming, line 128: what is this `*AXIS_DIM` even!?

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+			float32_t left = float32_t( idx ) - radius - 1.f;
+			float32_t right = float32_t( idx ) + radius;
+			const uint32_t last = AXIS_DIM - 1;
```
devshgraphicsprogramming, line 131: !?

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+{
+	return ( i * gl_WorkGroupSize().x ) + nbl::hlsl::glsl::gl_LocalInvocationIndex();
+}
+
+//TODO: glsl's mod(x, y) = x - y * floor(x/y). Would prolly need to implement a glsl mod
+void BoxBlur(
+	const uint32_t ch,
+	const float32_t inRadius,
+	const uint32_t wrapMode,
+	const float32_t4 borderColor,
```
devshgraphicsprogramming, line 94: can turn these into members of the struct or something (or just use the parameters you define in another header directly?

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+		if( idx < AXIS_DIM )
+		{
+			const float32_t radius = inRadius * AXIS_DIM;
+			float32_t left = float32_t( idx ) - radius - 1.f;
+			float32_t right = float32_t( idx ) + radius;
+			const uint32_t last = AXIS_DIM - 1;
+
+			float32_t result = 0.f;
+
+			if( right <= last )
```
devshgraphicsprogramming, line 135: this check is wrong, it should check if `right` is still within the texture/buffer, so you need `ceil(right)<float(itemCount)`

You could `assert(itemCount<MaxItemCount);` if the assert worked at all

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+void BoxBlur(
+	const uint32_t ch,
+	const float32_t inRadius,
+	const uint32_t wrapMode,
+	const float32_t4 borderColor,
+	BufferAccessor textureAccessor
+) {
+	float32_t blurred = textureAccessor.get( DataIndex( i ), ch );
+
+	for( uint32_t pass = 0u; pass < PASSES_PER_AXIS; ++pass )
```
devshgraphicsprogramming, line 99: the `PASSES_PER_AXIS` doesn't need to be a constexpr/define anywhere,  can be a regular paramter to the function or a member of the struct.

spv-opt will unroll the loop as long as you assign it a constexpr or literal

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+
+		uint32_t idx = DataIndex( i ); // ?
+		prefixSumsAccessor.set( idx, scanResult );
+
+		for( uint i = 0u; i < LOCAL_SPILLAGE; ++i )
+		{
+			uint32_t idx = DataIndex( i ); // ?
+			prefixSumsAccessor.set( idx, spill[ i ] );
+		}
+		prefixSumsAccessor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 122: this is actually what we want to get rid of

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+		float32_t spill[ LOCAL_SPILLAGE ];
+		for( uint32_t i = 0u; i < LOCAL_SPILLAGE; ++i )
+		{
+			float32_t scanResult = workgroupPrefixSum( blurred ) + previousBlockSum;
+			spill[ i ] = scanResult;
+			previousBlockSum = workgroupBroadcast( spill[ i ], gl_WorkGroupSize().x - 1u );
+		}
+
+		float32_t scanResult = workgroupPrefixSum( blurred ) + previousBlockSum;
+		previousBlockSum = workgroupBroadcast( scanResult, gl_WorkGroupSize().x - 1u );
```
devshgraphicsprogramming, line 112: unused result

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+
+							if( ( period & 0x1u ) == 1 )
+								v_ceiled = period * prefixSumsAccessor.get( last ) + prefixSumsAccessor.get( last ) - prefixSumsAccessor.get( last - uint32_t( fmod( d, AXIS_DIM ) ) );
+							else
+								v_ceiled = period * prefixSumsAccessor.get( last ) + prefixSumsAccessor.get( fmod( d - 1, AXIS_DIM ) );
+						}
+					}
+
+					result = lerp( v_floored, v_ceiled, frac( right ) );
+				} break;
```
devshgraphicsprogramming, line 211: The old code necessitated using `MaxWorkItems` shared memory.

you can see here that it expects the whole prefix sum to be available in shared memory for arbitrary addressing

devshgraphicsprogramming: Also the `AXIS_DIM` stuff is just weird here

include/nbl/builtin/hlsl/central_limit_blur/box_blur.hlsl
```diff
@@ -0,0 +1,297 @@
...
+	const float32_t4 borderColor,
+	BufferAccessor textureAccessor
+) {
+	float32_t blurred = textureAccessor.get( DataIndex( i ), ch );
+
+	for( uint32_t pass = 0u; pass < PASSES_PER_AXIS; ++pass )
+	{
+		float32_t previousBlockSum = 0.f;
+
+		float32_t spill[ LOCAL_SPILLAGE ];
```
devshgraphicsprogramming, line 103: btw you want to use `uint32_t` for the prefix sum and convert to-and-fro your own Fixed Point Unorm rep

PR #661, opened 2024-02-28
New spirv introspection

Przemog1:


include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -648,35 +648,30 @@ class NBL_API2 CSPIRVIntrospector : public core::Uncopyable
 			if (layout)
```
devshgraphicsprogramming, line 648: Above you need to improve the sanity checks on the `info` and results of other operations being successes

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -648,35 +648,30 @@ class NBL_API2 CSPIRVIntrospector : public core::Uncopyable
...
 			{
 				// TODO: merge CStageIntrospection into CPipelineIntrospection
+				pplnIntrospectData->merge(introspection.get());
```
devshgraphicsprogramming, line 651: you can probably hoist the `merge` above the if statement.

You need to handle the failure case of an unsuccessful `merge` and return nullptr.

devshgraphicsprogramming: then for the given `layout` you need to compare for compatibility, basically "are all the bindings in `pplnIntrospectData` present `layout` with same type and array size? (Also, if present, is the runtime sized/variable length array the last binding of both `pplnIntrospectData` and `layout`)"

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -648,35 +648,30 @@ class NBL_API2 CSPIRVIntrospector : public core::Uncopyable
...
-				//	pplnIntrospectData->createApproximateDescriptorSetLayoutFromIntrospection(1),
-				//	pplnIntrospectData->createApproximateDescriptorSetLayoutFromIntrospection(2),
-				//	pplnIntrospectData->createApproximateDescriptorSetLayoutFromIntrospection(3)
-				//);
+				layout = core::make_smart_refctd_ptr<ICPUPipelineLayout>(
+					pcRangesSpan,
+					pplnIntrospectData->createApproximateDescriptorSetLayoutFromIntrospection(0),
+					pplnIntrospectData->createApproximateDescriptorSetLayoutFromIntrospection(1),
+					pplnIntrospectData->createApproximateDescriptorSetLayoutFromIntrospection(2),
+					pplnIntrospectData->createApproximateDescriptorSetLayoutFromIntrospection(3)
```
devshgraphicsprogramming, line 668: we need to change the signature of `createApproximateDescriptorSetLayoutFromIntrospection` because not returning `nullptr` isnt really a criterion for success.

We should probably return a struct
```cpp
struct SSetLayoutCreateResult
{
  inline operator bool() const {return success;}

   core::smart_refctd_ptr<ICPUDescriptorSetLayout> layout;
   bool success;
};
```

then you can return a null layout if there are no bindings present with this set number and still count that as a success

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -150,32 +158,56 @@ NBL_API2 bool CSPIRVIntrospector::CPipelineIntrospectionData::merge(const CSPIRV
...
         .stageFlags = m_pushConstantBytes[0].value,
         .offset = 0,
         .size = 0
     };
-    for ()
+
+    SPushConstantRange current = {
+        .offset = 0,
+        .size = 0
+    };
```
devshgraphicsprogramming, line 175: I'd maybe keeptrack only of currentStage and currentOffset, the size (which is redundant) being kept around makes for difficult reading

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -109,20 +109,127 @@ static CSPIRVIntrospector::CStageIntrospectionData::VAR_TYPE spvcrossType2E_TYPE
...
+    for (const auto& binding : m_descriptorSetBindings[setID])
+    {
+        auto& outBinding = outBindings.emplace_back();
+
+        outBinding.binding = binding.binding;
+        outBinding.count = binding.count;
+        outBinding.type = binding.type;
+        outBinding.stageFlags = binding.stageMask;
+        // TODO outBinding.createFlags = 
+
```
devshgraphicsprogramming, line 227: sane default would be `ECF_NONE` we can't infer anything about them from the SPIR-V /  shader

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -109,20 +109,127 @@ static CSPIRVIntrospector::CStageIntrospectionData::VAR_TYPE spvcrossType2E_TYPE
...
+    }
+
+    core::smart_refctd_ptr<ICPUDescriptorSetLayout> output = core::make_smart_refctd_ptr<ICPUDescriptorSetLayout>(outBindings.data(), outBindings.data() + outBindings.size());
+
+    return output;
```
devshgraphicsprogramming, line 232: one request, make the function return `nullptr` if `outBindings.empty()`

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -484,6 +594,7 @@ NBL_API2 core::smart_refctd_ptr<const CSPIRVIntrospector::CStageIntrospectionDat
     {
         if (shaderStage == IShader::ESS_FRAGMENT)
         {
+            // TODO: desn't path here for frag_test1.hlsl, why?
             CSPIRVIntrospector::CStageIntrospectionData::SFragmentOutputInterface res =
```
devshgraphicsprogramming, line 598: ?

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -109,20 +109,127 @@ static CSPIRVIntrospector::CStageIntrospectionData::VAR_TYPE spvcrossType2E_TYPE
...
+    {
+        std::ostringstream debug;
+        debug << "ds ID: " << i << std::endl;
+        const auto& bindingInfos = stageData->getDescriptorSetInfo(i);
+        for (const auto& bindingInfo : bindingInfos)
+        {
+            CPipelineIntrospectionData::SDescriptorInfo descInfo;
+            descInfo.binding = bindingInfo.binding;
+            descInfo.type = bindingInfo.type;
+
```
devshgraphicsprogramming, line 124: you need to get the array-stride and the count.

Getting the count is mildly complicated:
1. if the `bindingInfo.count.isRuntimeSized()` then your `descInfo` needs to be runtime sized as well
2. else if the `bindingInfo.count.isSpecConstant` you need to look up the value in the `specConstant` map (if its nullptr, you need to return false and fail now) and substitute that for the `descInfo.count`
3. otherwise you're hitting the plain regular 99% of the time path, which is `!isSpecConstant` and you just take the `count.value` to become the `descInfo.count`

devshgraphicsprogramming: ah one more thing about (1), only the very last binding is allowed to be Runtime Sized (counterpart to variables in an SSBO but for descriptors) so if you have a runtime sized `bindingInfo` you need to check if its the last one in the `bindingInfos` and if its not, return false.

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -109,20 +109,127 @@ static CSPIRVIntrospector::CStageIntrospectionData::VAR_TYPE spvcrossType2E_TYPE
...
 NBL_API2 bool CSPIRVIntrospector::CPipelineIntrospectionData::merge(const CSPIRVIntrospector::CStageIntrospectionData* stageData, const ICPUShader::SSpecInfoBase::spec_constant_map_t* specConstants)
 {
-    return false;
+
+    // merge descriptors
+    for (uint32_t i = 0u; i < ICPUPipelineLayout::DESCRIPTOR_SET_COUNT; i++)
+    {
+        std::ostringstream debug;
+        debug << "ds ID: " << i << std::endl;
+        const auto& bindingInfos = stageData->getDescriptorSetInfo(i);
```
devshgraphicsprogramming, line 118: you need to be able to put `*this` back in its original state BEFORE the `merge` call if you encounter a failure.

So I'd take a full copy of the descriptor infos BEFORE entering the loop, then `std::move` the copies back into the members if you manage to make it to the other side without returning false.

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -109,20 +109,127 @@ static CSPIRVIntrospector::CStageIntrospectionData::VAR_TYPE spvcrossType2E_TYPE
...
+        for (const auto& bindingInfo : bindingInfos)
+        {
+            CPipelineIntrospectionData::SDescriptorInfo descInfo;
+            descInfo.binding = bindingInfo.binding;
+            descInfo.type = bindingInfo.type;
+            
+            const auto& foundBinding = m_descriptorSetBindings[i].find(descInfo);
+            if (foundBinding != m_descriptorSetBindings[i].end())
+            {
+                // TODO: validation
```
devshgraphicsprogramming, line 128: just check that the:
- type
- stride
- count
are matching.

An added condition (regardless of if you find a binding with same number) is that if your `descInfo.isRuntimeSized()` then either:
- `descInfo` has a higher or binding number than anything in `m_descriptorSetBindings[i]` (might pay off to have a `m_highestBindingNumber[4]` in the pipeline intro)
- `descInfo` is equal to `m_highestBindingNumber[i]` both need to be runtime sized -> but then you go into the `foundBinding != m_descriptorSetBindings[i].end()` case so the counts will be checked
- `m_descriptorSetBindings.empty()` 


overall a simple check of
```cpp
if (bindingInfo.count.isRuntimeSized() && bindingInfo.binding<m_highestBindingNumber[i])
    return false;
```
will do

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: (2) was in the old code by Criss

include/nbl/asset/utils/CSPIRVIntrospector.h
devshgraphicsprogramming: nvm what I wrote I'm dumb

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: Theoretically to avoid such copies,  you could split it into two loops, one that validates, and another that sets, but that would require you already know that the `stageData` is internally consistent, meaning there are no two bindings in there that would fail validation against themselves (i.e. no case of binding B being declared both as a UBO and an SSBO)

PR #643, opened 2024-01-10
Nahim shader defines

devshgraphicsprogramming:

PR #665, opened 2024-03-11
Global scan

kpentaris:
## Description
Initial implementation of the Global Scan in HLSL

## Testing 
Not yet finished

## TODO list:
Still need to test the HLSL code as well as the porting of the example code to the new API of vulkan_1_3 branch


include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -1,66 +1,63 @@
...
 #define _NBL_HLSL_SCAN_DECLARATIONS_INCLUDED_
 
 // REVIEW: Not sure if this file is needed in HLSL implementation
 
-#include "nbl/builtin/hlsl/scan/parameters_struct.hlsl"
+#include "nbl/builtin/hlsl/cpp_compat.hlsl"
 
+#ifndef NBL_BUILTIN_MAX_SCAN_LEVELS
+#define NBL_BUILTIN_MAX_SCAN_LEVELS 7
+#endif
```
devshgraphicsprogramming, line 14: do as `NBL_CONSTEXPR` instead of define

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -1,66 +1,63 @@
...
 {
 namespace hlsl
 {
 namespace scan
 {
+    // REVIEW: Putting topLevel second allows better alignment for packing of constant variables, assuming lastElement has length 4. (https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-packing-rules)
+	struct Parameters_t {
+		uint32_t lastElement[NBL_BUILTIN_MAX_SCAN_LEVELS/2+1];
+		uint32_t topLevel;
+		uint32_t temporaryStorageOffset[NBL_BUILTIN_MAX_SCAN_LEVELS/2];
```
devshgraphicsprogramming, line 26: use uint16_t for this

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -1,66 +1,63 @@
...
+	};
+    
 	Parameters_t getParameters();
```
devshgraphicsprogramming, line 29: no need for such a forward decl anymore

include/nbl/builtin/hlsl/scan/declarations.hlsl
devshgraphicsprogramming: also this better be formulated slightly differently, in terms of reduction passes, not reduction + downsweep

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -1,66 +1,63 @@
...
-namespace nbl
-{
-namespace hlsl
-{
-namespace scan
-{
 	template<typename Storage_t>
 	void setData(
-		in Storage_t data,
-		in uint levelInvocationIndex,
```
devshgraphicsprogramming, line 54: no more forward declarations, just let it be an accessor.

Also the `level` const-ref-args can be rolled up into a single struct `SDataIndex`

include/nbl/builtin/hlsl/scan/descriptors.hlsl
```diff
@@ -1,3 +1,112 @@
...
+    NBL_CONST_REF_ARG(uint32_t) pseudoLevel
+)
+{
+    const Parameters_t params = getParameters(); // defined differently for direct and indirect shaders
+    
+    uint32_t offset = levelInvocationIndex;
+    const bool notFirstOrLastLevel = bool(pseudoLevel);
+    if (notFirstOrLastLevel)
+		offset += params.temporaryStorageOffset[pseudoLevel-1u];
+
```
devshgraphicsprogramming, line 45: remove thsi whole file, it should be userspace

include/nbl/builtin/hlsl/scan/direct.hlsl
```diff
@@ -1,50 +1,90 @@
...
+        return glsl::atomicAdd(wgScratch[ix + offset], val);
+    }
+
+	void workgroupExecutionAndMemoryBarrier()
+	{
+		nbl::hlsl::glsl::barrier();
+		//nbl::hlsl::glsl::memoryBarrierShared(); implied by the above
+	}
+};
+static WGScratchProxy<0> accessor;
```
devshgraphicsprogramming, line 41: scratches are userspace

include/nbl/builtin/hlsl/scan/direct.hlsl
```diff
@@ -1,50 +1,90 @@
...
+/**
+ * Required since we rely on SubgroupContiguousIndex instead of 
+ * gl_LocalInvocationIndex which means to match the global index 
+ * we can't use the gl_GlobalInvocationID but an index based on 
+ * SubgroupContiguousIndex.
+ */
+uint32_t globalIndex()
+{
+	return nbl::hlsl::glsl::gl_WorkGroupID().x*WORKGROUP_SIZE+nbl::hlsl::workgroup::SubgroupContiguousIndex();
+}
```
devshgraphicsprogramming, line 64: we have this in a header already

include/nbl/builtin/hlsl/scan/direct.hlsl
```diff
@@ -1,50 +1,90 @@
...
+uint32_t3 nbl::hlsl::glsl::gl_WorkGroupSize() {return uint32_t3(WORKGROUP_SIZE,1,1);}
+
+struct ScanPushConstants
+{
+    nbl::hlsl::scan::Parameters_t scanParams;
+    nbl::hlsl::scan::DefaultSchedulerParameters_t schedulerParams;
+};
+
+[[vk::push_constant]]
+ScanPushConstants spc;
```
devshgraphicsprogramming, line 53: everything affecting the pipeline layout should be userspace

include/nbl/builtin/hlsl/scan/indirect.hlsl
```diff
@@ -13,36 +8,33 @@ namespace hlsl
...
-[numthreads(_NBL_HLSL_WORKGROUP_SIZE_, 1, 1)]
-void CSMain()
+[numthreads(WORKGROUP_SIZE,1,1)]
+void main()
 {
-	if (bool(nbl::hlsl::scan::getIndirectElementCount()))
-		nbl::hlsl::scan::main();
+    if(bool(nbl::hlsl::scan::getIndirectElementCount())) {
+        // TODO call main from virtual_workgroup.hlsl
+    }
```
devshgraphicsprogramming, line 38: all this should be userspace, also there will be very little difference between direct and indirect

include/nbl/builtin/hlsl/scan/virtual_workgroup.hlsl
```diff
@@ -6,87 +6,78 @@
...
-#if 0
 namespace nbl
 {
 namespace hlsl
 {
 namespace scan
 {
-	template<class Binop, class Storage_t>
-	void virtualWorkgroup(in uint treeLevel, in uint localWorkgroupIndex)
+	template<class Binop, typename Storage_t, bool isExclusive, uint16_t ItemCount, class Accessor, class device_capabilities=void>
```
devshgraphicsprogramming, line 18: why not template on the workgroup scan instead?

Within you can alias and extract:
- binop
- storage_t
- exclusive or not
- item count
- smem accessor
- device traits necessary

devshgraphicsprogramming:
Is this something I can actually checkout and try myself?

include/nbl/builtin/hlsl/scan/descriptors.hlsl
```diff
@@ -17,21 +17,21 @@ namespace hlsl
...
-template<uint32_t scratchElementCount=SCRATCH_SZ - 1> // (REVIEW): This should be externally defined. Maybe change the scratch buffer to RWByteAddressBuffer? Annoying to manage though...
+template<uint32_t scratchElementCount=SCRATCH_SZ - 1>
 struct Scratch
 {
     uint32_t workgroupsStarted;
     uint32_t data[scratchElementCount];
 };
 
-[[vk::binding(0 ,0)]] RWStructuredBuffer<uint32_t /*Storage_t*/> scanBuffer; // (REVIEW): Make the type externalizable. Decide how (#define?)
+[[vk::binding(0 ,0)]] RWStructuredBuffer<Storage_t> scanBuffer; // (REVIEW): Make the type externalizable. Decide how (#define?)
```
devshgraphicsprogramming, line 27: Use Buffer Device Address

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -30,8 +30,9 @@ namespace scan
 
     struct DefaultSchedulerParameters_t
     {
-        uint32_t finishedFlagOffset[NBL_BUILTIN_MAX_SCAN_LEVELS-1];
         uint32_t cumulativeWorkgroupCount[NBL_BUILTIN_MAX_SCAN_LEVELS];
+        uint32_t workgroupFinishFlagsOffset[NBL_BUILTIN_MAX_SCAN_LEVELS];
+        uint32_t lastWorkgroupSetCountForLevel[NBL_BUILTIN_MAX_SCAN_LEVELS];
```
devshgraphicsprogramming, line 35: If you use Buffer Device Address, you don't need to size the arrays, you just make them `bda::__ptr` instead

include/nbl/builtin/hlsl/scan/descriptors.hlsl
devshgraphicsprogramming: alternatively you can use accessor pattern

include/nbl/builtin/hlsl/scan/descriptors.hlsl
```diff
@@ -17,21 +17,21 @@ namespace hlsl
...
 [[vk::binding(1 ,0)]] RWStructuredBuffer<Scratch> /*globallycoherent (seems we can't use along with VMM)*/ scanScratchBuf; // (REVIEW): Check if globallycoherent can be used with Vulkan Mem Model
```
devshgraphicsprogramming, line 28: this one you need to use `BDA` 

globallycoherent can't be used with VMM, but DXc doesn't support/emit/upgrade to VMM

IIRC you can mark individual load/store as coherent even before VMM (just no acquire/release cause thats VMM) with spir-v intrinsics and that should be enough

Also because the scratch needs to be coherent, it only makes sense to come from a buffer, and you might as well use BDA for it

include/nbl/builtin/hlsl/scan/scheduler.hlsl
```diff
@@ -32,6 +32,11 @@ struct Scheduler
   template<class Accessor>
   bool getWork(NBL_REF_ARG(Accessor) accessor)
   {
+      
+glsl::memoryBarrierBuffer();
+if(scanScratchBuf[0u].workgroupsStarted >= schedParams.cumulativeWorkgroupCount[level])
+  return false;
```
devshgraphicsprogramming, line 38: watch out for indentation

.gitmodules
```diff
@@ -69,7 +69,7 @@
 	branch = main
 [submodule "examples_tests"]
 	path = examples_tests
-	url = git@github.com:Devsh-Graphics-Programming/Nabla-Examples-and-Tests.git
+	url = git@github.com:kpentaris/Nabla-Examples-and-Tests.git
```
devshgraphicsprogramming, line 72: undo this before merge

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -9,8 +9,8 @@
 
 #include "nbl/builtin/hlsl/cpp_compat.hlsl"
 
-#ifndef NBL_BUILTIN_MAX_SCAN_LEVELS
-#define NBL_BUILTIN_MAX_SCAN_LEVELS 7
+#ifndef NBL_BUILTIN_MAX_LEVELS
+#define NBL_BUILTIN_MAX_LEVELS 7
```
devshgraphicsprogramming, line 13: this should be a `NBL_CONSTEXPR_STATIC_INLINE` in `Parameters_t`

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -21,18 +21,18 @@ namespace scan
 {
     // REVIEW: Putting topLevel second allows better alignment for packing of constant variables, assuming lastElement has length 4. (https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-packing-rules)
```
devshgraphicsprogramming, line 22: we have `-fuse-scalar-layout` DXC option, which packs everything tightly

include/nbl/builtin/hlsl/scan/declarations.hlsl
devshgraphicsprogramming: I keep doing the same reviews
https://github.com/Devsh-Graphics-Programming/Nabla/pull/665#discussion_r1547511769

include/nbl/builtin/hlsl/scan/direct.hlsl
devshgraphicsprogramming: use accessors

include/nbl/builtin/hlsl/scan/virtual_workgroup.hlsl
```diff
@@ -6,87 +6,78 @@
...
+	void virtualWorkgroup(NBL_CONST_REF_ARG(uint32_t) treeLevel, NBL_CONST_REF_ARG(uint32_t) localWorkgroupIndex, NBL_REF_ARG(Accessor) accessor)
 	{
 		const Parameters_t params = getParameters();
-		const uint levelInvocationIndex = localWorkgroupIndex * _NBL_HLSL_WORKGROUP_SIZE_ + SubgroupContiguousIndex();
-		const bool lastInvocationInGroup = SubgroupContiguousIndex() == (_NBL_HLSL_WORKGROUP_SIZE_ - 1);
-
-		const uint lastLevel = params.topLevel << 1u;
-		const uint pseudoLevel = levelInvocationIndex <= params.lastElement[pseudoLevel];
+		const uint32_t levelInvocationIndex = localWorkgroupIndex * glsl::gl_WorkGroupSize().x + SubgroupContiguousIndex();
+		const bool lastInvocationInGroup = SubgroupContiguousIndex() == (gl_WorkGroupSize().x - 1);
```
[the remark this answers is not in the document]
devshgraphicsprogramming: most definitely

include/nbl/video/utilities/IUtilities.h
```diff
@@ -722,8 +727,9 @@ class NBL_API2 IUtilities : public core::IReferenceCounted
 
 #if 0 // TODO: port
         core::smart_refctd_ptr<CPropertyPoolHandler> m_propertyPoolHandler;
-        core::smart_refctd_ptr<CScanner> m_scanner;
 #endif
+        core::smart_refctd_ptr<CReduce> m_reducer;
+        core::smart_refctd_ptr<CScanner> m_scanner;
```
devshgraphicsprogramming, line 732: you can actually kick them out of the `IUtilities`

include/nbl/builtin/hlsl/scan/descriptors.hlsl
devshgraphicsprogramming: Even without VMM you can use Release/Acquire/SeqCst

The only difference between pre and post VMM, is because Volatile used to be a Memory Operand, and in VMM its a MEmory Semantic.

https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_memory_semantics_id

include/nbl/builtin/hlsl/scan/virtual_workgroup.hlsl
```diff
@@ -2,91 +2,85 @@
...
 
-		const uint lastLevel = params.topLevel << 1u;
-		const uint pseudoLevel = levelInvocationIndex <= params.lastElement[pseudoLevel];
+        // REVIEW: Right now in order to support REDUCE operation we need to set the max treeLevel == topLevel
+        // so that it exits after reaching the top?
+        
+        // Seems that even though it's a mem barrier is must NOT diverge!
+        // This was called inside getData but due to inRange it can possibly diverge 
+        // so it must be called outside! Not yet sure why a mem barrier must be uniformly called.
+        glsl::memoryBarrierBuffer(); // scanScratchBuf can't be declared as coherent due to VMM(?)
```
devshgraphicsprogramming, line 33: calls to pure-memory barriers (OpMemoryBarrier) can in-fact diverge, its just OpControlBarrier that needs to not diverge (same dynamic instance)
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpControlBarrier

also you want the accessor to define the memory barrier to use

include/nbl/builtin/hlsl/scan/virtual_workgroup.hlsl
```diff
@@ -2,91 +2,85 @@
...
+        
+        Storage_t data = Binop::identity; // REVIEW: replace Storage_t with Binop::type_t?
+        if(inRange)
+        {
+            getData<Storage_t, isExclusive>(data, levelInvocationIndex, levelWorkgroupIndex, treeLevel, pseudoLevel);
+        }
```
devshgraphicsprogramming, line 39: you're supposed to have an accessor for this instead of just a `getData` defined somewhere else

devshgraphicsprogramming: as stated before I'd take a whole `WorkgroupOp` and extract the:
- type_t
- Binop
- ItemCount

from that struct type

include/nbl/builtin/hlsl/scan/virtual_workgroup.hlsl
```diff
@@ -2,91 +2,85 @@
...
 namespace scan
 {
-	template<class Binop, class Storage_t>
-	void virtualWorkgroup(in uint treeLevel, in uint localWorkgroupIndex)
-	{
-		const Parameters_t params = getParameters();
-		const uint levelInvocationIndex = localWorkgroupIndex * _NBL_HLSL_WORKGROUP_SIZE_ + SubgroupContiguousIndex();
-		const bool lastInvocationInGroup = SubgroupContiguousIndex() == (_NBL_HLSL_WORKGROUP_SIZE_ - 1);
+    template<class Binop, typename Storage_t, bool isScan, bool isExclusive, uint16_t WorkgroupSize, class Accessor, class device_capabilities=void>
+    void virtualWorkgroup(NBL_CONST_REF_ARG(uint32_t) treeLevel, NBL_CONST_REF_ARG(uint32_t) levelWorkgroupIndex, NBL_REF_ARG(Accessor) accessor)
```
devshgraphicsprogramming, line 17: I'd take the following accessors:
- shared memory accessor for workgroup scan
- main accessor for scan input and result for level 0
- scratch accessor for scan input/output at levels higher than 0

include/nbl/builtin/hlsl/scan/virtual_workgroup.hlsl
```diff
@@ -2,91 +2,85 @@
...
+        const Parameters_t params = getParameters();
+        const DefaultSchedulerParameters_t schedulerParams = getSchedulerParameters();
+        Scheduler<WorkgroupSize> scheduler = Scheduler<WorkgroupSize>::create(params, schedulerParams);
+        // persistent workgroups
+        while (true)
+        {
+            if (!scheduler.getWork(accessor))
+            {
+                return;
+            }
```
devshgraphicsprogramming, line 75: 1. stop with the forward declared `get` and `set`
2. this a general pattern of running virtual workgroups, I'd move it out to a general header `virtual_workgroup_executor.hlsl` outside of `scan`

because of (2) you need to template on an opaque struct that wraps up everything you want to pass to a virtual workgroup. Basically turn the virtual workgroup into a stateful lambda.


```cpp
template<typename VirtualWorkgroup, typename Scheduler>
struct virtual_workgroup_executor
{
   void operator()(NBL_REF_ARG(Scheduler) scheduler);
};
```

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+		{
+			EO_AND = 0u,
+			EO_XOR,
+			EO_OR,
+			EO_ADD,
+			EO_MUL,
+			EO_MIN,
+			EO_MAX,
+			EO_COUNT
+		};
```
devshgraphicsprogramming, line 40: instead of inventing new enums, I'd template `CArithmeticOps` using
```cpp
template<template<class> class Binop, typename T>
```
and determine what you're doing and summing that way

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+			}
+            // given already computed tables of lastElement indices per level, number of levels, and storage offsets, tell us total auxillary buffer size needed
+			inline uint32_t getScratchSize(uint32_t ssboAlignment=256u)
+			{
+				uint32_t uint_count = 1u; // workgroup enumerator
+				uint_count += temporaryStorageOffset[MaxLevels/2u-1u]; // last scratch offset
+				uint_count += lastElement[topLevel]+1u; // and its size
+				return core::roundUp<uint32_t>(uint_count*sizeof(uint32_t),ssboAlignment);
+			}
+		};
```
devshgraphicsprogramming, line 73: this should not deviate from `nbl::hlsl::scan::Parameters_t` at all, if some of the constructors/methods you need on CPU are not there, move them there

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+				std::fill_n(cumulativeWorkgroupCount,Parameters::MaxLevels,0u);
+				std::fill_n(workgroupFinishFlagsOffset, Parameters::MaxLevels, 0u);
+				std::fill_n(lastWorkgroupSetCountForLevel,Parameters::MaxLevels,0u);
+			}
+            
+            // given the number of elements and workgroup size, figure out how many atomics we need
+            // also account for the fact that we will want to use the same scratch buffer both for the
+            // scheduler's atomics and the aux data storage
+			SchedulerParameters(Parameters& outScanParams, const uint32_t _elementCount, const uint32_t workgroupSize) : SchedulerParameters()
+			{
```
devshgraphicsprogramming, line 89: again, move to HLSL

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+					workgroupFinishFlagsOffset[i] = ((cumulativeWorkgroupCount[i] - 1u) >> hlsl::findMSB(workgroupSize - 1u)) + 1;
+					lastWorkgroupSetCountForLevel[i] = cumulativeWorkgroupCount[i] & (workgroupSize - 1u);
+				}
+
+				const auto wgFinishedFlagsSize = std::accumulate(workgroupFinishFlagsOffset, workgroupFinishFlagsOffset + Parameters::MaxLevels, 0u);
+				std::exclusive_scan(workgroupFinishFlagsOffset, workgroupFinishFlagsOffset + Parameters::MaxLevels, workgroupFinishFlagsOffset, 0u);
+				for (auto i=0u; i<sizeof(Parameters::temporaryStorageOffset)/sizeof(uint32_t); i++)
+					outScanParams.temporaryStorageOffset[i] += wgFinishedFlagsSize;
+				
+				std::inclusive_scan(cumulativeWorkgroupCount, cumulativeWorkgroupCount + Parameters::MaxLevels, cumulativeWorkgroupCount);
```
devshgraphicsprogramming, line 108: you can add equivalents of the `std::` functions to `algorithm.hlsl`, they'll just be single invocation utilities

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+            
+            // in case we scan very few elements, you don't want to launch workgroups that wont do anything
+			DispatchInfo(const IPhysicalDevice::SLimits& limits, const uint32_t elementCount, const uint32_t workgroupSize)
+			{
+				constexpr auto workgroupSpinningProtection = 4u; // to prevent first workgroup starving/idling on level 1 after finishing level 0 early
+				wg_count = limits.computeOptimalPersistentWorkgroupDispatchSize(elementCount,workgroupSize,workgroupSpinningProtection);
+			}
+
+			uint32_t wg_count;
+		};
```
devshgraphicsprogramming, line 133: move everything to HLSL, you can also have a look at PR #643 where the `computeOptimalPErsistentWorkgroupDispatchSize` may move to `device_capability_traits`

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+                    assert(false);
+                    break;
+            }
+            
+			return asset::CHLSLCompiler::createOverridenCopy(cpushader.get(), "#define WORKGROUP_SIZE %d\ntypedef %s Storage_t;\n#define BINOP nbl::hlsl::%s\n#define SCRATCH_SZ %d\n", m_workgroupSize, storageType, binop, scratchSz);
+        }
+
+		inline ILogicalDevice* getDevice() const {return m_device.get();}
+    protected:
+        // REVIEW: Does it need an empty destructor?
```
devshgraphicsprogramming, line 275: no a default one will be made

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+
+		core::smart_refctd_ptr<ILogicalDevice> m_device;
+		core::smart_refctd_ptr<IGPUDescriptorSetLayout> m_ds_layout;
```
devshgraphicsprogramming, line 278: you wont be using any descriptors, so no need for a layout

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+			infos[1].info.buffer = {scratch_range.offset,scratch_range.size};
+
+			video::IGPUDescriptorSet::SWriteDescriptorSet writes[2];
+			for (auto i=0u; i<2u; i++)
+			{
+				writes[i] = { .dstSet = set,.binding = i,.arrayElement = 0u,.count = 1u,.info = infos+i };
+			}
+
+			device->updateDescriptorSets(2, writes, 0u, nullptr);
+		}
```
devshgraphicsprogramming, line 179: no more descriptor sets

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+
+		inline auto getDefaultPipelineLayout() const { return m_pipeline_layout.get(); }
+
+		inline uint32_t getWorkgroupSize() const {return m_workgroupSize;}
+
+		inline void buildParameters(const uint32_t elementCount, DefaultPushConstants& pushConstants, DispatchInfo& dispatchInfo)
+		{
+			pushConstants.schedulerParams = SchedulerParameters(pushConstants.scanParams,elementCount,m_workgroupSize);
+			dispatchInfo = DispatchInfo(m_device->getPhysicalDevice()->getLimits(),elementCount,m_workgroupSize);
+		}
```
devshgraphicsprogramming, line 162: move to HLSL

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+		{
+			assert(core::isPoT(m_workgroupSize));
+			const asset::SPushConstantRange pc_range[] = {asset::IShader::ESS_COMPUTE,0u,sizeof(DefaultPushConstants)};
+			const IGPUDescriptorSetLayout::SBinding bindings[2] = {
+				{ 0u, asset::IDescriptor::E_TYPE::ET_STORAGE_BUFFER, IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_NONE, video::IGPUShader::ESS_COMPUTE, 1u, nullptr }, // main buffer
+				{ 1u, asset::IDescriptor::E_TYPE::ET_STORAGE_BUFFER, IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_NONE, video::IGPUShader::ESS_COMPUTE, 1u, nullptr } // scratch
+			};
+
+			m_ds_layout = m_device->createDescriptorSetLayout(bindings);
+			assert(m_ds_layout && "CArithmeticOps Descriptor Set Layout was not created successfully");
```
devshgraphicsprogramming, line 147: no bindings

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+                    nbl::system::ISystem::future_t<core::smart_refctd_ptr<nbl::system::IFile>> future;
+                    system->createFile(future, system::path(_path), core::bitflag(nbl::system::IFileBase::ECF_READ) | nbl::system::IFileBase::ECF_MAPPABLE);
+                    if (future.wait())
+                        return future.copy();
+                    return nullptr;
+                };
+
+                // hlsl = loadBuiltinData("nbl/builtin/hlsl/scan/indirect_reduce.hlsl"); ??
+                hlsl = loadBuiltinData(shaderFile);
+            }
```
devshgraphicsprogramming, line 217: this is super convoluted, why is the user supposed to know the path to the file you'll load?

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+                    break;
+                case EO_MIN:
+                    binop = "minimum";
+                    break;
+                default:
+                    assert(false);
+                    break;
+            }
+            
+			return asset::CHLSLCompiler::createOverridenCopy(cpushader.get(), "#define WORKGROUP_SIZE %d\ntypedef %s Storage_t;\n#define BINOP nbl::hlsl::%s\n#define SCRATCH_SZ %d\n", m_workgroupSize, storageType, binop, scratchSz);
```
devshgraphicsprogramming, line 270: it would be better to build the shader source instead of overriding some random shader

include/nbl/video/utilities/CArithmeticOps.h
```diff
@@ -0,0 +1,286 @@
...
+				constexpr auto workgroupSpinningProtection = 4u; // to prevent first workgroup starving/idling on level 1 after finishing level 0 early
+				wg_count = limits.computeOptimalPersistentWorkgroupDispatchSize(elementCount,workgroupSize,workgroupSpinningProtection);
+			}
+
+			uint32_t wg_count;
+		};
+
+		CArithmeticOps(core::smart_refctd_ptr<ILogicalDevice>&& device) : CArithmeticOps(std::move(device),core::roundDownToPoT(device->getPhysicalDevice()->getLimits().maxOptimallyResidentWorkgroupInvocations)) {}
+		
+        CArithmeticOps(core::smart_refctd_ptr<ILogicalDevice>&& device, const uint32_t workgroupSize) : m_device(std::move(device)), m_workgroupSize(workgroupSize)
```
devshgraphicsprogramming, line 137: if you don't plan to make `CArithmeticOps` be usable on its own, make the constructors `protected`

include/nbl/video/utilities/CReduce.h
```diff
@@ -0,0 +1,36 @@
...
+#include "nbl/video/utilities/CArithmeticOps.h"
+
+namespace nbl::video
+{
+
+    class CReduce final : public CArithmeticOps
+    {
+
+    public:
+        CReduce(core::smart_refctd_ptr<ILogicalDevice>&& device) : CReduce(std::move(device), core::roundDownToPoT(device->getPhysicalDevice()->getLimits().maxOptimallyResidentWorkgroupInvocations)) {}
```
devshgraphicsprogramming, line 17: don't repeat the `roundDownToPoT` everywhere, you should rely on `CArithmeticOps` to set that up for you

include/nbl/video/utilities/CScanner.h
```diff
@@ -8,15 +8,14 @@
 
 #include "nbl/video/IPhysicalDevice.h"
 #include "nbl/video/utilities/IDescriptorSetCache.h"
+#include "nbl/video/utilities/CArithmeticOps.h"
 
+#include "nbl/builtin/hlsl/scan/declarations.hlsl"
+static_assert(NBL_BUILTIN_MAX_LEVELS & 0x1, "NBL_BUILTIN_MAX_LEVELS must be odd!");
```
devshgraphicsprogramming, line 14: if we don't have our own `static_assert` in HLSL, then stick this inside a `ifndef __HLSL_VERSION` inside the HLSL header

include/nbl/video/utilities/CScanner.h
```diff
@@ -143,270 +142,43 @@ prefix sum with subgroup sized workgroups at peak Bandwidth efficiency in about
```
devshgraphicsprogramming, line 142: update the comments about the algo between lines 112-141 because we can't spinwait, we need to outline the new algo properly

include/nbl/video/utilities/CScanner.h
```diff
@@ -143,270 +142,43 @@ prefix sum with subgroup sized workgroups at peak Bandwidth efficiency in about
...
 {
 	public:		
 		enum E_SCAN_TYPE : uint8_t
 		{
 			 // computes output[n] = Sum_{i<=n}(input[i])
-			 EST_INCLUSIVE = _NBL_GLSL_SCAN_TYPE_INCLUSIVE_,
+			 EST_INCLUSIVE = 0u,
 			 // computes output[n] = Sum_{i<n}(input[i]), meaning first element is identity
-			 EST_EXCLUSIVE = _NBL_GLSL_SCAN_TYPE_EXCLUSIVE_,
+			 EST_EXCLUSIVE,
```
devshgraphicsprogramming, line 153: I'd template on a bool instead

src/nbl/video/utilities/CReduce.cpp
```diff
@@ -0,0 +1,56 @@
...
+        );
+    }
+    return m_pipelines[dataType][op].get();
+}
+
+core::smart_refctd_ptr<asset::ICPUShader> CReduce::createShader(const CArithmeticOps::E_DATA_TYPE dataType, const CArithmeticOps::E_OPERATOR op, const uint32_t scratchSz) const
+{
+    core::smart_refctd_ptr<asset::ICPUShader> base = createBaseShader("nbl/builtin/hlsl/scan/direct.hlsl", dataType, op, scratchSz);
+    return asset::CHLSLCompiler::createOverridenCopy(base.get(), "#define IS_EXCLUSIVE %s\n#define IS_SCAN %s\n", "false", "false");
+}
```
devshgraphicsprogramming, line 54: 1. the shader would need to have a  name `default_direct` instead of `direct` if it requires you define something for it to work
2. if possible its better to build a source with `ostringstream` without overriden copies, so you have this
```cpp
#include "nbl/builtin/hlsl/scan/default_direct.hlsl"

void main()
{
   nbl::hlsl::scan::default_direct<$Type,$Binop,$dataType,$scratchSize>::__call();
}
```
where Type is some template param that helps you tell apart, reduction, inclusive or exclusive scan

src/nbl/video/utilities/CScanner.cpp
```diff
@@ -1,52 +1,52 @@
...
+    return m_pipelines[scanType][dataType][op].get();
 }
-#endif
\ No newline at end of file
+    
+core::smart_refctd_ptr<asset::ICPUShader> CScanner::createShader(const char* shaderFile, const E_SCAN_TYPE scanType, const E_DATA_TYPE dataType, const E_OPERATOR op, const uint32_t scratchSz) const
+{
+    core::smart_refctd_ptr<asset::ICPUShader> base = createBaseShader("nbl/builtin/hlsl/scan/direct.hlsl", dataType, op, scratchSz);
+    return asset::CHLSLCompiler::createOverridenCopy(base.get(), "#define IS_EXCLUSIVE %s\n#define IS_SCAN %s\n", scanType == EST_EXCLUSIVE ? "true" : "false", "true");
+}
```
devshgraphicsprogramming, line 50: same comment as for CReduce

src/nbl/video/utilities/CScanner.cpp
```diff
@@ -1,52 +1,52 @@
...
+        cpuShader->setShaderStage(asset::IShader::ESS_COMPUTE);
+
+        auto gpushader = m_device->createShader(cpuShader.get());
+
+        m_specialized_shaders[scanType][dataType][op] = gpushader;
+    }
+    return m_specialized_shaders[scanType][dataType][op].get();
+}
 
-		if(indirect)
```
devshgraphicsprogramming, line 21: these 3 utilities are practically identical for CScan and CReduce, why not hoist them to CArithmeticOps?

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
+     * Parameter meaning:
+     * |> lastElement - the index of the last element of each Blelloch level in the scratch buffer
+     * |> topLevel - the top level the Blelloch Scan will have (this depends on the elementCount and the WorkgroupSize)
+     * |> temporaryStorageOffset - an offset array for each level of the Blelloch Scan. It is used when storing the REDUCTION result of each workgroup scan
+     * |> cumulativeWorkgroupCount - the sum-scan of all the workgroups that will need to be launched for each level of the Blelloch Scan (both upsweep and downsweep)
+     * |> finishedFlagOffset - an index in the scratch buffer where each virtual workgroup indicates that ALL its invocations have finished their work. This helps 
+     *                            synchronizing between workgroups with while-loop spinning.
+     */
+    void computeParameters(NBL_CONST_REF_ARG(uint32_t) elementCount, out Parameters_t _scanParams, out DefaultSchedulerParameters_t _schedulerParams)
+    {
```
devshgraphicsprogramming, line 44: The parameters are "married" to to the scheduling logic.

So make these methods of `DefaultSchedulerParameters_t` , this one is really just a constructor so turn it into a `static DefaultSchedulerParameters_t create(...)` method

Hint: whenever a function takes `inout` or `out` struct `T` as first parameter, it likely means that I wrote it in GLSL to be a pseudo-method (would have a C fn ptr signature matching a method). 

Bonus Hint: Whenever the first param is a `in T` it would have been a const method in C++.

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
-	 * |> cumulativeWorkgroupCount - the sum-scan of all the workgroups that will need to be launched for each level of the Blelloch Scan (both upsweep and downsweep)
-	 * |> finishedFlagOffset - an index in the scratch buffer where each virtual workgroup indicates that ALL its invocations have finished their work. This helps 
-	 *							synchronizing between workgroups with while-loop spinning.
-	 */
-	void computeParameters(in uint elementCount, out Parameters_t _scanParams, out DefaultSchedulerParameters_t _schedulerParams)
-	{
+    /**
+     * The CScanner.h parameter computation calculates the number of virtual workgroups that will have to be launched for the Scan operation
+     * (always based on the elementCount) as well as different offsets for the results of each step of the Scan operation, flag positions 
+     * that are used for synchronization etc.
```
devshgraphicsprogramming, line 23: The logic needs to change to "last one out closes the door"

The comments too.

we will now only launch `RoundUp(N/ItemsPerWorkgroup)` virtual workgroups, not the geometric series

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
+     * Remember that CScanner does a Blelloch Scan which works in levels. In each level of the Blelloch scan the array of elements is 
+     * broken down into sets of size=WorkgroupSize and each set is scanned using Hillis & Steele (aka Stone-Kogge adder). The result of 
+     * the scan is provided as an array element for the next level of the Blelloch Scan. This means that if we have 10000 elements and 
+     * WorkgroupSize=250, we will break the array into 40 sets and take their reduction results. The next level of the Blelloch Scan will 
+     * have an array of size 40. Only a single workgroup will be needed to work on that. After that array is scanned, we use the results 
+     * in the downsweep phase of Blelloch Scan.
+     * Keep in mind that each virtual workgroup executes a single step of the whole algorithm, which is why we have the cumulativeWorkgroupCount.
+     * The first virtual workgroups will work on the upsweep phase, the next on the downsweep phase.
```
devshgraphicsprogramming, line 31: Since we do "last workgroup takes over" we will now only launch `RoundUp(N/ItemsPerWorkgroup)` virtual workgroups, not the geometric series

So no need for the `cumulativeWorkgroupCount`.

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
+     * The intermediate results are stored in a scratch buffer. That buffer's size is is the sum of the element-array size for all the 
+     * Blelloch levels. Using the previous example, the scratch size should be 10000 + 40.
+     * 
+     * Parameter meaning:
+     * |> lastElement - the index of the last element of each Blelloch level in the scratch buffer
+     * |> topLevel - the top level the Blelloch Scan will have (this depends on the elementCount and the WorkgroupSize)
+     * |> temporaryStorageOffset - an offset array for each level of the Blelloch Scan. It is used when storing the REDUCTION result of each workgroup scan
+     * |> cumulativeWorkgroupCount - the sum-scan of all the workgroups that will need to be launched for each level of the Blelloch Scan (both upsweep and downsweep)
+     * |> finishedFlagOffset - an index in the scratch buffer where each virtual workgroup indicates that ALL its invocations have finished their work. This helps 
+     *                            synchronizing between workgroups with while-loop spinning.
```
devshgraphicsprogramming, line 41: amend comments as needed

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -21,18 +21,18 @@ namespace scan
...
     struct Parameters_t {
-        uint32_t lastElement[NBL_BUILTIN_MAX_SCAN_LEVELS/2+1];
+        uint32_t lastElement[NBL_BUILTIN_MAX_LEVELS/2+1];
         uint32_t topLevel;
-        uint32_t temporaryStorageOffset[NBL_BUILTIN_MAX_SCAN_LEVELS/2];
+        uint32_t temporaryStorageOffset[NBL_BUILTIN_MAX_LEVELS/2];
     };
     
     Parameters_t getParameters();
```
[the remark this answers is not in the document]
devshgraphicsprogramming: IMHO you need a constructor of `Parameters_t` that is able to work all those out from simple `elementCount` + `itemsPerWorkgroup`

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -1,66 +1,64 @@
...
-		in uint localWorkgroupIndex,
-		in uint treeLevel,
-		in uint pseudoLevel,
-		in bool inRange
-	);
+    struct DefaultSchedulerParameters_t
+    {
+        uint32_t cumulativeWorkgroupCount[NBL_BUILTIN_MAX_LEVELS];
+        uint32_t workgroupFinishFlagsOffset[NBL_BUILTIN_MAX_LEVELS];
+        uint32_t lastWorkgroupSetCountForLevel[NBL_BUILTIN_MAX_LEVELS];
```
devshgraphicsprogramming, line 35: not sure you need either `cumulativeWorkgroupCount` or `lastWorkgroupSetCountForLevel` for a "last one out closes the door" upsweep

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
-				_schedulerParams.finishedFlagOffset[5] = _schedulerParams.finishedFlagOffset[4]+WorkgroupCount(2);
-				
-				_scanParams.temporaryStorageOffset[0] = _schedulerParams.finishedFlagOffset[5]+WorkgroupCount(1);
-				_scanParams.temporaryStorageOffset[1] = _scanParams.temporaryStorageOffset[0]+WorkgroupCount(0);
-				_scanParams.temporaryStorageOffset[2] = _scanParams.temporaryStorageOffset[1]+WorkgroupCount(1);
-				break;
-			default:
-				break;
-#if NBL_BUILTIN_MAX_SCAN_LEVELS>7
+        const uint32_t workgroupSizeLog2 = firstbithigh(glsl::gl_WorkGroupSize().x);
```
devshgraphicsprogramming, line 46: modify it to not use Vulkan/SPIR-V environment builtins, that way we can use it from C++ too!

(So take workgroup size - or more accurately item per workgroup count - from the outside)

include/nbl/builtin/hlsl/scan/declarations.hlsl
```diff
@@ -1,66 +1,64 @@
...
-		in uint pseudoLevel
-	);
-}
-}
-}
-#define _NBL_HLSL_SCAN_GET_PADDED_DATA_DECLARED_
-#endif
+    // REVIEW: Putting topLevel second allows better alignment for packing of constant variables, assuming lastElement has length 4. (https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-packing-rules)
+    struct Parameters_t {
+        uint32_t lastElement[NBL_BUILTIN_MAX_LEVELS/2+1];
```
devshgraphicsprogramming, line 24: add documentation about what `lastElement` is used for

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
+                // climb down
+                _schedulerParams.finishedFlagOffset[3] = _schedulerParams.finishedFlagOffset[1]+2u;
+                
+                _scanParams.temporaryStorageOffset[0] = _schedulerParams.finishedFlagOffset[3]+WorkgroupCount(1);
+                _scanParams.temporaryStorageOffset[1] = _scanParams.temporaryStorageOffset[0]+WorkgroupCount(0);
+                break;
+            case 3u:
+                _schedulerParams.cumulativeWorkgroupCount[1] = _schedulerParams.cumulativeWorkgroupCount[0]+WorkgroupCount(1);
+                _schedulerParams.cumulativeWorkgroupCount[2] = _schedulerParams.cumulativeWorkgroupCount[1]+WorkgroupCount(2);
+                _schedulerParams.cumulativeWorkgroupCount[3] = _schedulerParams.cumulativeWorkgroupCount[2]+1u;
```
devshgraphicsprogramming, line 85: because due to forward progress guarantees you can only do upsweep or downsweep in a single dispatch, you don't need the "climb down"

you probably don't even need the cumulative workgroup counts, as the way you'd find out if your current workgroup truly is the last one, is by checking the return value of the finished flag atomic counter.

e.g. `markFinished`
```cpp
uint32_t howManyDone = flagsAccessor.atomicAdd(finishedOffset[level]+virtualWorkgroupIndexInThisLevel);
//uint32_t currentWorkgroupCount = currentElementCount>>ItemsPerWorkgroupLog2;
uint32_t nextLevelWorkgroupCount = currentWorkgroupCount>>ItemsPerWorkgroupLog2;
// NOT: either last workgroup, or in the last bundle of workgroups and last workgroup dynamically there
if (howManyDone!=(ItemsPerWorkgroup-1) && ((virtualWorkgroupIndexInThisLevel>>ItemsPerWorkgroupLog2)!=nextLevelWorkgroupCount || (virtualWorkgroupIndexInThisLevel&(ItemsPerWorkgroupLog2-1))!=howManyDone) )
    return;
// last one takes over
currentWorkgroupCount = nextLevelWorkgroupCount;
```

devshgraphicsprogramming: btw I'd try to rewrite this as a loop instead of a weird switch that was only written like this to avoid recursion in GLSL

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
+    }
+    
+    /**
+     * treeLevel - the current level in the Blelloch Scan
+     * localWorkgroupIndex - the workgroup index the current invocation is a part of in the specific virtual dispatch. 
+     * For example, if we have dispatched 10 workgroups and we the virtu    al workgroup number is 35, then the localWorkgroupIndex should be 5.
+     */
+    template<class Accessor>
+    bool getWork(NBL_CONST_REF_ARG(DefaultSchedulerParameters_t) params, NBL_CONST_REF_ARG(uint32_t) topLevel, NBL_REF_ARG(uint32_t) treeLevel, NBL_REF_ARG(uint32_t) localWorkgroupIndex, NBL_REF_ARG(Accessor) sharedScratch)
+    {
```
devshgraphicsprogramming, line 117: you can't do any flagging via sharedmemory because you want inter-workgroup communication, must be a BDA address!

devshgraphicsprogramming: even better a `bda::__ptr<uint32_t>` cause then you'll have `atomicAdd` methods

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
+                            dependentsCount -= scanParams.lastElement[treeLevel+1u] * glsl::gl_WorkGroupSize().x;
+                        }
+                    }
+                }
+                uint32_t dependentsFinishedFlagOffset = localWorkgroupIndex;
+                if (treeLevel>topLevel) // !(prevLevel<topLevel) TODO: merge with `else` above?
+                    dependentsFinishedFlagOffset /= glsl::gl_WorkGroupSize().x;
+                dependentsFinishedFlagOffset += params.finishedFlagOffset[prevLevel];
+                while (scanScratchBuf[0].data[dependentsFinishedFlagOffset]!=dependentsCount)
+                    glsl::memoryBarrierBuffer();
```
devshgraphicsprogramming, line 164: I expect a rewrite of this into "last one out closes the door"

include/nbl/builtin/hlsl/scan/default_scheduler.hlsl
```diff
@@ -35,187 +17,181 @@ namespace scan
...
+            if (treeLevel<topLevel)
+            {
+                finishedFlagOffset += localWorkgroupIndex/glsl::gl_WorkGroupSize().x;
+                glsl::atomicAdd(scanScratchBuf[0].data[finishedFlagOffset], 1u);
+            }
+            else if (treeLevel!=(topLevel<<1u))
+            {
+                finishedFlagOffset += localWorkgroupIndex;
+                scanScratchBuf[0].data[finishedFlagOffset] = 1u;
+            }
```
devshgraphicsprogramming, line 187: comments about what and why is happenning in the if statements

include/nbl/builtin/hlsl/scan/scheduler.hlsl
```diff
@@ -0,0 +1,87 @@
...
+  static Scheduler create(NBL_CONST_REF_ARG(Parameters_t) params, NBL_CONST_REF_ARG(DefaultSchedulerParameters_t) schedParams)
+  {
+      Scheduler<WorkgroupSize> scheduler;
+      scheduler.params = params;
+      scheduler.schedParams = schedParams;
+      return scheduler;
+  }
+    
+  template<class Accessor>
+  bool getWork(NBL_REF_ARG(Accessor) accessor)
```
devshgraphicsprogramming, line 33: whats the accessor for? needs a better name

Also why don't you store the accessor past the `create` as a member?

include/nbl/builtin/hlsl/scan/scheduler.hlsl
```diff
@@ -20,75 +20,88 @@ namespace scan
...
     {
-        return level == 0 ? schedParams.cumulativeWorkgroupCount[level] : schedParams.cumulativeWorkgroupCount[level] - schedParams.cumulativeWorkgroupCount[level-1u];
+        // could do scanScratchBuf[0u].workgroupsStarted[SubgroupContiguousIndex()] = 0u but don't know how many invocations are live during this call
+        if(workgroup::SubgroupContiguousIndex() == 0u)
+        {
+            for(uint32_t i = 0; i < params.topLevel; i++)
+            {
+                scanScratchBuf[0u].workgroupsStarted[i] = 0u;
+            }
+        }
```
devshgraphicsprogramming, line 98: don't you know the workgroup size?

kpentaris: My bad, this is actually not used anymore. I think it was being used inside an `inRage` at some point? Not sure. However we can't reset the `workgroupStarted` buffer within the shader at the end of the Reduce step (which was the initial goal) because it's possible that there are still workgroups that are "alive" where they won't be doing any work but will still increase the `workgroupStarted` buffer and some times we end up doing the reset **before** those WGs exit, ending up with some 1 values instead of all 0s.

PR #660, opened 2024-02-22
IES hot fixes

sunho:
fixer upper

PR #654, opened 2024-02-15
Continue with IES

devshgraphicsprogramming:
## TODO list:

Continues from : https://github.com/Devsh-Graphics-Programming/Nabla/pull/630

PR #671, opened 2024-03-22
IES loader rewrite

devshgraphicsprogramming:
## TODO list:

- [x] Fix comments from https://github.com/Devsh-Graphics-Programming/Nabla/pull/660
- [x] and https://github.com/Devsh-Graphics-Programming/Nabla/pull/654
- [x] add a `flatten` float property to `CElementEmissionProfile` similar to `normalizeEnergy`, with valid values in the `[0,1)` range... a straight up 1.0 would mean no IES profile so its not valid.

### How `flatten` is supposed to work

Basically we might want to make the IES profile slightly more emissive in angles where its "dim" to simulate dispersion through light fixtures and milky diffusers, such that the light emitter doesn't appear "too dark" to an observer outside the main cone.

This means we only have two options to modify the profile, because we want some minimal energy in each direction:
- `orig<K ? (orig>0 ? C:0):orig` minimal emission `C`
- `C*b+orig*(1-b)`

Further constraint is that the IES shoudln't emit light in the directions it didn't before, so whenever `original(dir)==0` then `modified(dir)==0` this is to ensure we don't turn hemispherical profiles into spherical, etc.

This "modification" shouldn't affect the overall energy outputted by an IES so:
```
Integral ModifiedProfile === Integral OriginalProfile
```

The last constraint complicates things for the first choice as we have to make sure that
```
Integral_{Domain where 0<orig<K} C dOmega == Integral_{Domain where 0<orig<K} orig(omega) dOmega
```
so basically C needs to equal the average emission over the domain where original emission was greater than 0 but less than `K`

The whole purpose was to make sure that the emitter is at least `C` bright to an observer outside the cone, however because `C` will be an average over the solid angles where `orig<K` we know that `C<K`

Basically to find out what `K` should be given `C` we would need to have a histogram of emission values weighted by their solid angle, then iterate through it until the average of entries before the iterators reaches `C`. However this introduces a problem because histogram of an interpolated IES is continuous, which would lead us to needing to resort to doing a bisection search with each step involving an integral. Furthermore this `clamp` cannot be modelled by any physical process.

The second option is better, because then we have
```
Integral_{Domain where orig>0} C*b+orig*(1-b) dOmega == Integral_{Sphere} orig dOmega
```
where we already know the RHS (TotalEnergy) as we store it in the metadata. Also
```
Integral_{Domain where orig>0} orig dOmega = Integral_{Sphere} orig dOmega == TotalEnergy
```

So then we have
```
C*b Integral_{Domain where orig>0} dOmega == TotalEnergy*b
```
and
```
C = TotalEnergy / NonZeroEmissionDomainSize
```
which just-so happens to be the "true average" emission.

Therefore `flatten` must be a blend between the original IES image and its average value.


src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,82 @@
...
+
+                core::vector<IES_STORAGE_FORMAT>& getHoriAngles() { return hAngles; }
+                const core::vector<IES_STORAGE_FORMAT>& getHoriAngles() const { return hAngles; }
+                core::vector<IES_STORAGE_FORMAT>& getVertAngles() { return vAngles; }
+                const core::vector<IES_STORAGE_FORMAT>& getVertAngles() const { return vAngles; }
+                core::vector<IES_STORAGE_FORMAT>& getData() { return data; }
+                const core::vector<IES_STORAGE_FORMAT>& getData() const { return data; }
+
+                IES_STORAGE_FORMAT getValue(size_t i, size_t j) const { return data[getVertAngles().size() * i + j]; }
+                IES_STORAGE_FORMAT getMaxValue() const { return *std::max_element(std::begin(data), std::end(data)); }
```
devshgraphicsprogramming, line 48: this should really be precomputed

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,82 @@
...
+                const IES_STORAGE_FORMAT& sample(IES_STORAGE_FORMAT vAngle, IES_STORAGE_FORMAT hAngle) const;
+                const IES_STORAGE_FORMAT& getIntegral() const;
+
+                //! Candlepower distribution curve plot as ICPUImageView
+                /*
+                    Creates 2D texture of CDC with width & height extent, zAngleDegreeRotation may be
+                    used to rotate normalized direction vector obtained from octahdronUVToDir utility with
+                */
+
+                core::smart_refctd_ptr<asset::ICPUImageView> createCDCTexture(const float& zAngleDegreeRotation = 0.f, const size_t& width = CDC_DEFAULT_TEXTURE_WIDTH, const size_t& height = CDC_DEFAULT_TEXTURE_HEIGHT) const;
```
devshgraphicsprogramming, line 67: get rid of the Z-rotation

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,153 @@
...
+    double dTheta = core::PI<double>() / numVSamples;
+    double dPhi = 2 * core::PI<double>() / numHSamples;
+    double dThetaInAngle = MAX_VANGLE / numVSamples;
+    double dPhiInAngle = MAX_HANGLE / numHSamples;
+    double res = 0;
+    for (size_t i = 0; i < numVSamples; i++) {
+        for (size_t j = 0; j < numHSamples; j++) {
+            res += dPhi * dTheta * sin(dTheta * i) * sample(i * dThetaInAngle, j * dPhiInAngle);
+        }
+    }
```
devshgraphicsprogramming, line 19: compute straight from the original irregular grid, no resampling

src/nbl/asset/utils/CIESProfile.h
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfile.h
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: done, but we have 0.089475341451795226 error, I need to think of it

devshgraphicsprogramming: that error is probably because the sampled thing was obviously inaccurate

src/nbl/asset/utils/CIESProfileParser.h
```diff
@@ -0,0 +1,33 @@
...
+{
+    namespace asset
+    {
+        class CIESProfileParser 
+        {
+            public:
+                CIESProfileParser(char* buf, size_t size) { ss << std::string(buf, size); }
+
+                const char* getErrorMsg() const { return errorMsg; }
+                bool parse(CIESProfile& result);
```
devshgraphicsprogramming, line 20: document the IES spherical coordinates to illuminant target, and whether the X-Y (cos(phi), sin(phi)) are LH or RH

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,89 @@
...
+
+                void setValue(size_t i, size_t j, IES_STORAGE_FORMAT val) { data[vAngles.size() * i + j] = val; }
+
+                const IES_STORAGE_FORMAT& sample(IES_STORAGE_FORMAT vAngle, IES_STORAGE_FORMAT hAngle) const;
+
+                PhotometricType type;
+                core::vector<IES_STORAGE_FORMAT> hAngles;   //! horizontal angles, represents "theta" spherical coordinate angle (in degrees) compoment with physics convention
+                IES_STORAGE_FORMAT hAOffset = {};           //! real hAngle value read from IES file is hAngles[x] + hAOffset
+                core::vector<IES_STORAGE_FORMAT> vAngles;   //! vertical angles, represents "phi" spherical coordinate angle (in degrees) compoment with physics convention
+                IES_STORAGE_FORMAT vAOffset = {};           //! real vAngle value read from IES file is vAngles[x] + vAOffset
```
devshgraphicsprogramming, line 79: or it can be here makes more sense

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,89 @@
...
+
+                core::smart_refctd_ptr<asset::ICPUImageView> createCDCTexture(const size_t& width = CDC_DEFAULT_TEXTURE_WIDTH, const size_t& height = CDC_DEFAULT_TEXTURE_HEIGHT) const;
+
+            private:
+                CIESProfile(PhotometricType type, size_t hSize, size_t vSize)
+                    : type(type), hAngles(hSize), vAngles(vSize), data(hSize* vSize) {}
+
+                // TODO for @Hazard, I would move it into separate file, we may use this abstraction somewhere too
+                static inline std::pair<float, float> sphericalDirToRadians(const core::vectorSIMDf& dir);
+                static inline core::vectorSIMDf octahdronUVToDir(const float& u, const float& v);
```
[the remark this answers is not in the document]
devshgraphicsprogramming: while we're at it we should write that Spherical coords in Nabla are always
```
x = cos(phi)*sin(theta)
y = sin(phi)*sin(theta)
z = cos(theta)
```

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -0,0 +1,132 @@
...
+        return false;
+    }
+    if (vAngles[0] != 0.0 && vAngles[0] != 90.0) {
+        errorMsg = "First angle must be 0 or 90 in type C";
+        return false;
+    }
+    if (vAngles[vSize - 1] != 90.0 && vAngles[vSize - 1] != 180.0) {
+        errorMsg = "Last angle must be 90 or 180 in type C";
+        return false;
+    }
```
devshgraphicsprogramming, line 98: reject 90,90 combo

check if the spec allows for 90-180 range

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -0,0 +1,132 @@
...
+                result.setValue(result.hAngles.size() - 1, j, result.getValue(i, j));
+        }
+    }
+    */
+
+    // to interprete angles pairs starting with (0, phi) for proper computations we shift the angles
+    result.hAOffset = result.hAngles[0];
+    if (result.hAOffset != 0)
+        for (auto& hAngle : result.hAngles)
+            hAngle = std::fmod((hAngle - result.hAOffset + 360), 360);
```
devshgraphicsprogramming, line 127: this makes no sense?

why are you adding 360? why are you going module 360?

aren't angles outside of 0,359.999 banned?

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,145 @@
...
+    for (size_t i = 0; i < numVSamples; i++) {
+        for (size_t j = 0; j < numHSamples; j++) {
+            auto val = dPhi * dTheta * std::sin(dTheta * i) * sample(i * dThetaInAngle, j * dPhiInAngle);
+            res += val;
+        }
+    }
+    return res;
+}
+
+const CIESProfile::IES_STORAGE_FORMAT& CIESProfile::sample(double vAngle, double hAngle) const {
```
devshgraphicsprogramming, line 24: why are you returning a reference to a temporary!?

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,145 @@
...
+    if (vAngle > vAngles.back()) return 0.0;
+    hAngle = hAngles.back() == 0.0
+        ? 0.0
+        : fmod(hAngle,
+            hAngles.back()); // when last horizontal angle is zero
+
+    // it's symmetric across all planes
```
devshgraphicsprogramming, line 31: check for `hAngles.size()<2` and do an if statement, and have a SPECIAL way to handle it, set
```
hAngle = hAngles.front();
```

devshgraphicsprogramming: `fmod` is not the correct operation, it radially repeats, the IES Viewer seems to mirror on a 90-270 degree test IES

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,89 @@
...
+                _NBL_STATIC_INLINE_CONSTEXPR IES_STORAGE_FORMAT MAX_HANGLE = 360.0;
+
+                enum PhotometricType : uint32_t 
+                {
+                    TYPE_NONE,
+                    TYPE_C,
+                    TYPE_B,
+                    TYPE_A,
+                };
+
```
devshgraphicsprogramming, line 33: extract and store (with enum) the spec version the file conforms to 1995 or 2002

src/nbl/asset/utils/CIESProfile.cpp
devshgraphicsprogramming: just stop with the FMOD, we'll patch 1995 profiles to 2002 spec outside

src/nbl/asset/utils/CIESProfileParser.cpp
devshgraphicsprogramming: > For Type C photometry, the first value shall always
be 0 degrees, and the last value shall be one of the
following:
1) 0 - in this case, there is only one horizontal angle,
and the luminaire is assumed to be laterally
symmetric in all planes.
2) 90 - the luminaire is assumed to be symmetric in
each quadrant.
3) 180 - the luminaire is assumed to be symmetric
about the 0 to 180 degree plane.
4) 360 - in this case, the luminaire is assumed to
exhibit no lateral symmet

devshgraphicsprogramming: do validation of this

Only if the version is 1995, allow also start at 90 and end at 270, assume symmetry about the 90 to 270 degree plane

devshgraphicsprogramming: need loader to spit out ICPUImageView + Sampler Wrap modes as metadata

src/nbl/asset/utils/CIESProfile.h
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfile.h
AnastaZIuk: removed

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: > 1995 header + [90,270] -> OTHER_HALF_SYMMETRIC, maps 89 -> 91 and 269 -> 271
I think you meant `271 -> 269`, right?

AnastaZIuk: > just stop with the FMOD, we need the cases turned into enums you handle with a switch

resolved anyway

> 1995 header + [90,270] -> OTHER_HALF_SYMMETRIC, maps 89 -> 91 and 269 -> 271
OTHER_HALF_SYMMETRIC horizontal angles patched to have HALF_SYMETRIC's range, 90 degrees offset is being saved for composition and order

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -39,21 +39,22 @@ const CIESProfile::IES_STORAGE_FORMAT CIESProfile::sample(IES_STORAGE_FORMAT the
...
+    auto lb = [](const core::vector<double>& angles, double angle) -> size_t
+    {
+        const size_t idx = std::upper_bound(std::begin(angles), std::end(angles), angle) - std::begin(angles);
+        return (size_t)std::max((int64_t)idx - 1, (int64_t)0);
+    };
+
+    auto ub = [](const core::vector<double>& angles, double angle) -> size_t
+    {
+        const size_t idx = std::upper_bound(std::begin(angles), std::end(angles), angle) - std::begin(angles);
+        return std::min<size_t>(idx, angles.size() - 1);
```
devshgraphicsprogramming, line 51: assert `!angles.empty()`

AnastaZIuk:
TODO main list completed

src/nbl/asset/utils/CIESProfile.h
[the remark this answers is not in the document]
AnastaZIuk: > while we're at it we should write that Spherical coords in Nabla are always
> 
> ```
> x = cos(phi)*sin(theta)
> y = sin(phi)*sin(theta)
> z = cos(theta)
> ```

these are cartesian coordinates obtained from spherical coordinates (we also assume radius is 1 here)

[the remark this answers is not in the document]
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfileParser.h
AnastaZIuk: resolved

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -172,4 +224,77 @@ core::smart_refctd_ptr<asset::ICPUImageView> CIESProfile::createCDCTexture(const
...
+{
+    if constexpr (policy._Parallelize)
+        static_assert(false, "meh it appears that filters with par policy are broken on this branch, use seq policy till fixed");
+
+    const bool inFlattenDomain = flatten >= 0.0 && flatten < 1.0; // [0, 1) range for blend equation, 1 is invalid
+
+    if (!inFlattenDomain)
+        return false;
+
+    auto outIES = inoutIES->getCreationParameters().image;
```
devshgraphicsprogramming, line 249: don't do the `flatten` handling in the Loader!

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -60,15 +64,46 @@ namespace nbl
...
+                const IES_STORAGE_FORMAT& getAvgEmmision() const { return avgEmmision; }
 
-                core::smart_refctd_ptr<asset::ICPUImageView> createCDCTexture(const size_t& width = CDC_DEFAULT_TEXTURE_WIDTH, const size_t& height = CDC_DEFAULT_TEXTURE_HEIGHT) const;
+                template<class ExecutionPolicy>
+                core::smart_refctd_ptr<asset::ICPUImageView> createIESTexture(ExecutionPolicy&& policy, const size_t& width = CDC_DEFAULT_TEXTURE_WIDTH, const size_t& height = CDC_DEFAULT_TEXTURE_HEIGHT) const;
+                core::smart_refctd_ptr<asset::ICPUImageView> createIESTexture(const size_t& width = CDC_DEFAULT_TEXTURE_WIDTH, const size_t& height = CDC_DEFAULT_TEXTURE_HEIGHT) const;
+                
+                template<class ExecutionPolicy>
+                static bool flattenIESTexture(ExecutionPolicy&& policy, const IES_STORAGE_FORMAT& avgEmmision, core::smart_refctd_ptr<asset::ICPUImageView> inoutIES, const float flatten = 0.0);
+                static bool flattenIESTexture(const IES_STORAGE_FORMAT& avgEmmision, core::smart_refctd_ptr<asset::ICPUImageView> inoutIES, const float flatten = 0.0);
```
devshgraphicsprogramming, line 75: this needs to go into the Mitsuba Loader / Material Frontend, where the IES profiles get cached and loaded

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -86,27 +87,59 @@ CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend:
...
-            if (normalizeEnergy == 0.0f) {
-                res->intensity *= meta->getMaxIntensity();
+
+            auto worldSpaceIESTransform = core::concatenateBFollowedByA(transform, _emitter->transform.matrix);
+            
+            // fill .w component of 1 & 2 row with 0 to perform normmalization on .xyz vectors for these rows bellow to get up & view vectors
+            worldSpaceIESTransform[1].w = 0;
+            worldSpaceIESTransform[2].w = 0;
+
+            _NBL_STATIC_INLINE_CONSTEXPR float THRESHOLD = std::numeric_limits<float>::epsilon();
```
devshgraphicsprogramming, line 114: use a higher threshold like `exp2(-14.f)`

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -86,27 +87,59 @@ CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend:
...
+            const auto det = core::determinant(worldSpaceIESTransform);
+
+            if (abs(det) < THRESHOLD) // protect us from determinant = 0 where inverse transform doesn't exist because the matrix is singular, also we don't want to be too much close to 0 because of the matrix conditioning index becoming higher and higher
+            {
+                os::Printer::log("ERROR: Emission profile rejected because determinant of transformation matrix does not exceed the minimum threshold and is too close or equal 0", ELL_ERROR);
+                return; // exit the lambda
             }
+
+            profile.right_hand = core::sign(det) > 0.0f;
```
devshgraphicsprogramming, line 123: the core::sign here is pointless

include/nbl/ext/MitsubaLoader/CElementEmissionProfile.h
```diff
@@ -58,7 +57,7 @@ struct CElementEmissionProfile : public IElement {
 	std::string getLogName() const override { return "emissionprofile "; }
 
 	std::string filename;
-	float normalizeEnergy;
+	float normalizeEnergy, flatten;
```
devshgraphicsprogramming, line 60: you're missing the logic/reg that loads the `flatten` from the XML, add it to a scene to test the parsing

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -253,105 +11,26 @@ CIESProfileLoader::loadAsset(io::IReadFile* _file,
...
-            double I = profile.sample(core::degrees(theta), core::degrees(phi));
-            uint16_t value = static_cast<uint16_t>(
-                std::clamp(I * maxValueRecip * 65535.0, 0.0, 65535.0));
-            *reinterpret_cast<uint16_t*>(
-                bufferPtr + i * bufferRowLength * texelBytesz + j * texelBytesz) =
-                value;
-        }
-    }
+    if (_params.loaderFlags & E_LOADER_PARAMETER_FLAGS::ELPF_LOAD_METADATA_ONLY)
+        cpuImageView = _override->findDefaultAsset<ICPUImageView>("nbl/builtin/image_view/dummy2d", loadContex, _hierarchyLevel).first; // note: we could also pass empty content, but this would require adjusting IAssetLoader source to not attempt to use all loaders to find the asset
```
devshgraphicsprogramming, line 31: what do other loaders return?

I think it could be acceptable to return an array of 1 nullptr image view, or create an IES Texture with no regions

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,137 @@
...
+{
+    namespace asset 
+    {
+        class CIESProfile 
+        {
+            public:
+                using IES_STORAGE_FORMAT = double;
+
+                _NBL_STATIC_INLINE_CONSTEXPR size_t CDC_DEFAULT_TEXTURE_WIDTH = 1024;
+                _NBL_STATIC_INLINE_CONSTEXPR size_t CDC_DEFAULT_TEXTURE_HEIGHT = 1024;
```
devshgraphicsprogramming, line 22: https://github.com/Devsh-Graphics-Programming/Nabla/pull/630#discussion_r1431351882
https://github.com/Devsh-Graphics-Programming/Nabla/pull/630#discussion_r1431357799

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,137 @@
...
+                    QUAD_SYMETRIC,              //! The luminaire is assumed to be symmetric in each quadrant
+                    HALF_SYMETRIC,              //! The luminaire is assumed to be symmetric about the 0 to 180 degree plane
+                    OTHER_HALF_SYMMETRIC,       //! HALF_SYMETRIC case for legacy V_1995 version where horizontal angles are in range [90, 270], in that case the parser patches horizontal angles to be HALF_SYMETRIC
+                    NO_LATERAL_SYMMET           //! The luminaire is assumed to exhibit no lateral symmet
+                };
+
+                CIESProfile() = default;
+                ~CIESProfile() = default;
+
+                const IES_STORAGE_FORMAT& getHAnglesOffset() const { return hAnglesOffset; }
```
devshgraphicsprogramming, line 57: why do you need to store `hAnglesOffset` wouldn't it be either 0 or 90 depending on `OTHER_HALF_SYMMETRIC` ?

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,137 @@
...
+                    when viewed as a texture, the net folds, and the apex where the seams join is Z+
+                */
+                
+                static inline core::vectorSIMDf octahdronUVToDir(const float& u, const float& v);
+                
+                void addHoriAngle(IES_STORAGE_FORMAT hAngle)
+                {
+                    hAngles.push_back(hAngle);
+                    data.resize(getHoriAngles().size() * vAngles.size());
+                }
```
devshgraphicsprogramming, line 113: hide this, you probably only need it at a single point in time, great candidate for being a lambda or something

src/nbl/asset/interchange/CIESProfileLoader.cpp
```diff
@@ -253,105 +11,26 @@ CIESProfileLoader::loadAsset(io::IReadFile* _file,
...
+    else
+        cpuImageView = profile.createIESTexture();
```
devshgraphicsprogramming, line 33: https://github.com/Devsh-Graphics-Programming/Nabla/pull/630#discussion_r1431351882

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+        const double horiInv = 1.0 / width;
+
+        const double dTheta = core::PI<double>() / height;
+        const double dPhi = 2 * core::PI<double>() / width;
+
+        integral = 0.0;
+        size_t nonZeroEmissionDomainSize = 0u;
+
+        std::atomic<IES_STORAGE_FORMAT> aIntegral = 0.0;
+        std::atomic<size_t> aNonZeroEmissionDomainSize = 0;
```
devshgraphicsprogramming, line 175: if you use atomics, you'll get negligible benefit from parallelization because of atomic contention.

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// this would not be required if we were to use c++20 on this branch
+template<typename T>
+T atomic_fetch_add(std::atomic<T>* obj, T arg) {
+    T expected = obj->load();
+    while (!std::atomic_compare_exchange_weak(obj, &expected, expected + arg));
+
+    return expected;
+}
```
devshgraphicsprogramming, line 113: the proper way to parallelize is for each thread/lane to allocate/grab its own temp `T` value and += into that by itself, and when the job completes, add up all the temps together, this is better because:
- no atomic contention
- IEEE754 issues with adding small values to a large accumulator are mitigated

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+template<class ExecutionPolicy>
+core::smart_refctd_ptr<asset::ICPUImageView> CIESProfile::createIESTexture(ExecutionPolicy&& policy, const size_t& width, const size_t& height) const
+{
+    if constexpr (policy._Parallelize)
+        static_assert(false, "meh it appears that filters with par policy are broken on this branch, use seq policy till fixed");
```
devshgraphicsprogramming, line 119: how?

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+        std::atomic<IES_STORAGE_FORMAT> aIntegral = 0.0;
+        std::atomic<size_t> aNonZeroEmissionDomainSize = 0;
+
+        auto fill = [&](uint32_t blockArrayOffset, core::vectorSIMDu32 position) -> void
+        {
+            const auto dir = octahdronUVToDir(((float)position.x + 0.5) * vertInv, ((float)position.y + 0.5) * horiInv);
+            const auto [theta, phi] = sphericalDirToRadians(dir);
+            const auto& intensity = sample(theta, phi);
+            const auto& value = intensity * maxValueRecip;
```
devshgraphicsprogramming, line 182: stop taking references to temporaries

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+                if constexpr (policy._Parallelize)
+                {
+                    atomic_fetch_add<IES_STORAGE_FORMAT>(&aIntegral, integrationV);
+                    aNonZeroEmissionDomainSize.fetch_add(1u);
+                }
+                else
+                {
+                    integral += integrationV;
+                    ++nonZeroEmissionDomainSize;
+                }
```
devshgraphicsprogramming, line 196: 1. The integral should be performed while parsing/loading the values in (so its independent of image size)
2. While parsing the irregular grid you can work out the texture resolution needed: https://github.com/Devsh-Graphics-Programming/Nabla/pull/630#discussion_r1431351882
3. You don't just do +1 on your domain size, you need to sum `dPhi*dTheta*sin(theta)` the differential element of the domain

Do everything here while parsing

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+            const uint16_t encodeV = static_cast<uint16_t>(std::clamp(value * UI16_MAX_D, 0.0, UI16_MAX_D));
```
devshgraphicsprogramming, line 198: remember about adding 0.5 to the value before mult by `UI16_MAX_D`

https://github.com/Devsh-Graphics-Programming/Nabla/pull/630#discussion_r1468267196

src/nbl/asset/utils/CIESProfile.cpp
devshgraphicsprogramming: another reason not to do it here, is that you can log failures properly and skip certain checks (like format)

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+    const auto& creationParams = outIES->getCreationParameters();
+    const bool validFormat = creationParams.format == CIESProfile::IES_TEXTURE_STORAGE_FORMAT;
+
+    if (!validFormat)
+        return false;
+
+    if (flatten > 0.0) // skip if 0 because its just copying texels
+    {
+        CFillImageFilter::state_type state;
+        state.outImage = outIES.get();
```
devshgraphicsprogramming, line 260: you can't modify the IES you've been passed, needs to be a new ICPUImage+View, with the view inserted into the cache with an extra suffix `?flatten=value` on the key

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+        const IImageFilter::IState::ColorValue::ReadMemoryInfo rInfo(creationParams.format, outIES->getBuffer()->getPointer());
+
+        auto fill = [&](uint32_t blockArrayOffset, core::vectorSIMDu32 position) -> void
+        {
+            state.fillValue.readMemory(rInfo, blockArrayOffset);
+
+            const auto decodeV = (IES_STORAGE_FORMAT)(*state.fillValue.asUShort) / CIESProfile::UI16_MAX_D;
+            const auto blendV = decodeV * (1.0 - flatten) + avgEmmision * flatten; //! blend the IES texture with "flatten"
+            const uint16_t encodeV = static_cast<uint16_t>(std::clamp(blendV * CIESProfile::UI16_MAX_D, 0.0, CIESProfile::UI16_MAX_D));
+
```
devshgraphicsprogramming, line 276: don't we have `encode` and `decode` functions for this format in `asset` namespace?

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -86,27 +87,59 @@ CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend:
...
+        auto fillProfile = [&]() -> void
+        {
+            if (_emitter->area.emissionProfile->flatten > 0)
+            {
+                auto clone = core::smart_refctd_ptr_static_cast<asset::ICPUImageView>(image->clone()); // TODO: check if its already a copy
+                if (!asset::CIESProfile::flattenIESTexture(meta->profile.getAvgEmmision(), clone, _emitter->area.emissionProfile->flatten))
+                {
+                    os::Printer::log("ERROR: Emission profile rejected because flatten operation on the IES texture failed!", ELL_ERROR);
+                    return; // exit the lambda
+                }
```
devshgraphicsprogramming, line 100: if there are any issues with `flatten` param being wrong (range, etc.) then log error but use the original unflattened IES

src/nbl/asset/utils/CIESProfile.h
devshgraphicsprogramming: ok there's one possibility, this could be a non-static method, then it could be merged with `createIESTexture` with `flatten` being the last parameter modifying the `sample` directly.

But for this you need the Max, Integral, etc. to be computed with parsing so all the values are ready.

src/nbl/asset/utils/CIESProfile.h
```diff
@@ -0,0 +1,137 @@
...
+                IES_STORAGE_FORMAT hAnglesOffset;                           //! The real horizontal angle provided by IES data is (hAngles[index] + hAnglesOffset) - the reason behind it is we patch 1995 IES OTHER_HALF_SYMETRIC case to be HALF_SYMETRIC
+                core::vector<IES_STORAGE_FORMAT> vAngles;                   //! Measurements in degrees of angular displacement measured counterclockwise in a horizontal plane for Type C photometry and clockwise for Type A and B photometry, a value represents spherical coordinate "phi" with physics convention
+                core::vector<IES_STORAGE_FORMAT> data;                      //! Candela values
+                IES_STORAGE_FORMAT maxValue = {};                           //! Max value from this->data vector
+                
+                mutable IES_STORAGE_FORMAT integral = {};                  //! Total energy emitted
+                mutable IES_STORAGE_FORMAT avgEmmision = {};                //! this->integral / <size of the emission domain where non zero values>
```
devshgraphicsprogramming, line 130: if you compute while parsing, they don't need to be mutable

devshgraphicsprogramming: also precompute `optimalResolution`

src/nbl/asset/utils/CIESProfileParser.h
```diff
@@ -0,0 +1,90 @@
...
+                    
+                    Therefore to obtain illuminant target candela value for vertical angle θ and horizontal angle φ you need to access 
+                    
+                    ● <2D texture grid of candela values pointer>[<<θ angles vector>.size()> * <index of θ angle> + <index of φ angle>]
+
+                    For more details see @CIESProfile class implementation
+
+                    Also note that
+
+                    ● a point (cos(φ), sin(φ)) on XY IES plane is denoted in LH cartesian coordinate system
```
devshgraphicsprogramming, line 73: more documentation.

Discord convo to jog your mind
```
Width is perpendicular to 0 degree direction
so Y+ is cos(phi) 
X- is sin(phi)

Nabla X+ to jest IES 270 deg
Y+ to IES 0 deg
```

dress it up more nicely

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -150,12 +157,58 @@ bool CIESProfileParser::parse(CIESProfile& result)
...
+    constexpr auto FULL_SOLID_ANGLE = 4.0f * core::PI<float>();
+
+    if (result.symmetry == CIESProfile::ISOTROPIC && result.vAngles.size() == 1) // not sure if any IES provides this kind of distribution but it should be possible to do so meaning fixed candela value for each direction
+    {
+        const auto candelaValue = result.getCandelaValue(0, 0);
+        constexpr auto solidAngle = FULL_SOLID_ANGLE; // spans whole sphere area
+
+        totalEmissionIntegral = candelaValue * solidAngle;
+        nonZeroEmissionDomainSize = solidAngle;
+    }
```
devshgraphicsprogramming, line 177: not possible, why:
- if you look at your own validation this would only be posible for the array {90}
- your `CIESProfile::sample` would return non-zero distribution only for `vAngle=90`, meaning a 0 integral

TL;DR we should reject as invalid, emission profiles with `vSize<2` when loading

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -150,12 +157,58 @@ bool CIESProfileParser::parse(CIESProfile& result)
...
+                        return core::PI<float>() * 2.0f; // spans whole circle
+                    else
+                        return core::radians<float>((i < result.hAngles.size() - 1) ? result.hAngles[i + 1] - result.hAngles[i] : result.hAngles[i] - result.hAngles[i - 1]);
+                };
+
+                const float thetaRad = core::radians<float>(result.vAngles[j]);
+                const float dThetaRad = core::radians<float>((j < result.vAngles.size() - 1) ? result.vAngles[j + 1] - result.vAngles[j] : result.vAngles[j] - result.vAngles[j - 1]);
+                const float dPhiRad = getDPhiRad();
+
+                const auto differentialSolidAngle = std::clamp<float>(std::sin(thetaRad) * dThetaRad * dPhiRad, 0.0, FULL_SOLID_ANGLE);
```
devshgraphicsprogramming, line 196: you need to integrate `sin(theta)` w.r.t there though, you can actually compute the solid angle of a [theta]x[phi] patch exactly

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -150,12 +157,58 @@ bool CIESProfileParser::parse(CIESProfile& result)
...
+            {
+                const auto candelaValue = result.getCandelaValue(i, j);
+
+                auto getDPhiRad = [&]()
+                {
+                    if (result.symmetry == CIESProfile::ISOTROPIC) // this could be nice templated lambda, TODO: correct once c++20 enters this branch
+                        return core::PI<float>() * 2.0f; // spans whole circle
+                    else
+                        return core::radians<float>((i < result.hAngles.size() - 1) ? result.hAngles[i + 1] - result.hAngles[i] : result.hAngles[i] - result.hAngles[i - 1]);
+                };
```
devshgraphicsprogramming, line 190: you can hoist `dPhi` computation outside the `vAngle` loop

devshgraphicsprogramming: as a matter of readability I'd prefer
```
i ? (hAngles[i]-hAngles[i-1]):(360.f-hAngles.back()+hAngles.front())
```
to
```
(i < result.hAngles.size() - 1) ? result.hAngles[i + 1] - result.hAngles[i] : result.hAngles[i] - result.hAngles[i - 1])
```

Also I think you have a bug when `i==hangles.size()-1` you're getting the phi difference of the wrong points (last and second to last)

src/nbl/asset/utils/CIESProfileParser.cpp
devshgraphicsprogramming: also clamping here is not necessary or beneficial, you can clamp the `nonZeroEmissionDomainSize` at the end instead

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -150,12 +157,58 @@ bool CIESProfileParser::parse(CIESProfile& result)
...
+
+                const float thetaRad = core::radians<float>(result.vAngles[j]);
+                const float dThetaRad = core::radians<float>((j < result.vAngles.size() - 1) ? result.vAngles[j + 1] - result.vAngles[j] : result.vAngles[j] - result.vAngles[j - 1]);
```
devshgraphicsprogramming, line 193: because to integrate you need to compute `cos(loTheta)-cos(hiTheta)` I'd get them individually.

Also because there's no grid wrapparound for this dimension, `j` needs to vary from 0 to `vAngles.size()-1` (one per gap)

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -150,12 +157,58 @@ bool CIESProfileParser::parse(CIESProfile& result)
...
+                    nonZeroEmissionDomainSize += differentialSolidAngle;
+                }
+            }
+
+    if (nonZeroEmissionDomainSize <= 0) // protect us from division by 0 (just in case, we should never hit it)
+        return false;
+
+    result.avgEmmision = totalEmissionIntegral / static_cast<decltype(totalEmissionIntegral)>(nonZeroEmissionDomainSize);
+    result.totalEmissionIntegral = totalEmissionIntegral * fluxMultiplier; // we use fluxMultiplier to calculate final total emission for case where we have some symmetry between planes (fluxMultiplier is 1.0f if ISOTROPIC or NO_LATERAL_SYMMET because they already have correct total emission integral calculated), also note it doesn't affect average emission at all
+    result.maxCandelaValue = *std::max_element(std::begin(result.data), std::end(result.data));
```
devshgraphicsprogramming, line 211: pointless to do the max_element now, could find it while integrating

src/nbl/asset/utils/CIESProfileParser.cpp
devshgraphicsprogramming: aaah I forgot about domain symmetries and that `hAngles.front()=DOMAIN_START` and `hAngles.back()=DOMAIN_END`

in this case all you need is
```cpp
result.symmetry!=ISOTROPIC ? (hAngles[i+1]-hAngles[i]):twoPI;
```

and for `i` to bary from 0 to `hAngles.size()-2`

src/nbl/asset/utils/CIESProfile.h
devshgraphicsprogramming: you can work out the Octagonal UV coordinates of the `(theta,phi)` points as you integrate and their Quadrilateral in UV space

src/nbl/asset/utils/CIESProfile.h
AnastaZIuk: I would keep separate `flattenIESTexture` while still allow to pass flatten into `createIESTexture` because it wouldn't be nice to resample just to flatten an ies, this way the utility can blend ready texels values from `inoutIES` but it's possible to flatten ies being created in place

src/nbl/asset/interchange/CIESProfileLoader.cpp
AnastaZIuk: also nobody has used `ELPF_LOAD_METADATA_ONLY` before, another solution would be to create dummy image without regions

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: sure

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: we do, will use them instead

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: we could either adjust them or reject as suggested, will take the second option

src/nbl/asset/utils/CIESProfile.h
AnastaZIuk: meh I will move it somewhere to Mitsuba as suggested, we don't have access to IES profile directly anyway but to meta only

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: also made it use asset:: encode

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -0,0 +1,300 @@
...
+template<class ExecutionPolicy>
+bool CIESProfile::flattenIESTexture(ExecutionPolicy&& policy, const IES_STORAGE_FORMAT& avgEmmision, core::smart_refctd_ptr<asset::ICPUImageView> inoutIES, const float flatten)
+{
+    if constexpr (policy._Parallelize)
+        static_assert(false, "meh it appears that filters with par policy are broken on this branch, use seq policy till fixed");
+
+    const bool inFlattenDomain = flatten >= 0.0 && flatten < 1.0; // [0, 1) range for blend equation, 1 is invalid
+
+    if (!inFlattenDomain)
+        return false;
```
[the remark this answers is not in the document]
AnastaZIuk: resolved in following commit

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: resolved in following commit

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: resolved in following commit

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: > also clamping here is not necessary or beneficial, you can clamp the `nonZeroEmissionDomainSize` at the end instead

resolved in following commit

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: > Also because there's no grid wrapparound for this dimension, j needs to vary from 0 to vAngles.size()-2 (one per gap)

resolved in following commit

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: now we host the utility in the profile which doesn't have access to cache, so we need to assume the input is already the new view inserted into cache with an extra key `?flatten=value` suffix

AnastaZIuk: if we really want to make new asset here with the key and insert into cache in the utility we need to move it into mitsuba sources probably

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -167,48 +171,37 @@ bool CIESProfileParser::parse(CIESProfile& result)
     float totalEmissionIntegral = 0.0, nonZeroEmissionDomainSize = 0.0;
     constexpr auto FULL_SOLID_ANGLE = 4.0f * core::PI<float>();
 
-    if (result.symmetry == CIESProfile::ISOTROPIC && result.vAngles.size() == 1) // not sure if any IES provides this kind of distribution but it should be possible to do so meaning fixed candela value for each direction
+    for (size_t i = 0; i < result.hAngles.size() - 2; i++)
```
devshgraphicsprogramming, line 174: -1 not -2

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -167,48 +171,37 @@ bool CIESProfileParser::parse(CIESProfile& result)
...
-        constexpr auto solidAngle = FULL_SOLID_ANGLE; // spans whole sphere area
+        const float dPhiRad = result.symmetry != CIESProfile::ISOTROPIC ? (hAngles[i + 1] - hAngles[i]) : core::PI<float>() * 2.0f;
 
-        totalEmissionIntegral = candelaValue * solidAngle;
-        nonZeroEmissionDomainSize = solidAngle;
-    }
-    else
-        for (size_t i = 0; i < result.hAngles.size(); i++)
-            for (size_t j = 0; j < result.vAngles.size(); j++)
+        for (size_t j = 0; j < result.vAngles.size() - 2; j++)
```
devshgraphicsprogramming, line 178: -1 not -2

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -167,48 +171,37 @@ bool CIESProfileParser::parse(CIESProfile& result)
...
+        {
+            const auto candelaValue = result.getCandelaValue(i, j);
+
+            if (result.maxCandelaValue < candelaValue)
```
devshgraphicsprogramming, line 182: where do you initialize to 0 , if you don't initialize to 0, its not given that `<` will pass on first try

src/nbl/asset/utils/CIESProfileParser.cpp
```diff
@@ -167,48 +171,37 @@ bool CIESProfileParser::parse(CIESProfile& result)
...
+                result.maxCandelaValue = candelaValue;
+
+            const float thetaRad = core::radians<float>(result.vAngles[j]);
+            const float dThetaRad = core::radians<float>((j < result.vAngles.size() - 1) ? result.vAngles[j + 1] - result.vAngles[j] : result.vAngles[j] - result.vAngles[j - 1]);
+
+            const auto differentialSolidAngle = std::sin(thetaRad) * dThetaRad * dPhiRad;
```
devshgraphicsprogramming, line 188: remember about proper integration of the `sin` as a `cos`

src/nbl/asset/utils/CIESProfileParser.cpp
devshgraphicsprogramming: or `<=` not `<`, you choose

src/nbl/asset/utils/CIESProfile.cpp
devshgraphicsprogramming: >here it will break it imo
see now that you use `asset::encode` we can fix this later XD

So mark as resolved after reading this message.

Because they're all broken :kek:
https://github.com/Devsh-Graphics-Programming/Nabla/blob/f0796aa799b3889ba96f7792c105c933bee0d98a/include/nbl/asset/format/encodePixels.h#L628

Basically if you don't add 0.5, you get this:
```
0.0 -> 0.0 -> 0
0.999... -> 65 534.999934465 -> 65 535
1.0 -> 65 535.0 -> 65 535
```
so 0 gets mapped to too frequently, and 65 535 almost never

This doesn't get fixed with even with a dither of +/-0.5 because anything below 0.5/(2^16-1) will get mapped to 0, 100% of the time

src/nbl/asset/interchange/CIESProfileLoader.cpp
devshgraphicsprogramming: >  another solution would be to create dummy image without regions

preferrable solution

src/nbl/asset/utils/CIESProfile.h
devshgraphicsprogramming: you'll probably never use both. 

If you want to sample only once because you'll flatten, you can load with `ELPF_LOAD_METADATA_ONLY` and `createIESTexture` with correct flatten.

I think sampling again and blending might be faster than decoding from R16_UNORM, because all `hAngles` and `vAngles` fit in cache and `data` is much sparser than your already sampled texture

AnastaZIuk: sure, so let's merge it to `createIESTexture`

src/nbl/asset/utils/CIESProfile.cpp
devshgraphicsprogramming: Yes this is why I want `flatten` to just be a parameter for the `CIESProfile createIESTexture` method and get rid of `flattenIESTexture` method

so we can do all this cache stuff outside, in the mitsuba loader

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: yeah I get what you mean here and why it's needed in our encodes but I'm not sure if you got what I mean at beginning

> about adding 0.5 to the **value**

so let's say I give you max intensity and value is 1.0f, you add 0.5f before multiplication and finally you end up with 1.5*65535 encode which is out of `uint16_t` scope, that's what I mean, it's just the addition would not be to the `value`

AnastaZIuk: this would also mean we map 0.0f -> 0.5f -> uint16_t(65535/2) which definitely should not take place

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: > remember about proper integration of the `sin` as a `cos`

we use [physics convention](https://en.wikipedia.org/wiki/Spherical_coordinate_system#/media/File:3D_Spherical.svg) for spherical coordinates hence differential solid angle is 

$$
{\displaystyle \mathrm {d} \Omega ={\frac {\mathrm {d} S_{r}}{r^{2}}}=\sin\theta*\mathrm {d} \theta *\mathrm {d} \varphi .}
$$

however if we were to swap polar with azimuth (mathematics convention) then we would use the `cos` end up with

$$
{\displaystyle \mathrm {d} \Omega ={\frac {\mathrm {d} S_{r}}{r^{2}}}=\cos\theta*\mathrm {d} \theta *\mathrm {d} \varphi .}
$$

AnastaZIuk: ah shit but 0 degrees is Z- IES so it makes sense to turn `sin` into `cos` as the part of this integration, okay okay

devshgraphicsprogramming: nooo, this is not what I meant

devshgraphicsprogramming: > > remember about proper integration of the `sin` as a `cos`
> 
> we use [physics convention](https://en.wikipedia.org/wiki/Spherical_coordinate_system#/media/File:3D_Spherical.svg) for spherical coordinates hence differential solid angle is
> 
> dΩ=dSrr2=sin⁡θ∗dθ∗dφ.
> 
> however if we were to swap polar with azimuth (mathematics convention) then we would use the `cos` & end up with
> 
> dΩ=dSrr2=cos⁡θ∗dθ∗dφ.

image

This is why it should be cos, not because of anything else.

alos tehta=0 being Z- or Z+ changes nothing, its either sin(tehta) or sin(pi-theta)

src/nbl/asset/utils/CIESProfile.cpp
devshgraphicsprogramming: ah yea I see I wrote "before" mistakenly. my bad

src/nbl/asset/utils/CIESProfileParser.cpp
devshgraphicsprogramming: you're not doing Monte Carlo or some other sampling thing, this is pure finite element stuff

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: `IES_STORAGE_FORMAT maxCandelaValue = {}; in header of the parser

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: resolved in following commit

src/nbl/asset/utils/CIESProfileParser.cpp
AnastaZIuk: resolved in following comment

AnastaZIuk: > ah one more thing, the `candelaValue` you multiply the solid angle with should be an average of `i,j` , `i+1,j`, `i,j+1`, and `i+1,j+1`

for non isotropic cases this interpolation is good, handled also isotropic case

src/nbl/asset/utils/CIESProfile.h
devshgraphicsprogramming: ok going to a later PR

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -104,8 +104,25 @@ inline std::pair<float, float> CIESProfile::sphericalDirToRadians(const core::ve
...
+    asset::IAssetLoader::SAssetLoadContext ctx = { {}, nullptr };
+
+    const asset::IAsset::E_TYPE types[]{ asset::IAsset::ET_IMAGE_VIEW, asset::IAsset::ET_TERMINATING_ZERO };
+    SAssetBundle bundle = aOverride->findCachedAsset(aHash, types, ctx, 0u);
+    {
+        auto contents = bundle.getContents();
+
+        if (!contents.empty() && bundle.getAssetType() == asset::IAsset::ET_IMAGE_VIEW)
+            return core::smart_refctd_ptr_static_cast<asset::ICPUImageView>(*contents.begin());
+    }
```
devshgraphicsprogramming, line 124: no this goes in the callee of `createIESTexture`

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: I decided to merge flatten texture creation into IES one **but** to keep it into profile because IES loader needs this util too, now we have more params to handle cache insertion

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -96,9 +96,10 @@ CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend:
...
             {
-                auto clone = core::smart_refctd_ptr_static_cast<asset::ICPUImageView>(image->clone()); // TODO: check if its already a copy
-                if (meta->profile.flattenIESTexture(clone, _emitter->area.emissionProfile->flatten))
-                    profile.texture = { clone, sampler, 1.f };
+                auto flattenIES = meta->profile.createIESTexture(m_loaderContext->override_, _emitter->area.emissionProfile->filename, _emitter->area.emissionProfile->flatten);
+
+                if (flattenIES.get())
+                    profile.texture = { flattenIES, sampler, 1.f };
                 else
                     os::Printer::log("ERROR: Failed to flatten an IES texture - using the original texture", ELL_ERROR);
```
devshgraphicsprogramming, line 104: if flatten fails you would have failed to create the original texture too, so the fallback needs to be no profile at all

src/nbl/asset/interchange/CIESProfileLoader.cpp
devshgraphicsprogramming: going in future PR

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: ah right

src/nbl/asset/utils/CIESProfileParser.h
devshgraphicsprogramming: this needs doing before merge

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: resolved in following commit

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: corrected in following commit

src/nbl/asset/utils/CIESProfile.cpp
```diff
@@ -200,7 +188,9 @@ core::smart_refctd_ptr<asset::ICPUImageView> CIESProfile::createIESTexture(Execu
     viewParams.subresourceRange.layerCount = 1u;
 
     auto imageView = ICPUImageView::create(std::move(viewParams));
-    bundle = asset::SAssetBundle(nullptr, { core::smart_refctd_ptr(imageView) });
+    const auto aHash = key + "?flatten=" + std::to_string(flatten);
+    asset::IAssetLoader::SAssetLoadContext ctx = { {}, nullptr };
+    SAssetBundle bundle = asset::SAssetBundle(nullptr, { core::smart_refctd_ptr(imageView) });
     aOverride->insertAssetIntoCache(bundle, aHash, ctx, 0u);
```
devshgraphicsprogramming, line 194: also don't insert, caller inserts!

don't want the `asset::IAssetLoader::IAssetLoaderOverride* aOverride, const std::string key` parameters here at all

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -92,19 +92,34 @@ CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend:
...
+            const bool inFlattenDomain = _emitter->area.emissionProfile->flatten >= 0.0 && _emitter->area.emissionProfile->flatten < 1.0; // [0, 1)
 
-            if (inFlattenDomain)
+            if (!inFlattenDomain)
             {
-                auto flattenIES = meta->profile.createIESTexture(m_loaderContext->override_, _emitter->area.emissionProfile->filename, _emitter->area.emissionProfile->flatten);
+                const auto cFlatten = core::min(core::abs(_emitter->area.emissionProfile->flatten), 1.f - std::numeric_limits<float>::epsilon());
+                os::Printer::log("WARNING: Flatten property = " + std::to_string(_emitter->area.emissionProfile->flatten) + " is outside it's [0, 1) domain - the flatten property will be clamped to " + std::to_string(cFlatten), ELL_WARNING);
+                _emitter->area.emissionProfile->flatten = cFlatten;
+            }
```
devshgraphicsprogramming, line 102: just do
```cpp
flatten = abs(flatten);
if (flatten>1.f)
{
   flatten = 1.f;
   error
}
else if (flatten>1.f-epsilon)
   warning
```

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -92,19 +92,34 @@ CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend:
...
-                    profile.texture = { flattenIES, sampler, 1.f };
-                else
-                    os::Printer::log("ERROR: Failed to flatten an IES texture - using the original texture", ELL_ERROR);
+            if (_emitter->area.emissionProfile->flatten > 0.0) // if 0 then it's just "image"
+            {
+                core::smart_refctd_ptr<asset::ICPUImageView> flattenIES = nullptr;
+
+                const auto aHash = _emitter->area.emissionProfile->filename + "?flatten=" + std::to_string(_emitter->area.emissionProfile->flatten);
+                const asset::IAsset::E_TYPE types[]{ asset::IAsset::ET_IMAGE_VIEW, asset::IAsset::ET_TERMINATING_ZERO };
+                asset::SAssetBundle bundle = m_loaderContext->override_->findCachedAsset(aHash, types, m_loaderContext->inner, 0u);
```
devshgraphicsprogramming, line 110: use your "metadata only" loading mode, and always go through this logic

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -92,19 +92,34 @@ CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend:
...
+                {
+                    auto contents = bundle.getContents();
+
+                    if (!contents.empty() && bundle.getAssetType() == asset::IAsset::ET_IMAGE_VIEW)
+                        flattenIES = core::smart_refctd_ptr_static_cast<asset::ICPUImageView>(*contents.begin());
+                }
+
+                if(!flattenIES)
+                    flattenIES = meta->profile.createIESTexture(m_loaderContext->override_, _emitter->area.emissionProfile->filename, _emitter->area.emissionProfile->flatten);
```
devshgraphicsprogramming, line 119: can you fail this?

src/nbl/asset/utils/CIESProfile.cpp
AnastaZIuk: I think we agreed the utility should do it, however resolved in following commit

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk:
```c++
else if (flatten>1.f-epsilon)
   warning
```

why would we allow for `flatten=1.f`?

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: can fail on `setBufferAndRegions` and throw nullptr

AnastaZIuk: corrected in following commit

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: then we don't need this check at all, if it returns any metadata it's the same `meta` we have from `getEmissionProfile`

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: 1.f should be error imo not a warning, we should not let it pass should we?

devshgraphicsprogramming: > ```c++
> else if (flatten>1.f-epsilon)
>    warning
> ```
> 
> why would we allow for `flatten=1.f`?

implied domain flatten mode, with a conical profile, makes it a uniform cone wheras the lack of an IES profile would make it a sphere or hemisphere

devshgraphicsprogramming: need a separate warning and error in next PR

AnastaZIuk: > need a separate warning and error in next PR

https://github.com/Devsh-Graphics-Programming/Nabla/commit/1b5131dbe039f59c7a0011d52a9a2a44757f87ae

src/nbl/asset/utils/CIESProfileParser.h
devshgraphicsprogramming: next PR

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
devshgraphicsprogramming: right now if `flatten==0` the IES profile goes missing !?

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
```diff
@@ -86,27 +87,99 @@ CMitsubaMaterialCompilerFrontend::EmitterNode* CMitsubaMaterialCompilerFrontend:
...
+                    {
+                        asset::IAssetLoader::SAssetLoadContext ctx = { {}, nullptr };
+                        asset::SAssetBundle bundle = asset::SAssetBundle(nullptr, { core::smart_refctd_ptr(flattenIES) });
+                        m_loaderContext->override_->insertAssetIntoCache(bundle, aHash, m_loaderContext->inner, 0u);
+                    }
+                    else
+                    {
+                        os::Printer::log("WARNING: Failed to flatten IES \"" + _emitter->area.emissionProfile->filename + "\" texture, using original one", ELL_WARNING);
+                        flattenIES = core::smart_refctd_ptr(image);
+                    }
```
devshgraphicsprogramming, line 132: else no IES profile at all, because we'll load metadata only on the `cacheEmissionProfile`

src/nbl/ext/MitsubaLoader/CMitsubaLoader.cpp
```diff
@@ -970,15 +970,11 @@ void CMitsubaLoader::cacheEmissionProfile(SContext& ctx, const CElementEmissionP
 		return;
 	
 	const asset::IAsset::E_TYPE types[]{ asset::IAsset::ET_IMAGE_VIEW, asset::IAsset::ET_TERMINATING_ZERO };
-	auto assetLoaded = ctx.override_->findCachedAsset(profile->filename, types, ctx.inner, 0u);
+	auto assetLoaded = interm_getAssetInHierarchy(m_assetMgr, profile->filename, ctx.inner.params, 0u, ctx.override_);
 	if (assetLoaded.getContents().empty())
```
devshgraphicsprogramming, line 974: this `interm_` is where you'd override to load metadata only

AnastaZIuk: resolved

src/nbl/ext/MitsubaLoader/CMitsubaMaterialCompilerFrontend.cpp
AnastaZIuk: resolved

PR #673, opened 2024-03-23
Builtin Resources new constexpr storage data updates

AnastaZIuk:


AnastaZIuk:
there was also an attempt of making per built-in resource data compilations I want to save https://github.com/Devsh-Graphics-Programming/Nabla/commit/d5fd97a9539708cbe2da905f3db2d25c621349a3, but decided to not go into this because it needs us to generate another PCH/reuse & even with that it's just slow and not worth it

include/nbl/core/xxHash256.h
```diff
@@ -132,7 +133,92 @@ inline void XXHash_256(const void* input, size_t len, uint64_t* out)
 	out[3] += v4;
 }
 
+constexpr std::array<uint64_t, 4> XXHash_256(const uint8_t* input, const size_t len)
+{
```
devshgraphicsprogramming, line 137: 1. missing namesapce
2. implement the non-constexpr in terms of this one

src/nbl/builtin/builtinHeaderGen.py
```diff
@@ -1,94 +1,119 @@
...
+        std::tm modified = {{ // Absolute time since 1970 in UTC+0
+            .tm_sec = 6,
+            .tm_min = 9,
+            .tm_hour = 6,
+            .tm_mday = 9,
+            .tm_mon = 6,
+            .tm_year = 9,
+            .tm_isdst = 0
+        }};
+    }};
```
devshgraphicsprogramming, line 83: no reason for every archive namespace to re-define this struct, slap it in `nbl::system` somewhere

include/nbl/core/xxHash256.h
AnastaZIuk: resolved in following commit

src/nbl/builtin/builtinHeaderGen.py
AnastaZIuk: resolved

PR #676, opened 2024-04-13
Shader preprocessor cache

Fletterio:
## Description
Most of the shader cache logic implemented.

## Testing 
Ran examples fine

## TODO list:
Can be merged as-is, but the Cache isn't yet being used


include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -13,15 +13,17 @@
...
 {
 
 class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 {
 	public:
 		//string to be replaced with all "#" except those in "#include"
-		static constexpr const char* PREPROC_DIRECTIVE_DISABLER = "_this_is_a_hash_";
-		static constexpr const char* PREPROC_DIRECTIVE_ENABLER = PREPROC_DIRECTIVE_DISABLER;
+		static constexpr const char* PREPROC_DIRECTIVE_DISABLER = "_this_is_a_hash_"; // TODO: remove/move to GLSL Compiler
+		static constexpr const char* PREPROC_DIRECTIVE_ENABLER = PREPROC_DIRECTIVE_DISABLER; // TODO: remove/move to GLSL Compiler
```
devshgraphicsprogramming, line 26: thats your TODO

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -30,6 +32,7 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 				{
 					system::path absolutePath = {};
 					std::string contents = {};
+					std::optional<std::array<uint64_t, 4>> hash = std::nullopt; // TODO: we're not yet using IFile::getPrecomputedHash(), so for builtins we can maybe use that in the future
```
devshgraphicsprogramming, line 35: since you return `contents` and you actually needed to load it in, then the `hash` shouldn't be optional (cause you had to touch the contents anyway)

It would only be the `IFile` that has an optional hash

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -13,15 +13,17 @@
 #include "nbl/asset/ICPUShader.h"
 #include "nbl/asset/utils/ISPIRVOptimizer.h"
 
+#include "nbl/core/xxHash256.h"
+
+#include <nlohmann/json.hpp>
+using json = nlohmann::json;
```
devshgraphicsprogramming, line 19: thou shalt not use `using` in a public header! ever.

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 		*/
 		struct SCompilerOptions
 		{
+			inline void setCommonData(const SCompilerOptions& opt)
+			{
+				(*this) = opt;
+			}
```
devshgraphicsprogramming, line 167: regular
```cpp
static_cast<SCompilerOptions&>(lhs) = rhs;
```
does the job, unsure why you added this

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+
+			virtual IShader::E_CONTENT_TYPE getCodeContentType() const { return IShader::E_CONTENT_TYPE::ECT_UNKNOWN; };
```
devshgraphicsprogramming, line 169: and now you have a v-table, which I wanted to avoid

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+		private:
+			// we only do lookups based on main file contents + compiler options
+			struct Hash
+			{
+				inline size_t operator()(const SEntry& entry) const noexcept
+				{
+					std::vector<char> hashable = entry.compilerData.getHashable();
+					std::string hashableString(hashable.begin(), hashable.end());
+					hashableString += entry.mainFileContents;
+					return std::hash<std::string>{}(hashableString);
```
devshgraphicsprogramming, line 391: no, you're supposed to precompute the hash and store it in `SEntry` for later retrieval here

because otherwise everytime you lookup, you end up hashing it

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+				
+			};
+
+			
+
+			// json serialization methods for the unordered multiset go here since I need access to Hash and KeyEquals 
+
+			core::unordered_multiset<SEntry, Hash, KeyEqual> m_container;
+			// If provided at creation time, this will hold the json representing entries in m_container + all SPIRV shaders
+			core::smart_refctd_ptr<ICPUBuffer> storageBuffer = nullptr;
```
devshgraphicsprogramming, line 411: after you deserialize from an input buffer, you should drop it and not hold onto it

because `SEntry` should keep all the data it needs

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+			// Specifies the size in bytes of the container json. This is so we know the offset into the buffer where shader bytecode starts
+			uint64_t containerJsonSize = 0;
```
devshgraphicsprogramming, line 413: again, why hold onto it?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
-		virtual core::smart_refctd_ptr<ICPUShader> compileToSPIRV(const char* code, const SCompilerOptions& options) const = 0;
+		inline core::smart_refctd_ptr<ICPUShader> compileToSPIRV(const char* code, const SCompilerOptions& options, core::smart_refctd_ptr<CCache> cache = nullptr) const
+		{
+			if (!code)
+				return nullptr;
+			return compileToSPIRV({code,strlen(code)},options,cache);
+		}
 
-		inline core::smart_refctd_ptr<ICPUShader> compileToSPIRV(system::IFile* sourceFile, const SCompilerOptions& options) const
+		inline core::smart_refctd_ptr<ICPUShader> compileToSPIRV(system::IFile* sourceFile, const SCompilerOptions& options, core::smart_refctd_ptr<CCache> cache = nullptr) const
```
devshgraphicsprogramming, line 440: dont pass arguments to functions as smart pointers that you're not going to own later after the function returns, use a raw (dumb) pointer for the `cache`

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+						return found->value;
+					}	
+				}
+				return nullptr;
+			}
+
+			CCache() {}
+
+			// TODO: add methods as needed, e.g. to serialize and deserialize to/from a pointer
+			std::vector<uint8_t> serialize();
```
devshgraphicsprogramming, line 378: can you somehow work out an upper bound on the serialization size?

If so you could just have a "serialize to a pointer" method

If you can't then return an `ICPUBuffer` with an allocator that hijacks a `core::vector<uint8_t>` or `core::vector<uint64_t>`

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+			static core::smart_refctd_ptr<CCache> deserialize(core::smart_refctd_ptr<ICPUBuffer> serializedCache);
+			static core::smart_refctd_ptr<CCache> deserialize(std::span<uint8_t> serializedCache);
```
devshgraphicsprogramming, line 380: ok so this is basically your factory method, I'd only provide the `span` one

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+				dependency_container_t dependencies;
+				mutable CPUShaderCreationParams shaderParams;
+				bool serialized = false;
+				mutable core::smart_refctd_ptr<asset::ICPUShader> value = nullptr;
+			};
+
+			inline void insert(SEntry&& entry)
+			{
+				m_container.insert(std::move(entry));
 			}
```
devshgraphicsprogramming, line 334: I'd also add a 
```cpp
inline void merge(const CCache* other)
```

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
 
-			virtual IShader::E_CONTENT_TYPE getCodeContentType() const { return IShader::E_CONTENT_TYPE::ECT_UNKNOWN; };
+			// can move to .cpp and have it not inline
```
devshgraphicsprogramming, line 335: good idea to move it

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+			inline core::smart_refctd_ptr<asset::ICPUShader> find(const SEntry& mainFile, CIncludeFinder* finder)
```
devshgraphicsprogramming, line 336: `find` should be `const` method

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+					// Perf note: hashing while preprocessor lexing is likely to be slower than just hashing the whole array like this 
+					inline SPreprocessingDependency(const system::path& _requestingSourceDir, const std::string_view& _identifier, const std::string_view& _contents) :
+						requestingSourceDir(_requestingSourceDir), identifier(_identifier), contents(_contents)
+					{
+						assert(!_contents.empty());
+						const auto reqDirStr = requestingSourceDir.make_preferred().string();
+						std::vector<char> hashable;
+						hashable.insert(hashable.end(), reqDirStr.data()[0], reqDirStr.data()[reqDirStr.size()]);
+						hashable.insert(hashable.end(), identifier.data()[0], identifier.data()[identifier.size()]);
+						hashable.insert(hashable.end(), _contents.data()[0], _contents.data()[_contents.size()]);
```
devshgraphicsprogramming, line 198: what on earth are you doing here?

I'm sure that insert takes position, then a begin and end iterator of the source, you're passing random out of bound chars. Some Rust-isms that don't translate to C++?

P.S. `hashable` should be created with the correct size of `reqDirStr.size()+identifier.size()+_contents.size()` so it does not need to reallocate

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+
+					SPreprocessingDependency(SPreprocessingDependency&) = delete;
+					SPreprocessingDependency& operator=(SPreprocessingDependency&) = delete;
+				
+					inline SPreprocessingDependency(SPreprocessingDependency&&) = default;
+					inline SPreprocessingDependency& operator=(SPreprocessingDependency&&) = default;
+
+					inline bool operator==(const SPreprocessingDependency& other) const
+					{
+						return hash == other.hash && identifier == identifier && contents == contents;
```
devshgraphicsprogramming, line 211: also check for `standardInclude` you don't want to cache match <> with ""

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+					// path or identifier
+					system::path requestingSourceDir = "";
+					std::string identifier = "";
+					// file contents
+					std::string contents = "";
+					// hash of the contents
+					std::array<uint64_t, 4> hash = {};
+					// If true, then `getIncludeStandard` was used to find, otherwise `getIncludeRelative`
+					bool standardInclude = false;
+					nbl::system::IFileBase::time_point_t lastWriteTime = {};
```
devshgraphicsprogramming, line 226: comment out if not currently using the timestamp

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+					inline bool operator==(const SPreprocessingDependency& other) const
+					{
+						return hash == other.hash && identifier == identifier && contents == contents;
+					}
+
+					// Needed for json vector serialization I believe
+					SPreprocessingDependency() {}
+
+					// path or identifier
+					system::path requestingSourceDir = "";
```
devshgraphicsprogramming, line 218: are we using `requestingSourceDir` for something?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+						sortedExtraDefines.assign(extraDefines.begin(), extraDefines.end());
+						// Sort them by identifier so the hash is not order-sensitive!
+						std::sort(sortedExtraDefines.begin(), sortedExtraDefines.end(), [](const SMacroData& lhs, const SMacroData& rhs) {return lhs.identifier < rhs.identifier; });
+						for (const auto& defines : sortedExtraDefines) {
+							hashable.insert(hashable.end(), defines.identifier.data()[0], defines.identifier.data()[defines.identifier.size()]);
+							hashable.insert(hashable.end(), defines.definition.data()[0], defines.definition.data()[defines.definition.size()]);
+						}
+
+						return hashable;
+					}
```
devshgraphicsprogramming, line 260: make this struct final, this method private and `SCompilerData` a friend

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+						auto versionString = std::to_string(static_cast<uint32_t>(targetSpirvVersion));
+						hashable.insert(hashable.end(), versionString.data()[0], versionString.data()[versionString.size()]);
+						auto debugString = std::to_string(static_cast<uint8_t>(debugInfoFlags.value));
+						hashable.insert(hashable.end(), debugString.data()[0], debugString.data()[debugString.size()]);
+						for (auto pass : optimizerPasses) {
+							auto passString = std::to_string(pass);
+							hashable.insert(hashable.end(), passString.data()[0], passString.data()[passString.size()]);
+						}
+						return hashable;
+					}
```
devshgraphicsprogramming, line 290: don't even make this method, you need a precomputed hash, so:
- make the struct `final`
- make the members private
- provide a constructor, this constructor should do what `getHashable` does and store the xxHAsh256 result for later
- check the hash first in the `operator==` so its faster

P.S. Rememver that `stage` can be changed by the preprocessor (a `#pragma`) so may need to be updated after construction before the insertion.

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: tbf if the hasing gets done in `SCompilerData` ctor only, then you don't even need the method

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+					std::vector<ISPIRVOptimizer::E_OPTIMIZER_PASS> optimizerPasses;
+					core::bitflag<E_DEBUG_INFO_FLAGS> debugInfoFlags;
+					SPreprocessorData preprocessorData;
+				};
+
+				// The ordering is important here, the dependencies MUST be added to the array IN THE ORDER THE PREPROCESSOR INCLUDED THEM!
+				using dependency_container_t = std::vector<SPreprocessingDependency>;
+				inline SEntry(std::string_view _mainFileContents, dependency_container_t&& _dependencies, SCompilerData&& compilerData) : mainFileContents(std::move(std::string(_mainFileContents))), dependencies(std::move(_dependencies))
+				{
+				}
```
devshgraphicsprogramming, line 303: precompute the hash for lookup here (no deps)

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+						requestingSourceDir(_requestingSourceDir), identifier(_identifier), contents(_contents)
+					{
+						assert(!_contents.empty());
+						const auto reqDirStr = requestingSourceDir.make_preferred().string();
+						std::vector<char> hashable;
+						hashable.insert(hashable.end(), reqDirStr.data()[0], reqDirStr.data()[reqDirStr.size()]);
+						hashable.insert(hashable.end(), identifier.data()[0], identifier.data()[identifier.size()]);
+						hashable.insert(hashable.end(), _contents.data()[0], _contents.data()[_contents.size()]);
+						// Can't static cast here?
+						hash = nbl::core::XXHash_256((uint8_t*)(hashable.data()), hashable.size() * (sizeof(char) / sizeof(uint8_t)));
```
devshgraphicsprogramming, line 200: `reinterpret_cast<T*>(p)` is preferred to `(T*)(p)`

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+				inline SEntry& operator=(SEntry&&) = default;
+
+				// This next bit is a bit of a Frankenstein. We serialize shader creation parameters into a json, while the actual shader code goes in another file
+				struct CPUShaderCreationParams {
+					IShader::E_SHADER_STAGE stage;
+					IShader::E_CONTENT_TYPE contentType; //I think this one could be skipped since it's always going to be SPIR-V
+					std::string filepathHint;
+					uint64_t codeByteSize = 0;
+					uint64_t offset = 0; // Offset into the serialized .bin for the Cache where code starts
+				};
```
devshgraphicsprogramming, line 317: this is a JSON serialization specific struct, an implementation detail.

Should not be in a public header anywhere.

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+
+				std::string mainFileContents;
+				SCompilerData compilerData;
+				// Keeping this one commented out. Could be useful for lazy-loading of dependencies if it takes up too much memory
+				// uint64_t entryID;
```
devshgraphicsprogramming, line 322: don't want to keep a reference to the cache backing storage after the CCache is deserialized

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+
+				dependency_container_t dependencies;
+				mutable CPUShaderCreationParams shaderParams;
+				bool serialized = false;
+				mutable core::smart_refctd_ptr<asset::ICPUShader> value = nullptr;
```
devshgraphicsprogramming, line 327: what are the first two members doing, and why is `value` mutable?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+				struct CPUShaderCreationParams {
+					IShader::E_SHADER_STAGE stage;
+					IShader::E_CONTENT_TYPE contentType; //I think this one could be skipped since it's always going to be SPIR-V
+					std::string filepathHint;
+					uint64_t codeByteSize = 0;
+					uint64_t offset = 0; // Offset into the serialized .bin for the Cache where code starts
+				};
+
+				std::string mainFileContents;
+				SCompilerData compilerData;
```
devshgraphicsprogramming, line 320: maybe instead of `Data` use `Args` or `Params`

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: btw why the `CIncludeFinder` isn't a const pointer?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -157,32 +161,291 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+						if (!found->value) { // Load shader if not already loaded
+							// If the Cache has no storageBuffer, then it's fresh new (not picked up from serialization): all compiled shaders should be loaded
+							// If codeByteSize is 0, then the Entry found lives in the Cache but not in the serialization, so the Entry was generated in the current runtime:
+							// This means the shader should be loaded
+							assert(storageBuffer && found->shaderParams.codeByteSize != 0);
+							auto code = core::make_smart_refctd_ptr<ICPUBuffer>(found->shaderParams.codeByteSize);
+							memcpy(code->getPointer(), (uint8_t*)storageBuffer->getPointer() + CONTAINER_JSON_SIZE_BYTES + containerJsonSize + found->shaderParams.offset, found->shaderParams.codeByteSize);
+							found->value = core::make_smart_refctd_ptr<ICPUShader>(std::move(code), found->shaderParams.stage, found->shaderParams.contentType, std::move(std::string(found->shaderParams.filepathHint)));
+						}
+						return found->value;
```
devshgraphicsprogramming, line 369: you do realize the only reason you're getting "double storage" is because you are creating a new `ICPUBuffer` with a default allocator for the `found->value` and copying the data across while holding onto the old serialization CPUBuffer

You could simply just create the ICPUBuffers when deserializing and not hold onto the buffer with serialization data.

Or if you truly wanted no RAM-copies, then you can have a special method to deserialize from an `IFile` where you create your own `CMappedFileCPUBuffer : public ICPUBuffer` that holds onto the source memory mapped `IFile` and returns the offsetted file mapped pointer with `getPointer()`

include/nbl/asset/utils/shaderCompiler_serialization.h
```diff
@@ -0,0 +1,200 @@
+#ifndef  _NBL_ASSET_SHADER_COMPILER_SERIALIZATION_H_INCLUDED_
+#define _NBL_ASSET_SHADER_COMPILER_SERIALIZATION_H_INCLUDED_
+
+#include "nbl/asset/utils/IShaderCompiler.h"
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+using SEntry = nbl::asset::IShaderCompiler::CCache::SEntry;
+
+
```
devshgraphicsprogramming, line 10: this can all do into `src` not `include`

Nobody should care about how we serialize our cache.

include/nbl/asset/utils/shaderCompiler_serialization.h
```diff
@@ -0,0 +1,200 @@
...
+            { "standardInclude", dependency.standardInclude },
+            { "lastWriteTimeTicks", ticks },
+        };
+    }
+
+    inline void from_json(const json& j, SEntry::SPreprocessingDependency& dependency)
+    {
+        uint64_t ticks;
+        j.at("lastWriteTimeTicks").get_to(ticks);
+        dependency.lastWriteTime = std::chrono::utc_clock::time_point(std::chrono::utc_clock::duration(ticks));
```
devshgraphicsprogramming, line 130: what's going on here?

include/nbl/system/IFile.h
```diff
@@ -2,6 +2,7 @@
 #define _NBL_SYSTEM_I_FILE_H_INCLUDED_
 
 #include "nbl/system/ISystem.h"
+#include "nbl/asset/ICPUBuffer.h"
```
devshgraphicsprogramming, line 5: circular dependency from `system` to `asset` not allowed, your `IFileBackedCPUBuffer` needs its own header in `nbl/asset`

include/nbl/system/IFile.h
```diff
@@ -107,4 +108,17 @@ class IFile : public IFileBase, private ISystem::IFutureManipulator
 
 }
 
+namespace nbl::asset {
+	class IFileBackedCPUBuffer : public ICPUBuffer {
+
+		IFileBackedCPUBuffer(core::smart_refctd_ptr<nbl::system::IFile> _file)
+			: ICPUBuffer(_file->getSize(), file->getMappedPointer()), file(_file) {}
```
devshgraphicsprogramming, line 115: needs a factory `static core::smart_refctd_ptr<IFileBackedCPUBuffer> create` method, because if the file is not mappable (creation flags don't include it) the creation needs to fail and a nullptr needs to be returned

Also you're forwarding to the contructor wrong, just passing the mapped pointer will still cause a copy of the memory to be allocated

You need to refine the constructor, and provide overloads for `getPointer` (both const and non-const), and a bunch of other methods, like convertToDummy etc.

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -358,29 +360,46 @@ std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADE
     if(context.get_hooks().m_pragmaStage != IShader::ESS_UNKNOWN)
         stage = context.get_hooks().m_pragmaStage;
 
-
-
+    // Context does not have access to the system, so it's up to us to add the lastWriteTime now
```
devshgraphicsprogramming, line 363: you could just pass `m_system.get()` to the `context`'s constructor though, right?

src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -358,29 +360,46 @@ std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADE
...
+                system::ISystem::future_t<core::smart_refctd_ptr<system::IFile>> future;
+                m_system->createFile(future, absPath.c_str(), system::IFile::ECF_READ);
+                if (!future.wait())
+                    return {};
+                future.acquire().move_into(f);
+            }
+            if (!f)
+                return {};
+            (*dependencies)[i].lastWriteTime = f->getLastWriteTime();
+        }
```
devshgraphicsprogramming, line 380: why not make the `getInclude` and friends return the last modified time too ?

devshgraphicsprogramming: also what you wrote here will fail for any "generated" include for which no `IFile` can be opened

src/nbl/asset/utils/ISPIRVOptimizer.cpp
```diff
@@ -92,3 +92,8 @@ nbl::core::smart_refctd_ptr<ICPUBuffer> ISPIRVOptimizer::optimize(const ICPUBuff
 
     return optimize(spirv, count, logger);
 }
+
+const std::initializer_list<ISPIRVOptimizer::E_OPTIMIZER_PASS> nbl::asset::ISPIRVOptimizer::getPasses() const
+{
```
devshgraphicsprogramming, line 97: why not inline, and why `initializer_list` !?

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +214,57 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+    uint64_t dumpedContainerJsonLength = dumpedContainerJson.size();
+    // first 8 entries are the size of the json, stored byte by byte
+    std::vector<uint8_t> retVal;
+    retVal.reserve(CONTAINER_JSON_SIZE_BYTES + dumpedContainerJsonLength + shadersBuffer.size());
+    for (size_t i = 0; i < CONTAINER_JSON_SIZE_BYTES; i++) {
+        uint8_t byte = static_cast<uint8_t>((dumpedContainerJsonLength >> (8 * (CONTAINER_JSON_SIZE_BYTES - i - 1))) & 0xFF);
+        retVal.push_back(byte);
+    }
+    retVal.insert(retVal.end(), std::make_move_iterator(dumpedContainerJson.begin()), std::make_move_iterator(dumpedContainerJson.end()));
+    retVal.insert(retVal.end(), std::make_move_iterator(shadersBuffer.begin()), std::make_move_iterator(shadersBuffer.end()));
```
devshgraphicsprogramming, line 246: rewrite and don't hold onto `storageBuffer`

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +214,57 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+    json containerJson = json::parse(containerJsonString);
+    from_json(containerJson, retVal->m_container);
+    
+    return retVal;
+}
+
+core::smart_refctd_ptr<IShaderCompiler::CCache> IShaderCompiler::CCache::deserialize(std::span<uint8_t> serializedCache)
+{
+    auto cacheBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(serializedCache.size());
+    memcpy(cacheBuffer->getPointer(), serializedCache.data(), serializedCache.size());
```
devshgraphicsprogramming, line 268: ooof, I'd rather not have this method around

include/nbl/system/IFile.h
devshgraphicsprogramming: Also make the `IFileBackedCPUBuffer` constructor take an offset (default 0) and size, so we can create "views" over sub-sections of the file

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: and a `clone` that does a "deep copy" of the cache (with new/default `ICPUBuffer`s)

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: When I deserialize the Cache I don't immediately load all shaders, they're retrieved from the buffer on demand when someone tries to find its matching entry the first time (they will be loaded on subsequent times or if the shader was compiled at runtime and not fetched from the Cache). Changing `found->value` further down in the code required the method to be non-const I believe. 

I could just change it to load all shaders when the Cache is deserialized otherwise

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: Move all this code to the `SCompilerData` hash method then?

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: The first struct I use to retrieve the shader bytecode from the `storageBuffer` and provide paramaters to call the `ICPUShader` constructor. `serialized` can probably be obviated by using value to check whether the shader is loaded or not. `value` was made mutable so that I could load the shader's code only when someone tried to compile it (it tells the compiler that Hash and KeyEquals are not affected by changing this value so the unordered multiset stays the same, I think)

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: I'll change it to load everything at Cache deserialization then

src/nbl/asset/utils/CHLSLCompiler.cpp
Fletterio: I see. I'll move this to the preprocessing then

src/nbl/asset/utils/ISPIRVOptimizer.cpp
Fletterio: The optimizer holds an initializer list iirc. Apparently it's a light wrapper around a pointer + size a la `std::span`, so I just returned that (according to cppref the copy constructor does a shallow copy).

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: If you want on-demand use `IFileBackedCPUBuffer` with offsets

src/nbl/asset/utils/ISPIRVOptimizer.cpp
devshgraphicsprogramming: eh, we need to fix and use a vector... cause initializer_list is only meant for arrays of compile time constants

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: I couldn't work out an upper bound, but I'm not getting what was meant by the latter.

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: The PS poses an issue: if the stage is changed by the preprocessor, then the next time compilation is asked for with the same compiler options there won't be a match (because I'd need a preprocess to find out). I'm leaving it as it is right now, since if the stage is changed by a pragma then the contents of the file will still match

devshgraphicsprogramming: you are right, make a comment about why its that way

(we're hashing serializing compiler inputs.... not results)

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -223,108 +218,155 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+					hashable.insert(hashable.end(), versionString.begin(), versionString.end());
+					hashable.insert(hashable.end(), debugString.begin(), debugString.end());
+					for (auto pass : compilerArgs.optimizerPasses) {
+						hashable.push_back(static_cast<char>(pass));
+					}
+
+					// Now add the mainFileContents and produce both lookup and early equality rejection hashes
+					hashable.insert(hashable.end(), mainFileContents.begin(), mainFileContents.end());
+					std::string hashableString(std::move_iterator(hashable.begin()), std::move_iterator(hashable.end()));
+					lookupHash = std::hash<std::string>{}(hashableString);
```
devshgraphicsprogramming, line 354: you don't need to convert everything to a string and use `std::hash` you were supposed to use the xxx256Hash because its quite a bit faster and can be applied over any byte array (doesn't care about `\0` string terminators)

devshgraphicsprogramming: Also `std::hash` is size_t which is 64bit, but xx256Hash is 256 bit so less chance of collision

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -397,47 +426,40 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
 		{
-			/*core::vector<CCache::SEntry::SPreprocessingDependency> dependencies;
-			if (options.cache)
-			{
-				const auto& mainDep = dependencies.emplace_back(code,options.hash());
-				auto found = options.cache->find(mainDep);
+			CCache::SEntry entry;
+			std::vector<CCache::SEntry::SPreprocessingDependency> dependencies;
+			if (cache) {
+				entry = std::move(CCache::SEntry(code, options));
```
devshgraphicsprogramming, line 442: a bit of C++ knowledge, when you newly create an object its already an `r-value` reference ergo does not need `std::move`

include/nbl/asset/utils/IShaderCompiler.h
AnastaZIuk: [xxHash256](https://github.com/Devsh-Graphics-Programming/Nabla/blob/44c706dea466a546d6abaea3cede0dba4c52c924/include/nbl/core/xxHash256.h#L150)

devshgraphicsprogramming: the 256bit hash could then be passed through std::hash to obtain a 64bit value for the functor that `unordered_map` needs

AnastaZIuk: [xxHash256](https://github.com/Devsh-Graphics-Programming/Nabla/blob/44c706dea466a546d6abaea3cede0dba4c52c924/include/nbl/core/xxHash256.h#L150)

include/nbl/system/IFile.h
devshgraphicsprogramming: looking at your 2 commits, it seems you forgot to add,commit and push some new files?

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -106,14 +106,15 @@ IShaderCompiler::CIncludeFinder::CIncludeFinder(core::smart_refctd_ptr<system::I
...
     IShaderCompiler::IIncludeLoader::found_t retVal;
-    if (auto contents = tryIncludeGenerators(includeName)) 
+    if (auto contents = tryIncludeGenerators(includeName))
         retVal = std::move(contents);
-    else if (auto contents = trySearchPaths(includeName)) 
-        retVal = std::move(contents);
-    else retVal = std::move(m_defaultFileSystemLoader->getInclude(requestingSourceDir.string(),includeName));
+    else if (auto contents = trySearchPaths(includeName))
+            retVal = std::move(contents);
+    else retVal = std::move(m_defaultFileSystemLoader->getInclude(requestingSourceDir.string(), includeName));
```
devshgraphicsprogramming, line 117: unless a function returns a `T&` (l-value reference) an `std::move` applied on a return value does nothing

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -247,24 +259,36 @@ std::vector<uint8_t> IShaderCompiler::CCache::serialize()
...
-    uint64_t* bufferStart = (uint64_t*)retVal->storageBuffer->getPointer();
-    retVal->containerJsonSize = bufferStart[0];
-    uint8_t* containerJsonStart = (uint8_t*)(bufferStart + 1);
-    std::string containerJsonString(containerJsonStart, containerJsonStart + retVal->containerJsonSize);
+
+    // First get the size of the json in the buffer, stored in the first 8 bytes
+    uint64_t* cacheStart = reinterpret_cast<uint64_t*>(serializedCache.data());
+    uint64_t containerJsonSize = cacheStart[0];
+    // Next up get the json that stores the container data
+    std::string containerJsonString(serializedCache.begin() + CONTAINER_JSON_SIZE_BYTES, serializedCache.begin() + CONTAINER_JSON_SIZE_BYTES + containerJsonSize);
```
devshgraphicsprogramming, line 270: btw learn about `std::string_view` it prevents copies, just casts any memory range to act as a string

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: are you going to make a `deserialize` from an `IFile` or you're leaving it as a TODO for someone else?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -397,47 +426,40 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+				auto found = cache->find(entry, options.preprocessorOptions.includeFinder);
 				if (found)
-					return core::smart_refctd_ptr<ICPUShader>(found);
+					return found;
+			}
+			auto retVal = compileToSPIRV_impl(code, options, cache ? &dependencies : nullptr);
+			if (cache) {
+				entry.dependencies = std::move(dependencies);
+				entry.value = retVal;
```
devshgraphicsprogramming, line 450: forgot to insert into the cache?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -192,25 +193,19 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
-				
+					inline SPreprocessingDependency(SPreprocessingDependency&) = default;
+					inline SPreprocessingDependency& operator=(SPreprocessingDependency&) = default;
 					inline SPreprocessingDependency(SPreprocessingDependency&&) = default;
 					inline SPreprocessingDependency& operator=(SPreprocessingDependency&&) = default;
 
-					inline bool operator==(const SPreprocessingDependency& other) const
-					{
-						return hash == other.hash && identifier == identifier && contents == contents;
-					}
```
devshgraphicsprogramming: why did you get rid of the explicit equality operator?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -223,108 +218,155 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+				struct SCompilerArgs; // Forward declaration for SPreprocessorArgs's friend declaration
+
+				struct SPreprocessorArgs final {
 
-					inline bool operator==(const SPreprocessorData& other) const {
+					friend class SCompilerArgs;
+					friend class SEntry;
+
+					// Public default needed for json serialization
+					SPreprocessorArgs() {};
```
devshgraphicsprogramming, line 237: you can make the json serializaiton functions friends

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -223,108 +218,155 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 					std::array<uint64_t, 4> hash = {};
```
devshgraphicsprogramming, line 218: if there are fields that need to be properly initialized, then only the constructor should be public

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -223,108 +218,155 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+
 						if (extraDefines.size() != other.extraDefines.size()) return false;
-						for (auto definesIt = extraDefines.begin(), otherDefinesIt = other.extraDefines.begin(); definesIt != extraDefines.end(); definesIt++, otherDefinesIt++) {
+						core::vector<SMacroData> sortedExtraDefines, otherSortedExtraDefines;
+
+						sortedExtraDefines.assign(extraDefines.begin(), extraDefines.end());
+						std::sort(sortedExtraDefines.begin(), sortedExtraDefines.end(), [](const SMacroData& lhs, const SMacroData& rhs) {return lhs.identifier < rhs.identifier; });
+						
+						otherSortedExtraDefines.assign(other.extraDefines.begin(), other.extraDefines.end());
+						std::sort(otherSortedExtraDefines.begin(), otherSortedExtraDefines.end(), [](const SMacroData& lhs, const SMacroData& rhs) {return lhs.identifier < rhs.identifier; });
```
devshgraphicsprogramming, line 249: this is painful (memory allocation for a simple op)

You should instead sort the defines in the constructor  so this is amortized.

You will also be able to sort them in-place on the `extraDefines` vector instead of allocating new temporary memory.

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -223,108 +218,155 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+				struct SCompilerArgs final {
 
-					inline bool operator==(const SCompilerData& other) const {
-						if (stage != other.stage || targetSpirvVersion != other.targetSpirvVersion || debugInfoFlags != other.debugInfoFlags || preprocessorData != other.preprocessorData) return false;
-						if (optimizerPasses.size() != other.optimizerPasses.size()) return false;
+					friend class SEntry;
+					friend void to_json(nlohmann::json&, const SCompilerArgs&);
+					friend void from_json(const nlohmann::json&, SCompilerArgs&);
+					// Public default needed for json serialization
+					SCompilerArgs() {}
```
devshgraphicsprogramming, line 279: which method exactly?

Fletterio: The unordered multiset gets stored by the json library as an array (all iterables get stored as a json array). To be able to retrieve a struct from an array the library seems to require a default constructor. `SEntry` is actually the one that needs to have a (public) default constructor, but that requires SCompilerArgs to have one as well. I could alternatively make only `SEntry` have default public constructor and make this one be private, since `SEntry` is friend

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: This seems reasonable

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: std::hash does not take an array, but boost::hash does. Do we have that available?

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: It was never used - its contents are only compared against a `found_t` struct when the `find`method is called from within the Cache when trying to see if a matching entry exists

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: As in, deserialize the whole Cache from an IFile at once, or the one that creates some sort of mapped file buffer and retrieves shaders on demand? I could do the first soon, the other I left for later for now. Do I add a commented out TODO with that method?

src/nbl/CMakeLists.txt
```diff
@@ -469,6 +469,18 @@ target_include_directories(Nabla PUBLIC
...
+
+# nlohmann_json
+set(JSON_BuildTests OFF CACHE INTERNAL "")
+add_dependencies(Nabla nlohmann_json)
+if(NBL_STATIC_BUILD)
+	target_link_libraries(Nabla INTERFACE nlohmann_json::nlohmann_json)
+else()
+	target_link_libraries(Nabla PRIVATE nlohmann_json::nlohmann_json)
+endif()
+target_include_directories(Nabla PUBLIC ${THIRD_PARTY_SOURCE_DIR}/nlohmann_json/include)
```
AnastaZIuk, line 481: make it

```cmake
target_include_directories(Nabla PUBLIC $<TARGET_PROPERTY:nlohmann_json::nlohmann_json,INCLUDE_DIRECTORIES>)
```
 this way if a target uses a set of include directories which are auto-generated or depend on the configuration you will catch them all properly

AnastaZIuk:
please use our fork version of [the json library](https://github.com/Devsh-Graphics-Programming/json) and not the [original remote](https://github.com/nlohmann/json)

src/nbl/CMakeLists.txt
```diff
@@ -469,6 +469,18 @@ target_include_directories(Nabla PUBLIC
 	"${THIRD_PARTY_SOURCE_DIR}/dxc/dxc/external/SPIRV-Tools/include"
 	"${THIRD_PARTY_SOURCE_DIR}/dxc/dxc/external/SPIRV-Headers/include"
 )
+
+# nlohmann_json
+set(JSON_BuildTests OFF CACHE INTERNAL "")
```
AnastaZIuk, line 474: move it [here](https://github.com/Devsh-Graphics-Programming/Nabla/pull/676/files#diff-e286d8c6322ab030a049019a13e4c648f1a84a7e0f51ca0fb82e04e6f902065aR7) and set as ordinary variable, no need to use cache

Fletterio: move what where, sorry

src/nbl/CMakeLists.txt
Fletterio: I'm not really understanding this setting. If I do this it won't find the headers

src/nbl/CMakeLists.txt
AnastaZIuk:
```cmake
set(JSON_BuildTests OFF)
```

[here](https://github.com/Devsh-Graphics-Programming/Nabla/pull/676/files#diff-e286d8c6322ab030a049019a13e4c648f1a84a7e0f51ca0fb82e04e6f902065aR7) before `add_subdirectory` invocation

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: do I add default move, deleted copy contructors?

Fletterio: What do I do with constructors/assignment? Deleted copy, default move?

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: So I'll extend ICPUBuffer with some custom CVectorBuffer or something similar, that holds its own vector. What methods do I need to override in this case?

The same question goes for the CFileBackedCPUBuffer

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: do you use the implicit `==` at all ?

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: Only the methods which are virtual can be overriden so
```cpp
template<typename T, typename Allocator>
class CVectorCPUBuffer final : public ICPUBuffer
{
   public:
       inline CVectorCPUBuffer(core::vector<T,Allocator>&& _vec) : m_vec(std::move(_vec))
       {
           data = m_vec.data();
       }
   
   protected:
       virtual inline ~CVectorCPUBuffer()
       {
           freeData();
       }
       inline void freeData() override
       {
           m_vec.clear();
           ICPUBuffer::data = nullptr;
       }
   
       core::vector<T,Allocator> m_vec;
};
```

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: things that inherit from `IReferenceCounted` always live on the heap, therefore they have no copy or move assignment nor constructors

Furthermore if their creation could fail, they need a non-public constructor and all the potentially failing parts of creation hoisted out to either a `static core::smart_refctd_ptr<this_t> create` factory method or some other facotry class method

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: I think not - should I delete it?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -353,8 +354,14 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
-				inline SEntry(SEntry&) = default;
-				inline SEntry& operator=(SEntry&) = default;
+				// Making the copy constructor deep-copy
+				inline SEntry(const SEntry& other) 
+					: mainFileContents(other.mainFileContents), compilerArgs(other.compilerArgs), hash(other.hash), lookupHash(other.lookupHash), 
+					  dependencies(other.dependencies), value(core::smart_refctd_ptr_static_cast<ICPUShader, IAsset>(other.value->clone())) {}
+				
+				inline SEntry& operator=(const SEntry& other) {
+					*this = SEntry(other);
+				}
```
devshgraphicsprogramming, line 364: bad bad bad idea to have deep copy on everything.

You shouldn't deep copy the `shader` its a shared pointer, its okay to share a reference
P.S. `value` could actually be `core::smart_refctd_ptr<const ICPUShader>` cause its readonly

if you truly need then make

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -375,14 +382,16 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 			// Alternatively, adding the time an SEntry entered the cache could also serve this purpose
 			inline void merge(const CCache* other) {
 				for (auto& entry : other->m_container) {
-					m_container.insert(entry);
+					SEntry entryCopy(entry);
+					m_container.insert(entryCopy);
```
devshgraphicsprogramming, line 386: you end up with 2 copies here, better approach would be to `std::move(entryCopy)`, best would be to use `emplace(entry)` which forwards the arguments to a constructor and does a zero copy initialization

devshgraphicsprogramming: same thing happens later on

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: > Do I add a commented out TODO with that method?

yes
```cpp
/*
// `serializedCache` must be mapped with at least a read flag
static core::smart_refctr_ptr<CCache> deserialize(const system::IFile* serializedCache);
*/
```

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +219,116 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+
+        retVal->insert(std::move(entries[i]));
+    }
+
+    return retVal;
+}
+
+core::smart_refctd_ptr<IShaderCompiler::CCache> IShaderCompiler::CCache::deserialize(core::smart_refctd_ptr<const system::IFile> serializedCache) {
+    const void* fileBufferPointer = serializedCache->getMappedPointer();
+    assert(fileBufferPointer); // Check the file is readable
```
devshgraphicsprogramming, line 329: just return `nullptr`, asserts make applications crash, they're for abnormal errors/assumptions

Also check the mapping flags for reading

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +219,116 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+    
+    size_t fileSize = serializedCache->getSize();
+    std::span cacheSpan{ reinterpret_cast<const uint8_t*>(fileBufferPointer), fileSize };
+    return IShaderCompiler::CCache::deserialize(std::span{ reinterpret_cast<const uint8_t*>(fileBufferPointer), fileSize });
```
devshgraphicsprogramming, line 333: now you want a comment (if you want to leave as TODO) that this deserialization should work differently, basically that the `ICPUBuffer` of the spir-v code should be created differently

include/nbl/asset/utils/ISPIRVOptimizer.h
```diff
@@ -42,6 +42,7 @@ class ISPIRVOptimizer final : public core::IReferenceCounted
 
         core::smart_refctd_ptr<ICPUBuffer> optimize(const uint32_t* _spirv, uint32_t _dwordCount, system::logger_opt_ptr logger) const;
         core::smart_refctd_ptr<ICPUBuffer> optimize(const ICPUBuffer* _spirv, system::logger_opt_ptr logger) const;
+        const std::initializer_list<E_OPTIMIZER_PASS> getPasses() const;
```
devshgraphicsprogramming, line 45: you still need to change the container for `m_passes`

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -13,15 +13,16 @@
 #include "nbl/asset/ICPUShader.h"
 #include "nbl/asset/utils/ISPIRVOptimizer.h"
 
+#include "nbl/core/xxHash256.h"
+
+#include <nlohmann/json.hpp>
```
devshgraphicsprogramming, line 18: do you need to leak this to public headers?

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +219,116 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+        // Now create the CPU shader creation parameters struct
+        CPUShaderCreationParams params;
+        params.stage = entry.value->getStage();
+        params.contentType = entry.value->getContentType();
+        params.filepathHint = entry.value->getFilepathHint();
+        params.offset = shadersBuffer.size();
+        params.codeByteSize = entry.value->getContent()->getSize();
+
+        // And add it to the shader creation parameters array
+        shaderCreationParams.push_back(params);
```
devshgraphicsprogramming, line 271: use `emplace_back` whenever you can, and if you can't `std::move` this saves strings/paths being copied and reallocated

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +219,116 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+
+        // Finally, insert the shader bytecode into the shaders buffer
+        shadersBuffer.reserve(shadersBuffer.size() + params.codeByteSize);
+        shadersBuffer.insert(shadersBuffer.end(), (uint8_t*)entry.value->getContent()->getPointer(), (uint8_t*)entry.value->getContent()->getPointer() + params.codeByteSize);
+    }
+    json containerJson{
+        { "entries", entries },
+        { "shaderCreationParams", shaderCreationParams },
```
devshgraphicsprogramming, line 279: `std::move` probably won't harm you here

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +219,116 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+        params.contentType = entry.value->getContentType();
+        params.filepathHint = entry.value->getFilepathHint();
+        params.offset = shadersBuffer.size();
+        params.codeByteSize = entry.value->getContent()->getSize();
+
+        // And add it to the shader creation parameters array
+        shaderCreationParams.push_back(params);
+
+        // Finally, insert the shader bytecode into the shaders buffer
+        shadersBuffer.reserve(shadersBuffer.size() + params.codeByteSize);
```
devshgraphicsprogramming, line 274: you could do a quick loop and work out the correct reservation size so you don't run the risk of multiple reallocations here

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +219,116 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+        shadersBuffer.insert(shadersBuffer.end(), (uint8_t*)entry.value->getContent()->getPointer(), (uint8_t*)entry.value->getContent()->getPointer() + params.codeByteSize);
```
devshgraphicsprogramming, line 275: IMHO a `resize`+`memcpy` instead of a `reserve`+`insert` may be faster here

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -289,3 +219,116 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
...
+    }
+    json containerJson{
+        { "entries", entries },
+        { "shaderCreationParams", shaderCreationParams },
+    };
+    std::string dumpedContainerJson = std::move(containerJson.dump());
+    uint64_t dumpedContainerJsonLength = dumpedContainerJson.size();
+    core::vector<uint8_t> retVal;
```
devshgraphicsprogramming, line 283: if you saved the shader data size and rawarray first, you could only have one `vector` and two `resize` calls

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: yes for safety/performance's sake

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: you can forward declare `nlohmann::json` instead btw just a
```cpp
namespace nlohmann
{
class json;
}
```

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: This was to be used in the cache clone. I could alternatively call this `SEntry::clone()` instead of making it the copy constructor. 

When cloning then I can just +1 the refctd shader then, right? No need to deep copy that one

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: Right, I thought I had put a move here, maybe I lost it in the refactor. I'll try the best approach or eventually the netter one if I can't make the best one work easily

src/nbl/asset/utils/IShaderCompiler.cpp
Fletterio: Oh no this one does the same as the other one - it directly loads the whole cache at once from a File. The Cache that's streamed I'm planning to make it a different class, because some methods (such as the merging or the finding) need to be different. I might make those virtual, have the `CStreamedCache` extend `CCache` and override those methods

include/nbl/asset/utils/IShaderCompiler.h
Fletterio: What about the hashing? Do I move it to the .cpp?

src/nbl/CMakeLists.txt
AnastaZIuk: also change scope from `PUBLIC` to `PRIVATE` in `target_include_directories`

include/nbl/asset/utils/IShaderCompiler.h
AnastaZIuk: > What about the hashing? Do I move it to the .cpp?

yes, also while you are on it please do also [this](https://github.com/Devsh-Graphics-Programming/Nabla/pull/676/files#r1580487872)

src/nbl/CMakeLists.txt
AnastaZIuk: > I'm not really understanding this setting

this is a [generator expression](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#manual:cmake-generator-expressions(7)) which gets [`INCLUDE_DIRECTORIES`](https://cmake.org/cmake/help/latest/prop_tgt/INCLUDE_DIRECTORIES.html#prop_tgt:INCLUDE_DIRECTORIES) target property storing `nlohmann_json::nlohmann_json`'s target set of include search directories, the expression is evaluated during build system generation & can handle configuration specific defined paths (it may have paths which are different depending on a configuration). Doing so is handy because it's a target responsibility to define the includes not ours

include/nbl/asset/utils/IShaderCompiler.h
AnastaZIuk: one more thing

```c++
#include "nlohmann/json.hpp"
```

not

```c++
#include <nlohmann/json.hpp>
```

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: Either have a `clone()` method or a non-public copy ctor, because you can accidentally end up with lots of redundant copies because of how C++ syntax works https://github.com/Devsh-Graphics-Programming/Nabla/pull/676#discussion_r1580122994

You don't +1 manually, every time you create a `smart_refctd_ptr` the refcount increases, works exactly like `std::shared_ptr` for a class `T` that inherits from `std::enable_shared_from_this<T>`

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: yeah if you can't figure it out in 30 minutes, do the second best thing

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: > > What about the hashing? Do I move it to the .cpp?
> 
> yes, also while you are on it please do also [this](https://github.com/Devsh-Graphics-Programming/Nabla/pull/676/files#r1580487872)

no stays where it is, was supposed to be a public header anyway.

AnastaZIuk: > > > What about the hashing? Do I move it to the .cpp?
> > 
> > 
> > yes, also while you are on it please do also [this](https://github.com/Devsh-Graphics-Programming/Nabla/pull/676/files#r1580487872)
> 
> no stays where it is, was supposed to be a public header anyway.

I meant to move `#include` part to .cpp to not expose headers, sry

include/nbl/asset/utils/ISPIRVOptimizer.h
devshgraphicsprogramming: yes but then remember to return a `span<const E_OPTIMIZER_PASS>` for this method

src/nbl/asset/utils/IShaderCompiler.cpp
devshgraphicsprogramming: ok, `CStreamedCache` would be a whole new task, which we don't have time for. So open a GH issue and we'll do it at some point later.

Thinking about it, we don't need a `deserialize` method from an `IFile` here.

To do such a thing "properly" we'd need an LZMA (7z) CArchive where:
- each shader is a virtual IFile thats LZMA compressed
- there's a JSON header IFile thats also LZMA compressed
- you decompress from LZMA on the fly when requesting a file

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: write this up in a GH issue you'll open about a streaming cache

include/nbl/system/IFile.h
devshgraphicsprogramming: write this up in a GH issue you'll open about a streaming cache

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -283,8 +284,8 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 						return retVal;
 					}
 
-					inline SCompilerArgs(SCompilerArgs&) = default;
-					inline SCompilerArgs& operator=(SCompilerArgs&) = default;
+					inline SCompilerArgs(const SCompilerArgs&) = default;
+					inline SCompilerArgs& operator=(const SCompilerArgs&) = default;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: done?

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: still outstanding

src/nbl/asset/utils/IShaderCompiler.cpp
devshgraphicsprogramming: write this up in a GH issue you'll open about a streaming cache

src/nbl/CMakeLists.txt
devshgraphicsprogramming: @AnastaZIuk review if fixed

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: @AnastaZIuk review if fixed

devshgraphicsprogramming:
Ok this looks ready to merge, first merge `master` to this branch and its associated Examples branch, then tell me when to merge.

src/nbl/CMakeLists.txt
AnastaZIuk: will correct myself later, resolving

include/nbl/asset/utils/IShaderCompiler.h
AnastaZIuk: > one more thing
> 
> ```c++
> #include "nlohmann/json.hpp"
> ```
> 
> not
> 
> ```c++
> #include <nlohmann/json.hpp>
> ```

still needs fixing

Fletterio: If I forward declare `nlohmann::json` I get `nlohmann::json is ambiguous` compilation errors, even if I don't include the `json.h` here.  As for the hashing, since it's constexpr there is no forwarding, so I'm moving anything that hashes out of the .h and into the .cpp

Fletterio:
I think it's ready to be merged

devshgraphicsprogramming:
> I think it's ready to be merged

please write up the GH issue about streaming cache

include/nbl/asset/utils/IShaderCompiler.h
devshgraphicsprogramming: btw the reason it was ambiguous is because `nlohmann::json` was a `using` alias, and not a struct or class, fixing by myself.

PR #677, opened 2024-04-17
New spirv introspection

Przemog1:


include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -13,172 +17,657 @@
 #include "nbl/asset/ICPUImageView.h"
 #include "nbl/asset/ICPUComputePipeline.h"
 #include "nbl/asset/ICPURenderpassIndependentPipeline.h"
-#include "nbl/asset/utils/ShaderRes.h"
+//#include "nbl/asset/utils/ShaderRes.h"
```
devshgraphicsprogramming, line 20: does that file even exist anymore/is useful for anything?

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -13,172 +17,657 @@
...
-// podzielic CIntrospectionData na dwie klasy
-// jedna bez inputOutput i bez push constant blocka `CIntrospectionData`
-// druga dziedziczy z pierwszej i dodaje te 2 rzeczy `CStageIntrospectionData`
+// TODO: put this somewhere useful in a separate header
+namespace nbl::core
+{
+
+template<typename T>
+struct based_offset
+{
```
devshgraphicsprogramming, line 39: `nbl/core/based_offset.h` ?

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -13,172 +17,657 @@
...
+					{
+						uint32_t index;
+					};
+
+					inline bool isArray() const {return !count.empty();}
+					inline bool isRunTimeSized() const {return isArray() ? count[0].value == 0 : false;}
+
+					//! Note: for SSBOs and UBOs it's the block name
+					span_t<char,Mutable> name = {};
+					span_t<SArrayInfo,Mutable> count = {};
```
devshgraphicsprogramming, line 389: maybe we'll need to use different type than `SArrayInfo` if there's an ifff relationship between `OpRuntimeArray` and `VARIABLE_DESCRIPTOR_COUNT_BIT`
https://discord.com/channels/427551838099996672/427951526967902218/1230618569931227146

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -13,172 +17,657 @@
...
+
+				//! TODO: Add getters for all the other members!
+				inline const auto& getDescriptorSetInfo(const uint8_t set) const {return m_descriptorSetBindings[set];}
+				inline const auto& getInputs() const { return m_input; }
+				inline const core::vector<SFragmentOutputInterface>& getFragmentShaderOutputs() const
+				{
+					if (m_shaderStage != IShader::ESS_FRAGMENT)
+					{
+						// TODO: log error
+						return {};
```
devshgraphicsprogramming, line 438: you don't want to fail/bail from a method/function that returns a reference, by returning a reference to a temporary

you either return a pointer, or if the function happens to return a `const&` like here, you return a reference to a `const static`

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -13,172 +17,657 @@
...
+		};
+		class CPipelineIntrospectionData final : public CIntrospectionData
 		{
-			std::string entryPoint;
-			core::smart_refctd_ptr<const ICPUShader> cpuShader;
+			public:
+				struct SDescriptorInfo final : CIntrospectionData::SDescriptorInfo
+				{
+					inline bool isArray() const {return stride;}
+					inline bool isRuntimeSized() const {return isArray() && count==0;}
```
devshgraphicsprogramming, line 564: if there's no iff relationship https://discord.com/channels/427551838099996672/427951526967902218/1230618569931227146 then we need to track `isruntimeSized` separately from `count` as `count` would just become a lower bound on the size of the array then

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -188,43 +677,56 @@ class NBL_API2 CSPIRVIntrospector : public core::Uncopyable
...
-		void shaderMemBlockIntrospection(spirv_cross::Compiler& _comp, impl::SShaderMemoryBlock& _res, uint32_t _blockBaseTypeID, uint32_t _varID, const mapId2SpecConst_t& _sortedId2sconst) const;
-		size_t calcBytesizeforType(spirv_cross::Compiler& _comp, const spirv_cross::SPIRType& _type) const;
-
+#endif	
 	private:
+		core::smart_refctd_ptr<const CStageIntrospectionData> doIntrospection(const CStageIntrospectionData::SParams& params);
+		size_t calcBytesizeForType(spirv_cross::Compiler& comp, const spirv_cross::SPIRType& type) const;
+		// TODO: hash map instead
+		using OutputVecT = core::vector<CSPIRVIntrospector::CStageIntrospectionData::SOutputInterface>;
+		using FragmentOutputVecT = core::vector<CSPIRVIntrospector::CStageIntrospectionData::SFragmentOutputInterface>;
```
devshgraphicsprogramming, line 686: outstanding I guess?

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+        return CSPIRVIntrospector::CStageIntrospectionData::VAR_TYPE::UNKNOWN_OR_STRUCT;
+    }
 }
 
-bool CSPIRVIntrospector::introspectAllShaders(core::smart_refctd_ptr<const CIntrospectionData>* outIntrospections, const std::span<const ICPUShader::SSpecInfo> infos)
+// returns true if successfully added all the info to self, false if incompatible with what's already in our pipeline or incomplete (e.g. missing spec constants)
+NBL_API2 bool CSPIRVIntrospector::CPipelineIntrospectionData::merge(const CSPIRVIntrospector::CStageIntrospectionData* stageData, const ICPUShader::SSpecInfoBase::spec_constant_map_t* specConstants)
 {
-    auto it = outIntrospections;
-    for (auto info : infos)
```
devshgraphicsprogramming, line 92: you should protect against `!stageData`

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+
+    for (const auto& binding : m_descriptorSetBindings[setID])
+    {
+        auto& outBinding = outBindings.emplace_back();
+
+        outBinding.binding = binding.binding;
+        outBinding.count = binding.count;
+        outBinding.type = binding.type;
+        outBinding.stageFlags = binding.stageMask;
+        outBinding.createFlags = ICPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_NONE;
```
devshgraphicsprogramming, line 296: if its a runtime sized binding , you need the variable count flag here.

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+        // TODO: outBinding.samplers
```
devshgraphicsprogramming, line 297: you cannot use immutable samplers, because HLSL doesn't give you any info about what such a sampler should be.

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
-        {EF_R64_SFLOAT, EF_R64G64_SFLOAT, EF_R64G64B64_SFLOAT, EF_R64G64B64A64_SFLOAT},
-        {EF_R32_SFLOAT, EF_R32G32_SFLOAT, EF_R32G32B32_SFLOAT, EF_R32G32B32A32_SFLOAT}
+    const uint32_t descSet = comp.get_decoration(r.id,spv::DecorationDescriptorSet);
+    if (descSet> DESCRIPTOR_SET_COUNT)
+        return nullptr; // TODO: log fail
+
+    auto a = comp.get_name(r.id);
+    const spirv_cross::SPIRType& type = comp.get_type(r.type_id);
+    const auto arrDim = type.array.size();
+    // assuming only 1D arrays because i don't know how desc set layout binding is constructed when it's let's say 2D array (e.g. uniform sampler2D smplr[4][5]; is it even legal?)
```
devshgraphicsprogramming, line 340: definitely nnot legal in HLSL2021, just checked with all resource types except UBO, Input and Output

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
 
-                //all formats returned by glslType2E_FORMAT() already are multiple-of-4 bytes so no need to pad
-                reloffset += getTexelOrBlockBytesize(static_cast<E_FORMAT>(attr.format));
+        member.rowMajor = _comp.get_member_decoration(_parentType.self, m, spv::DecorationRowMajor); // ?
 
-                vtxInput.enabledAttribFlags |= (1u << io.location);
-            }
+            //if (member.mtxColCnt > 1u)
+            //    member.mtxStride = _comp.type_struct_member_matrix_stride(_parentType, m);
+#endif
```
devshgraphicsprogramming, line 424: where do you handle the matrix info?

Przemog1: nowhere yet

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+    // assuming only 1D arrays because i don't know how desc set layout binding is constructed when it's let's say 2D array (e.g. uniform sampler2D smplr[4][5]; is it even legal?)
+    if (arrDim>1)
+        return nullptr; // TODO: log fail
+
+
+    CStageIntrospectionData::SDescriptorVarInfo<true> res = {
+        {
+            .binding = comp.get_decoration(r.id,spv::DecorationBinding),
+            .type = restype//,
+            //.descCount = 1
```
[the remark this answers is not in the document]
Przemog1: thats mistake, thanks for pointing that out

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
 
-template<E_SHADER_RESOURCE_TYPE restype>
-static std::pair<bool, IImageView<ICPUImage>::E_TYPE> imageInfoFromResource(const SShaderResource<restype>& _res)
-{
-    return {_res.shadow, _res.viewType};
-}
+                if (arrInfo.isSpecConstant)
+                {
+                    if (!specConstants)
+                        return 0u;
```
devshgraphicsprogramming, line 128: you need to document that this lambda returning 0 indicates failure

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -13,172 +17,657 @@
...
+				NBL_API2 core::smart_refctd_ptr<ICPUDescriptorSetLayout> createApproximateDescriptorSetLayoutFromIntrospection(const uint32_t setID);
+				NBL_API2 core::smart_refctd_ptr<ICPUPipelineLayout> createApproximatePipelineLayoutFromIntrospection(core::smart_refctd_ptr<const CStageIntrospectionData>& introspection);
+
+			protected:
+				// ESS_UNKNOWN on a byte means its not declared in any shader merged so far
+				std::array<core::bitflag<ICPUShader::E_SHADER_STAGE>,MaxPushConstantsSize> m_pushConstantBytes;
+				//
+				struct Hash
+				{
+					inline size_t operator()(const SDescriptorInfo& item) const {return item.binding;}
```
devshgraphicsprogramming, line 592: you need to put `item.binding` through `std::hash`, its necessary for the hashmap to throw the buckets around nicely

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
-                r->reserve(100u);
-                auto& members = pc.info.members;
-                r->push_back({shaderStage, members.array[0].offset, members.array[0].size});
-                for (uint32_t i = 1u; i < members.count; ++i)
+                if (stageIntroBindingInfo.isRunTimeSized())
+                {
+                    descInfo.count = 0u;
+                    descInfo.stride = std::accumulate(stageIntroBindingInfo.count.begin(), stageIntroBindingInfo.count.end(), 1u, multiplyArrayDimensions);
+                    if (descInfo.stride == 0u)
+                        return false;
```
devshgraphicsprogramming, line 154: maybe since the check is the same in both branches, you want to hoist it out and move it after the if-else branch

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
 }
 
-bool CSPIRVIntrospector::introspectAllShaders(core::smart_refctd_ptr<const CIntrospectionData>* outIntrospections, const std::span<const ICPUShader::SSpecInfo> infos)
+// returns true if successfully added all the info to self, false if incompatible with what's already in our pipeline or incomplete (e.g. missing spec constants)
+NBL_API2 bool CSPIRVIntrospector::CPipelineIntrospectionData::merge(const CSPIRVIntrospector::CStageIntrospectionData* stageData, const ICPUShader::SSpecInfoBase::spec_constant_map_t* specConstants)
 {
-    auto it = outIntrospections;
-    for (auto info : infos)
+    // validate if descriptors are compatible
+    DescriptorSetBindings descriptorsToMerge[ICPUPipelineLayout::DESCRIPTOR_SET_COUNT];
```
devshgraphicsprogramming, line 113: you need to initialize those to a copy of the already present `m_descriptorSetBindings` because you're supposed to check cross-stage-introspection collisions as well, not just self

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+                descInfo.stride = 0u;
+            }
+
+            const auto& pplnIntroDataFoundBinding = m_descriptorSetBindings[i].find(descInfo);
+            if (pplnIntroDataFoundBinding != m_descriptorSetBindings[i].end())
+            {
+                if (pplnIntroDataFoundBinding->type != stageIntroBindingInfo.type)
+                    return false;
+                if (pplnIntroDataFoundBinding->count != descInfo.count)
+                    return false;
```
devshgraphicsprogramming, line 176: having non-matching counts is okay, you just `max` their counts

Whats important is to be in agreement whether the binding is runtime sized or not 
Unless there's no iff relationship, then you need to handle "upgrading" a regular binding to a runtime sized one, while also max-ing the counts (because the introspection returned count is now a lower-bound on the size of the runtime array).

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
-            rngToPush.stageFlags = static_cast<ICPUShader::E_SHADER_STAGE>(stageFlags);
-            merged.push_back(rngToPush);
-            stageFlags = 0u;
-
-            rngToPush.offset = i;
-            rngToPush.size = sizeof(uint32_t);
+            if (descriptor.binding < highestBindingsTmp[i] && descriptor.isRuntimeSized())
+                return false;
         }
     }
```
devshgraphicsprogramming, line 202: this should be done somewhere around line 172

include/nbl/asset/utils/CSPIRVIntrospector.h
devshgraphicsprogramming: I read the spec https://registry.khronos.org/vulkan/specs/1.3/html/vkspec.html#interfaces-resources-setandbinding a bit and found:

> The array must be a single-dimensional array and size of the array must be no larger than the number of descriptors in the binding.

Therefore we shouldn't use `span_t<SArrayInfo>` and just a single `SArrayInfo`

It would be good to leave this spec quote here and refer to section 15.5.3

devshgraphicsprogramming: We can still use `SArrayInfo` as the stage introspection doesn't need to worry about the case of a Vulkan Layout SBinding having a runtime flag while also having a SPIR-V module that defines the binding as a static sized array

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -13,172 +17,657 @@
...
-			std::string entryPoint;
-			core::smart_refctd_ptr<const ICPUShader> cpuShader;
+			public:
+				struct SDescriptorInfo final : CIntrospectionData::SDescriptorInfo
+				{
+					inline bool isArray() const {return stride;}
+					inline bool isRuntimeSized() const {return isArray() && count==0;}
+
+					uint32_t count : 21 = 0;
+					uint32_t stride : 11 = 0;
```
devshgraphicsprogramming, line 567: Ok because descriptor arrays can only be:
- 1 dimensional
- runtime array or fixed size
and non-arrays are treated as `[1]` this means we don't need `stride` at all.

However if SPIR-V modules don't need to agree whether a binding is runtime sized, we need
```cpp
uint32_t count : 31 = 0;
uint32_t runtimeSizedCompatible : = 0;
```

Basically then this means that `count` is the lower bound on the binding size (some SPIR-V Stage Module exists that uses the binding as `T[count]`) and `runtimeSizedCompatible` means that some SPIR-V Stage Module exists that uses `T[]`

As a sidenote we should probably store a `uint32_t minByteSize;` for SSBO and UBO descriptor types, which tells us how big we should make the buffer that backs the resource to keep all stages/SPIR-V Modules happy.

Remember that the `minByteSize` does not count runtime arrays toward itself, so 
```glsl
buffer Block {
  ... some other data ...
  T runtimeData[];
} myBuffers[6];
// minByteSize = offsetof(Block,runtimeData[0])
```

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: Also you're not checking whether all the Stage's bindings agree with itself (because there can be resource aliasing) , this why you either need to:
- copy initialize `descriptorsToMerge = m_descriptorSetBindings` and do all checks against that (then you'll check further stage bindings against previous stage bindings), or
- turn this validation from `find` to `debug <<` into a lambda and do two lookups, one in `decriptorsToMerge` and one in `m_descriptorSetBindings`

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
-        for (uint32_t stg=0u; stg<infos.size(); stg++)
-            for (const SPushConstantRange& rng : ranges[stg])
-                if (curr.overlap(rng))
-                    tmpFlags |= rng.stageFlags;
+        const auto& introBindingInfos = stageData->getDescriptorSetInfo(i);
 
-        if (!i)
-            stageFlags = tmpFlags;
+        for (const auto& stageIntroBindingInfo : introBindingInfos)
+            highestBindingsTmp[i] = std::max<const int32_t>(highestBindingsTmp[i], stageIntroBindingInfo.binding);
```
devshgraphicsprogramming, line 195: you could continually increase this value while doing `descriptorsToMerge[i].insert(descInfo);` in the loop above, saves you two iterations

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
-    }
 
-    core::smart_refctd_dynamic_array<SPushConstantRange> rngsArray;
-    if (merged.size())
+    // validation successfull, update `CPipelineIntrospectionData` contents
+    m_highestBindingNumbers = highestBindingsTmp;
+    for (uint32_t i = 0u; i < ICPUPipelineLayout::DESCRIPTOR_SET_COUNT; ++i)
+        m_descriptorSetBindings[i].merge(descriptorsToMerge[i]);
+
+    // TODO: possible push constants validation
```
devshgraphicsprogramming, line 209: there's no validation to do except whether the `offset+size<=MAX_PUSH_CONSTANTS_SIZE` , but IIRC you might already be doing that when creating the stage introspections

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+    // can only be success now
+    const auto& pc = stageData->getPushConstants();
+    auto a = pc.size;
+    if (pc.present())
     {
-        rngsArray = core::make_refctd_dynamic_array<decltype(rngsArray)>(merged.size());
-        memcpy(rngsArray->data(),merged.data(),rngsArray->size()*sizeof(SPushConstantRange));
+        std::span<core::bitflag<ICPUShader::E_SHADER_STAGE>> pcRangesSpan(
+            &m_pushConstantBytes[pc.offset],
+            &m_pushConstantBytes[pc.offset + pc.size]
```
devshgraphicsprogramming, line 217: heads up, with a `T data[256]` doing `&data[256]` is bad practice because `operator[]` and `&` might actually perform some operation (because of overrides) on an out-of-bound memory location, and address sanitizers might moan

the correct syntax here is `pcRangesSpan = {m_pushCosntantBytes+pc.offset,pc.size}`

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
-        if (introBnd.descCountIsSpecConstant)
-        {
-            auto& specInfo = refShader->getSpecializationInfo();
-            auto val = specInfo.getSpecializationByteValue(binding.count);
-            assert(val.second == 4ull);
-            memcpy(&binding.count, val.first, 4ull);
-        }
+    if (prev.stageFlags)
+    {
+        prev.size = current.offset - prev.offset;
```
devshgraphicsprogramming, line 269: after reaching the end of the loop `currentByteOffset==MaxPushConstantsSize` which means taht `current.offset` will be `MaxPushCosntantsSize-1` as well

This is why you need to set the size to `MaxPushConstantsSize-prev.offset`

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+    }
+
+    if(tmp.size() == 0)
+        return nullptr;
 
-        bindings.push_back(binding);
+    std::ostringstream debug;
+    for (auto it = tmp.begin(); it != tmp.end(); it++)
+        debug << "Stage flags: " << std::bitset<32>(it->stageFlags) << " offset: " << it->offset << " size: " << it->size << std::endl;
+    std::cout << debug.str();
```
devshgraphicsprogramming, line 279: would be cool to use a logger for this and make a `CPipelineIntrospectionData::debugPrint(system::ILogger* logger) const` method for all this

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: can leave a comment about section 15.5.3 of Vulkan spec saying descirptor arrays have to be 1D

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: they need to be done same way as `memberSizes`, offsets and strides

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+            auto memberBaseType = comp.get_type(memberId).basetype;
 
-    //all except vertex input are defaulted
-    SBlendParams blending;
-    SPrimitiveAssemblyParams primAssembly;
-    SRasterizationParams raster;
+            auto typeName = getBaseTypeNameFromEnum(spvcrossType2E_TYPE(memberBaseType));
 
-    auto layout = createApproximatePipelineLayoutFromIntrospection_impl(introspections, { infos.begin(),infos.end() });
+            getTypeStore()->typeName = addString(getBaseTypeNameFromEnum(spvcrossType2E_TYPE(memberBaseType)));
```
devshgraphicsprogramming, line 440: why the duplicate code?

devshgraphicsprogramming: anyway we were supposed to delete `getBaseTypeNameFromEnum` when we're done

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+        }
+        else
+        {
+            getTypeStore()->typeName = addString("TODO");
```
devshgraphicsprogramming, line 444: You could actually find the name of a type because whenever one defines a type in SPIR-V they use `OpType...` which gives them a result ID (integer)
```glsl

in vec4 color1;
in vec4 multiplier;
noperspective in vec4 color2;
out vec4 color;

struct S {
    bool b;
    vec4 v[5];
    int i;
};

uniform blockName {
    S s;
    bool cond;
};
```
this is the SPIR-V that generates them
```
          %6 = OpTypeFloat 32                         ; 32-bit float
          %7 = OpTypeVector %6 4                      ; vec4

         %10 = OpConstant %6 1
         %11 = OpConstant %6 2

         %13 = OpTypeInt 32 0                         ; 32-bit int, sign-less
         %14 = OpConstant %13 5
         %15 = OpTypeArray %7 %14
         %16 = OpTypeInt 32 1
         %17 = OpTypeStruct %13 %15 %16
         %18 = OpTypeStruct %17 %13
```
and in the debug info you'll have
```
               OpName %17 "S"
               OpMemberName %17 0 "b"
               OpMemberName %17 1 "v"
               OpMemberName %17 2 "i"
               OpName %18 "blockName"
               OpMemberName %18 0 "s"
               OpMemberName %18 1 "cond"
```

Note that depending on the compiler, anything thats an `OpTypeFloat`, `OpTypeInt`, `OpTypeVector` or `OpTypeArray` might not get debug names so we should probably build outselves.

You don't need to do all this right now, because I want you to progress to other tasks, but would be good to write it up here and open a GH issue about it for later fixing.

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,528 @@ namespace nbl::asset
...
+    stageIntroData->m_shaderStage = shaderStage;
 
-    _comp.set_entry_point(entryPoint, stage);
+    comp.set_entry_point(params.entryPoint, stage);
 
     // spec constants
-    spirv_cross::SmallVector<spirv_cross::SpecializationConstant> sconsts = _comp.get_specialization_constants();
-    mapId2SpecConst_t mapId2SpecConst;
-    introData->specConstants.reserve(sconsts.size());
+        // TODO: hash map instead
```
devshgraphicsprogramming, line 526: do it

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -504,210 +560,333 @@ core::smart_refctd_ptr<const CSPIRVIntrospector::CIntrospectionData> CSPIRVIntro
...
+    else
+        stageIntroData->m_output = OutputVecT();
+
     for (const spirv_cross::Resource& r : resources.stage_outputs)
     {
-        SShaderInfoVariant& res = addInfo_common(r, ESIT_STAGE_OUTPUT);
-        res.glslType = getStageIOtype(r.base_type_id);
+        if (shaderStage == IShader::ESS_FRAGMENT)
+        {
+            // TODO: desn't path here for frag_test1.hlsl, why?
```
devshgraphicsprogramming, line 688: someone didn't figure out  the shader stage? or is a stale TODO?

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -504,210 +560,333 @@ core::smart_refctd_ptr<const CSPIRVIntrospector::CIntrospectionData> CSPIRVIntro
...
-        m.rowMajor = false;
-        m.type = EGVT_UNKNOWN_OR_STRUCT;
-        m.members.array = nullptr;
-        m.members.count = 0u;
-        return m;
+    auto addBaseAndConvertExtentToImmutable = [basePtr](std::span<const CIntrospectionData::SArrayInfo>& count)->void
+    {
+        count = reinterpret_cast<const core::based_span<CIntrospectionData::SArrayInfo>&>(count)(basePtr);
     };
+    // TODO: spec constant finalization
```
devshgraphicsprogramming, line 738: done?

src/nbl/asset/utils/CSPIRVIntrospector.cpp
Przemog1: `debugPrint` as a distinct function?

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: also spec constants should be in an `unordered_set`, see the TODO in the header

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -504,210 +560,333 @@ core::smart_refctd_ptr<const CSPIRVIntrospector::CIntrospectionData> CSPIRVIntro
...
-            _pushStack.push({member.members, mtype, member.offset});
-        else
-		{
-            member.mtxRowCnt = mtype.vecsize;
-            member.mtxColCnt = mtype.columns;
-            if (member.mtxColCnt > 1u)
-                member.mtxStride = _comp.type_struct_member_matrix_stride(_parentType, m);
+    addBaseAndConvertBlockToImmutable(reinterpret_cast<SPushConstantInfo<true>&>(m_pushConstants));
+    addBaseAndConvertStringToImmutable(m_pushConstants.name);
+    std::ostringstream debug = {};
```
devshgraphicsprogramming, line 764: time to move all the debug print to a dedicated method that takes a `system::ILogger*`

src/nbl/asset/utils/CSPIRVIntrospector.cpp
Przemog1: we are getting rid of `stride` anyway

devshgraphicsprogramming: ah yes! need a new way to indicate failure I guess?

Przemog1: done

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -348,7 +333,7 @@ CSPIRVIntrospector::CStageIntrospectionData::SDescriptorVarInfo<true>* CSPIRVInt
             //.descCount = 1
         },
         /*.name = */addString(r.name),
-        /*.count = */addCounts(arrDim,type.array.data(),type.array_size_literal.data())
+        /*.count = *///addCounts(arrDim,type.array.data(),type.array_size_literal.data())
```
devshgraphicsprogramming, line 336: noting so we don't forget to undo

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -333,7 +333,6 @@ CSPIRVIntrospector::CStageIntrospectionData::SDescriptorVarInfo<true>* CSPIRVInt
             //.descCount = 1
         },
         /*.name = */addString(r.name),
-        /*.count = *///addCounts(arrDim,type.array.data(),type.array_size_literal.data())
```
devshgraphicsprogramming: ok so where's the count being set now?

include/nbl/asset/utils/CSPIRVIntrospector.h
Przemog1: it does exist, idk if it is used anywhere

devshgraphicsprogramming: remove it and see

include/nbl/asset/utils/CSPIRVIntrospector.h
Przemog1: would leave it until we can test gpu pipeline introspection

devshgraphicsprogramming: ok

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -315,24 +315,25 @@ CSPIRVIntrospector::CStageIntrospectionData::SDescriptorVarInfo<true>* CSPIRVInt
...
-    const auto arrDim = type.array.size();
-    // assuming only 1D arrays because i don't know how desc set layout binding is constructed when it's let's say 2D array (e.g. uniform sampler2D smplr[4][5]; is it even legal?)
+    const size_t arrDim = type.array.size();
+    // only 1D arrays allowed
     if (arrDim>1)
         return nullptr; // TODO: log fail
 
-
+    const uint32_t descriptorArraySize = type.array.empty() ? 1u : *type.array.data();    
+    const bool isArrayTypeLiteral = descriptorArraySize <= 1u ? true : type.array_size_literal[0]; // can be set to true for runtime sized descriptors, doesn't make sense but also doesn't matter
```
devshgraphicsprogramming, line 329: i think you have a bug here, a spec constant beign used would give you `descriptorArraySize = specConstantID`

then a spec constant Id of 0 or 1 would register as `isArrayTypeLiteral`

I think you should check `typ.array_size_literal[0]` first 

Or did you mean to check `arrDim!=0` here isntead?

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -526,6 +527,30 @@ class NBL_API2 CSPIRVIntrospector : public core::Uncopyable
 
 					return range;
 				}
+				inline SDescriptorArrayInfo addDescriptorCount(const uint32_t size, const bool sizeIsLiteral)
+				{
+					const auto descCount = alloc<SDescriptorArrayInfo>(1u);
+					auto descriptorArraySize = descCount(m_memPool.data());
```
devshgraphicsprogramming, line 533: don't allocate anything, no point allocating fixed-size member, the alloc pool stuff is only for variable lenght things like strings, and arrays of unknown size

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -526,6 +527,30 @@ class NBL_API2 CSPIRVIntrospector : public core::Uncopyable
...
+
+					// the API for this spec constant checking is truly messed up
+					if (size == 0u)
+					{
+						descriptorArraySize[0].count = size;
+						descriptorArraySize[0].countMode = SDescriptorArrayInfo::DESCRIPTOR_COUNT::RUNTIME;
+					}
+					else if (sizeIsLiteral)
```
devshgraphicsprogramming, line 541: `sizeIsLiteral` is the first thing you should check, because right not spec constant ID of 0 would falsely register as a runtime array

```cpp
if (sizeIsLiteral)
{
    descriptorArraySize.count = size;
    descriptorArraySize.countMode = size!=0 ? STATIC:RUNTIME;
}
else
{
    descriptorArraySize.count = size;
    descriptorArraySize.countMode = SPEC_CONSTANT;
}
```

src/nbl/asset/utils/CSPIRVIntrospector.cpp
Przemog1: stale

include/nbl/asset/utils/CSPIRVIntrospector.h
```diff
@@ -630,6 +548,17 @@ class NBL_API2 CSPIRVIntrospector : public core::Uncopyable
 					std::fill(m_highestBindingNumbers.begin(), m_highestBindingNumbers.end(), -1);
 				}
 
+				inline bool canSpecializationlesslyCreateDescSetFrom() const
+				{
+					for (const auto& descSet : m_descriptorSetBindings)
+					{
+						auto found = std::find_if(descSet.begin(),descSet.end(),[](const SDescriptorInfo& bnd)->bool{ return bnd.isRuntimeSized();});
+						if (found!=descSet.end())
+							return false;
```
devshgraphicsprogramming, line 557: its not about being runtime sized, its about having the count from spec-constants

include/nbl/asset/utils/CSPIRVIntrospector.h
devshgraphicsprogramming: returning a span would probably work better

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,511 @@ namespace nbl::asset
...
+                        const auto& specConstantFound = specConstants->find(count.specID);
+                        if (specConstantFound == specConstants->end())
+                            return false;
 
-    SPushConstantRange rngToPush; rngToPush.offset = 0u; rngToPush.size = 0u;
-    uint32_t stageFlags = 0u;
-    for (uint32_t i = 0u; i < ICPUMeshBuffer::MAX_PUSH_CONSTANT_BYTESIZE/sizeof(uint32_t); i += sizeof(uint32_t))
-    {
-        SPushConstantRange curr; curr.offset = i; curr.size = sizeof(uint32_t);
+                        descInfo.count = count.count;
```
devshgraphicsprogramming, line 146: no, you're supposed to use `specConstantFound`'s value

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,511 @@ namespace nbl::asset
...
-            rngToPush.stageFlags = static_cast<ICPUShader::E_SHADER_STAGE>(stageFlags);
-            merged.push_back(rngToPush);
-            stageFlags = 0u;
+            const auto& pplnIntroDataFoundBinding = m_descriptorSetBindings[i].find(descInfo);
+            if (pplnIntroDataFoundBinding != m_descriptorSetBindings[i].end())
+            {
+                if (pplnIntroDataFoundBinding->type != stageIntroBindingInfo.type)
+                    return false;
+                descInfo.count = std::max(pplnIntroDataFoundBinding->count, descInfo.count);
+            }
```
devshgraphicsprogramming, line 171: either you look up both in `m_descriptorSetBindings` AND `descriptorsToMerge` or you initialize `descriptorsToMerge` to be a copy of `m_descriptorSetBindings` as per https://github.com/Devsh-Graphics-Programming/Nabla/pull/677#discussion_r1572639296

Just add
```cpp
const auto& selfIntersectionFound = descriptorsToMerge[i].find(descInfo);
if (selfIntersectionFound!=descriptorsToMerge[i].end() && selfIntersectionFound->type!=stageIntroBindingInfo.type)
    return false;
```

Right now you don't check "self collisions" i.e. bindings that repeat with different types in the same SPIR-V module (which would make the module invalid ofc)

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -13,472 +13,511 @@ namespace nbl::asset
...
+    for (uint32_t i = 0u; i < ICPUPipelineLayout::DESCRIPTOR_SET_COUNT; ++i)
     {
-        rngToPush.stageFlags = static_cast<ICPUShader::E_SHADER_STAGE>(stageFlags);
-        merged.push_back(rngToPush);
+        const auto& introBindingInfos = stageData->getDescriptorSetInfo(i);
+
+        for (const auto& descriptor : descriptorsToMerge[i])
+        {
+            if (descriptor.binding < highestBindingsTmp[i] && descriptor.isRuntimeSized())
+                return false;
```
devshgraphicsprogramming, line 186: ok here you're checking if the new descriptors you're adding are not runtime sized if they're not the last.

But also, if you add new descriptors and `m_descriptorSetBindings` has an old runtime descriptor, you need to check whether it can still be the runtime descriptor

you can do this by looking up `m_highestBindingsNumbers[i]` (BEFORE you assign the new value) in `m_descriptorSetBindings`, checking if that descriptor was runtime sized and if it was you need to check if `highestBindingsTmp[i]` is higher

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: add this as a block comment  on this line

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -858,10 +933,14 @@ void CSPIRVIntrospector::CStageIntrospectionData::debugPrint(system::ILogger* lo
         debug << "SPEC CONSTATS:\n";
 
         for (auto& specConstant : m_specConstants)
-            debug << " name: " << std::string_view(specConstant.name.begin(), specConstant.name.end()) << ' '
-            << "TODO: type "
-            << "id: " << specConstant.id << " byte size: " << "TODO: specConstant.defaultValue" << '\n';
-    }
+        {
+            // TODO: it gives weird errors, debug
```
devshgraphicsprogramming, line 943: maybe doesn't have `\0` at end of span OR it was forgotten about during the mutable->imutable conversion from based_span to regular span

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -255,17 +260,19 @@ NBL_API2 bool CSPIRVIntrospector::CPipelineIntrospectionData::merge(const CSPIRV
...
         }
+
+        if (m_highestBindingNumbers[i].isRunTimeSized && m_highestBindingNumbers[i].binding < highestBindingsTmp[i].binding)
+            return false;
     }
 
-    // validation successfull, update `CPipelineIntrospectionData` contents
-    m_highestBindingNumbers = highestBindingsTmp;
+    //// validation successfull, update `CPipelineIntrospectionData` contents
+    //m_highestBindingNumbers = highestBindingsTmp;
```
devshgraphicsprogramming, line 272: you still need to copy the results over into `m_highestBindingNumbers`

include/nbl/asset/utils/CSPIRVIntrospector.h
Przemog1: Nabla does compile without it, deleting

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -473,8 +469,23 @@ void CSPIRVIntrospector::CStageIntrospectionData::shaderMemBlockIntrospection(co
...
-//            comp.get_declared_struct_size();
-//            comp.get_declared_struct_size_runtime_array();
+
+            SType<true>::MatrixInfo matrixInfo = {
+                .rowMajor = comp.get_member_decoration(parentType.self, entry.memberIndex, spv::DecorationRowMajor),
+                .rows = parentType.vecsize,
+                .columns = parentType.columns
+            };
+
+            getParentTypeStore()->memberMatrixInfos()(m_memPool.data())[entry.memberIndex] = reinterpret_cast<uint32_t&>(matrixInfo);
```
devshgraphicsprogramming, line 479: I wouldn't do the reinterpret_cast to `uint32_t&`

src/nbl/asset/utils/CSPIRVIntrospector.cpp
```diff
@@ -470,13 +470,14 @@ void CSPIRVIntrospector::CStageIntrospectionData::shaderMemBlockIntrospection(co
             getParentTypeStore()->memberOffsets()(m_memPool.data())[entry.memberIndex] = comp.type_struct_member_offset(parentType,entry.memberIndex);
             getParentTypeStore()->memberStrides()(m_memPool.data())[entry.memberIndex] = getTypeStore()->isArray() ? comp.type_struct_member_array_stride(parentType,entry.memberIndex):0u;
 
+            _NBL_DEBUG_BREAK_IF(parentType.columns > 1);
```
devshgraphicsprogramming, line 473: why do you debug break if columns>1 ?

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: remove the TODO and correct the comment

src/nbl/asset/utils/CSPIRVIntrospector.cpp
Przemog1: forgot to delete

tools/nsc/.profiles/0.json
```diff
@@ -1,31 +1,31 @@
...
-  "profile" : 
-  {
-    "backend" : "vulkan",
-    "buildModes" : [],
-    "gpuArchitectures" : [],
-    "platform" : "windows",
-    "runConfiguration" : "Release"
-  },
-  "scriptPath" : "../test/test.py",
-  "threadsPerBuildProcess" : 2
```
devshgraphicsprogramming, line 30: remove these files, they're temp

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: please add the concents of the first comment as block comment on this line

src/nbl/asset/utils/CSPIRVIntrospector.cpp
devshgraphicsprogramming: fixed?

devshgraphicsprogramming:
you've accidentally added the OpenSSL module back

PR #693, opened 2024-05-03
Separable samplers

Fletterio:
## Description
Writing the necessary stuff so that we can have decoupled image/samplers in bindings

## Testing 
I changed example 07 a little bit and it runs normally. If you remove the sampler from the equation it doesn't forward any to the vk write

## TODO list:
If you still provide a sampler in the DescWrite, it will be forwarded to Vulkan (the current way stuff works is that if the type's category is image and you provide samplers, they will be sent). I *think* this is not an issue because if the `VkDescriptorType` of the write is `VK_SAMPLED_IMAGE` the samplers of the image info are ignored. However it makes no sense to give samplers to an image info for a sampled image descriptor. If necessary I can make it generate a warning or fail


devshgraphicsprogramming:
>  However it makes no sense to give samplers to an image info for a sampled image descriptor. If necessary I can make it generate a warning or fail

Its necessary for us to ignore the sampler as well, because otherwise we write it into the lifetime tracking of the pool.

Fletterio:
I'll see when it gets lifetime tracked then and have it be ignored at that point

src/nbl/video/IDescriptorPool.cpp
```diff
@@ -14,7 +14,7 @@ IDescriptorPool::IDescriptorPool(core::smart_refctd_ptr<const ILogicalDevice>&&
...
     m_mutableSamplerStorage = std::make_unique<core::StorageTrivializer<core::smart_refctd_ptr<video::IGPUSampler>>[]>(m_creationParameters.maxDescriptorCount[static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_COMBINED_IMAGE_SAMPLER)]);
     m_storageImageStorage = std::make_unique<core::StorageTrivializer<core::smart_refctd_ptr<IGPUImageView>>[]>(m_creationParameters.maxDescriptorCount[static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_STORAGE_IMAGE)] + m_creationParameters.maxDescriptorCount[static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_INPUT_ATTACHMENT)]);
     m_UBO_SSBOStorage = std::make_unique<core::StorageTrivializer<core::smart_refctd_ptr<IGPUBuffer>>[]>(m_creationParameters.maxDescriptorCount[static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_UNIFORM_BUFFER)] + m_creationParameters.maxDescriptorCount[static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_STORAGE_BUFFER)] + m_creationParameters.maxDescriptorCount[static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_UNIFORM_BUFFER_DYNAMIC)] + m_creationParameters.maxDescriptorCount[static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_STORAGE_BUFFER_DYNAMIC)]);
```
devshgraphicsprogramming, line 20: btw we should probably leverage the newfound knowledge that you can't have an UPDATE_AFTER_BIND pool and descriptors that have DYNAMIC types, probably in `ILogicalDevice::createDescriptorPool`

devshgraphicsprogramming: also one line down there's a typo/bug, its supposed to be UNIFORM_TEXEL_BUFFER + STORAGE_BUFFER_DYNAMIC

include/nbl/video/ILogicalDevice.h
```diff
@@ -889,6 +889,7 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
             std::span<const IGPUDescriptorSet::SWriteDescriptorSet> writes;
             std::span<const IGPUDescriptorSet::SCopyDescriptorSet> copies;
             const asset::IDescriptor::E_TYPE* pWriteTypes;
+            uint32_t samplerCount = 0u;
             uint32_t bufferCount = 0u;
             uint32_t bufferViewCount = 0u;
             uint32_t imageCount = 0u;
```
devshgraphicsprogramming, line 895: should add a comment that `imageCount` counts both COMBINED and separatae images

devshgraphicsprogramming: ah I thought I was looking at a different struct, nvm what I said

include/nbl/asset/ICPUDescriptorSet.h
```diff
@@ -101,35 +101,7 @@ class NBL_API2 ICPUDescriptorSet final : public IDescriptorSet<ICPUDescriptorSet
 	private:
 		static inline IDescriptor::E_CATEGORY getCategoryFromType(const IDescriptor::E_TYPE type)
```
devshgraphicsprogramming, line 102: I'll just remove the function all together

include/nbl/asset/IDescriptorSet.h
```diff
@@ -92,31 +93,31 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
 				}
 				~SDescriptorInfo()
 				{
-					if (desc && desc->getTypeCategory()==IDescriptor::EC_IMAGE)
+					if (desc and (desc->getTypeCategory()==IDescriptor::EC_IMAGE or desc->getTypeCategory() == IDescriptor::EC_SAMPLER))
```
devshgraphicsprogramming, line 96: no when you are `EC_SAMPLER` then your `desc` is the sampler, and `sampler` is not supposed to be used

include/nbl/asset/IDescriptorSet.h
```diff
@@ -92,31 +93,31 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
...
 						info.image.sampler = nullptr;
 				}
 
 				inline SDescriptorInfo& operator=(const SDescriptorInfo& other)
 				{
-					if (desc && desc->getTypeCategory()==IDescriptor::EC_IMAGE)
+					if (desc and (desc->getTypeCategory()==IDescriptor::EC_IMAGE or desc->getTypeCategory() == IDescriptor::EC_SAMPLER))
```
devshgraphicsprogramming, line 102: same as https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/files#r1648875709

include/nbl/asset/IDescriptorSet.h
```diff
@@ -47,9 +47,10 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
                 };
                 struct SImageInfo
                 {
-					// This will be ignored if the DS layout already has an immutable sampler specified for the binding.
-                    core::smart_refctd_ptr<typename layout_t::sampler_type> sampler;
-                    IImage::LAYOUT imageLayout;
+					IImage::LAYOUT imageLayout;
+					// If the binding is COMBINED, this will be ignored if the DS layout already has an immutable sampler specified for the binding
+					// If the binding is an immutable SAMPLER, whatever's here doesn't matter because the write validation will fail
```
devshgraphicsprogramming, line 52: if the binding is `SAMPLER` then its the `desc` thats the sampler, not this `sampler`

include/nbl/asset/IDescriptorSet.h
devshgraphicsprogramming: see https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/files#r1648876515

include/nbl/asset/IDescriptorSet.h
devshgraphicsprogramming: this variable is only relevant for `COMBINED` and Mutable bindings, otherwise gets ignored or fails validation

include/nbl/asset/IDescriptorSet.h
```diff
@@ -92,31 +93,31 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
...
 
 				inline SDescriptorInfo& operator=(const SDescriptorInfo& other)
 				{
-					if (desc && desc->getTypeCategory()==IDescriptor::EC_IMAGE)
+					if (desc and (desc->getTypeCategory()==IDescriptor::EC_IMAGE or desc->getTypeCategory() == IDescriptor::EC_SAMPLER))
 						info.image.sampler = nullptr;
 					desc = other.desc;
 					const auto type = desc->getTypeCategory();
-					if (type!=IDescriptor::EC_IMAGE)
+					if (type!=IDescriptor::EC_IMAGE and type!=IDescriptor::EC_SAMPLER)
```
devshgraphicsprogramming, line 106: again https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/files#r1648875709

so you need a `switch (type)` probably:
```
case EC_BUFFER:
   info.buffer = other.info.buffer;
   break;
case EC_IMAGE:
   info.image= other.info.image;
   break;
default: // the rest don't have layouts, extra samplers, or buffer ranges (even buffer view
   break;
```

include/nbl/asset/IDescriptorSet.h
devshgraphicsprogramming: `imageLayout` ofc not ignored for `SAMPLED_IMAGE`

include/nbl/asset/IDescriptorSet.h
```diff
@@ -92,31 +93,31 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
...
 						info.buffer = other.info.buffer;
 					else
 						info.image = other.info.image;
 					return *this;
 				}
 				inline SDescriptorInfo& operator=(SDescriptorInfo&& other)
 				{
-					if (desc && desc->getTypeCategory()==IDescriptor::EC_IMAGE)
+					if (desc and (desc->getTypeCategory() == IDescriptor::EC_IMAGE or desc->getTypeCategory() == IDescriptor::EC_SAMPLER))
```
devshgraphicsprogramming, line 114: no, only for `EC_IMAGE`

include/nbl/asset/IDescriptorSet.h
```diff
@@ -92,31 +93,31 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
...
 						info.image = {nullptr,IImage::LAYOUT::UNDEFINED};
 					desc = std::move(other.desc);
 					if (desc)
 					{
 						const auto type = desc->getTypeCategory();
-						if (type!=IDescriptor::EC_IMAGE)
+						if (type!=IDescriptor::EC_IMAGE and type!=IDescriptor::EC_SAMPLER)
```
devshgraphicsprogramming, line 120: again https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/files#r1648880223

include/nbl/asset/IDescriptorSetLayout.h
```diff
@@ -396,9 +406,9 @@ class IDescriptorSetLayout : public virtual core::IReferenceCounted  // TODO: tr
 
 	CBindingRedirect m_descriptorRedirects[static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_COUNT)];
 	CBindingRedirect m_immutableSamplerRedirect;
-	CBindingRedirect m_mutableSamplerRedirect;
+	CBindingRedirect m_mutableCombinedSamplerRedirect;
 
-	core::smart_refctd_dynamic_array<core::smart_refctd_ptr<sampler_type>> m_samplers = nullptr;
+	core::smart_refctd_dynamic_array<core::smart_refctd_ptr<sampler_type>> m_immutableSamplers = nullptr;
```
devshgraphicsprogramming, line 411: how do you know which part of `m_immutableSamplers` is for COMBINED's immutable samplers and for SAMPLER immutables?

devshgraphicsprogramming: or do you now need to know ?

include/nbl/asset/IDescriptorSet.h
```diff
@@ -47,9 +47,10 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
                 };
                 struct SImageInfo
                 {
-					// This will be ignored if the DS layout already has an immutable sampler specified for the binding.
-                    core::smart_refctd_ptr<typename layout_t::sampler_type> sampler;
-                    IImage::LAYOUT imageLayout;
+					IImage::LAYOUT imageLayout;
+					// If the binding is COMBINED, this will be ignored if the DS layout already has an immutable sampler specified for the binding
+					// If the binding is an immutable SAMPLER, whatever's here doesn't matter because the write validation will fail
+					core::smart_refctd_ptr<typename layout_t::sampler_type> sampler;
```
devshgraphicsprogramming, line 53: btw have you established if its invalid to have `sampler` non-nullptr and `desc` non-nullptr when `desc` is a SAMPLER ?

include/nbl/asset/IDescriptorSetLayout.h
```diff
@@ -340,14 +342,14 @@ class IDescriptorSetLayout : public virtual core::IReferenceCounted  // TODO: tr
 
 	inline const CBindingRedirect& getDescriptorRedirect(const IDescriptor::E_TYPE type) const { return m_descriptorRedirects[static_cast<uint32_t>(type)]; }
 	inline const CBindingRedirect& getImmutableSamplerRedirect() const { return m_immutableSamplerRedirect; }
-	inline const CBindingRedirect& getMutableSamplerRedirect() const { return m_mutableSamplerRedirect; }
+	inline const CBindingRedirect& getMutableCombinedSamplerRedirect() const { return m_mutableCombinedSamplerRedirect; }
 
 	inline core::SRange<const core::smart_refctd_ptr<sampler_type>> getImmutableSamplers() const
```
devshgraphicsprogramming, line 347: for the `getImmutable...` don't you need to distinguish between immutables from COMBINED and SAMPLER bindings?

include/nbl/asset/IDescriptorSetLayout.h
```diff
@@ -359,34 +361,42 @@ class IDescriptorSetLayout : public virtual core::IReferenceCounted  // TODO: tr
...
-				if (b.samplers)
-					buildInfo_immutableSamplers.emplace_back(b.binding, b.createFlags, b.stageFlags, b.count);
-				else
-					buildInfo_mutableSamplers.emplace_back(b.binding, b.createFlags, b.stageFlags, b.count);
+			switch (b.type) {
+				case IDescriptor::E_TYPE::ET_SAMPLER:
+					if (b.samplers)
+						buildInfo_immutableSamplers.emplace_back(b.binding, b.createFlags, b.stageFlags, b.count);
+					else 
+						buildInfo_descriptors[static_cast<uint32_t>(b.type)].emplace_back(b.binding, b.createFlags, b.stageFlags, b.count);
```
devshgraphicsprogramming, line 369: you already know the type, so use that instead of `static_cast<uint32_t>(b.type)`

include/nbl/video/IDescriptorPool.h
```diff
@@ -218,7 +222,8 @@ class NBL_API2 IDescriptorPool : public IBackendObject
         std::unique_ptr<IGPUDescriptorSet* []> m_allocatedDescriptorSets = nullptr; // This array might be sparse.
 
         std::unique_ptr<core::StorageTrivializer<core::smart_refctd_ptr<video::IGPUImageView>>[]> m_textureStorage;
-        std::unique_ptr<core::StorageTrivializer<core::smart_refctd_ptr<video::IGPUSampler>>[]> m_mutableSamplerStorage;
+        std::unique_ptr<core::StorageTrivializer<core::smart_refctd_ptr<video::IGPUSampler>>[]> m_mutableCombinedSamplerStorage;
+        std::unique_ptr<core::StorageTrivializer<core::smart_refctd_ptr<video::IGPUSampler>>[]> m_MutableStandaloneSamplerStorage;
```
devshgraphicsprogramming, line 226: change the capitalization of `M` to `m`

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -140,7 +157,7 @@ void IGPUDescriptorSet::dropDescriptors(const IGPUDescriptorSet::SDropDescriptor
     const auto descriptorType = getBindingType(drop.binding);
 
 	auto* dstDescriptors = drop.dstSet->getDescriptors(descriptorType, drop.binding);
-	auto* dstSamplers = drop.dstSet->getMutableSamplers(drop.binding);
+	auto* dstSamplers = drop.dstSet->getMutableCombinedSamplers(drop.binding);
```
devshgraphicsprogramming, line 160: you should skip this call and the contents of the if statement on line 166 if the `descriptorType` is not COMBINED

(there's a binary search or hashmap lookup involved, its potentially expensive)

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -152,7 +169,10 @@ void IGPUDescriptorSet::dropDescriptors(const IGPUDescriptorSet::SDropDescriptor
 
 	// we only increment the version to detect UPDATE-AFTER-BIND and automagically invalidate descriptor sets
 	// so, only if we do the path that writes descriptors, do we want to increment version
-    if (getOriginDevice()->getEnabledFeatures().nullDescriptor)
+    auto& bindingRedirect = m_layout->getDescriptorRedirect(descriptorType);
```
devshgraphicsprogramming, line 172: if you look at https://github.com/Devsh-Graphics-Programming/Nabla/blob/3df38a45978t510f9823c97adfbc3bea4c90e8aee/include/nbl/video/IGPUDescriptorSet.h#L89

it performs a `getDescriptorRedirect(type)` then calls https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/asset/IDescriptorSetLayout.h#L161

which means it calls `findBindingStorageIndex` alreayd

I'd like to amortize this lookup of a storage_offset` and use it to both `getCreateFlags` and get your destination arraydescriptor array

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -152,7 +169,10 @@ void IGPUDescriptorSet::dropDescriptors(const IGPUDescriptorSet::SDropDescriptor
...
+    auto bindingCreateFlags = bindingRedirect.getCreateFlags(bindingRedirect.findBindingStorageIndex({ drop.binding }));
+    bool shouldIncrementVersion = not (bindingCreateFlags.hasFlags(IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_UPDATE_AFTER_BIND_BIT) or bindingCreateFlags.hasFlags(IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_UPDATE_UNUSED_WHILE_PENDING_BIT));
```
devshgraphicsprogramming, line 174: its significantly less clutter to write
```cpp
using binding_flags_t = IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS;
const bool shouldIncrementVersion = not (bindingCreateFlags.hasFlags(binding_flags_t::ECF_UPDATE_AFTER_BIND_BIT) or bindingCreateFlags.hasFlags(binding_flags_t::ECF_UPDATE_UNUSED_WHILE_PENDING_BIT));
```

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -152,7 +169,10 @@ void IGPUDescriptorSet::dropDescriptors(const IGPUDescriptorSet::SDropDescriptor
...
+    if (getOriginDevice()->getEnabledFeatures().nullDescriptor or shouldIncrementVersion)
```
devshgraphicsprogramming, line 175: AND, not OR

alternatively for consistent behaviour, you can just do `shouldIncrementVersion` and not care about `nullDescriptor` feature

src/nbl/video/IGPUDescriptorSet.cpp
devshgraphicsprogramming: also why not make some static method, i.e. `IGPUDescriptorSetLayout::writeIncrementsVersion(const core::bitflag<SBinding::E_CREATE_FLAGS>)` that you can reuse?

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -172,8 +192,8 @@ bool IGPUDescriptorSet::validateCopy(const IGPUDescriptorSet::SCopyDescriptorSet
         auto* srcDescriptors = copy.srcSet->getDescriptors(type, copy.srcBinding);
         auto* dstDescriptors = copy.dstSet->getDescriptors(type, copy.dstBinding);
 
-        auto* srcSamplers = copy.srcSet->getMutableSamplers(copy.srcBinding);
-        auto* dstSamplers = copy.dstSet->getMutableSamplers(copy.dstBinding);
+        auto* srcSamplers = copy.srcSet->getMutableCombinedSamplers(copy.srcBinding);
+        auto* dstSamplers = copy.dstSet->getMutableCombinedSamplers(copy.dstBinding);
```
devshgraphicsprogramming, line 196: again, I'd try to claw back some perf by doing
```cpp
type!=COMBINED ?  nullptr:...Set->getMutableCombinedSamplers(copy...Binding);
```

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -201,18 +223,24 @@ void IGPUDescriptorSet::processCopy(const IGPUDescriptorSet::SCopyDescriptorSet&
...
         assert(!(!srcSamplers != !dstSamplers));
 
         if (srcDescriptors && dstDescriptors)
             std::copy_n(srcDescriptors, copy.count, dstDescriptors);
 
         if (srcSamplers && dstSamplers)
             std::copy_n(srcSamplers, copy.count, dstSamplers);
+
+        auto& bindingRedirect = m_layout->getDescriptorRedirect(type);
+        auto bindingCreateFlags = bindingRedirect.getCreateFlags(bindingRedirect.findBindingStorageIndex({ copy.srcBinding }));
```
devshgraphicsprogramming, line 237: again amortize the lookup of the storage redirect from the binding, like here https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/files#r1649030085

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -201,18 +223,24 @@ void IGPUDescriptorSet::processCopy(const IGPUDescriptorSet::SCopyDescriptorSet&
...
+        if (not (bindingCreateFlags.hasFlags(IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_UPDATE_AFTER_BIND_BIT) or bindingCreateFlags.hasFlags(IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_UPDATE_UNUSED_WHILE_PENDING_BIT)))
+            shouldIncrementVersion = true;
```
devshgraphicsprogramming, line 239: https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/files#r1649046291

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -201,18 +223,24 @@ void IGPUDescriptorSet::processCopy(const IGPUDescriptorSet::SCopyDescriptorSet&
         auto* dstDescriptors = copy.dstSet->getDescriptors(type, copy.dstBinding);
         assert(!(!srcDescriptors != !dstDescriptors));
```
devshgraphicsprogramming, line 224: btw the reason we have the `t` loop, is because we have to "guess the type"

if would be useful to `continue` (skip rest of loop) if `!srcDescriptors` like
```cpp
if (!srcDescriptors)
{
   assert(!dstDescriptors);
   continue;
}
assert(dstDescriptors);
```

src/nbl/video/IGPUDescriptorSet.cpp
devshgraphicsprogramming: also it would be useful to `break;` out of the loop after one successful run

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -129,8 +144,10 @@ void IGPUDescriptorSet::processWrite(const IGPUDescriptorSet::SWriteDescriptorSe
         if (mutableSamplers)
             mutableSamplers[j] = write.info[j].info.image.sampler;
     }
-
-    incrementVersion();
+    auto& bindingRedirect = m_layout->getDescriptorRedirect(type);
+    auto bindingCreateFlags = bindingRedirect.getCreateFlags(bindingRedirect.findBindingStorageIndex({ write.binding }));
+    if (not (bindingCreateFlags.hasFlags(IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_UPDATE_AFTER_BIND_BIT) or bindingCreateFlags.hasFlags(IGPUDescriptorSetLayout::SBinding::E_CREATE_FLAGS::ECF_UPDATE_UNUSED_WHILE_PENDING_BIT)))
```
devshgraphicsprogramming, line 149: https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/files#r1649046291

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -129,8 +144,10 @@ void IGPUDescriptorSet::processWrite(const IGPUDescriptorSet::SWriteDescriptorSe
         if (mutableSamplers)
             mutableSamplers[j] = write.info[j].info.image.sampler;
     }
-
-    incrementVersion();
+    auto& bindingRedirect = m_layout->getDescriptorRedirect(type);
+    auto bindingCreateFlags = bindingRedirect.getCreateFlags(bindingRedirect.findBindingStorageIndex({ write.binding }));
```
devshgraphicsprogramming, line 148: https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/files#r1649030085

include/nbl/asset/IDescriptorSetLayout.h
Fletterio: I don't recall needing to know, I think I just treat all immutable samplers equally

include/nbl/asset/IDescriptorSet.h
Fletterio: No, and in fact I just realized both `sampler` and `desc` are either pointing to the same thing pointlessly, or could point to different things. 

The latter is unwanted since it's flatout wrong so I'll refactor so it doesn't happen. I think I'll opt for forcing `sampler` to be nullptr when the descriptor is a SAMPLER, since most predicated for behaviour are done by checking the category of `desc`

Fletterio: Forcing `sampler` and `desc` to point to the same object in such a case is also an option - I get an extra useless smart pointer but the code ends up being simpler I think: no edge cases for the standalone sampler type

Fletterio: Okay I think I can have both: I give `SDescriptorInfo` some sort of `getSampler()` method that I can use throughout the rest of the code so I don't have edge cases, and also force SAMPLER to have its `info.sampler` be nullptr or at least just ignore that field

Fletterio: So now, a question about write/copy validation: I expect to only be using `desc` for descriptors of type SAMPLER, and ignore whatever is given under the `info` field. Should I, however, do anything about the info field? Like do I force it to be nullptr and drop a smart pointer there, or is that a problem of the user?

include/nbl/asset/IDescriptorSetLayout.h
Fletterio: Not sure, I don't see why I would need to. I need a refresher on most of this but I don't recall needing to

include/nbl/asset/ICPUDescriptorSet.h
devshgraphicsprogramming: todo

include/nbl/asset/IDescriptorSetLayout.h
devshgraphicsprogramming: yeah, reading the code it looks like you don't need to know, will resolve after I read everything

include/nbl/asset/IDescriptorSet.h
devshgraphicsprogramming: yeah it seems like `info.sampler` should be expected to be `nullptr` when the `desc` is `SAMPLER`

and nobody should ever be providing a non-null `info.sampler` for `COMBINED` binding with immutable sampler, or any sbindinginfo at all for a sampler binding that has an immutable sampler

You can also rename the `samplers` to `immutableSamplers` here for clarity https://github.com/Devsh-Graphics-Programming/Nabla/blob/3df38a45978510f9823c97adfbc3bea4c90e8aee/include/nbl/asset/IDescriptorSetLayout.h#L62

devshgraphicsprogramming: > Okay I think I can have both: I give `SDescriptorInfo` some sort of `getSampler()` method that I can use throughout the rest of the code so I don't have edge cases, and also force SAMPLER to have its `info.sampler` be nullptr or at least just ignore that field

> So now, a question about write/copy validation: I expect to only be using desc for descriptors of type SAMPLER, and ignore whatever is given under the info field. Should I, however, do anything about the info field? Like do I force it to be nullptr and drop a smart pointer there, or is that a problem of the user?

its a union with a buffer, its just sanest to ignore them like we do for buffers.

For semantical clarity you can define everything as:
```cpp
struct SDescriptorInfo
{
  struct SBufferInfo
  {
    size_t offset;
    size_t size;//in Vulkan it's called `range` but IMO it's misleading so i changed to `size`
    
    auto operator<=>(const SBufferInfo&) const = default;
  };
  struct SImageInfo
  {
    IImage::LAYOUT imageLayout;
  };
  struct SCombinedImageSamplerInfo : SImageInfo
  {
    core::smart_refctd_ptr<typename layout_t::sampler_type> sampler;
  };

  core::smart_refctd_ptr<IDescriptor> desc;
  union SBufferImageInfo
  {
    SBufferImageInfo()
    {
      memset(&buffer, 0, core::max<size_t>(sizeof(buffer), sizeof(combinedImageSampler)));
    };

    ~SBufferImageInfo() {};

    SBufferInfo buffer;
    SImageInfo image;
    SCombinedImageSamplerInfo combinedImageSampler;
  } info;
};
```

Fletterio: This would break a lot of existing code potentially, is it okay to do this?

devshgraphicsprogramming: > This would break a lot of existing code potentially, is it okay to do this?

how would it break? its still a union and bytewise the members are still in the same place.

only `image` turns into `combinedImageSampler` but thats an easy `Ctrl+H` job

.gitmodules
```diff
@@ -70,6 +70,7 @@
 [submodule "examples_tests"]
 	path = examples_tests
 	url = git@github.com:Devsh-Graphics-Programming/Nabla-Examples-and-Tests.git
+	branch = master
```
devshgraphicsprogramming, line 73: don't want this in `gitmodules`

include/nbl/video/IGPUDescriptorSet.h
```diff
@@ -70,6 +70,20 @@ class IGPUDescriptorSet : public asset::IDescriptorSet<const IGPUDescriptorSetLa
             return asset::IDescriptor::E_TYPE::ET_COUNT;
         }
 
+        // Same as above, but also retrieving the binding's storage index for its corresponsing redirect
+        inline asset::IDescriptor::E_TYPE getBindingType(const uint32_t binding, uint32_t& bindingStorageIndex) const
```
devshgraphicsprogramming, line 74: then implement `getBindingType(const uint32_t binding)` as 
```cpp
inline asset::IDescriptor::E_TYPE getBindingType(const uint32_t binding) const
{
   uint32_t dummy;
   return getBindingType(binding,dummy);
}
```

include/nbl/video/IGPUDescriptorSet.h
```diff
@@ -99,6 +113,20 @@ class IGPUDescriptorSet : public asset::IDescriptorSet<const IGPUDescriptorSetLa
             return descriptors+localOffset;
         }
 
+        // Same as above, but amortizes lookup if you already have an index
+        inline core::smart_refctd_ptr<asset::IDescriptor>* getDescriptorsIndexed(const asset::IDescriptor::E_TYPE type, const uint32_t bindingStorageIndex) const
```
devshgraphicsprogramming, line 117: we made the `storage_range_index_t` for a reason, so you couldn't shove a random integer as argument and loose the semantics

note that `binding_number_t` and `storage_range_index_t` both contain a `uint32_t` byt have vastly different meanings.

btw why does the above not use `binding_number_t` and takes `uint32_t` as argument?

devshgraphicsprogramming: also why not implement the above in terms of this one

include/nbl/video/IGPUDescriptorSet.h
```diff
@@ -112,6 +140,20 @@ class IGPUDescriptorSet : public asset::IDescriptorSet<const IGPUDescriptorSetLa
...
+            const auto localOffset = getLayout()->getMutableCombinedSamplerRedirect().getStorageOffset(redirect_t::storage_range_index_t{ bindingStorageIndex }).data;
+            if (localOffset == getLayout()->getMutableCombinedSamplerRedirect().Invalid)
+                return nullptr;
+
+            auto* samplers = getAllMutableCombinedSamplers();
+            if (!samplers)
+                return nullptr;
+
+            return samplers + localOffset;
+        }
```
devshgraphicsprogramming, line 155: comments same as for https://github.com/Devsh-Graphics-Programming/Nabla/pull/693/commits/8aef486c514ea26f35361afcf74fa153c1320da0#r1659354882

include/nbl/asset/ICPUDescriptorSet.h
devshgraphicsprogramming: todo

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -45,9 +45,10 @@ asset::IDescriptor::E_TYPE IGPUDescriptorSet::validateWrite(const IGPUDescriptor
     const char* debugName = getDebugName();
 
     // screw it, we'll need to replace the descriptor writing with update templates of descriptor buffer soon anyway
-    const auto descriptorType = getBindingType(write.binding);
+    uint32_t descriptorRedirectBindingIndex;
```
devshgraphicsprogramming, line 48: its super important to use typed indices, instead of an unsafe `uint32_t` type
https://github.com/Devsh-Graphics-Programming/Nabla/pull/693#discussion_r1659353745

include/nbl/video/IGPUDescriptorSet.h
devshgraphicsprogramming: btw using typed arguments, would allow you to use the same method name

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -215,34 +210,38 @@ void IGPUDescriptorSet::processCopy(const IGPUDescriptorSet::SCopyDescriptorSet&
...
     for (uint32_t t = 0; t < static_cast<uint32_t>(asset::IDescriptor::E_TYPE::ET_COUNT); ++t)
     {
         const auto type = static_cast<asset::IDescriptor::E_TYPE>(t);
+        auto& bindingRedirect = m_layout->getDescriptorRedirect(type);
 
         auto* srcDescriptors = copy.srcSet->getDescriptors(type, copy.srcBinding);
-        auto* dstDescriptors = copy.dstSet->getDescriptors(type, copy.dstBinding);
-        assert(!(!srcDescriptors != !dstDescriptors));
+        // We can do this because dstSet === this as asserted at the start
+        auto descriptorRedirectBindingIndex = bindingRedirect.findBindingStorageIndex({ copy.dstBinding }).data;
```
devshgraphicsprogramming, line 220: I'd add `dst` to the `descriptorRedirectBindingIndex` variable name

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -681,6 +681,16 @@ void CVulkanLogicalDevice::updateDescriptorSets_impl(const SUpdateDescriptorSets
...
             switch (asset::IDescriptor::GetTypeCategory(type))
             {
+                case asset::IDescriptor::EC_SAMPLER:
+                {
+                    outWrite->pImageInfo = outImageInfo;
+                    for (auto j = 0u; j < write.count; j++, outImageInfo++)
+                    {
+                        const auto& imageInfo = infos[j].info.image;
+                        // Write was validated so we can just get the sampler
+                        outImageInfo->sampler = static_cast<const CVulkanSampler*>(imageInfo.sampler.get())->getInternalObject();
```
devshgraphicsprogramming, line 691: you need to get the `desc` as per https://github.com/Devsh-Graphics-Programming/Nabla/pull/693#discussion_r1648903341

devshgraphicsprogramming:
btw your dxc submodule is not inline with the one on `master`, please fix so there's no difference

include/nbl/asset/IDescriptorSet.h
```diff
@@ -93,34 +95,32 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
 				}
 				~SDescriptorInfo()
 				{
-					if (desc and (desc->getTypeCategory()==IDescriptor::EC_IMAGE or desc->getTypeCategory() == IDescriptor::EC_SAMPLER))
-						info.image.sampler = nullptr;
```
devshgraphicsprogramming: you just broke my code, look at what the original did

include/nbl/asset/IDescriptorSet.h
```diff
@@ -93,34 +95,32 @@ class IDescriptorSet : public virtual core::IReferenceCounted // TODO: try to re
...
 				}
 
 				inline SDescriptorInfo& operator=(const SDescriptorInfo& other)
 				{
-					if (desc and (desc->getTypeCategory()==IDescriptor::EC_IMAGE or desc->getTypeCategory() == IDescriptor::EC_SAMPLER))
-						info.image.sampler = nullptr;
+					if (desc and desc->getTypeCategory()==IDescriptor::EC_IMAGE)
+						info.combinedImageSampler.sampler = nullptr;
 					desc = other.desc;
 					const auto type = desc->getTypeCategory();
```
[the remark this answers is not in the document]
devshgraphicsprogramming: see the `&&` operator handling of nullptr new desc

include/nbl/asset/IDescriptorSet.h
Fletterio: Since the `SDescriptorInfo` owns the `SBufferImageInfo info` struct I thought this would happen regardless on destruction. Rolling it back

devshgraphicsprogramming: no because its in a union, so you loose any implicit construction/destruction of a member!

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -61,49 +71,52 @@ asset::IDescriptor::E_TYPE IGPUDescriptorSet::validateWrite(const IGPUDescriptor
         return asset::IDescriptor::E_TYPE::ET_COUNT;
     }
 
-    core::smart_refctd_ptr<video::IGPUSampler>* mutableSamplers = nullptr;
+    // Possible TODO: ensure the types are the same, not just categories! Requires IDescriptor to provide a virtual getType() method
```
devshgraphicsprogramming, line 74: remove nonsensical comment

PR #587, opened 2023-10-30
Nahim devicefeatures

devshgraphicsprogramming:

PR #698, opened 2024-06-12
FFT Workgroup Ops

Fletterio:
## Description
Adds the option of doing Forward/Inverse FFT on Workgroup-Sized complex data

## Testing 
On the `nahim_complex_tests` branch, the example 05 is a bit hacked into shape and the tests were run there. Tested that FFT(IFFT(x)) = x = IFFT(FFT(x))

## TODO list:
`nahim_complex` branch should be merged to master before this one


include/nbl/builtin/hlsl/bit.hlsl
```diff
@@ -94,14 +94,30 @@ uint16_t clz(uint64_t N)
...
+    return (hi<<count)|shifted_masked_value|lo;
+}
+
+} //namespace impl
+
+uint32_t bitfieldInsert(uint32_t base, uint32_t value, uint32_t offset, uint32_t count)
+{
+    const uint32_t shifted_masked_value = (value&((1u<<count)-1))<<offset;
+    return impl::bitfieldInsert(base,shifted_masked_value,base&((1u<<offset)-1),count);
+}
```
devshgraphicsprogramming, line 119: these are GLSL equivalents, should go into the `glsl` namespace and `glsl_compat` header

btw does GLSL/HLSL/C++ guarantee to wipe high bits of your `value`? If not we could assert that `value>>count == 0` and not bother with the first masking before `<<offset`

P.S. we need to template this stuff
```cpp
template<Unsigned>
enable_if_t<is_unsigned_v<Unsigned>,Unsigned> bitfieldInsert(Unsigned base, Unsigned value, uint16_t offset, uint16_t count)
```

devshgraphicsprogramming: Also because its broken into two functions, its kinda hard to read.

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -32,6 +32,7 @@ template<typename Scalar, uint32_t SubgroupSize, bool inverse>
 void FFT(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi)
```
devshgraphicsprogramming, line 32: you shouldn't template on the subgroup size, just use `glsl::gl_SubgroupSize()` for your loops

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -32,6 +32,7 @@ template<typename Scalar, uint32_t SubgroupSize, bool inverse>
 void FFT(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi) 
 {
     const uint32_t doubleSubgroupSize = SubgroupSize << 1;
+    
     // special first iteration
     if (! inverse)
         fft::DIF<Scalar>::radix2(fft::twiddle<Scalar, false>(glsl::gl_SubgroupID(), doubleSubgroupSize), lo, hi);
```
devshgraphicsprogramming, line 38: gl_SiubgroupID is not the ordinal of the invocation in the subgroup, but the index of the whole subgroup, did you mean to do that?

devshgraphicsprogramming: P.S. one more thing about the code later on, you should divide `hi` by scalar instead of dividing by complex, it will be faster (just how you divide `lo`)

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    sharedmemAccessor.set(threadID + 3 * WorkgroupSize, hi.imag());    
+}
+
+template<typename Accessor, typename Scalar, uint32_t WorkgroupSize>
+void load(NBL_CONST_REF_ARG(Accessor) sharedmemAccessor, uint32_t threadID, NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi) {
+    lo.real(sharedmemAccessor.get(threadID + 0 * WorkgroupSize));
+    hi.real(sharedmemAccessor.get(threadID + 1 * WorkgroupSize));
+    lo.imag(sharedmemAccessor.get(threadID + 2 * WorkgroupSize));  
+    hi.imag(sharedmemAccessor.get(threadID + 3 * WorkgroupSize));    
+}
```
devshgraphicsprogramming, line 28: Use this little baby:
https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/hlsl/memory_accessor.hlsl

Shove the original Accessor inside and you're done.

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    sharedmemAccessor.set(lo_x_ix, lo.real());
+    sharedmemAccessor.set(hi_x_ix, hi.real());
+    sharedmemAccessor.set(lo_y_ix, lo.imag());
+    sharedmemAccessor.set(hi_y_ix, hi.imag());     
+}
+
+// When doing a 1D FFT of size N, we call N/2 threads to do the work, where each thread of index 0 <= k < N/2 is in charge of computing two elements
+// in the output array, those of index k and k + N/2. The first is "lo" and the latter is "hi"
+template<typename Accessor, typename Scalar, uint32_t WorkgroupSize, uint32_t SubgroupSize, bool inverse>
+void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi){
```
devshgraphicsprogramming, line 55: instead of taking `lo` and `hi` take a `preloaded[K]` array where `K=N/WorkgroupSize` and must be PoT (for now, we can experiment later with specializations mixing radices to get K=3,5,6,7)

devshgraphicsprogramming: you also can't template on the SubgroupSize :(

devshgraphicsprogramming: also you need two accessors:
- SharedMemoryAccessor (for workgroup FFT)
- Accessor (for getting/setting outputs)

Both you shall wrap in  `MemoryAdaptor` to turn `uint32_t` gets/sets into `vector<Scalar>` sets and gets

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    const uint32_t doubleWorkgroupSize = WorkgroupSize << 1;
+    const uint32_t threadID = glsl::gl_LocalInvocationIndex();
```
devshgraphicsprogramming, line 57: Neither the SPIR-V or Vulkan spec say that `LocalInvocationIndex` bijects with `(SubgroupID,SubgroupInvocationIndex)` in any specific way

Therefore you need to use (`workgroup::SubgroupContiguousIndex`)[https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/hlsl/workgroup/basic.hlsl#L26] instead of `glsl::gl_LocalInvocationIndex` to ensure the packing you rely on here.

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: Also having free floating functions just called `store` and `load` is super meh

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+        store<Accessor, Scalar, WorkgroupSize>(sharedmemAccessor, threadID, lo, hi);             
+
+        // The bigger steps
+        for (uint32_t step = SubgroupSize << 1; step < WorkgroupSize; step <<= 1)
+            FFT_loop<Accessor, Scalar, WorkgroupSize, true>(sharedmemAccessor, step, threadID);
+
+        // Get values back from smem
+        sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
+        load<Accessor, Scalar, WorkgroupSize>(sharedmemAccessor, threadID, lo, hi); 
+        sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 75: you'll need to wrap this in a loop of `for (uint32_t k=0; k<K; k+=2)` w.r.t. https://github.com/Devsh-Graphics-Programming/Nabla/pull/698/files#r1641173444

Good news is that each Workgroup-sized FFT can run independently of each other :)
So no extra smem needed, just some registers/temporaries to store them.

devshgraphicsprogramming: btw right now you don't need the first and last `workgroupExecutionAndMemoryBarrier` because we always assume the accessors is un-aliased and fresh.


When you move to doing multiple workgroup FFTs after each other, then you only need one barrier, and it should be at the top, guarded by `if (k!=0)`

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    hi.imag(sharedmemAccessor.get(threadID + 3 * WorkgroupSize));    
+}
+
+template<typename Accessor, typename Scalar, uint32_t WorkgroupSize, bool inverse>
+void FFT_loop(NBL_REF_ARG(Accessor) sharedmemAccessor, uint32_t step, uint32_t threadID){
+    const uint32_t sub_ix = threadID & (step - 1);
+    const uint32_t lo_x_ix = impl::bitfieldInsert(threadID, 0u ,sub_ix, 1u);
+    const uint32_t hi_x_ix = lo_x_ix | step;
+    const uint32_t lo_y_ix = lo_x_ix + 2 * WorkgroupSize;
+    const uint32_t hi_y_ix = hi_x_ix + 2 * WorkgroupSize;
```
devshgraphicsprogramming, line 36: do `<<1` instead of `2*` and `|` instead of `+`

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    lo.real(sharedmemAccessor.get(threadID + 0 * WorkgroupSize));
+    hi.real(sharedmemAccessor.get(threadID + 1 * WorkgroupSize));
+    lo.imag(sharedmemAccessor.get(threadID + 2 * WorkgroupSize));  
+    hi.imag(sharedmemAccessor.get(threadID + 3 * WorkgroupSize));    
+}
+
+template<typename Accessor, typename Scalar, uint32_t WorkgroupSize, bool inverse>
+void FFT_loop(NBL_REF_ARG(Accessor) sharedmemAccessor, uint32_t step, uint32_t threadID){
+    const uint32_t sub_ix = threadID & (step - 1);
+    const uint32_t lo_x_ix = impl::bitfieldInsert(threadID, 0u ,sub_ix, 1u);
```
devshgraphicsprogramming, line 33: don't use `impl::bitfieldInsert` this is super hard to read.

Also needs comments, I remember whats supposed to happen, but nobody else will

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    {
+        // special first iteration
+        fft::DIF<Scalar>::radix2(fft::twiddle<Scalar, false>(threadID, doubleWorkgroupSize), lo, hi); 
+
+        // Put values into workgroup shared smem
+        sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
+        store<Accessor, Scalar, WorkgroupSize>(sharedmemAccessor, threadID, lo, hi);
+        
+        // Run bigger steps until Subgroup-sized
+        for (uint32_t step = WorkgroupSize >> 1; step > SubgroupSize; step >>= 1)
```
devshgraphicsprogramming, line 93: how does it make sense to start at `step=WorkgroupSize*2` ?

That step is makes you step over the whole  array

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+        // special last iteration
+        fft::DIT<Scalar>::radix2(fft::twiddle<Scalar, true>(threadID, doubleWorkgroupSize), lo, hi); 
+        divides_assign< complex_t<Scalar> > divAss;
+        divAss(lo, WorkgroupSize / SubgroupSize);
+        divAss(hi, WorkgroupSize / SubgroupSize);
+    }
+    else
+    {
+        // special first iteration
+        fft::DIF<Scalar>::radix2(fft::twiddle<Scalar, false>(threadID, doubleWorkgroupSize), lo, hi);
```
devshgraphicsprogramming, line 86: this special iteration only makes sense when `K>=4`, cause your workgroup FFT already takes `2*WorkgroupSize` inputs

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: I'd do this
```cpp
template<uint16_t K, bool Inverse, class device_capabilities=void>
struct FFT;

// 2 items per invocation forward specialization
template<class device_capabilities>
struct FFT<2,false,device_capabilities>
{
    template<typename Accessor, typename ShaderMemoryAccessor>
    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(ShaderMemoryAccessor) sharedmemAccessor);
};
// 2 items per invocation inverse specialization
template<class device_capabilities>
struct FFT<2,true,device_capabilities>
{
    template<typename Accessor, typename ShaderMemoryAccessor>
    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(ShaderMemoryAccessor) sharedmemAccessor);
};

// then define 4,8,16 in terms of calling the FFT<2> and doing the special radix steps before/after
template<uint16_t K, bool Inverse, class device_capabilities>
struct FFT
{
    template<typename Accessor, typename ShaderMemoryAccessor>
    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(ShaderMemoryAccessor) sharedmemAccessor)
    {
        if (!Inverse)
        {
           ... special steps ...
        }
        FFT<2,Inverse,device_capabilities>::template __call<Accessor,SharedMemoryAccessor>(access,sharedMemAccessor);
        if (Inverse)
        {
           ... special steps ...
        }
    }
};
```

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    complex_t<Scalar> hi = {sharedmemAccessor.get(hi_x_ix), sharedmemAccessor.get(hi_y_ix)};
+
+    fft::DIX<Scalar, inverse>::radix2(fft::twiddle<Scalar, inverse>(sub_ix, step << 1), lo, hi);
+
+    // Share results between threads
+    sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
+    sharedmemAccessor.set(lo_x_ix, lo.real());
+    sharedmemAccessor.set(hi_x_ix, hi.real());
+    sharedmemAccessor.set(lo_y_ix, lo.imag());
+    sharedmemAccessor.set(hi_y_ix, hi.imag());
```
devshgraphicsprogramming, line 49: I think there's a way to optimize the shared memory traffic, not only for less load/store, but less smem usage overall.

Lets assume SubgroupSize 32 and Workgroup Size 1024

## When you do a forward FFT of size N=WorkgroupSize*2

Your first step has stride WorkgroupSize and requires no smem transfer.

### Step WorkgroupSize/2

But the second step steps with `WorkgroupSize/2` meaning half of the subgroups load whole workgroup.

Subgroups [0,16) don't actually need to trade their `lo` items, they can stay in registers. You'd be only storing them to load them again. They should only store their `hi` items from previous step.

Subgroups [16,32) don't need to trade their `hi` items, they should only store their `lo` items from previous step.

### Step WorkgroupSize/4

Subgroups [0,8) and [16,24) can stay where they are, and keep their `lo` items, while subgroups [8,16) and [24,32) "move" but can keep their `hi` items.

...

I think you can start to see the pattern emerging.

Bottom "half" of an iteration step bundle stores `hi` to memory, while Top stores `lo`.

Then the Top half gets reassigned.

include/nbl/builtin/hlsl/subgroup/fft.hlsl
devshgraphicsprogramming: you need to change the template signature to
```cpp
template<bool Inverse, typename Scalar>
```
the reasons is you need to put non-deductible template parameters first (otherwise I need to type out the `Scalar` type. Also the subgroup size as above, can't be a constexpr.

Furthermore I'd experiment with rolling up the FFT and FFT_loop into a single struct with an `operator()` to keep convention with the subgroup arithmetic operations

include/nbl/builtin/hlsl/subgroup/fft.hlsl
Fletterio: whoops, I thought I had changed that back, I was testing that the operator overloading was working properly there.

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -192,6 +192,38 @@ T bitfieldExtract( T val, uint32_t offsetBits, uint32_t numBits )
...
+    const Unsigned allOnes = ~ Unsigned(0);
+    // Produce a mask with `count` ones starting at the `offset` bit (counting from LSB), everything else is 0
+    Unsigned mask = ~(allOnes << count) << offset;
+    // Offset `value` by `offset`, then `&` with the mask. This puts the value's first `count` LSB starting at the offset, everything else is 0
+    value = (value << offset) & mask 
+    // Next invert the mask so all those positions are 0, and `&` the base with the new mask to set those bits in the base to 0
+    base &= ~mask;
+    // Finally, insert the bits
+    return base | value;
+}
```
devshgraphicsprogramming, line 206: look at the `bitfieldExtract` its a SPIR-V intrinsic, implement the same way

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -12,94 +12,6 @@ namespace hlsl
...
-    fft::DIT<Scalar>::radix2(fft::twiddle<Scalar, true>(glsl::gl_SubgroupInvocationID(), doubleSubgroupSize), lo, hi);
-    divides_assign< complex_t<Scalar> > divAss;
-    divAss(lo, doubleSubgroupSize);
-    divAss(hi, doubleSubgroupSize);            
-}   
-
-*/
-
 // -----------------------------------------------------------------------------------------------------------------------------------------------------------------
 template<uint16_t K, bool Inverse, typename Scalar, class device_capabilities=void>
```
devshgraphicsprogramming, line 16: don't call it `K` call it `Radix` cause otherwise people can get confused what the `K` is

devshgraphicsprogramming: also move it to AFTER `device_capabilities`, and default it to `2`

Otherwise user has to specify its `radix=2` every time, which they probably don't want (only inverse vs. forward)

devshgraphicsprogramming: btw you don't use the `device_capabilities` at all, so whats the point of taking them?

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -12,49 +12,94 @@ namespace hlsl
...
-template<typename Scalar, uint32_t SubgroupSize, bool inverse>
-void FFT(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi) 
+template<typename Scalar, class device_capabilities>
+struct FFT<2, false, Scalar, device_capabilities>
 {
-    const uint32_t doubleSubgroupSize = SubgroupSize << 1;
+    static void FFT_loop(uint32_t stride, NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi)
+    {
+        const bool topHalf = (glsl::gl_SubgroupInvocationID() & stride) != 0;
+        const vector <Scalar, 2> toTrade = topHalf ? vector <Scalar, 2>(lo.real(), lo.imag()) : vector <Scalar, 2>(hi.real(), hi.imag());
```
devshgraphicsprogramming, line 30: btw `?` should work on regular structs, packing to/from `vector<Scalar,2>` to use in `subgroupShuffleXor` should happen after the ternary for readability

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -12,49 +12,94 @@ namespace hlsl
...
+    static void __call(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi) 
+    {
+        const uint32_t subgroupSize = glsl::gl_SubgroupSize();  //This is N/2
+        const uint32_t doubleSubgroupSize = subgroupSize << 1;  //This is N
     
-    // special first iteration
-    if (! inverse)
-        fft::DIF<Scalar>::radix2(fft::twiddle<Scalar, false>(glsl::gl_SubgroupID(), doubleSubgroupSize), lo, hi);                                                                                   
+        // special first iteration
+        fft::DIF<Scalar>::radix2(fft::twiddle<false, Scalar>(glsl::gl_SubgroupInvocationID(), doubleSubgroupSize), lo, hi);
```
devshgraphicsprogramming, line 52: why don't you define the `radix2` so it asks for `halfN` instead of `N` ? then we can skip a `<<1` at every iteration just to get your `N`

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -12,49 +12,94 @@ namespace hlsl
...
+template<typename Scalar, class device_capabilities>
+struct FFT<2, true, Scalar, device_capabilities>
+{
+    static void FFT_loop(uint32_t stride, NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi)
+    {
+        // Get twiddle with k = subgroupInvocation mod stride, N = 2 * stride
+        fft::DIT<Scalar>::radix2(fft::twiddle<true, Scalar>(glsl::gl_SubgroupInvocationID() & (stride - 1), stride << 1), lo, hi);   
+
+        const bool topHalf = (glsl::gl_SubgroupInvocationID() & stride) != 0;
+        const vector <Scalar, 2> toTrade = topHalf ? vector <Scalar, 2>(lo.real(), lo.imag()) : vector <Scalar, 2>(hi.real(), hi.imag());
```
devshgraphicsprogramming, line 72: same as https://github.com/Devsh-Graphics-Programming/Nabla/pull/698/files/86f33e80cf9ed8c4e3ce19fd9e310c8112af5d06..37df1798e5fc929c0cc5880c5fb6eab7738b0605#r1648756422

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
     }
 }
 
+*/
+
+// ----------------------------------------- ABOVE MARKED FOR DELETION -----------------------------------------------------
+
+template<uint16_t K, bool Inverse, typneame Scalar, class device_capabilities=void>
+struct FFT;
```
devshgraphicsprogramming, line 115: Move `K` to AFTER device_capabilities, and default it to 2

Otherwise user has to specify its radix=2 every time, which they probably don't want (only inverse vs. forward)

Also don't call it `K` call it `Radix` cause otherwise people can get confused what the `K` is

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
...
+
+// For the FFT methods below, we assume:
+//      - Accessor is a global memory accessor to an array fitting 2 * _NBL_HLSL_WORKGROUP_SIZE_ elements of type complex_t<Scalar>, used to get inputs / set outputs of the FFT
```
devshgraphicsprogramming, line 118: you only need `NBL_HLSL_WORKGROUP_SIZE` complex elements
cause you only trade half of your entries at every step

include/nbl/builtin/hlsl/subgroup/fft.hlsl
devshgraphicsprogramming: ah ok you can use `device_capabilities` after #587 gets merged to `[unroll(device_traits<device_capabilities>::subgroupMinSize)]` your for loops

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
...
+
+
+    template<typename Accessor, typename SharedMemoryAccessor>
+    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
+    {
+        // Set up the MemAdaptors
+        MemoryAdaptor<Accessor> memAdaptor;
+        memAdaptor.accessor = accessor;
+        MemoryAdaptor<SharedMemoryAccessor> sharedmemAdaptor;
+        sharedmemAdaptor.accessor = sharedmemAccessor;
```
devshgraphicsprogramming, line 161: because of the statefullness, you need to move them back before you return from the function
```
accessor = memAdaptor.accessor;
sharedmemAccessor = sharedmemAdaptor.accessor;
```

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
...
```
devshgraphicsprogramming, line 231: https://github.com/Devsh-Graphics-Programming/Nabla/pull/698/files/86f33e80cf9ed8c4e3ce19fd9e310c8112af5d06..37df1798e5fc929c0cc5880c5fb6eab7738b0605#r1648812713

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
...
+        const uint32_t threadID = uint32_t(SubgroupContiguousIndex());
+
+        // Read lo, hi values from global memory
+        vector <Scalar, 4> loHiPacked;
+        memAdaptor.get(threadID, loHiPacked);
+        complex_t<Scalar> lo = {loHiPacked.x, loHiPacked.y};  
+        complex_t<Scalar> hi = {loHiPacked.z, loHiPacked.w};
+
+        // special first iteration
+        fft::DIF<Scalar>::radix2(fft::twiddle<false, Scalar>(threadID, _NBL_HLSL_WORKGROUP_SIZE_ << 1), lo, hi);
```
devshgraphicsprogramming, line 173: you need to predicate this on `_NBL_HLSL_WORKGROUP_SIZE>SubgroupSize` otherwise you end up doing the special step twice when your workgroup-size == subgroup size

(you don't need to worry about WorkgroupSize < SubgroupSize because that goes against `comptueFullSubgroups` pipeline creation option that you'll require)

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
...
+        // Run bigger steps until Subgroup-sized
+        for (uint32_t stride = _NBL_HLSL_WORKGROUP_SIZE_ >> 1; stride > glsl::gl_SubgroupSize(); stride >>= 1)
+            FFT_loop<SharedMemoryAccessor>(stride, lo, hi, threadID, sharedmemAdaptor);
+  
+        // Subgroup-sized FFT
+        subgroup::FFT<2, false, Scalar, device_capabilities>::__call(lo, hi);
+
+        // Put values back in global mem
+        loHiPacked = vector <Scalar, 4>(lo.real(), lo.imag(), hi.real(), hi.imag());
+        memAdaptor.set(threadID, loHiPacked);
```
devshgraphicsprogramming, line 184: again, store `lo` and `hi` separately

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
...
+
+        // special last iteration
+        fft::DIT<Scalar>::radix2(fft::twiddle<Scalar, true>(threadID, _NBL_HLSL_WORKGROUP_SIZE_ << 1), lo, hi); 
+        divides_assign< complex_t<Scalar> > divAss;
+        divAss(lo, _NBL_HLSL_WORKGROUP_SIZE_ / glsl::gl_SubgroupSize());
+        divAss(hi, _NBL_HLSL_WORKGROUP_SIZE_ / glsl::gl_SubgroupSize());
+        
+        // Put values back in global mem
+        loHiPacked = vector <Scalar, 4>(lo.real(), lo.imag(), hi.real(), hi.imag());
+        memAdaptor.set(threadID, loHiPacked);
```
devshgraphicsprogramming, line 257: my comments are the same as for the forward FFT function

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
...
+
+// 2 items per invocation forward specialization
+template<typename Scalar, class device_capabilities>
+struct FFT<2,false,device_capabilities>
+{
+    template<typename SharedMemoryAccessor>
+    static void FFT_loop(uint32_t stride, NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi, uint32_t threadID, NBL_REF_ARG(MemoryAdaptor<SharedMemoryAccessor>) sharedmemAdaptor)
+    {
+        const bool topHalf = (threadID & stride) != 0;
+        vector <Scalar, 2> toTrade = topHalf ? vector <Scalar, 2>(lo.real(), lo.imag()) : vector <Scalar, 2>(hi.real(), hi.imag());
```
devshgraphicsprogramming, line 129: I think you can just `bool(threadId&stride)` instead of comparing against 0

Also again, the ternary operator probably works on structs IIRC

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,6 +107,197 @@ void FFT(NBL_REF_ARG(Accessor) sharedmemAccessor, NBL_REF_ARG(complex_t<Scalar>)
...
+        
+        // Block memory writes until all threads are done with previous work
+        sharedmemAdaptor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 132: you don't need the barrier on the first call to `FFT_loop` because there's no previous work (that you know about)

We have a convention in our HLSL where we assume that "accessor is always clean and doesn't alias" and if it does get reused/aliased its the responsability of the caller to ensure proper barriers are issued.

So in like with this convention you can "hoist" this barrier out of `FFT_loop` and into the `for` loop where you can check for the first iteration being performed

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: update the comment

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -10,6 +10,17 @@ namespace nbl
...
+namespace glsl 
+{
+
+// Define this method from glsl_compat/core.hlsl 
+uint32_t3 gl_WorkGroupSize() {
+    return uint32_t3(_NBL_HLSL_WORKGROUP_SIZE_, 1, 1);
+}
+
+} //namespace glsl
+
```
devshgraphicsprogramming, line 24: noooo, this is ALWAYS user-space

devshgraphicsprogramming: think what happens if every workgroup header tried to define this!?

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -57,7 +68,7 @@ struct FFT<2,false, Scalar, device_capabilities>
     static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
     {
         // Set up the MemAdaptors
-        MemoryAdaptor<Accessor> memAdaptor;
+        MemoryAdaptor<Accessor, _NBL_HLSL_WORKGROUP_SIZE_ << 1> memAdaptor;
```
devshgraphicsprogramming, line 78: why are you using a big stride for gmem inputs/outputs?

include/nbl/builtin/hlsl/subgroup/fft.hlsl
devshgraphicsprogramming: does it not work on structs or what?

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -24,12 +25,37 @@ uint32_t3 gl_WorkGroupSize() {
 namespace workgroup
 {
 
+// ---------------------------------- Utils -----------------------------------------------
+
+template<typename SharedMemoryAccessor, typename Scalar>
+void exchangeValues(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi, uint32_t threadID, uint32_t stride, NBL_REF_ARG(MemoryAdaptor<SharedMemoryAccessor>) sharedmemAdaptor)
```
devshgraphicsprogramming, line 31: namespace/scope it better, right now its hanging around in `workgroup` namespace

include/nbl/builtin/hlsl/workgroup/shuffle.hlsl
```diff
@@ -0,0 +1,40 @@
+#ifndef _NBL_BUILTIN_HLSL_WORKGROUP_SHUFFLE_INCLUDED_
+#define _NBL_BUILTIN_HLSL_WORKGROUP_SHUFFLE_INCLUDED_
+
+#include "nbl/builtin/hlsl/memory_accessor.hlsl"
+
+// TODO: Add other shuffles
+// TODO: Consider adding an enable_if or static assert that 1 <= N <= 4 and that Scalar is a proper scalar type
+// TODO: Consider adding version that doesn't take a precomputed threadID and instead calls workgroup::SubgroupContiguousIndex
+
+// Unlike subgroups we pass a precomputed threadID so we don't go around recomputing it every time
```
devshgraphicsprogramming, line 10: make it consistent with other workgroup ops, and a struct

common subexpression elimination (CSE) during spirv-opt should take care of duplicate `threadID` computation, provide a variant that does not ask for `threadID` to be consistent with other workgroup ops (ballot, arithmetic, etc)

IMO you could implement it in terms of a shuffle

include/nbl/builtin/hlsl/workgroup/shuffle.hlsl
```diff
@@ -0,0 +1,40 @@
...
+namespace nbl 
+{
+namespace hlsl
+{
+namespace workgroup
+{
+
+
+template<typename SharedMemoryAccessor, typename Scalar, uint32_t N = 1>
+void shuffleXor(NBL_REF_ARG(vector <Scalar, N>) value, uint32_t mask, uint32_t threadID, NBL_REF_ARG(MemoryAdaptor<SharedMemoryAccessor>) sharedmemAdaptor)
```
devshgraphicsprogramming, line 22: drop the `Scalar` and `N` just make it generic on `T

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -166,12 +167,22 @@ struct FFT<2,true, Scalar, device_capabilities>
...
+        {   
+            // If we enter this for loop, then the special first workgroup shuffle went through, so wait on that
+            sharedmemAdaptor.workgroupExecutionAndMemoryBarrier(); 
             FFT_loop<SharedMemoryAccessor>(stride, lo, hi, threadID, sharedmemAdaptor);
+        }
 
-        // special last iteration
-        if (_NBL_HLSL_WORKGROUP_SIZE_ > glsl::gl_SubgroupSize())
+        // special last iteration - only if workgroupsize > subgroupsize
+        if (_NBL_HLSL_WORKGROUP_SIZE_ ^ glsl::gl_SubgroupSize())
```
devshgraphicsprogramming, line 185: you do realize you can put all this code under a single `NBL_HLSL_WORKGROUP_SIZE > SubgroupSize`  ?

Also whats with the bitwise XOR instead of a `>` ?

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -84,20 +91,34 @@ struct FFT<2,false, Scalar, device_capabilities>
...
+            if ( !(stride & (_NBL_HLSL_WORKGROUP_SIZE_ >> 1)) )
+                sharedmemAdaptor.workgroupExecutionAndMemoryBarrier(); 
             FFT_loop<SharedMemoryAccessor>(stride, lo, hi, threadID, sharedmemAdaptor);
-  
+        }
+
+        // special last workgroup-shuffle - only if workgroupsize > subgroupsize
+        if (_NBL_HLSL_WORKGROUP_SIZE_ ^ glsl::gl_SubgroupSize()) 
+        {
+            // Wait for all threads to be done with reads in the last loop before writing to shared mem
```
devshgraphicsprogramming, line 110: again, same as https://github.com/Devsh-Graphics-Programming/Nabla/pull/698/files/dcd78d64fbdda32f25efeea29d5320d542c34482..74b978af5275b1b6d9b6a4fe9a39c023ebfb1899#r1655570564 can stuff everything into same `if (`

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -84,20 +91,34 @@ struct FFT<2,false, Scalar, device_capabilities>
...
+        // special first iteration - only if workgroupsize > subgroupsize
+        if (_NBL_HLSL_WORKGROUP_SIZE_ ^ glsl::gl_SubgroupSize())
             fft::DIF<Scalar>::radix2(fft::twiddle<false, Scalar>(threadID, _NBL_HLSL_WORKGROUP_SIZE_), lo, hi); 
 
         // Run bigger steps until Subgroup-sized
         for (uint32_t stride = _NBL_HLSL_WORKGROUP_SIZE_ >> 1; stride > glsl::gl_SubgroupSize(); stride >>= 1)
+        {   
+            // If at least one loop was executed, we must wait for all threads to get their values before we write to shared mem again
+            if ( !(stride & (_NBL_HLSL_WORKGROUP_SIZE_ >> 1)) )
+                sharedmemAdaptor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 103: put the barrier AFTER the `FFT_loop` call, you don't need to barrier before because you always start with a "fresh, uncontented" accessor

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -84,20 +91,34 @@ struct FFT<2,false, Scalar, device_capabilities>
...
+                sharedmemAdaptor.workgroupExecutionAndMemoryBarrier(); 
             FFT_loop<SharedMemoryAccessor>(stride, lo, hi, threadID, sharedmemAdaptor);
-  
+        }
+
+        // special last workgroup-shuffle - only if workgroupsize > subgroupsize
+        if (_NBL_HLSL_WORKGROUP_SIZE_ ^ glsl::gl_SubgroupSize()) 
+        {
+            // Wait for all threads to be done with reads in the last loop before writing to shared mem      
+            sharedmemAdaptor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 111: if you barrier after like in https://github.com/Devsh-Graphics-Programming/Nabla/pull/698/files/dcd78d64fbdda32f25efeea29d5320d542c34482..74b978af5275b1b6d9b6a4fe9a39c023ebfb1899#r1655572151

then you don't need a barrier here

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: you branch multiple times on same condition, move all under one if

include/nbl/builtin/hlsl/subgroup/fft.hlsl
```diff
@@ -11,6 +11,8 @@ namespace hlsl
 {
 namespace subgroup
 {
+namespace fft
+{
```
devshgraphicsprogramming, line 15: I did not mean for you to add an `fft` namespace to the `FFT` struct, just the `exchange` function!

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -246,5 +221,6 @@ struct FFT
 }
 }
 }
+}
```
devshgraphicsprogramming, line 224: namespace `fft` was only supposed to be for `exchangeValues`

include/nbl/builtin/hlsl/workgroup/shuffle.hlsl
```diff
@@ -18,8 +18,29 @@ namespace workgroup
...
+    {
+        sharedmemAdaptor.set(threadID, value);
+        
+        // Wait until all writes are done before reading
+        sharedmemAdaptor.workgroupExecutionAndMemoryBarrier();
+    
+        sharedmemAdaptor.get(threadID ^ mask, value);
+    }
+
+    static void __call(NBL_REF_ARG(T) value, uint32_t mask, NBL_REF_ARG(MemoryAdaptor<SharedMemoryAccessor>) sharedmemAdaptor)
```
devshgraphicsprogramming, line 34: you're not supposed to require that the type being passed in is a `MemoryAdaptor<AccessorT>`, keep it as `AccessorT` always

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -11,28 +11,18 @@ namespace nbl
...
+{
 // ---------------------------------- Utils -----------------------------------------------
 
 template<typename SharedMemoryAccessor, typename Scalar>
 void exchangeValues(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi, uint32_t threadID, uint32_t stride, NBL_REF_ARG(MemoryAdaptor<SharedMemoryAccessor>) sharedmemAdaptor)
 {
     const bool topHalf = bool(threadID & stride);
     vector <Scalar, 2> toExchange = topHalf ? vector <Scalar, 2>(lo.real(), lo.imag()) : vector <Scalar, 2>(hi.real(), hi.imag());
-    shuffleXor<SharedMemoryAccessor, Scalar, 2>(toExchange, stride, threadID, sharedmemAdaptor);
+    shuffleXor<SharedMemoryAccessor, vector <Scalar, 2> >::__call(toExchange, stride, sharedmemAdaptor);
```
devshgraphicsprogramming, line 25: P.S. its okay to rely on template argument deduction in this case

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -67,60 +57,59 @@ struct FFT<2,false, Scalar, device_capabilities>
...
+		const uint32_t hiIx = loIx + _NBL_HLSL_WORKGROUP_SIZE_;
 
         // Read lo, hi values from global memory
         vector <Scalar, 2> loVec;
         vector <Scalar, 2> hiVec;
-        memAdaptor.get(threadID, loVec);
-        memAdaptor.get(threadID + _NBL_HLSL_WORKGROUP_SIZE_, hiVec);
+        // TODO: if we get rid of the Memory Adaptor on the accessor and require comples getters and setters, then no `2*`
+        memAdaptor.get(2 * loIx , loVec);
+        memAdaptor.get(2 * hiIx, hiVec);
```
devshgraphicsprogramming, line 83: ok lets do it, rely that `Accessor` has `get(uint32_t index, NBL_REF(complex_t<Scalar>) val)` and `set` methods

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -18,43 +17,41 @@ namespace fft
...
 template<typename SharedMemoryAccessor, typename Scalar>
-void exchangeValues(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi, uint32_t threadID, uint32_t stride, NBL_REF_ARG(MemoryAdaptor<SharedMemoryAccessor>) sharedmemAdaptor)
+void exchangeValues(NBL_REF_ARG(complex_t<Scalar>) lo, NBL_REF_ARG(complex_t<Scalar>) hi, uint32_t threadID, uint32_t stride, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
 {
     const bool topHalf = bool(threadID & stride);
+    // Ternary won't take structs so we use this aux variable
     vector <Scalar, 2> toExchange = topHalf ? vector <Scalar, 2>(lo.real(), lo.imag()) : vector <Scalar, 2>(hi.real(), hi.imag());
-    shuffleXor<SharedMemoryAccessor, vector <Scalar, 2> >::__call(toExchange, stride, sharedmemAdaptor);
+    complex_t<Scalar> toExchangeComplex = {toExchange.x, toExchange.y};
+    shuffleXor<SharedMemoryAccessor, complex_t<Scalar> >::__call(toExchangeComplex, stride, sharedmemAccessor);
```
devshgraphicsprogramming, line 26: but then you don't use the vector at all?

Btw I never knew that subgroup shuffle works on structs, seems fishy (even subgroup extended types only mentions vectors)

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -18,43 +17,41 @@ namespace fft
...
-//        that is, one "lo" and one "hi" complex numbers per thread, essentially 4 Scalars per thread. The data layout is assumed to be a whole array of real parts 
-//        followed by a whole array of imaginary parts. So it would be something like
-//        [x_0, x_1, ..., x_{2 * _NBL_HLSL_WORKGROUP_SIZE_}, y_0, y_1, ..., y_{2 * _NBL_HLSL_WORKGROUP_SIZE_}] 
-//      - SharedMemoryAccessor accesses a shared memory array that can fit _NBL_HLSL_WORKGROUP_SIZE_ elements of type complex_t<Scalar>, so 2 * _NBL_HLSL_WORKGROUP_SIZE_ Scalars 
+//        that is, one "lo" and one "hi" complex numbers per thread, essentially 4 Scalars per thread. 
+//        There are no assumptions on the data layout: we just require the accessor to provide get and set methods for complex_t<Scalar>.
+//      - SharedMemoryAccessor accesses a shared memory array that can fit _NBL_HLSL_WORKGROUP_SIZE_ elements of type complex_t<Scalar>, with get and set 
+//        methods for complex_t<Scalar>. It benefits from coalesced accesses
```
devshgraphicsprogramming, line 45: Shared Memory Accessor is still supposed to only have `uint32_t` methods because SMEM is usually banked in 32 banks

include/nbl/builtin/hlsl/workgroup/shuffle.hlsl
```diff
@@ -17,41 +14,25 @@ namespace hlsl
...
     {
-        sharedmemAdaptor.set(threadID, value);
+        sharedmemAccessor.set(threadID, value);
         
         // Wait until all writes are done before reading
-        sharedmemAdaptor.workgroupExecutionAndMemoryBarrier();
+        sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
     
-        sharedmemAdaptor.get(threadID ^ mask, value);
+        value = sharedmemAccessor.get(threadID ^ mask);
```
devshgraphicsprogramming, line 27: do not change this!!!!

include/nbl/builtin/hlsl/workgroup/shuffle.hlsl
```diff
@@ -17,41 +14,25 @@ namespace hlsl
 namespace workgroup
 {
 
-
 template<typename SharedMemoryAccessor, typename T>
 struct shuffleXor
 {
-    static void __call(NBL_REF_ARG(T) value, uint32_t mask, uint32_t threadID, NBL_REF_ARG(MemoryAdaptor<SharedMemoryAccessor>) sharedmemAdaptor)
+    static void __call(NBL_REF_ARG(T) value, uint32_t mask, uint32_t threadID, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
```
devshgraphicsprogramming, line 20: UNDO

include/nbl/builtin/hlsl/subgroup/fft.hlsl
devshgraphicsprogramming: ok I see that `?` not supported on structs

include/nbl/builtin/hlsl/memory_accessor.hlsl
```diff
@@ -257,6 +257,36 @@ struct MemoryAdaptor<BaseAccessor, 0>
...
+struct OffsetAccessor
+{
+    BaseAccessor accessor;
+
+    void set(uint32_t idx, NBL_REF_ARG(AccessorType) x) {accessor.set(idx + Offset, x);}
+
+    AccessorType get(uint32_t idx) {return accessor.get(idx + Offset);}
+
+    // TODO: figure out the `enable_if` syntax for this
+    void workgroupExecutionAndMemoryBarrier() {accessor.workgroupExecutionAndMemoryBarrier();}
```
devshgraphicsprogramming, line 272: see the member testers Nipun generates in #643 for device traits

include/nbl/builtin/hlsl/memory_accessor.hlsl
```diff
@@ -257,6 +257,36 @@ struct MemoryAdaptor<BaseAccessor, 0>
...
+// ---------------------------------------------- Offset Accessor ----------------------------------------------------
+
+template<class BaseAccessor, class AccessorType, uint32_t Offset>
+struct OffsetAccessor
+{
+    BaseAccessor accessor;
+
+    void set(uint32_t idx, NBL_REF_ARG(AccessorType) x) {accessor.set(idx + Offset, x);}
+
+    AccessorType get(uint32_t idx) {return accessor.get(idx + Offset);}
```
devshgraphicsprogramming, line 269: you can work out the types using `using value_type = decltype(declval<BaseAccessor>().get(0xdeadbeefu))`

although due to the presence of overloads, I'd make the set and get possibly templated?

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -157,28 +159,101 @@ struct FFT<2,true, Scalar, device_capabilities>
...
-    template<typename Accessor, typename ShaderMemoryAccessor>
-    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(ShaderMemoryAccessor) sharedmemAccessor)
+    template<typename Accessor, typename SharedMemoryAccessor>
+    static enable_if_t< (mpl::is_pot_v<K> && K > 2), void > __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
     {
-        if (!Inverse)
+        static const uint32_t virtualThreadCount = K >> 1;
+        static const uint16_t passes = mpl::log2<K>::value - 1;
+        uint32_t stride = K >> 1;
+        //[unroll(passes)]
```
devshgraphicsprogramming, line 172: DXC doesn't like the syntax or?

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -157,28 +159,101 @@ struct FFT<2,true, Scalar, device_capabilities>
...
         {
-           ... special steps ...
+            //[unroll(K/2)]
+            for (uint32_t virtualThread = 0; virtualThread < virtualThreadCount; virtualThread++)
+            {
+                const uint32_t virtualThreadID = (virtualThread << _NBL_HLSL_WORKGROUP_SIZE_LOG_2_) + SubgroupContiguousIndex();
+
+                const uint32_t lsb = virtualThread & (stride - 1);
+                const uint32_t loIx = ((virtualThread ^ lsb) << 1) | lsb;
+                const uint32_t hiIx = loIx | stride;
```
devshgraphicsprogramming, line 182: this all seems like waaay too much math

how about you start with `for (uint32_t virtualThreadID=gl_LocalInvocationIndex; virtualThreadID<(K<<(WORKGROUPSIZE_LOG2-1)); virtualThreadID+=WORKGROUPSIZE*2)` and work out the rest from that?

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -157,28 +159,101 @@ struct FFT<2,true, Scalar, device_capabilities>
...
+                
+                complex_t<Scalar> lo = accessor.get(loIx << _NBL_HLSL_WORKGROUP_SIZE_LOG_2_);
+                complex_t<Scalar> hi = accessor.get(hiIx << _NBL_HLSL_WORKGROUP_SIZE_LOG_2_);
+                
+                hlsl::fft::DIF<Scalar>::radix2(hlsl::fft::twiddle<false,Scalar>(virtualThreadID & ((stride << _NBL_HLSL_WORKGROUP_SIZE_LOG_2_) - 1), stride << _NBL_HLSL_WORKGROUP_SIZE_LOG_2_),lo,hi);
```
devshgraphicsprogramming, line 187: the indexing here is pretty hard-core too

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -157,28 +159,101 @@ struct FFT<2,true, Scalar, device_capabilities>
...
-    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(ShaderMemoryAccessor) sharedmemAccessor)
+    template<typename Accessor, typename SharedMemoryAccessor>
+    static enable_if_t< (mpl::is_pot_v<K> && K > 2), void > __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
     {
-        if (!Inverse)
+        static const uint32_t virtualThreadCount = K >> 1;
+        static const uint16_t passes = mpl::log2<K>::value - 1;
+        uint32_t stride = K >> 1;
+        //[unroll(passes)]
+        for (uint16_t pass = 0; pass < passes; pass++)
```
devshgraphicsprogramming, line 173: alternative way to formulate is also `for (uint32_t stride = K<<(WORKGROUP_SIZE_LOG2-1); stride>WORKGROUP_SIZE; stride>>=1)`

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: why not count `stride` properly ?

why not count everything with the <<WG_SZ_LOG2 already done?

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: subgroup shuffle cannot work on structs, I just got confirmation

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -157,28 +160,94 @@ struct FFT<2,true, Scalar, device_capabilities>
...
+                
+                complex_t<Scalar> lo = accessor.get(loIx);
+                complex_t<Scalar> hi = accessor.get(hiIx);
+                
+                hlsl::fft::DIF<Scalar>::radix2(hlsl::fft::twiddle<false,Scalar>(virtualThreadID & (stride - 1), stride),lo,hi);
+                
+                accessor.set(loIx, lo);
+                accessor.set(hiIx, hi);
+            }
+            accessor.memoryBarrier(); // no execution barrier just making sure writes propagate to accessor
```
devshgraphicsprogramming, line 186: let me just do one quick verification. We need to check that the same invocation still is responsible for computing all elements of a butterfly (because of no execution barrier & we want to be able to use a "register accessor").

Let's say you have a 512 wg, and 8196 fft size.

devshgraphicsprogramming: First stride is 4196, so invocation i touches: i,i+4096, i+512,i+4618, etc. at offsets this is 0,4096,512,4618,1024,5120,...

Then stride 2048, so `i` touches at offsets: 0,2048,512,2560,1024,3084,1536,3596, 4096,6144,4618,6656,....

Ok I'm seeing a pattern, we go through all multiples of `WORKGROUP_SIZE` one way or another, just in different orders. The register optimization will work.

Mark as resolved as soon as you read.

include/nbl/builtin/hlsl/subgroup/fft.hlsl
devshgraphicsprogramming: device traits PR is now merged to master

include/nbl/builtin/hlsl/memory_accessor.hlsl
devshgraphicsprogramming: why do you make the `get` signature different from all the above!?

include/nbl/builtin/hlsl/memory_accessor.hlsl
```diff
@@ -133,6 +257,36 @@ struct MemoryAdaptor
...
+// Dynamic offset version
+template<class BaseAccessor, class AccessorType>
+struct DynamicOffsetAccessor
+{
+    BaseAccessor accessor;
+    uint32_t offset;
+
+    void set(uint32_t idx, NBL_REF_ARG(AccessorType) x) {accessor.set(idx + offset, x);}
+
+    AccessorType get(uint32_t idx) {return accessor.get(idx + offset);}
```
devshgraphicsprogramming, line 284: keep the signature the same for `get` as other accessors!

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: nvm, forgot its a workgroup shuffle, it can work on anything as long as you can decompose into 4 byte chunks for banking

include/nbl/builtin/hlsl/workgroup/shuffle.hlsl
devshgraphicsprogramming: (the signature of `get`)

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: nvm new math is good

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: new math is good

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: new math is good

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -157,28 +160,94 @@ struct FFT<2,true, Scalar, device_capabilities>
...
+        for (uint32_t k = 0; k < K; k += 2)
+        {
+            if (k)
+            sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
+            offsetAccessor.offset = _NBL_HLSL_WORKGROUP_SIZE_*k;
+            FFT<2,true, Scalar, device_capabilities>::template __call(offsetAccessor,sharedmemAccessor);
+        }
+        accessor = offsetAccessor.accessor;
+      
+        for (uint32_t stride = _NBL_HLSL_WORKGROUP_SIZE_ << 1; stride <= (K >> 1) * _NBL_HLSL_WORKGROUP_SIZE_; stride <<= 1)
```
devshgraphicsprogramming, line 222: `stride<K*WORKGROUP_SIZE` would be more legible than `<=` and `K>>1`

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -157,28 +160,94 @@ struct FFT<2,true, Scalar, device_capabilities>
...
+                {
+                    divides_assign< complex_t<Scalar> > divAss;
+                    divAss(lo, K >> 1);
+                    divAss(hi, K >> 1);  
+                }
+
+                accessor.set(loIx, lo);
+                accessor.set(hiIx, hi);
+            }
+            accessor.memoryBarrier(); // no execution barrier just making sure writes propagate to accessor
```
devshgraphicsprogramming, line 246: you don't need a memory barrier after the last iteration of this.

But you need a memory barrier between the first iteration of this loop and the mini-FFTs you did before.

Therefore more the barrier to the start of the scope (before the unrolled loop)

devshgraphicsprogramming:
@Fletterio can we merge this before you continue working on the FFT Bloom example?

include/nbl/builtin/hlsl/bit.hlsl
```diff
@@ -34,9 +34,8 @@ namespace hlsl
 {
 
 template<class T, class U>
-T bit_cast(U val)
+enable_if_t<sizeof(T) <= sizeof(U), T> bit_cast(U val)
```
devshgraphicsprogramming, line 37: should be == instead of <=

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -7,6 +7,10 @@
 #include "nbl/builtin/hlsl/workgroup/shuffle.hlsl"
 #include "nbl/builtin/hlsl/mpl.hlsl"
 #include "nbl/builtin/hlsl/memory_accessor.hlsl"
+#include "nbl/builtin/hlsl/bit.hlsl"
+
+// Caveats
+// - Sin and Cos in HLSL take 32-bit floats. Using this library with 64-bit floats works perfectly fine, but DXC will emit warnings
```
devshgraphicsprogramming, line 13: also the sin and cos will only truly be float32 precision but in a float64

include/nbl/builtin/hlsl/bit.hlsl
devshgraphicsprogramming: done myself

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -103,10 +104,19 @@ struct FFT;
...
-//      - SharedMemoryAccessor accesses a shared memory array that can fit _NBL_HLSL_WORKGROUP_SIZE_ elements of type complex_t<Scalar>, with get and set 
-//        methods for complex_t<Scalar>. It benefits from coalesced accesses   
+//        that is, one "lo" and one "hi" complex numbers per thread, essentially 4 Scalars per thread. The arrays it accesses with `get` and `set` can optionally be
+//        different, if you don't want the FFT to be done in-place. 
+//        The Accessor MUST provide the following methods:
+//            * void get(uint32_t index, inout complex_t<Scalar> value);
+//            * void set(uint32_t index, in complex_t<Scalar> value);
+//            * void memoryBarrier();
+//        You might optionally want to provide a `workgroupExecutionAndMemoryBarrier()` method on it to wait on to be sure the whole FFT pass is done
+
```
devshgraphicsprogramming, line 114: but you don't use it though?

PR #699, opened 2024-06-13
Better templated Promote struct

Fletterio:
## Description
Changed the Promote struct so that it only works with proper Scalars and saving a lot of boilerplate specialization code

## Testing 
Ran fine on the complex FFT tests (coming to a new example on the examples repo)

### DIFF
--- a/include/nbl/builtin/hlsl/cpp_compat/promote.hlsl
+++ b/include/nbl/builtin/hlsl/cpp_compat/promote.hlsl
@@ -1,6 +1,8 @@
 #ifndef _NBL_BUILTIN_HLSL_CPP_COMPAT_PROMOTE_INCLUDED_
 #define _NBL_BUILTIN_HLSL_CPP_COMPAT_PROMOTE_INCLUDED_
 
+#include "nbl/builtin/hlsl/type_traits.hlsl"
+
 namespace nbl
 {
 namespace hlsl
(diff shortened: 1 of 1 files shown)


include/nbl/builtin/hlsl/cpp_compat/promote.hlsl
```diff
@@ -20,76 +23,43 @@ struct Promote
...
-template<typename U>
-struct Promote<float64_t1,U>
+template<typename Scalar, typename U>
+struct Promote<vector <Scalar, 1>, U>
 {
-    float64_t1 operator()(U v)
+    enable_if_t<is_scalar<Scalar>::value && is_scalar<U>::value, vector <Scalar, 1> > operator()(U v)
     {
-        return float64_t1(v);
+        vector <Scalar, 1> promoted = {Scalar(v)};
```
devshgraphicsprogramming, line 31: I think you should be able to just write `vector<Scalar,1>(v)` the conversion from U to Scalar is implicit here

devshgraphicsprogramming:
Any tests done to make sure it doesn't break existing examples and actually works?

Fletterio:
Ran grep to check which builtins used promote. It's two files under `builtin/hlsl/colorspace`. In the examples folder, no example shader directly uses promote, and the only test importing the aforementioned colorspace builtins is `22_CppCompat`, which runs just fine. I tested all the other examples just in case too (those that compiled in master, at least)

PR #703, opened 2024-07-03
Separable image sampler amortized lookups

Fletterio:
## Description
When validating writes/copies of descriptor sets, we do a binary search via `CBindingRedirect`. This same search is then repeated later in the processing of the write/copy, which is unnecessary. These changes avoid the duped searches

## Testing 
Examples that wrote to descriptor sets work fine. I couldn't find any examples that had a copy (there is one in the examples repo but not under `build/examples_tests` so I assume it's disabled).


include/nbl/video/IGPUDescriptorSet.h
```diff
@@ -87,14 +87,14 @@ class IGPUDescriptorSet : public asset::IDescriptorSet<const IGPUDescriptorSetLa
 	private:
         inline void incrementVersion() { m_version.fetch_add(1ull); }
 
+        using redirect_t = IGPUDescriptorSetLayout::CBindingRedirect;
         friend class ILogicalDevice;
-        asset::IDescriptor::E_TYPE validateWrite(const IGPUDescriptorSet::SWriteDescriptorSet& write) const;
-        void processWrite(const IGPUDescriptorSet::SWriteDescriptorSet& write, const asset::IDescriptor::E_TYPE type);
-        bool validateCopy(const IGPUDescriptorSet::SCopyDescriptorSet& copy) const;
-        void processCopy(const IGPUDescriptorSet::SCopyDescriptorSet& copy);
+        asset::IDescriptor::E_TYPE validateWrite(const IGPUDescriptorSet::SWriteDescriptorSet& write, redirect_t::storage_range_index_t& descriptorRedirectBindingIndex, redirect_t::storage_range_index_t& mutableSamplerRedirectBindingIndex) const;
```
devshgraphicsprogramming, line 92: instead of doing 1 return value + multiple out pointers or references, why not wrap all 3 into a struct? e.g.
```cpp
struct SValidationResult
{
   asset::IDescriptor::E_TYPE type;
   redirect_t::storage_range_index_t descriptorRedirect;
   redirect_t::storage_range_index_t mutableSamplerRedirect;
};
```

and pass this by `const&` to the `processWrite` ?

Also oa similar thing for the copies.

src/nbl/video/IGPUDescriptorSet.cpp
```diff
@@ -216,46 +232,26 @@ bool IGPUDescriptorSet::validateCopy(const IGPUDescriptorSet::SCopyDescriptorSet
...
+        incrementVersion();
 
-        if (srcSamplers && dstSamplers)
-            std::copy_n(srcSamplers, copy.count, dstSamplers);
+    std::copy_n(srcDescriptors, copy.count, dstDescriptors);
 
-        break;
-    }
+    if (srcSamplers && dstSamplers)
+        std::copy_n(srcSamplers, copy.count, dstSamplers);
```
devshgraphicsprogramming, line 254: very nice code cleanup I applaud

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -469,8 +472,13 @@ bool ILogicalDevice::updateDescriptorSets(const std::span<const IGPUDescriptorSe
                 return false;
         }
         outCategory++;
+        i++;
     }
 
+    core::vector<redirect_t::storage_range_index_t> copySrcDescriptorRedirectIndices(descriptorCopies.size()), copySrcMutableSamplerRedirectIndices(descriptorCopies.size());
+    core::vector<redirect_t::storage_range_index_t> copyDstDescriptorRedirectIndices(descriptorCopies.size()), copyDstMutableSamplerRedirectIndices(descriptorCopies.size());
+    core::vector<asset::IDescriptor::E_TYPE> copyTypes(descriptorCopies.size());
```
devshgraphicsprogramming, line 480: normally SoA perform better than AoS, except when the arrays are backed by separate dynamic allocations/containers

Hence moving to a Array of Structures and having the validation parameters in one struct will benefit you

PR #711, opened 2024-07-14
Flatten regions stream hash image filter

devshgraphicsprogramming:


include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+		
+		class CState : public IImageFilter::IState
+		{
+			public:
+				virtual ~CState() {}
+
+				using hash_t = std::array<size_t, 4u>;
+
+				const ICPUImage*					inImage = nullptr;
+				hash_t								outHash = { {0} };
```
devshgraphicsprogramming, line 39: this needs to be a `core::blake3_hash_t`

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+				ScratchMemory						scratchMemory;
+		};
+
+		using state_type = CState;
+
+		static inline ScratchMemory allocateScratchMemory(const asset::ICPUImage* inImage)
+		{
+			ScratchMemory scratch;
+			scratch.flatten = asset::IAsset::castDown<asset::ICPUImage>(inImage->clone());
```
devshgraphicsprogramming, line 48: you need to do a level 0 clone, otherwise you get a buffer cope as well

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+
+			const auto& parameters = scratch.flatten->getCreationParameters();
+			scratch.heap = core::make_smart_refctd_ptr<asset::ICPUBuffer>(parameters.mipLevels * parameters.arrayLayers * sizeof(CState::outHash));
+
```
devshgraphicsprogramming, line 52: now you need to set regions on the cloned image, like flatten region does

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+			if (!state)
+				return false;
+
+			if(!state->inImage)
+				return false;
+
+			if(!state->scratchMemory.flatten)
+				return false;
+
+			if (state->scratchMemory.flatten->getBuffer()->getSize() != state->inImage->getBuffer()->getSize())
```
devshgraphicsprogramming, line 67: this will not necessarily be true if the input has gaps

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+					rit->imageSubresource.layerCount = parameters.arrayLayers;
+					rit->imageOffset = { 0u,0u,0u };
+					rit->imageExtent = { localExtent.x,localExtent.y,localExtent.z };
+					auto levelSize = info.roundToBlockSize(localExtent);
+					auto memsize = levelSize[0] * levelSize[1] * levelSize[2] * parameters.arrayLayers * bytesPerPixel;
+
+					assert(memsize.getNumerator() % memsize.getDenominator() == 0u);
+					bufferSize += memsize.getIntegerApprox();
+				}
+			}
```
devshgraphicsprogramming, line 152: this needs to happen in `allocateScratchMemory`

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+
+		using state_type = CState;
+
+		static inline ScratchMemory allocateScratchMemory(const asset::ICPUImage* inImage)
+		{
+			ScratchMemory scratch;
+			scratch.flatten = asset::IAsset::castDown<asset::ICPUImage>(inImage->clone());
+
+			const auto& parameters = scratch.flatten->getCreationParameters();
+			scratch.heap = core::make_smart_refctd_ptr<asset::ICPUBuffer>(parameters.mipLevels * parameters.arrayLayers * sizeof(CState::outHash));
```
devshgraphicsprogramming, line 51: other filters just ask for a pointer and tell you the size it should back

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+			auto executePerMipLevel = [&](const uint32_t miplevel)
+			{
+				/*
+					we stream-hash texels per given mip level & layer
+				*/
+
+				auto* hashers = _NBL_NEW_ARRAY(blake3_hasher, parameters.arrayLayers);
+
+				for (auto layer = 0u; layer < parameters.arrayLayers; ++layer)
+					blake3_hasher_init(&hashers[layer]);
```
devshgraphicsprogramming, line 163: you can just grab them out of scratch

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+
+				CBasicImageFilterCommon::executePerRegion(policy, proxy.flatten.outImage.get(), hash, regions->begin(), regions->end(), clipFunctor); // fire the hasher for layers with specified execution policy, yes you can use parallel policy here if you want at it will work
+
+				for (auto layer = 0u; layer < parameters.arrayLayers; ++layer)
+				{
+					auto* out = reinterpret_cast<uint8_t*>(reinterpret_cast<CState::hash_t*>(state->scratchMemory.heap->getPointer()) + (miplevel * parameters.arrayLayers) + layer);
+					blake3_hasher_finalize(hashers + layer, out, sizeof(CState::hash_t)); // finalize hash for layer + put it to heap for given mip level
+				}
+
+				_NBL_DELETE_ARRAY(hashers, parameters.arrayLayers);
```
devshgraphicsprogramming, line 183: you need to combine the per-layer hashes into a mip level hash here

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
```diff
@@ -0,0 +1,220 @@
...
+
+			blake3_hasher hasher;
+			blake3_hasher_init(&hasher);
+			{
+				for (auto miplevel = 0u; miplevel < parameters.mipLevels; ++miplevel)
+					for (auto layer = 0u; layer < parameters.arrayLayers; ++layer)
+					{
+						auto* hash = reinterpret_cast<CState::hash_t*>(state->scratchMemory.heap->getPointer()) + (miplevel * parameters.arrayLayers) + layer;
+						blake3_hasher_update(&hasher, hash->data(), sizeof(CState::hash_t));
+					}
```
devshgraphicsprogramming, line 204: here take hashes of mip levels and turn into one hash

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: ok but remember this is created from `master` where it doesn't exist yet

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: does it make any difference? It will make my scratch smaller but I would like to store each hash in the scratch for our [tool & testing purposes](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/tree/14b34ffab28112de092b56d17be52e32d450b834/64_StreamHash#streamhashapp) (however I could add template parameter which toggles whether I store all hashes or only for mips to do what you have in mind - by default would do it)

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: https://github.com/Devsh-Graphics-Programming/Nabla/pull/711#discussion_r1677514207

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: right, okay

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: okay will adjust

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: the copy was intended, but yes it's a dumb copy

AnastaZIuk: done

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: done

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: done

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: done

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: done

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: I won't change it, it can run both mips & layers with parallel policy then it would break, must finalize at the end but now using hashers from scratch memory

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: I leave it how it is, want to keep all data in scratch unless you want template I can add it

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: this thing doesn't exist on current `HEAD`, once you merge I can adjust

AnastaZIuk: you have it on your branch I believe

devshgraphicsprogramming: time to merge `asset_converter3` to this branch then

include/nbl/core/hash/blake.h
```diff
@@ -12,15 +12,19 @@ struct blake3_hash_t
 {
 	inline bool operator==(const blake3_hash_t&) const = default;
 
-	uint8_t data[BLAKE3_OUT_LEN];
+	uint8_t data[BLAKE3_OUT_LEN] = { {0} };
 };
 
+static_assert(sizeof(blake3_hash_t) == BLAKE3_OUT_LEN);
+
+// TODO, Arek: why not in source and limit to translation unit as part of Nabla library, again we make a wrapper just to expose 3rdparty to our public headers and with blake case we need to link it when in static mode anyway because it's not header only library
```
AnastaZIuk, line 20: ok actually those blake3 structs we use are built with basic types only & all in headers

include/nbl/asset/filters/CFlattenRegionsStreamHashImageFilter.h
AnastaZIuk: done

include/nbl/asset/IDescriptorSetLayout.h
```diff
@@ -467,7 +467,6 @@ class IDescriptorSetLayout : public IDescriptorSetLayoutBase
 				if ((b.type == IDescriptor::E_TYPE::ET_SAMPLER or b.type == IDescriptor::E_TYPE::ET_COMBINED_IMAGE_SAMPLER) and b.immutableSamplers)
 				{
 					const auto localOffset = m_immutableSamplerRedirect.getStorageOffset(typename CBindingRedirect::binding_number_t(b.binding)).data;
-					assert(bool(localOffset));
```
devshgraphicsprogramming, line 470: why did you remove this?

tad1: when `localOffset==0` (copying from the beginning) it aborts

tad1: I don't know if that was the intent

devshgraphicsprogramming: that would be because we have a `.data` there.

I know how to change it, my bug/typo

src/nbl/asset/interchange/CGLILoader.cpp
```diff
@@ -187,22 +188,34 @@ namespace nbl
...
 				const auto layerSize = getFullSizeOfLayer(mipLevel);
+				const auto mipOffset = (mipLevel * imageInfo.arrayLayers);
+
 				for (uint16_t layer = 0; layer < imageInfo.arrayLayers; ++layer)
 				{
 					const auto layersData = getCurrentGliLayerAndFace(layer);
 					const auto gliLayer = layersData.first;
 					const auto gliFace = layersData.second;
+					
+					const auto pOffset = mipOffset + layer;
```
devshgraphicsprogramming, line 204: unused variable?

src/nbl/asset/ICPUImage.cpp
```diff
@@ -0,0 +1,319 @@
...
+		return execute(core::execution::seq, state);
+	}
+
+private:
+
+	struct ScratchMap
+	{
+		asset::SBufferRange<asset::ICPUBuffer> hashes; // hashes, single hash is obtained from given miplevel & layer, full hash for an image is a hash of this hash buffer
+		asset::SBufferRange<asset::ICPUBuffer> hashers; // hashers, used to produce a hash
+		asset::SBufferRange<asset::ICPUBuffer> flatten; // tightly packed texels from input, no memory gaps
```
devshgraphicsprogramming, line 297: its weird to use `ICPUBuffer` which is an asset, to represent what is essentially a span<uint8_t>

devshgraphicsprogramming: actually a `std::span<hash_t>`, `std::span<blake3_hasher>` , the last one can stay an ICPUBuffer

src/nbl/asset/ICPUImage.cpp
```diff
@@ -0,0 +1,319 @@
...
+		*/
+
+		ScratchMap scratch;
+		{
+			auto buffer = getScratchAsBuffer(state->scratch.size);
+			const auto product = parameters.mipLevels * parameters.arrayLayers;
+
+			scratch.hashes = { .offset = 0u, .size = product * sizeof(CState::outHash), .buffer = buffer };
+			scratch.hashers = { .offset = scratch.hashes.size, .size = product * sizeof(blake3_hasher), .buffer = buffer };
+			scratch.flatten = { .offset = scratch.hashers.offset + scratch.hashers.size, .size = state->scratch.size - scratch.hashers.size - scratch.hashes.size, .buffer = buffer };
```
devshgraphicsprogramming, line 138: this is the only thing that should be an `ICPUBuffer` (and not really a range)

src/nbl/asset/ICPUImage.cpp
```diff
@@ -0,0 +1,319 @@
...
+			auto out = ICPUImage::create(IImage::SCreationParams(parameters));
+			out->setBufferAndRegions(core::smart_refctd_ptr(getScratchAsBuffer(scratch.flatten.size, scratch.flatten.offset)), std::move(regions));
+
+			auto& flatten = proxy.flatten;
+			flatten.inImage = state->inImage;
+			flatten.outImage = out;
+			flatten.preFill = true;
+			memset(flatten.fillValue.pointer, 0, sizeof(flatten.fillValue.pointer));
+
+			assert(CFlattenRegionsImageFilter::execute(policy, &proxy.flatten)); // this should never fail, at this point we are already validated
```
devshgraphicsprogramming, line 191: @AnastaZIuk @tad1 I don't know if you already came across this, but any code in an `assert(blah)` just disappears in a non-Debug build (cause assert is a macro thats empty in a Release build)

src/nbl/asset/ICPUImage.cpp
```diff
@@ -0,0 +1,319 @@
...
+			}
+
+			auto out = ICPUImage::create(IImage::SCreationParams(parameters));
+			out->setBufferAndRegions(core::smart_refctd_ptr(getScratchAsBuffer(scratch.flatten.size, scratch.flatten.offset)), std::move(regions));
+
+			auto& flatten = proxy.flatten;
+			flatten.inImage = state->inImage;
+			flatten.outImage = out;
+			flatten.preFill = true;
+			memset(flatten.fillValue.pointer, 0, sizeof(flatten.fillValue.pointer));
```
devshgraphicsprogramming, line 189: question, does flatten work with Compressed Texture Formats?

(since we can't encode into those texel blocks)

src/nbl/asset/ICPUImage.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+		const auto* const image = isFullyFlatten ? state->inImage : proxy.flatten.outImage.get();
+		const auto texelOrBlockByteSize = asset::getTexelOrBlockBytesize(parameters.format);
+		const uint8_t* inData = reinterpret_cast<const uint8_t*>(image->getBuffer()->getPointer());
+
+		std::vector<uint32_t> layers(parameters.arrayLayers);
+		std::iota(layers.begin(), layers.end(), 0);
+
+		std::vector<uint32_t> levels(parameters.mipLevels);
+		std::iota(levels.begin(), levels.end(), 0);
```
devshgraphicsprogramming, line 209: make your own custom iterators over a range of integers (someone probably has done it already), instead of using `iota`, the number of arrayLayers can be stupidly high

src/nbl/asset/ICPUImage.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+		/*
+			we stream-hash texels per given mip level & layer
+		*/
+
+		auto* const hashes = reinterpret_cast<CState::hash_t*>(getScratchAsBuffer(scratch.hashes.size, scratch.hashes.offset)->getPointer());
+		auto* const hashers = reinterpret_cast<blake3_hasher*>(getScratchAsBuffer(scratch.hashers.size, scratch.hashers.offset)->getPointer());
```
devshgraphicsprogramming, line 216: see this would make more sense with `span<>` instead of buffers

src/nbl/asset/ICPUImage.cpp
```diff
@@ -0,0 +1,319 @@
...
+				const bool performNullHash = regions.empty();
+
+				if (performNullHash)
+				{
+					const auto mipExtentInBlocks = info.convertTexelsToBlocks(image->getMipSize(miplevel));
+					const auto zeroLength = info.getBlockByteSize() * mipExtentInBlocks.x;
+					auto zeroArray = std::make_unique<uint8_t[]>(zeroLength);
+					for (auto z = 0; z < mipExtentInBlocks.z; z++)
+						for (auto y = 0; y < mipExtentInBlocks.y; y++)
+							blake3_hasher_update(hasher, zeroArray.get(), zeroLength);
```
devshgraphicsprogramming, line 249: actually you have my blessing to give this a special treatment at the start, you can do a hash of a zero-length array if there are no regions (it will be much faster)

and assert() in the filter that there are SOME regions

include/nbl/asset/IDescriptorSetLayout.h
devshgraphicsprogramming: fixed on `asset-conversion-v3` branch

src/nbl/asset/ICPUImage.cpp
tad1: mentioned in #731 - it works, but `CFlattenRegionsImageFilter` does not set content hash for the output image

devshgraphicsprogramming: explain?

PR #716, opened 2024-07-17
Clustered rendering

AnastaZIuk:


devshgraphicsprogramming:
Ok now make it point at master, and remove all the stuff that conflicts with the `examples_tests` submodule.

PR #719, opened 2024-07-23
Fix subgroupArithmetic error

nipunG314:


devshgraphicsprogramming:
no need for special PR, we'll update the submodule pointer to master HEAD eventually

PR #720, opened 2024-07-23
update getRegions() to return std::span<> instead of core::SRange<>

tad1:
## Description
shallowly updates `getRegions()` in `ICPUImage.h` (_and methods explicit usages_) to use `std::span<>`

to obtain raw pointers (required by some methods) `<span>.data()` and `<span>.data() + <span>.size()` is used, since it's the only method to access raw pointers without dereferencing iterators
> dereferencing `.end()` iterator may cause address violation segfault
## Testing 
passes 24_ColorSpaceTest (requires a patch - see [0aa4c5](https://github.com/tad1/Nabla-Examples-and-Tests/commit/0aa4c5e7f3a5e3ce6680cbdac7ba022d1d99e8be))

## TODO list:
- [x] decide if methods like `CBasicImageFilterCommon::executePerRegion` should use iterators **or** raw pointers (_current_)

### DIFF
--- a/include/nbl/asset/filters/dithering/CPrecomputedDither.h
+++ b/include/nbl/asset/filters/dithering/CPrecomputedDither.h
@@ -77,7 +77,7 @@ namespace nbl
 								asset::encodePixels<decodeFormat>(outData + offset, decodeBuffer);
 							};
 
-							CBasicImageFilterCommon::executePerRegion(flattenDitheringImage.get(), decode, flattenDitheringImage->getRegions(chosenMipmap).begin(), flattenDitheringImage->getRegions(chosenMipmap).end());
+							CBasicImageFilterCommon::executePerRegion(flattenDitheringImage.get(), decode, flattenDitheringImage->getRegions(chosenMipmap).data(), flattenDitheringImage->getRegions(chosenMipmap).data() + flattenDitheringImage->getRegions(chosenMipmap).size());
 
 							auto decodeCreationParams = creationParams;
 							decodeCreationParams.format = decodeFormat;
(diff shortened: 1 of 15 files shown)


include/nbl/asset/filters/dithering/CPrecomputedDither.h
```diff
@@ -77,15 +77,15 @@ namespace nbl
 								asset::encodePixels<decodeFormat>(outData + offset, decodeBuffer);
 							};
 
-							CBasicImageFilterCommon::executePerRegion(flattenDitheringImage.get(), decode, flattenDitheringImage->getRegions(chosenMipmap).begin(), flattenDitheringImage->getRegions(chosenMipmap).end());
+							CBasicImageFilterCommon::executePerRegion(flattenDitheringImage.get(), decode, flattenDitheringImage->getRegions(chosenMipmap).data(), flattenDitheringImage->getRegions(chosenMipmap).data() + flattenDitheringImage->getRegions(chosenMipmap).size());
```
devshgraphicsprogramming, line 80: could you fix the double calls, and save the span into a temporary variable? Currentl:
1. its horrible line clutter
2. `getRegions` actually involves a binary search, so I'd rather not perform it twice

devshgraphicsprogramming:
Maybe make the `executePerRegion` take a span as well?

include/nbl/asset/filters/dithering/CPrecomputedDither.h
tad1: resolved

tad1:
> Maybe make the `executePerRegion` take a span as well?
done in [42c1f12](https://github.com/Devsh-Graphics-Programming/Nabla/pull/720/commits/42c1f12db372136b5995a9b9e094021274309d3f)

PR #726, opened 2024-08-06
Add logging for nbl::video device calls

### LINKED ISSUE #684: Logging nbl::video device call failures
## Problem
Many device functions return a boolean to indicate a successful call to lower level api (Vulkan) function.
But the return value is often ignored making it hard to debug the cause of the problem.
Even if not ignored, simply returning false on a failure doesn't help much pinpointing the exact cause. 
There is also `createXXX` functions that may return nullptr due to reasons such as invalid creation params
for example inside `ILogicalDevice::createFramebuffer` 
```c
if (!params.validate())
    return nullptr;
```
may fail due to 30+ reasons specified in vulkan specification and handled manually in `IFramebuffer::SCreationParams::validate`

that's why we need to "error log" the exact cause of the failure each time we `return false` on device calls

## Solution
use `system::logger_opt_ptr m_logger` in ILogicalDevice` to log every possible scenario that fails.

tad1:
## Description
Add a bunch of error logs for `ILogicalDevice` methods to disambiguate failures.
Add `_NBL_DEBUG_BREAK_IF` on DEVICE_LOST
- resolves #684

### testing 
compiles, checked formatting
tested regression in:
- 24th ColorSpace example
- 62th CAD example

### coverage
- [x] `ILogicalDevice.h`
- [x] `ILogicalDevice.cpp`
- [X] `IQueue.cpp`


include/nbl/video/ILogicalDevice.h
```diff
@@ -105,16 +105,19 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
...
                         if (barrier.dep.srcStageMask || barrier.dep.srcAccessMask)
+                            m_logger.log("Ownership operation ACQUIRE requires, srcStageMask and srcAccessMask being set to 0, in  [%s - %s:%p]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
                             return false;
                         break;
                     case IGPUCommandBuffer::SOwnershipTransferBarrier::OWNERSHIP_OP::RELEASE:
                         if (barrier.dep.dstStageMask || barrier.dep.dstAccessMask)
+                            m_logger.log("Ownership operation RELEASE requires, dstStageMask and dstAccessMask being set to 0, in  [%s - %s:%p]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
                             return false;
                         break;
                     default:
```
devshgraphicsprogramming, line 123: remove this whole switch, I messed up, those masks are actually useful for ownership transfers

include/nbl/video/ILogicalDevice.h
```diff
@@ -105,16 +105,19 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
                 // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier2-srcStageMask-03854
                 constexpr auto HostBit = asset::PIPELINE_STAGE_FLAGS::HOST_BIT;
                 if (barrier.dep.srcStageMask.hasFlags(HostBit)||barrier.dep.dstStageMask.hasFlags(HostBit))
+                    m_logger.log("Source and destination queue family index should be equal, HOST_BIT is set [%s - %s:%p]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
```
devshgraphicsprogramming, line 108: better log message, source and dest queue indices are a vulkan thing, we only have an other queue index and a ownership op type

the real reason is that you can't do an ownership transfer that syncs against a host stage

include/nbl/video/ILogicalDevice.h
tad1: resolved

include/nbl/video/ILogicalDevice.h
tad1: changed to `"Invalid barrier, ownership transfer with host is not allowed"`

tad1:
@devshgraphicsprogramming I think this covers the scope of #684

devshgraphicsprogramming:
@Erfan-Ahmadi review rest and press the green button

include/nbl/video/ILogicalDevice.h
```diff
@@ -104,22 +104,11 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
                 // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkBufferMemoryBarrier2-srcStageMask-03851
                 // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageMemoryBarrier2-srcStageMask-03854
                 constexpr auto HostBit = asset::PIPELINE_STAGE_FLAGS::HOST_BIT;
-                if (barrier.dep.srcStageMask.hasFlags(HostBit)||barrier.dep.dstStageMask.hasFlags(HostBit))
+                if (barrier.dep.srcStageMask.hasFlags(HostBit) || barrier.dep.dstStageMask.hasFlags(HostBit)) {
```
Erfan-Ahmadi, line 107: If you see our codebase, for multi-line control blocks (if/else/while/switch), position for open braces is usually on the next line.
I personally don't rely on and disabled auto-formatting, but if you're using one. Select the positiin of open braces to be on a `new line`
image

include/nbl/video/ILogicalDevice.h
```diff
@@ -334,8 +335,10 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
         // Create an ImageView that can actually be used by shaders (@see ICPUImageView)
         inline core::smart_refctd_ptr<IGPUImageView> createImageView(IGPUImageView::SCreationParams&& params)
         {
-            if (!params.image->wasCreatedBy(this))
+            if (!params.image->wasCreatedBy(this)) {
+                m_logger.log("Invalid image was given [%s - %s:%d]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
```
Erfan-Ahmadi, line 339: Improved message:
"The image was not created by this Device"

Erfan-Ahmadi:
Go to `IQueue.cpp` improve the logging there to have the same format as what you currently have.
also add additional logging based on submit result near the end of the function (line 87), that can have DEVICE_LOST as well

include/nbl/video/ILogicalDevice.h
```diff
@@ -915,38 +1027,54 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
         template<typename CreationParams, typename ExtraLambda>
         inline CreationParams::SSpecializationValidationResult commonCreatePipelines(IGPUPipelineCache* const pipelineCache, const std::span<const CreationParams> params, ExtraLambda&& extra)
         {
-            if (pipelineCache && !pipelineCache->wasCreatedBy(this))
+            if (pipelineCache && !pipelineCache->wasCreatedBy(this)) {
+                m_logger.log("Invalid pipelineCache was given [%s - %s:%d]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
```
Erfan-Ahmadi, line 1031: pipelineCache wasn't created by this device

include/nbl/video/ILogicalDevice.h
```diff
@@ -134,12 +123,15 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
...
+            if (!range.buffer || range.size == 0u) {
+                m_logger.log("No buffer was specified [%s - %s:%d]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
                 return false;
+            }
             // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkBufferMemoryBarrier2-offset-01187
             // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkBufferMemoryBarrier2-offset-01189
             const size_t remain = range.size!=IGPUCommandBuffer::SBufferMemoryBarrier<ResourceBarrier>{}.range.size ? range.size:1ull;
             if (range.offset+remain>range.buffer->getSize())
+                m_logger.log("Invalid range was specified [%s - %s:%d]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
                 return false;
```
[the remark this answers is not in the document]
tad1: resolved

include/nbl/video/ILogicalDevice.h
tad1: yeah, that's better

tad1: resolved

tad1:
> Go to `IQueue.cpp` improve the logging there to have the same format as what you currently have. also add additional logging based on submit result near the end of the function (line 87), that can have DEVICE_LOST as well

done

src/nbl/video/IQueue.cpp
```diff
@@ -83,8 +83,17 @@ auto IQueue::submit(const std::span<const SSubmitInfo> _submits) -> RESULT
...
-    if (result!=RESULT::SUCCESS)
+    if (result == RESULT::DEVICE_LOST)
+    {
+        logger->log("Device lost [%s - %s:%d]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
+        _NBL_DEBUG_BREAK_IF(true);
+    }
+
+    if (result != RESULT::SUCCESS)
+    {
+        logger->log("Failed submit command buffers to the queue [%s - %s:%d]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
```
Erfan-Ahmadi, line 94: your logic should be:

```c++
if (result != SUCCESS)
{
if (result == DEVICE_LOST)
 // log Device LOST
else if (result == ....) 
 // log. ...
else
 // log = "Failed submitting command buffers to queue..
 return result.
}
```

include/nbl/video/ILogicalDevice.h
```diff
@@ -177,6 +173,12 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
                     for (const auto& queue : *m_queues)
                         queue->cullResources(info.semaphore);
                 }
+                if (waitStatus == retval_t::DEVICE_LOST)
+                {
+                    m_logger.log("Device lost [%s - %s:%d]", system::ILogger::ELL_ERROR, __FUNCTION__, __FILE__, __LINE__);
```
Erfan-Ahmadi, line 178: @devshgraphicsprogramming What do you think about a a `LOG_ERROR` macro.
It allows us to change to format of how we log in the future and add `GIT_COMMIT_HASH`

Erfan-Ahmadi: I say macro because __FUNCTION__, __LINE__ need to work :D

devshgraphicsprogramming: yes but such a macro should not be in a header shared between multiple cpp, just each of the CPPs

alternatively it can be defined as a so-called X-macro (a short header you include, but need to define some defines first for the macros inside to work, e.g. `m_logger` for the logger object)

src/nbl/video/IQueue.cpp
```diff
@@ -3,6 +3,10 @@
 #include "nbl/video/ILogicalDevice.h"
 #include "nbl/video/TimelineEventHandlers.h"
 
+#define LOG_FUNCTION logger->log
+#define LOG(SEVERITY, FORMAT, ...) LOG_FUNCTION("FORMAT [%s - %s:%d]", SEVERITY, __VA_ARGS__, __FUNCTION__, __FILE__, __LINE__);
+#define LOG_ERROR(FORMAT, ...) LOG(nbl::system::ILogger::ELL_ERROR, FORMAT, __VA_ARGS__)
```
devshgraphicsprogramming: thought about moving the macros into two headers (X-macros) one to define them and another to undef?

src/nbl/video/IQueue.cpp
tad1: resolved

include/nbl/logging_macros.h
```diff
@@ -0,0 +1,8 @@
+#ifdef __GIT_INFO_H_INCLUDED__
```
devshgraphicsprogramming, line 1: emit error if not there

include/nbl/logging_macros.h
```diff
@@ -0,0 +1,8 @@
...
+// when "git_info.h" is included
+#	define LOG(SEVERITY, FORMAT, ...) LOG_FUNCTION(FORMAT" [%s][%s - %s:%d]", SEVERITY, __VA_ARGS__, nbl::gtml::nabla_git_info.commitShortHash, __FUNCTION__, __FILE__, __LINE__);
+#	define LOG_ERROR(FORMAT, ...) LOG(nbl::system::ILogger::ELL_ERROR, FORMAT, __VA_ARGS__)
+#else
+#	define LOG(SEVERITY, FORMAT, ...) LOG_FUNCTION(FORMAT" [%s - %s:%d]", SEVERITY, __VA_ARGS__, __FUNCTION__, __FILE__, __LINE__);
+#	define LOG_ERROR(FORMAT, ...) LOG(nbl::system::ILogger::ELL_ERROR, FORMAT, __VA_ARGS__)
```
devshgraphicsprogramming, line 7: prefix with NBL, and check for double define

src/nbl/video/IQueue.cpp
```diff
@@ -3,6 +3,10 @@
 #include "nbl/video/ILogicalDevice.h"
 #include "nbl/video/TimelineEventHandlers.h"
 
+#include "git_info.h"
```
devshgraphicsprogramming, line 6: why is `git_info` on its own weird path + why its header guard is `__GIT_INFO_H_INCLUDED__` and not `_NBL_GIT_INFO_H_INCLUDED_`

include/nbl/logging_macros.h
```diff
@@ -1,8 +1,8 @@
...
+#ifdef _GIT_INFO_H_INCLUDED_
 // when "git_info.h" is included
-#	define LOG(SEVERITY, FORMAT, ...) LOG_FUNCTION(FORMAT" [%s][%s - %s:%d]", SEVERITY, __VA_ARGS__, nbl::gtml::nabla_git_info.commitShortHash, __FUNCTION__, __FILE__, __LINE__);
-#	define LOG_ERROR(FORMAT, ...) LOG(nbl::system::ILogger::ELL_ERROR, FORMAT, __VA_ARGS__)
+#	define NBL_LOG(SEVERITY, FORMAT, ...) NBL_LOG_FUNCTION(FORMAT" [%s][%s - %s:%d]", SEVERITY, __VA_ARGS__, nbl::gtml::nabla_git_info.commitShortHash, __FUNCTION__, __FILE__, __LINE__);
+#	define NBL_LOG_ERROR(FORMAT, ...) NBL_LOG(nbl::system::ILogger::ELL_ERROR, FORMAT, __VA_ARGS__)
 #else
-#	define LOG(SEVERITY, FORMAT, ...) LOG_FUNCTION(FORMAT" [%s - %s:%d]", SEVERITY, __VA_ARGS__, __FUNCTION__, __FILE__, __LINE__);
-#	define LOG_ERROR(FORMAT, ...) LOG(nbl::system::ILogger::ELL_ERROR, FORMAT, __VA_ARGS__)
+#	define NBL_LOG(SEVERITY, FORMAT, ...) NBL_LOG_FUNCTION(FORMAT" [%s - %s:%d]", SEVERITY, __VA_ARGS__, __FUNCTION__, __FILE__, __LINE__);
```
devshgraphicsprogramming: check for double define

PR #728, opened 2024-08-13
Autoexposure example restoration

devshgraphicsprogramming:


include/nbl/builtin/hlsl/math/morton.hlsl
```diff
@@ -0,0 +1,283 @@
...
+    {
+        x = (x | (x << 6)) & morton4d_mask<T>(1);
+    }
+    x = (x | (x << 3)) & morton4d_mask<T>(0);
+
+    return x;
+}
+#else
+template <typename T>
+constexpr T morton2d_mask(uint8_t _n)
```
devshgraphicsprogramming, line 146: why two different versions for HLSL and C++ ?

include/nbl/builtin/hlsl/math/morton.hlsl
```diff
@@ -0,0 +1,283 @@
...
+{
+    const static uint64_t mask[4] =
+    {
+        0x1111111111111111ull,
+        0x0303030303030303ull,
+        0x000F000F000F000Full,
+        0x000000FF000000FFull
+    };
+    return (T)mask[_n];
+}
```
devshgraphicsprogramming, line 61: couldn't you make them into
```cpp
namespace morton
{
namespace impl
{
template<uint16_t Dims, typename T>
struct mask;

// now the partial specializations
}
}
```
with the masks as `NBL_CONSTEXPR` member variables

This way you can have
```cpp
template<uint16_t Dims, typename T, uint16_t BitDepth=1+(sizeof(T)-1)/Dims>
enable_if_t<is_scalar_v<T>&&is_integral_v<T>,T> decode(T x)
{
   static_assert(BitDepth*Dims<=sizeof(T));

   x = x & mask<Dims,T>::value[0];
   
   .. rest of if-else statements
}
```

include/nbl/builtin/hlsl/math/morton.hlsl
```diff
@@ -0,0 +1,283 @@
...
+T morton2d_decode_x(T _morton) { return impl::morton2d_decode<T, bitDepth>(_morton); }
+template<typename T, uint32_t bitDepth = sizeof(T) * 8u>
+T morton2d_decode_y(T _morton) { return impl::morton2d_decode<T, bitDepth>(_morton >> 1); }
+
+template<typename T, uint32_t bitDepth = sizeof(T) * 8u>
+T morton2d_encode(T x, T y) { return impl::separate_bits_2d<T, bitDepth>(x) | (impl::separate_bits_2d<T, bitDepth>(y) << 1); }
+template<typename T, uint32_t bitDepth = sizeof(T) * 8u>
+T morton3d_encode(T x, T y, T z) { return impl::separate_bits_3d<T, bitDepth>(x) | (impl::separate_bits_3d<T, bitDepth>(y) << 1) | (impl::separate_bits_3d<T, bitDepth>(z) << 2); }
+template<typename T, uint32_t bitDepth = sizeof(T) * 8u>
+T morton4d_encode(T x, T y, T z, T w) { return impl::separate_bits_4d<T, bitDepth>(x) | (impl::separate_bits_4d<T, bitDepth>(y) << 1) | (impl::separate_bits_4d<T, bitDepth>(z) << 2) | (impl::separate_bits_4d<T, bitDepth>(w) << 3); }
```
devshgraphicsprogramming, line 278: imho these should be `template<uint16_t Dims, typename T, uint16_t BitDepth=sizeof(T)*8>`

because that way you only need to write the encode and decode once
```cpp
template<uint16_t Dims, typename T, uint16_t BitDepth=sizeof(T)*8>
vector<T,Dims> decode(const T _morton)
{
    vector<T,Dims> retval;
    for (uint16_t i=0; i<Dims; i++)
       retval[i] = impl::decode<Dims,T,BitDepth>(_morton>>i);
    return retval;
}

template<uint16_t Dims, typename T, uint16_t BitDepth=sizeof(T)*8>
T encode(const vector<T,Dims> coord)
{
    T retval = impl::separate_bits<Dims,T,BitDepth>(coord[0]);
    for (uint16_t i=1; i<Dims; i++)
       retval |= impl::separate_bits<Dims,T,BitDepth>(coord[1])<<i;
    return retval;
}
```

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,93 @@
...
+	float32_t rcpWhite2;
+};
+
+struct ACESParams
+{
+	using this_t = ACESParams;
+	static this_t create(float EV, float key = 0.18f, float Contrast = 1.f) {
+		this_t retval;
+		retval.gamma = Contrast;
+		retval.exposure = EV + log2(key * 0.77321666f);
```
devshgraphicsprogramming, line 38: did the GLSL have any comment (or git blame) about what the `0.77321666f` is?

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,93 @@
...
+		retval.keyAndManualLinearExposure = key * exp2(EV);
+		retval.rcpWhite2 = 1.f / (WhitePointRelToEV * WhitePointRelToEV);
+		return retval;
+	}
+
+	float32_t keyAndManualLinearExposure;
+	float32_t rcpWhite2;
+};
+
+struct ACESParams
```
devshgraphicsprogramming, line 32: 1. drop `Params` from the name
2. put the actual tonemapping as an `operator()`

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,93 @@
...
+float32_t3 reinhard(ReinhardParams params, float32_t3 rawCIEXYZcolor)
+{
+	float32_t exposureFactors = params.keyAndManualLinearExposure;
+	float32_t exposedLuma = rawCIEXYZcolor.y * exposureFactors;
+	float32_t colorMultiplier = (exposureFactors * (1.0 + exposedLuma * params.rcpWhite2) / (1.0 + exposedLuma));
+	return rawCIEXYZcolor * colorMultiplier;
+}
+
+float32_t3 aces(ACESParams params, float32_t3 rawCIEXYZcolor)
+{
```
devshgraphicsprogramming, line 56: these should be `operator()(const float32_t3 rawCIEXYZcolor)` inside the respective Reinhard and ACES structs

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,93 @@
...
+	const float32_t3x3 XYZ_RRT_Input = float32_t3x3(
+		float32_t3(1.594168310, -0.262608051, -0.231993079),
+		float32_t3(-0.6332771780, 1.5840380200, 0.0164147373),
+		float32_t3(0.00892840419, 0.03648501260, 0.87711471300)
+	);
+
+	// this is obviously fitted to some particular simulated sensor/film and display
+	float32_t3 v = mul(XYZ_RRT_Input, tonemapped);
+	float32_t3 a = v * (v + float32_t3(0.0245786, 0.0245786, 0.0245786)) - float32_t3(0.000090537, 0.000090537, 0.000090537);
+	float32_t3 b = v * (v * float32_t3(0.983729, 0.983729, 0.983729) + float32_t3(0.4329510, 0.4329510, 0.4329510)) + float32_t3(0.238081, 0.238081, 0.238081);
```
devshgraphicsprogramming, line 72: btw we have some `promote` thing made by @Przemog1 you can use for making vectors full of the same scalar

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+{
+	float32_t2 meteringWindowScale;
+	float32_t2 meteringWindowOffset;
+};
+
+template<uint32_t GroupSize, typename ValueAccessor, typename SharedAccessor, typename TexAccessor>
+struct geom_luma_meter {
+    using this_t = geom_luma_meter<GroupSize, ValueAccessor, SharedAccessor, TexAccessor>;
+
+    static this_t create(NBL_REF_ARG(LumaMeteringWindow) window, float32_t lumaMinimum, float32_t lumaMaximum)
```
devshgraphicsprogramming, line 36: CONST_REF nor just REF

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+        NBL_REF_ARG(SharedAccessor) sdata,
+        uint32_t2 sampleCount,
+        float32_t2 viewportSize
+    )
+    {
+        uint32_t2 coord = {
+            morton2d_decode_x(glsl::gl_LocalInvocationIndex()),
+            morton2d_decode_y(glsl::gl_LocalInvocationIndex())
+        };
+        uint32_t tid = workgroup::SubgroupContiguousIndex();
```
devshgraphicsprogramming, line 81: you should use the `tid` instead of LocalInvocationIndex for the morton code

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+
+        uint32_t2 sampleIndex = coord * GroupSize + float32_t2(glsl::gl_SubgroupID() + 1, glsl::gl_SubgroupInvocationID() + 1);
+        float32_t luma = 0.0f;
+
+        if (sampleIndex.x <= sampleCount.x && sampleIndex.y <= sampleCount.y) {
```
devshgraphicsprogramming, line 86: the computation of `sampleIndex` makes no sense here.

Also you shouldn't be bound testing anything, every invocation must sample otherwise your histogram/average is wrong.

Please compute a normalized `uv` coordinate relative to a full texture like this
```cpp
const float32_t2 uv  = (tileOffset+float32_t2(coord))*window.scale/totalInvocationsInDispatch+window.offset;
```
where `tileOffset` is basically `gl_WorkgroupID*gl_WorkgroupSize` but passed in from the outside

and `window.scale/totalInvocationsInDispatch` is precomputed obviously

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+template<uint32_t GroupSize, typename ValueAccessor, typename SharedAccessor, typename TexAccessor>
+struct geom_luma_meter {
+    using this_t = geom_luma_meter<GroupSize, ValueAccessor, SharedAccessor, TexAccessor>;
+
+    static this_t create(NBL_REF_ARG(LumaMeteringWindow) window, float32_t lumaMinimum, float32_t lumaMaximum)
+    {
+        this_t retval;
+        retval.window = window;
+        retval.minLuma = lumaMinimum;
+        retval.maxLuma = lumaMaximum;
```
devshgraphicsprogramming, line 41: typo, you've set the min and max equal to each other

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+    float32_t computeLuma(
+        NBL_REF_ARG(TexAccessor) tex,
+        uint32_t2 sampleCount,
+        uint32_t2 sampleIndex,
+        float32_t2 viewportSize
+    )
+    {
+        float32_t2 stride = window.meteringWindowScale / (sampleCount + float32_t2(1.0f, 1.0f));
+        float32_t2 samplePos = stride * sampleIndex;
+        float32_t2 uvPos = (samplePos + float32_t2(0.5f, 0.5f)) / viewportSize;
```
devshgraphicsprogramming, line 60: the windowOffset should have had the 0.5 texel shift inside itself, so viewportSize not needed

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+        float32_t3 color = colorspace::oetf::sRGB(tex.get(uvPos));
```
devshgraphicsprogramming, line 61: this is wrong, the HDR image you're reading (the noises example with FLOAT16 format) is scene referred in linear space

let the TextureAccessor worry about applying an EOTF if there is one needed.

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+        float32_t luma = dot(colorspace::sRGBtoXYZ[1], color);
```
devshgraphicsprogramming, line 62: actually require/expect a `TexAccessor::toXYZ` static member `float32_t3x3` variable :P

think of it as the `TexAccessor` telling everyone what the colorspace of the image is

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
devshgraphicsprogramming: P.S. its also more useful to take a precomputed `minLumaLog2` and `lumaLog2Range` (diff between log of max and log of min)

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+
+        luma = clamp(luma, minLuma, maxLuma);
+
+        return log2(luma / minLuma) / log2(maxLuma / minLuma);
```
devshgraphicsprogramming, line 66: you want to just do a `max(log2(luma)-minLumaLog2,0.f)` here without any normalization

also rename the function to `computeLumaLog2` or something

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+        uint32_t tid = workgroup::SubgroupContiguousIndex();
+
+        uint32_t2 sampleIndex = coord * GroupSize + float32_t2(glsl::gl_SubgroupID() + 1, glsl::gl_SubgroupInvocationID() + 1);
+        float32_t luma = 0.0f;
+
+        if (sampleIndex.x <= sampleCount.x && sampleIndex.y <= sampleCount.y) {
+            luma = computeLuma(tex, sampleCount, sampleIndex, viewportSize);
+            float32_t lumaSum = reduction(luma, sdata);
+
+            sdata.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 90: not sure you need the barrier, the last invocation will have the reduction value already

devshgraphicsprogramming: actually every invocation will

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+
+                uint32_t3 workgroupSize = glsl::gl_WorkGroupSize();
+                uint32_t workgroupIndex = dot(uint32_t3(workgroupSize.y * workgroupSize.z, workgroupSize.z, 1), glsl::gl_WorkGroupID());
+
+                val.atomicAdd(workgroupIndex & ((1 << glsl::gl_SubgroupSizeLog2()) - 1), lumaSumBitPattern);
+            }
+        }
+    }
+
+    LumaMeteringWindow window;
```
devshgraphicsprogramming, line 106: the window shouldn't be stored, just fed to `gatherLuma`, see my point about `gatherLuma` naming

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+            if (tid == GroupSize - 1) {
+                uint32_t3 workGroupCount = glsl::gl_NumWorkGroups();
+                uint32_t fixedPointBitsLeft = 32 - uint32_t(ceil(log2(workGroupCount.x * workGroupCount.y * workGroupCount.z))) + glsl::gl_SubgroupSizeLog2();
+
+                uint32_t lumaSumBitPattern = uint32_t(clamp((lumaSum - log2(minLuma)) * (log2(maxLuma) - log2(minLuma)), 0.f, float32_t((1 << fixedPointBitsLeft) - 1)));
+
+                uint32_t3 workgroupSize = glsl::gl_WorkGroupSize();
+                uint32_t workgroupIndex = dot(uint32_t3(workgroupSize.y * workgroupSize.z, workgroupSize.z, 1), glsl::gl_WorkGroupID());
+
+                val.atomicAdd(workgroupIndex & ((1 << glsl::gl_SubgroupSizeLog2()) - 1), lumaSumBitPattern);
```
devshgraphicsprogramming, line 101: there's a number of bugs here that we went over on discord.

1. your `lumaSum` should be just a sum of `max(log2(luma),minLumaLog2)`
2. `fixedPointBitsLeft` needs a better name and needs to come from the outside (precomputed)
3. the bitpattern needs to map `lumaSum==minLumaLog2*WorkgroupSize` to 0 and `lumaSum==maxLumaLog2*WorkgroupSize` to `(1<<fixedPointBitsLeft)-1`

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+        float32_t2 uvPos = (samplePos + float32_t2(0.5f, 0.5f)) / viewportSize;
+        float32_t3 color = colorspace::oetf::sRGB(tex.get(uvPos));
+        float32_t luma = dot(colorspace::sRGBtoXYZ[1], color);
+
+        luma = clamp(luma, minLuma, maxLuma);
+
+        return log2(luma / minLuma) / log2(maxLuma / minLuma);
+    }
+
+    void gatherLuma(
```
devshgraphicsprogramming, line 69: I'd rename this to `sampleLuma` because you want another method thats called either `gather` (because whole workgroup gathers partial averages or histograms) or `get` to actually obtain your measurement in the second dispatch

I recommend having the getter in the same struct, because you need much of the state and precomputed constants to be able to joint and decode the measurements to produce a single EV value.
(you can copy paste your struct initialization from push constants or whatever in both the measurement and tonemapping shader)

include/nbl/builtin/hlsl/math/morton.hlsl
nipunG314: HLSL doesn't have `constexpr`, right?

devshgraphicsprogramming: add a `#define NBL_CONSTEXPR_FUNC` to https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/hlsl/cpp_compat.hlsl

then you can do an empty define of `NBL_CONSTEXPR_FUNC` for HLSL_VERSION

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,122 @@
...
+
+struct LumaMeteringWindow
+{
+	float32_t2 meteringWindowScale;
+	float32_t2 meteringWindowOffset;
+};
+
+template<uint32_t GroupSize, typename ValueAccessor, typename SharedAccessor, typename TexAccessor>
+struct geom_luma_meter {
+    using this_t = geom_luma_meter<GroupSize, ValueAccessor, SharedAccessor, TexAccessor>;
```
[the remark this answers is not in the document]
nipunG314: Done

include/nbl/builtin/hlsl/math/morton.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
nipunG314: Fixed

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Fixed

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,113 @@
...
+
+    float32_t computeLuma(
+        NBL_REF_ARG(TexAccessor) tex,
+        uint32_t2 sampleCount,
+        uint32_t2 sampleIndex,
+        float32_t2 viewportSize
+    )
+    {
+        float32_t2 stride = window.meteringWindowScale / (sampleCount + float32_t2(1.0f, 1.0f));
+        float32_t2 samplePos = stride * sampleIndex;
```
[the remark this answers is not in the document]
nipunG314: Fixed

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Done

devshgraphicsprogramming:
What are the action points before completion?

devshgraphicsprogramming:
you might want to check your submodules if you've fast forwarded them Or inadvertently rewound them

include/nbl/builtin/hlsl/spirv_intrinsics/core.hlsl
```diff
@@ -61,37 +61,45 @@ pointer_t<StorageClass,T> copyObject([[vk::ext_reference]] T v);
...
+[[vk::ext_capability(spv::CapabilityPhysicalStorageBufferAddresses)]]
 [[vk::ext_capability(spv::CapabilityInt64Atomics)]]
 [[vk::ext_instruction(spv::OpAtomicISub)]]
 enable_if_t<is_same_v<T,uint64_t> || is_same_v<T,int64_t>, T> atomicISub([[vk::ext_reference]] T ptr, uint32_t memoryScope, uint32_t memorySemantics, T value);
 
 template<typename T, typename Ptr_T> // DXC Workaround
+[[vk::ext_capability(spv::CapabilityPhysicalStorageBufferAddresses)]]
 [[vk::ext_capability(spv::CapabilityInt64Atomics)]]
 [[vk::ext_instruction(spv::OpAtomicISub)]]
 enable_if_t<is_spirv_type_v<Ptr_T> && (is_same_v<T,uint64_t> || is_same_v<T,int64_t>), T> atomicISub(Ptr_T ptr, uint32_t memoryScope, uint32_t memorySemantics, T value);
```
devshgraphicsprogramming, line 105: no, not every pointer is a BDA pointer, see the `master` changes with `is_spirv_type`
https://github.com/Devsh-Graphics-Programming/Nabla/blob/0ecda6a8bcfaf8bba023d9cefcdc51be5d6a19b5/include/nbl/builtin/hlsl/type_traits.hlsl#L341C34-L341C50

devshgraphicsprogramming: you can now even check for `Ptr_T` being a pointer
https://github.com/Devsh-Graphics-Programming/Nabla/blob/0ecda6a8bcfaf8bba023d9cefcdc51be5d6a19b5/include/nbl/builtin/hlsl/spirv_intrinsics/core.hlsl#L101

include/nbl/builtin/hlsl/math/morton.hlsl
devshgraphicsprogramming: @Fletterio will take over

include/nbl/builtin/hlsl/math/morton.hlsl
devshgraphicsprogramming: @Fletterio will take over

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,100 @@
...
+{
+namespace hlsl
+{
+namespace tonemapper
+{
+
+template<typename T = float32_t>
+struct Reinhard
+{
+	using float_t = enable_if_t<is_floating_point<T>::value, T>;
```
devshgraphicsprogramming, line 21: with new concepts, the constraint should be applied via `NBL_PRIMARY_REQUIRES`

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,100 @@
...
+		this_t retval;
+		retval.keyAndManualLinearExposure = key * exp2(EV);
+		retval.rcpWhite2 = 1.f / (WhitePointRelToEV * WhitePointRelToEV);
+		return retval;
+	}
+
+	float_t3 operator()(float_t3 rawCIEXYZcolor) {
+		float_t exposureFactors = keyAndManualLinearExposure;
+		float_t exposedLuma = rawCIEXYZcolor.y * exposureFactors;
+		float_t colorMultiplier = (exposureFactors * (1.0 + exposedLuma * rcpWhite2) / (1.0 + exposedLuma));
```
devshgraphicsprogramming, line 35: the 1.0 needs to be dealt with in a funny way by making sure its the correct type constant 
```
const float_t unit = 1.0;
```
because in C++ a literal 1.0 is a double and in HLSL its a float32_t

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,100 @@
...
+	float_t keyAndManualLinearExposure;
+	float_t rcpWhite2;
+};
+
+template<typename T = float32_t>
+struct ACES
+{
+	using float_t = enable_if_t<is_floating_point<T>::value, T>;
+	using float_t3 = typename conditional<is_same_v<float_t, float32_t>, float32_t3, float16_t3>::type;
+	using float_t3x3 = typename conditional<is_same_v<float_t, float32_t>, float32_t3x3, float16_t3x3>::type;
```
devshgraphicsprogramming, line 48: again `vector<>` and `matrix<>` are things in HLSL

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,100 @@
...
+		this_t retval;
+		retval.gamma = Contrast;
+		const float_t reinhardMatchCorrection = 0.77321666f; // middle grays get exposed to different values between tonemappers given the same key
+		retval.exposure = EV + log2(key * reinhardMatchCorrection);
+		return retval;
+	}
+
+	float_t3 operator()(float_t3 rawCIEXYZcolor) {
+		float_t3 tonemapped = rawCIEXYZcolor;
+		if (tonemapped.y > 1.175494351e-38)
```
devshgraphicsprogramming, line 61: use `bit_cast<float_t>(numeric_limits<float_t>::min)`

src/nbl/builtin/CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "glsl/barycentric/utils.glsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/bda/__ref.hlsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/bda/__ptr.hlsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/bda/bda_accessor.hlsl")
+# luma metering
+LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/luma_meter/common.hlsl")
+LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/luma_meter/luma_meter.hlsl")
+# tonemapper
+LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/tonemapper/operators.hlsl")
```
devshgraphicsprogramming, line 41: I'd create `tonemapper/operators/reinhardt` and `tonemapper/operators/aces` instead of slapping everything into one file

src/nbl/builtin/CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "glsl/barycentric/utils.glsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/bda/__ref.hlsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/bda/__ptr.hlsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/bda/bda_accessor.hlsl")
+# luma metering
+LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/luma_meter/common.hlsl")
+LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/luma_meter/luma_meter.hlsl")
```
devshgraphicsprogramming, line 39: similarly, `luma_meter/luma_meter` is a tautology, a file each like `luma_meter/histogram` and `luma_meter/geom_avg` would be nice

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+{
+namespace hlsl
+{
+namespace luma_meter
+{
+
+template<uint32_t GroupSize, typename ValueAccessor, typename SharedAccessor, typename TexAccessor>
+struct geom_meter {
+    using float_t = typename SharedAccessor::type;
+    using float_t2 = typename conditional<is_same_v<float_t, float32_t>, float32_t2, float16_t2>::type;
```
devshgraphicsprogramming, line 27: even if doing color computation in `float16_t` this doesn't free you from doing texture coordinate calc in `float32_t`

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+        );
+
+        uint32_t3 workGroupCount = glsl::gl_NumWorkGroups();
+        uint32_t fixedPointBitsLeft = 32 - uint32_t(ceil(log2(workGroupCount.x * workGroupCount.y * workGroupCount.z))) + glsl::gl_SubgroupSizeLog2();
+
+        return (luma / (1 << fixedPointBitsLeft)) / sampleCount;
+    }
+
+    float_t sampleCount;
+    float_t2 lumaMinMax;
```
devshgraphicsprogramming, line 142: don't do weird things we used to do in GLSL (due to no scalar layout), have a separate variable for min and max

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+        sdata.workgroupExecutionAndMemoryBarrier();
+
+        uint32_t percentile40, percentile60;
+        sdata.get(BinCount * 0.4, percentile40);
+        sdata.get(BinCount * 0.6, percentile60);
+
+        return (int2Float(percentile40, lumaMinMax.x, lumaMinMax.y - lumaMinMax.x) + int2Float(percentile60, lumaMinMax.x, lumaMinMax.y - lumaMinMax.x)) / 2;
+    }
+
+    float_t sampleCount;
```
devshgraphicsprogramming, line 285: there's no use for this variable in the geometric luminosity meter

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
devshgraphicsprogramming: still outstanding for the geom meter

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+            )
+        );
+
+        uint32_t3 workGroupCount = glsl::gl_NumWorkGroups();
+        uint32_t fixedPointBitsLeft = 32 - uint32_t(ceil(log2(workGroupCount.x * workGroupCount.y * workGroupCount.z))) + glsl::gl_SubgroupSizeLog2();
+
+        return (luma / (1 << fixedPointBitsLeft)) / sampleCount;
+    }
+
+    float_t sampleCount;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: that was the purpose of the `rcpFirstPassWGCount` variable in the old GLSL

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+            template __call <SharedAccessor>(value, sdata);
+    }
+
+    float_t computeLumaLog2(
+        NBL_CONST_REF_ARG(MeteringWindow) window,
+        NBL_REF_ARG(TexAccessor) tex,
+        float_t2 shiftedCoord
+    )
+    {
+        float_t2 uvPos = shiftedCoord * window.meteringWindowScale + window.meteringWindowOffset;
```
devshgraphicsprogramming, line 51: precompute a scale and offset from the Metering Window + the number of workgroups + the workgroup size to apply to a `uint16_t2` unnormalized coordinate

right now you have waaay too many variables:
- tile Offset
- viewport size
- metering window 
which are being manipulated per-pixel

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+        float_t luma = (float_t)val_accessor.get(index & ((1 << glsl::gl_SubgroupSizeLog2()) - 1));
+        return luma / rangeLog2 + minLog2;
+    }
+
+    void sampleLuma(
+        NBL_CONST_REF_ARG(MeteringWindow) window,
+        NBL_REF_ARG(ValueAccessor) val,
+        NBL_REF_ARG(TexAccessor) tex,
+        NBL_REF_ARG(SharedAccessor) sdata,
+        float_t2 tileOffset,
```
devshgraphicsprogramming, line 92: why is tile Offset being provided from the outside? its a byproduct of your workgroupID, and workgroupSize-1 decoded as morton +1 in each dimension

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+        float_t2 shiftedCoord
+    )
+    {
+        float_t2 uvPos = shiftedCoord * window.meteringWindowScale + window.meteringWindowOffset;
+        float_t3 color = tex.get(uvPos);
+        float_t luma = (float_t)TexAccessor::toXYZ(color);
+
+        luma = clamp(luma, lumaMinMax.x, lumaMinMax.y);
+
+        return max(log2(luma), log2(lumaMinMax.x));
```
devshgraphicsprogramming, line 57: why `max` you already clamped!

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+    {
+        uint32_t tid = workgroup::SubgroupContiguousIndex();
+        uint32_t2 coord = {
+            morton2d_decode_x(tid),
+            morton2d_decode_y(tid)
+        };
+
+        float_t luma = 0.0f;
+        float_t2 shiftedCoord = (tileOffset + (float32_t2)(coord)) / viewportSize;
+        luma = computeLumaLog2(window, tex, shiftedCoord);
```
devshgraphicsprogramming, line 104: `luma` should be called `lumaLog2`

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+        float_t lumaSum = reduction(luma, sdata);
+
+        if (tid == GroupSize - 1) {
```
devshgraphicsprogramming, line 107: its somewhat semantically cleaner to pick the first, instead of last, esp since its a reduction you performed before

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+        float_t luma = (float_t)TexAccessor::toXYZ(color);
+
+        luma = clamp(luma, lumaMinMax.x, lumaMinMax.y);
+
+        return max(log2(luma), log2(lumaMinMax.x));
+    }
+
+    void uploadFloat(
+        NBL_REF_ARG(ValueAccessor) val_accessor,
+        uint32_t index,
```
devshgraphicsprogramming, line 62: don't give bad names to variables, this needs to be called `workGroupIndex`

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+        };
+
+        float_t luma = 0.0f;
+        float_t2 shiftedCoord = (tileOffset + (float32_t2)(coord)) / viewportSize;
+        luma = computeLumaLog2(window, tex, shiftedCoord);
+        float_t lumaSum = reduction(luma, sdata);
+
+        if (tid == GroupSize - 1) {
+            uint32_t3 workgroupCount = glsl::gl_NumWorkGroups();
+            uint32_t workgroupIndex = (workgroupCount.x * workgroupCount.y * workgroupCount.z) / 64;
```
devshgraphicsprogramming, line 109: you're computing the wrong thing, every workgroup gets the same index 🤦 

Also the original code was touching the same address with every for the MEAN meter mode

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+
+    void uploadFloat(
+        NBL_REF_ARG(ValueAccessor) val_accessor,
+        uint32_t index,
+        float_t val,
+        float_t minLog2,
+        float_t rangeLog2
+    )
+    {
+        uint32_t3 workGroupCount = glsl::gl_NumWorkGroups();
```
devshgraphicsprogramming, line 68: take the workgroup count and workgroup XYZ coordinate (or workgroup index)  from outside (as function arguments) otherwise in the presence of solutions such as virtual workgroups or persistent threads, this whole thing will fall apart

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
devshgraphicsprogramming: also you should have the min and max  precomputed with `log2` already applied

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
devshgraphicsprogramming: btw if you have the `log2` already precomputed then you can do
```cpp
return min(spirv::nMax(log2(luma),lumaMinLog2),lumaMaxLog2);
```

`nMin` is a special SPIR-V version of min that will return the other operand if another is NaN (which happens on log of negative value or 0)

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+        uint32_t fixedPointBitsLeft = 32 - uint32_t(ceil(log2(workGroupCount.x * workGroupCount.y * workGroupCount.z))) + glsl::gl_SubgroupSizeLog2();
+
+        uint32_t lumaSumBitPattern = uint32_t(clamp((val - minLog2) * rangeLog2, 0.f, float32_t((1 << fixedPointBitsLeft) - 1)));
```
devshgraphicsprogramming, line 71: lets write some docs for this....

The `val` was produced by a workgroup reduction is performed of values in the `[MinLog2,MaxLog2]` range

Which makes the `scaledLogLuma` (the variable that should hold `(val-minLog2)*rangeLog2`) is between 0 and WorkGroupSize

This value is atomic added by N workgroups

You now want to represent it in Fixed Point during the atomic add, but not be vulnerable to overflow, this means the worst case is adding N times WorkGroupSize.

This means that we need to multiply the  by `(2^32-1)/N` precomputed as a float or if you must round up `N` to PoT and see how many bits are left (512 workgroups, means 9 bits, so 23 are left). To avoid rounding precision errors, the PoT method is chosen.

**I have no clue where you're getting `+SubgroupSizeLog2` from.**

Also the value of `(1<<fixedPointBitsLeft)-1` must be precomputed in `create` and stored as a member

IT should be as easy as
```cpp
const uint32_t scaledLumaLog2BitPattern = uint32_t((val-lumaMinLog2)*maxIncrement_over_lumaRangeLog2+float_t(0.5));
```

where `maxIncrement = (0x1u<<(32u-uint32_t(ceil(log2(WorkGroupCount*WorkGroupSize)))))-1;`

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+
+        val_accessor.atomicAdd(index & ((1 << glsl::gl_SubgroupSizeLog2()) - 1), lumaSumBitPattern);
```
devshgraphicsprogramming, line 73: no, always the same address should be added to, if you wanted to stagger, then you should stagger based on modulo of the workgroup index

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
```diff
@@ -0,0 +1,100 @@
...
+namespace hlsl
+{
+namespace tonemapper
+{
+
+template<typename T = float32_t>
+struct Reinhard
+{
+	using float_t = enable_if_t<is_floating_point<T>::value, T>;
+	using float_t3 = typename conditional<is_same_v<float_t, float32_t>, float32_t3, float16_t3>::type;
```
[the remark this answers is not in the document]
nipunG314: Done

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+
+    float_t downloadFloat(
+        NBL_REF_ARG(ValueAccessor) val_accessor,
+        uint32_t index,
+        float_t minLog2,
+        float_t rangeLog2
+    )
+    {
+        float_t luma = (float_t)val_accessor.get(index & ((1 << glsl::gl_SubgroupSizeLog2()) - 1));
+        return luma / rangeLog2 + minLog2;
```
devshgraphicsprogramming, line 84: again, you're getting random floats based on workgroup index which _thankfully_ was always the same (rare case of two wrongs making a right)

Again if you wanted to stagger, you should use entire subgroup to load the values, then subgroup reduce them

just converting to `float_t` is not the correct way to decode, you should divide by

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+            downloadFloat(
+                val,
+                tid,
+                log2(lumaMinMax.x),
+                log2(lumaMinMax.y / lumaMinMax.x)
+            )
+        );
+
+        uint32_t3 workGroupCount = glsl::gl_NumWorkGroups();
+        uint32_t fixedPointBitsLeft = 32 - uint32_t(ceil(log2(workGroupCount.x * workGroupCount.y * workGroupCount.z))) + glsl::gl_SubgroupSizeLog2();
```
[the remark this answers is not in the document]
devshgraphicsprogramming: You must precompute the `fixedPointsBitsLeft` in the `create` method (and it needs to know how many invocations you'll be running the sample step)

include/nbl/builtin/hlsl/tonemapper/operators.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
```diff
@@ -0,0 +1,293 @@
...
+    float_t downloadFloat(
+        NBL_REF_ARG(ValueAccessor) val_accessor,
+        uint32_t index,
+        float_t minLog2,
+        float_t rangeLog2
+    )
+    {
+        float_t luma = (float_t)val_accessor.get(index & ((1 << glsl::gl_SubgroupSizeLog2()) - 1));
+        return luma / rangeLog2 + minLog2;
+    }
```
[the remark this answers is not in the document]
nipunG314: Done

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Fixed

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Fixed

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Fixed

include/nbl/builtin/hlsl/luma_meter/luma_meter.hlsl
nipunG314: Fixed

include/nbl/builtin/hlsl/luma_meter/histogram.hlsl
```diff
@@ -0,0 +1,320 @@
...
+                vid,
+                histo.get(vid)
+            );
+        }
+        sdata.workgroupExecutionAndMemoryBarrier();
+
+        int_t lower, upper;
+        if (tid == 0)
+        {
+            const uint32_t lowerPercentile = uint32_t(BinCount * lowerBoundPercentile);
```
AnastaZIuk: use `totalSamples * percentile` and `totalSamples` can be `histo[BinCount-1]`

`lowerPercentile` is in bin units while histo prefix sums are in samples, you compare different units hence the threshold is tiny and the median collapses to the first bins, your EV moves toward `log2(lumaMin)` and tonemapping is biased dark

counterexample https://godbolt.org/z/49WjcazMr

PR #734, opened 2024-08-21
Add some default barriers to BdaAccessor

Fletterio:
## Description
Adds default barriers to bda accessors

## Testing 
None


include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -93,15 +93,19 @@ struct exchangeValues<SharedMemoryAdaptor, float64_t>
     }
 };
 
+// Get the required size (in number of uint32_t elements) of the workgroup shared memory array needed for the FFT
+template <typename scalar_t, uint32_t WorkgroupSize>
+NBL_CONSTEXPR uint32_t sharedMemSize = 2 * WorkgroupSize * (sizeof(scalar_t) / sizeof(uint32_t));
```
devshgraphicsprogramming, line 98: what namespace are you in?

Fletterio: `workgroup::fft`, why?

Fletterio: okay I see this mehtod should be user-facing while exchange values in the same namespace should not be user-facing. Where should I put this instead?

devshgraphicsprogramming: okay if you want to make a template variable (before this wasn't possible, which is why workgroup scan stuff is structs) thats fine, can be in `workgroup::fft` namespace

Then have it as it is `SharedMemorySize` and make sure to report in bytes, or call it `SharedMemoryDWORDs` if you want to count in multiples of `uint32_t`

Furthermore `sizeof(complex<scalar_t>)` is better than `2*sizeof(scalar_t)` for understanding a readability.

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -224,9 +223,29 @@ struct bitfieldInsert
...
+template<typename T>
+struct bitfieldReverse
+{
+    static T __call( T base )
+    {
+        return spirv::bitFieldReverse<T>( base );
+    }
+};
+
+} //namespace impl
```
devshgraphicsprogramming, line 243: at this point, what does going through a struct for insert and reverse give you?

Fletterio: I was replicating what bitfield extract was doing but now that you called it out I realize it's only doing that to have partial specializations. Fixing now

PR #735, opened 2024-08-27
device_features: actually disable `alphaToOne`

alichraghi:
## Description
in https://github.com/Devsh-Graphics-Programming/Nabla/commit/21d6979d44837d5f31ba7552236e9d2a1b6b1826, `alphaToOne` was set to `false` but somehow regressed later.

## Testing 
I was able to run helloswapchain example after this change.

### DIFF
--- a/src/nbl/video/device_capabilities/device_features.json
+++ b/src/nbl/video/device_capabilities/device_features.json
@@ -141,7 +141,7 @@
         {
             "type": "bool",
             "name": "alphaToOne",
-            "value": true,
+            "value": false,
             "comment": ["Some AMD don't support"]
         },
         {


devshgraphicsprogramming:
@nipunG314 check and merge

@alichraghi why are you force pushing to your branches?

alichraghi:
> why are you force pushing to your branches?

I didn't know about the commit signing thing at first.

nipunG314:
@alichraghi This change happened when we transitioned from a handwritten header file to a generated one. In the original header, the value was `true`.

alichraghi:
But either way this should be set to `false` because even some semi new-gen AMD GPUs doesn't support it. (@Erfan-Ahmadi can verify this).

PR #736, opened 2024-08-28
video: remove `*_BUFFER_DYNAMIC` descriptors and make bindings always have the  `UPDATE_AFTER_BIND_BIT` flag

### LINKED ISSUE #694: Remove _DYNAMIC descriptors and make descriptor bindings alway have the `UPDATE_AFTER_BIND_BIT` flag
## Description

Bindings with `UPDATE_AFTER_BIND_BIT` and `DYNAMIC` SSBO and UBO are not allowed within the same Descriptor Set Layout.

The extra array of offsets during binding a set for DYNAMIC bindings is annoying unclear API. The way the offsets sum with the Descriptor Set's Buffer Range.

SSBOs are obsoleted by BDA anyway, and WHOLE_BUFFER never worked with dynamic offsets.

If we can have 500k UBOs which can be indexed, then we can create lots of bindings at different offsets to index instead of playing with dynamic offsets.

Also `UPDATE_AFTER_BIND_BIT` relaxes external synchronization requirements, making descrptor sets more thread safe.

alichraghi:
## Description
Closes #694 

## Testing 
I Ran every example that used descriptor sets but some of them didn't compile for being outdated.


devshgraphicsprogramming:
This interferes with a lot of stuff, need to wait before merging

PR #740, opened 2024-09-04
Content hashes for loader-external assets

tad1:
## Description

Grant/invalidate hashes for `ICPUImage`, `ICPUBuffer`, `ICPUShader` in case of creation/modification. With exception to: internal useage, `ICPUBuffer`'s non-asset use, code marked as deprecated.

## Testing 

compiles and runs Nabla + 24th example. 
tested small subset of filters.

## TODO list:

- [x] decide if hash for  an `ICPUImage`'s buffer should be calculated, or invalidated


src/nbl/asset/interchange/CImageLoaderJPG.cpp
```diff
@@ -339,8 +339,12 @@ asset::SAssetBundle CImageLoaderJPG::loadAsset(system::IFile* _file, const asset
...
 	image->setBufferAndRegions(std::move(buffer), regions);
 
-	auto hash = contentHasher.finalizeSeq();
-	image->setContentHash(hash);
+	{
+		auto hash = contentHasher.finalizeSeq();
+		auto* buffer = image->getBuffer();
+		buffer->setContentHash(buffer->computeContentHash());
+		image->setContentHash(hash);
+	}
```
tad1, line 347: >  decide if hash for an ICPUImage's buffer should be calculated, or invalidated

Here's an example code. 
If computing hash for an `ICPUImage`'s buffer is unnecessary, we can invalidate it instead.

devshgraphicsprogramming: the hash of the buffer is unecessary indeed (ICPUImage never looks at the buffer data hash, because it has no bearing on the image data uniqueness).

include/nbl/asset/filters/CRegionBlockFunctorFilter.h
```diff
@@ -59,6 +59,11 @@ class CRegionBlockFunctorFilter : public CImageFilter<CRegionBlockFunctorFilter<
 
 			CBasicImageFilterCommon::executePerBlock<ExecutionPolicy,Functor>(std::forward<ExecutionPolicy>(policy),state->image,*state->regionIterator,state->functor);
 
+			if constexpr(!ConstImage)
+			{
+				state->image->setContentHash(core::INVALID_HASH);
+			}
+
```
devshgraphicsprogramming, line 66: does the FlattenRegions filter implement stuff in terms of `execute` defined here ?

include/nbl/core/hash/blake.h
```diff
@@ -21,6 +21,8 @@ struct blake3_hash_t final
 	uint8_t data[BLAKE3_OUT_LEN];
 };
 
+const blake3_hash_t INVALID_HASH = {};
+
```
devshgraphicsprogramming, line 25: either call it `EmptyHash` or move to `IAsset`

tad1: moved to `IPreHashed`

include/nbl/asset/filters/CRegionBlockFunctorFilter.h
tad1: FlattenRegions don't use RegionBlockFunctor

devshgraphicsprogramming: ok no problem then

include/nbl/core/hash/blake.h
devshgraphicsprogramming: you only moved to a header, but you didn't move it to become a static const variable!

include/nbl/asset/IPreHashed.h
```diff
@@ -10,13 +10,14 @@
...
 
-const core::blake3_hash_t INVALID_HASH = {};
 
 //! Sometimes an asset is too complex or big to be hashed, so we need a hash to be set explicitly.
 //! Meant to be inherited from in conjunction with `IAsset`
 class IPreHashed : public IAsset
 {
 	public:
+		constexpr static inline core::blake3_hash_t INVALID_HASH = {};
+
```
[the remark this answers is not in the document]
devshgraphicsprogramming: ok discord convo reveals its a bad idea, needs to be `constexpr` but different

PR #741, opened 2024-09-06
add git info to nabla

tad1:
## Description

Adds git meta to source code.

## Testing 

tested on dynamic build (feat. 24th example).
on git commit, project recompile only `git_info.cpp` - as intended

## Todo
- [x] fix vars


tad1:
Works internally, if we need to export that from `Nabla.dll` add `NBL_API2 GitInfo getGitInfo();`

src/nbl/CMakeLists.txt
```diff
@@ -653,6 +658,20 @@ glue_source_definitions(Nabla _NBL_SOURCE_DEFINITIONS_)
 set(_NBL_DEFINE_FILE_WRAPPER_ ${CMAKE_CURRENT_BINARY_DIR}/include/define.h)
 write_source_definitions("${_NBL_DEFINE_FILE_WRAPPER_}" "${_NBL_SOURCE_DEFINITIONS_}")
 
+add_custom_target(CheckGit COMMAND ${CMAKE_COMMAND}
+	-DWORKING_DIRECTORY:PATH=${NBL_ROOT_PATH}
+	-DGIT_EXECUTABLE:PATH=${GIT_EXECUTABLE}
+	-DINPUT_FILE:PATH="${NBL_ROOT_PATH}/src/nbl/git_info.cpp.in"
+	-DOUTPUT_FILE:PATH="${CMAKE_BINARY_DIR}/src/nbl/git_info.cpp"
+	-P ${NBL_ROOT_PATH}/cmake/scripts/nbl/gitInfo.cmake
```
AnastaZIuk, line 666: `"${NBL_ROOT_PATH}/cmake/scripts/nbl/gitInfo.cmake"`

src/nbl/CMakeLists.txt
```diff
@@ -653,6 +658,20 @@ glue_source_definitions(Nabla _NBL_SOURCE_DEFINITIONS_)
...
+	BYPRODUCTS ${CMAKE_BINARY_DIR}/src/nbl/git_info.cpp
+)
+
+set(INPUT_FILE "${NBL_ROOT_PATH}/src/nbl/git_info.cpp.in")
+set(OUTPUT_FILE "${CMAKE_BINARY_DIR}/src/nbl/git_info.cpp")
+set(WORKING_DIRECTORY ${NBL_ROOT_PATH})
+include("${NBL_ROOT_PATH}/cmake/scripts/nbl/gitInfo.cmake")
```
AnastaZIuk, line 673: `include(scripts/nbl/gitInfo)`

cmake/scripts/nbl/gitInfo.cmake
```diff
@@ -0,0 +1,18 @@
+# A configure-time and build-time script, that produces source with current git commit info
```
AnastaZIuk, line 1: > (...) build-time script

actually we should not proxy pass the script to be executed at build time nor create a build target for that purpose - it can be all done nicely at configure time without introducing special target to the build system

we could just `configure_file` and that will do the job, I would remove the script + `CheckGit` target and move the script's content directly where you include it

by the way if it complains about `"${OUTPUT_FILE}"` being not `GENERATED` when you create the library with it as part of your sources then you can simply set [this property](https://cmake.org/cmake/help/latest/prop_sf/GENERATED.html#prop_sf:GENERATED) for the output

tad1: With build-time check we make sure that: for any "bug report" (from someone who builds Nabla on their own) will have correct git-hash.

In case of configure-time only: checking out different commit/branch without re-configuring CMake, we will keep the old hash.  For releases (i.e. `Nabla.dll`) config-time only is good enough.

AnastaZIuk: > checking out different commit/branch without re-configuring CMake, may keep the old hash

okay fair argument, but if we want to make it well then integrate [this](https://github.com/andrew-hardin/cmake-git-version-tracking/tree/master) repo as 3rdparty instead writing our own git check actually (it does exactly what you do now but with a few improvements regarding regenerating the .cpp if the git hash has actually changed + much more metadata there is already written we just plug it & it works -> you can read [this](https://jonathanhamberg.com/post/cmake-embedding-git-hash/) article for details why it's better and what they do extra)

AnastaZIuk: okay I forked the [repo](https://github.com/Devsh-Graphics-Programming/cmake-git-version-tracking) - use it & add short version of commit sha, the only thing to do then is to link the target to Nabla afterwards

tad1: > with a few improvements regarding regenerating the .cpp if the git hash has actually changed

Actually we don't need their solution for this issue. I used CMake 3.10+ feature for this issue (our minimal: 3.29). This gives much cleaner script (see [gitInfo.cmake](https://github.com/Devsh-Graphics-Programming/Nabla/blob/16f83bf555146637528e11c2c2d9d13a3afb3c95/cmake/scripts/nbl/gitInfo.cmake)).

AnastaZIuk: explained in discord, we stay with the fork + introduce a small modification to it

AnastaZIuk: - add short sha
- modify the template cpp & header to keep our meta data in nice structure
- we need to use NBL_API2 for the output header's structure, so we just have to include our define.h
- link to Nabla, profit

AnastaZIuk: sorry, the `NBL_API2` is in [BuildConfigOptions.h](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a10ab3d0ce1626a1d5ea8ed7d2e8a0aca4665462/include/nbl/config/BuildConfigOptions.h.in#L91) but it's okay, lets give the meta library target proper [include search directory](https://cmake.org/cmake/help/latest/command/target_include_directories.html) and include the config

AnastaZIuk: as you are on it you can fix this [`// DDL exports`](https://github.com/Devsh-Graphics-Programming/Nabla/blob/a10ab3d0ce1626a1d5ea8ed7d2e8a0aca4665462/include/nbl/config/BuildConfigOptions.h.in#L90C1-L90C15) typo xD

3rdparty/CMakeLists.txt
```diff
@@ -275,6 +275,10 @@ add_library(spirv_cross OBJECT
 )
 target_compile_definitions(spirv_cross PUBLIC SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS)
 
+# cmake-git-version-tracking
+add_subdirectory(git-version-tracking EXCLUDE_FROM_ALL)
```
devshgraphicsprogramming, line 279: wont this just produce a `git hash` info of the submodule and not nabla parent repo itself?

include/nabla.h
```diff
@@ -18,7 +18,7 @@
 #define __NABLA_H_INCLUDED__
 
 // meta info
-#include "nbl/git_info.h"
+#include "git_info.h"
```
devshgraphicsprogramming, line 21: can you show me what's in this header after generation

3rdparty/CMakeLists.txt
```diff
@@ -275,6 +275,11 @@ add_library(spirv_cross OBJECT
 )
 target_compile_definitions(spirv_cross PUBLIC SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS)
 
+# cmake-git-version-tracking
+set(GIT_WORKING_DIR ${NBL_ROOT_PATH})
```
AnastaZIuk, line 279: `"${NBL_ROOT_PATH"`

include/nabla.h
tad1: git_info.cpp after generation
```cpp
#include "git_info.h"

namespace nbl{
    const GitInfo git_info = {
    .isPopulated = true,
    .hasUncommittedChanges = true,
    .commitAuthorName =  "tad1",
    .commitAuthorEmail = "tadziobrzeski@gmail.com",
    .commitHash = "84ff6ebd8a2db2f88de01e5bc7d142ebb205aa7d",
    .commitShortHash = "84ff6ebd8",
    .commitDate = "2024-09-09 19:18:37 +0200",
    .commitSubject = "use NBL_ROOT_PATH as git wd",
    .commitBody = "",
    .describe = "shared_build_msvc_assemblies-3959-g84ff6ebd8",
    .branchName = "tad_add_git_commit_hash",
    .latestTag = "v0.4-delta-2681-g84ff6ebd8",
    .latestTagName = "v0.4-delta"
    };

    const GitInfo getGitInfo() {
        return git_info;
    };
} // namespace nbl
```

internal useage
```cpp
#include "git_info.h" //or #include "nabla.h" 
// ....
m_logger.log("last commit %s:%s", system::ILogger::ELL_ERROR, nbl::git_info.commitShortHash, nbl::git_info.commitSubject);
// ....
```

exterrnal `.dll` useage
```cpp
#include "nabla.h" 
// ....
m_logger->log("last commit %s:%s", ILogger::ELL_ERROR, nbl::getGitInfo().commitShortHash, nbl::getGitInfo().commitSubject);
// ....
```

AnastaZIuk:
nice, lets make it a real tool (only really a few small changes required)

their top CMake you turn into:

```cmake
cmake_minimum_required(VERSION 3.2...3.30)
project(cmake_git_version_tracking LANGUAGES C)

set(PRE_CONFIGURE_FILE "git_info.cpp.in")

function(NBL_ADD_GIT_TRACKING_META_LIBRARY NBL_TARGET_NAME POST_CONFIGURE_FILE GIT_WORKING_DIR <more args if needed>)
  set(NBL_GIT_CHECK_TARGET_NAME "${NBL_TARGET_NAME}_check_git") # modify the include file for it to create ${NBL_GIT_CHECK_TARGET_NAME}" target

  include(git_watcher.cmake)

  add_library(${NBL_TARGET_NAME} STATIC ${POST_CONFIGURE_FILE})
  target_include_directories(${NBL_TARGET_NAME} SYSTEM PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
  add_dependencies(${NBL_TARGET_NAME} "${NBL_GIT_CHECK_TARGET_NAME}")
endfunction()
```

you should also take care of the template header probably (also should be immutable & output you should be able to setup as for cpp, could set the post output header file with the function args), then the usage is that we add this subdirectory and

```cmake
NBL_ADD_GIT_TRACKING_META_LIBRARY(nabla-repo-meta "<binary output for Nabla post source cpp/header meta file>" "${NBL_ROOT_PATH}")
NBL_ADD_GIT_TRACKING_META_LIBRARY(dxc-repo-meta "<binary output for DXC post source cpp/header meta file>" "<dxc top path>")
```

we create as many meta libs as we want, then we just link them to Nabla in which we have

```c++
enum E_GIT_REPO_META : uint8_t
{
  EGRM_NABLA,
  // EGRM_DXC,
  // ..
};

// ...

nbl::getGitInfo(E_GIT_REPO_META repo = EGRM_NABLA) { return <array with git info?>[repo]; }
```

now onto `<array with git info?>` - it doesn't matter how you store those git info structs for me, it could be an array to gather them all but if betta idea then go ahead

the only requirements are that
- tool works with "similar" logic to our builtin resources, so it basically outputs single cpp & header file per the util call
- `nbl::getGitInfo(E_GIT_REPO_META)` returns pointer/reference to constexpr struct

AnastaZIuk:
remember that the only meta we add now is `EGRM_NABLA`, its just that the design allows to easily extend to more repos which meta can be easily embeded into Nabla with minimum lines of CMake code

3rdparty/CMakeLists.txt
AnastaZIuk: back then probably, now Tadzio fixed it and with the [tool ](https://github.com/Devsh-Graphics-Programming/Nabla/pull/741#issuecomment-2338841799) we control it per single invocation

include/nabla.h
devshgraphicsprogramming: looks good

AnastaZIuk:
> * tool works with "similar" logic to our builtin resources (single cpp, header and namespace file per the util call)

okay we agreed in DM to do this

3rdparty/CMakeLists.txt
```diff
@@ -275,6 +275,13 @@ add_library(spirv_cross OBJECT
 )
 target_compile_definitions(spirv_cross PUBLIC SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS)
 
+# cmake-git-version-tracking
+set(GIT_WORKING_DIR "${NBL_ROOT_PATH}")
```
AnastaZIuk, line 279: you don't need it anymore, we have `NBL_ADD_GIT_TRACKING_META_LIBRARY` now

PR #745, opened 2024-09-25
Shader caching fixes

Erfan-Ahmadi:
Shader Caching Fixes:
- Writing to writeCache in shader cache hit scenario
- ISystem `deleteFile` function


include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -399,7 +399,7 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 					return retVal;
 				}
 
-				NBL_API2 core::smart_refctd_ptr<asset::ICPUShader> find(const SEntry& mainFile, const CIncludeFinder* finder) const;
+				NBL_API2 SEntry find(const SEntry& mainFile, const CIncludeFinder* finder) const;
```
devshgraphicsprogramming, line 402: why do you return the whole entry?

PR #747, opened 2024-10-01
HLSL-Preprocessor fixes

AnastaZIuk:

### DIFF
--- a/src/nbl/asset/utils/CHLSLCompiler.cpp
+++ b/src/nbl/asset/utils/CHLSLCompiler.cpp
@@ -333,7 +333,7 @@ std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADE
         for (auto i=context.begin(); i!=context.end(); i++) {
             auto value = i->get_value();
             stream << i->get_value();
-            
+
             // TODO: replace this hack with `support_option_emit_contnewlines` flag when Boost::Wave respect it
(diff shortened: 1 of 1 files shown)


src/nbl/asset/utils/CHLSLCompiler.cpp
```diff
@@ -328,8 +329,19 @@ std::string CHLSLCompiler::preprocessShader(std::string&& code, IShader::E_SHADE
...
             stream << i->get_value();
+            
+            // TODO: replace this hack with `support_option_emit_contnewlines` flag when Boost::Wave respect it
+            // See also https://github.com/Devsh-Graphics-Programming/Nabla/issues/746
+            if (emit_nl_after != -1) {
+                emit_nl_after--;
+                if (emit_nl_after == 0) stream << "\n";
+            }
+            if (boost::ends_with(stream.str(), "#pragma shader_stage(")) emit_nl_after = 2;
+        }
```
devshgraphicsprogramming, line 344: don't we use Boost Wave in GLSL compiler as well?

alichraghi: no we don't

devshgraphicsprogramming: yeah I just foundout with @nipunG314 today XD

devshgraphicsprogramming:
@Erfan-Ahmadi merge if it pleases you.

PR #748, opened 2024-10-04
shader cache polishing

alichraghi:


include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -185,7 +185,7 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 				// Used to check compatibility of Caches before reading
 				constexpr static inline std::string_view VERSION = "1.0.0";
 
-				using hash_t = std::array<uint64_t,4>;
+				using hash_t = std::array<uint8_t,32>;
```
devshgraphicsprogramming, line 188: there's a `core::blake3_hash_t` type we use for this

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -33,7 +33,7 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 				{
 					system::path absolutePath = {};
 					std::string contents = {};
-					std::array<uint64_t, 4> hash = {}; // TODO: we're not yet using IFile::getPrecomputedHash(), so for builtins we can maybe use that in the future
+					std::array<uint8_t, 32> hash = {}; // TODO: we're not yet using IFile::getPrecomputedHash(), so for builtins we can maybe use that in the future
```
devshgraphicsprogramming, line 36: use the blake3 type

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -351,9 +346,12 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 
 						// Now add the mainFileContents and produce both lookup and early equality rejection hashes
 						hashable.insert(hashable.end(), mainFileContents.begin(), mainFileContents.end());
-						hash = nbl::core::XXHash_256(hashable.data(), hashable.size());
-						lookupHash = hash[0];
-						for (auto i = 1u; i < 4; i++) {
+
+						core::blake3_hasher hasher;
+						hasher.update(hashable.data(), hashable.size());
+						hash = { *static_cast<core::blake3_hash_t>(hasher).data };
```
devshgraphicsprogramming, line 352: just use the `core::blake3_hash_t` type

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -351,9 +346,12 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
+						lookupHash = std::bit_cast<uint64_t, uint8_t[8]>({hash[0], hash[1], hash[2], hash[3], hash[4], hash[5], hash[6], hash[7]});
+						for (auto i = 8u; i < 32; i++) {
 							core::hash_combine<uint64_t>(lookupHash, hash[i]);
```
devshgraphicsprogramming, line 355: there's a `std::hash` specialization for `core::blake3_hash_t` that lets you make a STD-compatible size_T hash from the large hash

devshgraphicsprogramming:
Did you test the hash cache works nicely on example 23 (arithmetic test) and  62 CAD ?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -240,6 +239,8 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 								return true;
 							}
 
+							std::string sourceIdentifier;
+
```
devshgraphicsprogramming, line 243: why is this now public?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -283,6 +283,9 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 								return retVal;
 							}
 
+							IShader::E_SHADER_STAGE stage;
+							SPreprocessorArgs preprocessorArgs;
+
```
devshgraphicsprogramming, line 288: why is this now public?

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -362,20 +360,23 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
-						  dependencies(other.dependencies), cpuShader(other.cpuShader) {}
+						dependencies(other.dependencies), spirv(other.spirv), uncompressedSize(other.uncompressedSize) {}
 				
 					inline SEntry& operator=(SEntry& other) = delete;
 					inline SEntry(SEntry&& other) = default;
 					// Used for late initialization while looking up a cache, so as not to always initialize an entry even if caching was not requested
 					inline SEntry& operator=(SEntry&& other) = default;
 
+					core::smart_refctd_ptr<ICPUShader> decodeShader() const;
+
```
devshgraphicsprogramming, line 371: nitpick, `decompress` better naming than `decode`

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -22,6 +25,73 @@ IShaderCompiler::IShaderCompiler(core::smart_refctd_ptr<system::ISystem>&& syste
...
+    }
+
+    auto retVal = compileToSPIRV_impl(code, options, options.writeCache ? &dependencies : nullptr);
+    // compute the SPIR-V shader content hash
+    {
+        auto backingBuffer = retVal->getContent();
+        const_cast<ICPUBuffer*>(backingBuffer)->setContentHash(backingBuffer->computeContentHash());
+    }
+
+    std::cout << "writeCache: " << options.writeCache << "\n";
```
devshgraphicsprogramming, line 59: we have our own logger interface for a reason

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -224,7 +294,7 @@ auto IShaderCompiler::CIncludeFinder::tryIncludeGenerators(const std::string& in
 
 core::smart_refctd_ptr<asset::ICPUShader> IShaderCompiler::CCache::find(const SEntry& mainFile, const IShaderCompiler::CIncludeFinder* finder) const
 {
-    return find_impl(mainFile, finder)->cpuShader;
+    return find_impl(mainFile, finder)->decodeShader();
```
devshgraphicsprogramming, line 297: ooof, seems like nobody checked (even before) if the thing was even found before dereferencing an iterator

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -273,10 +343,10 @@ core::smart_refctd_ptr<ICPUBuffer> IShaderCompiler::CCache::serialize() const
         // We keep a copy of the offsets and the sizes of each shader. This is so that later on, when we add the shaders to the buffer after json creation
         // (where the params array has been moved) we don't have to read the json to get the offsets again
         offsets[i] = shaderBufferSize;
-        sizes[i] = entry.cpuShader->getContent()->getSize();
+        sizes[i] = entry.spirv->getSize();
 
         // And add the params to the shader creation parameters array
-        shaderCreationParams.emplace_back(entry.cpuShader->getStage(), entry.cpuShader->getContentType(), entry.cpuShader->getFilepathHint(), sizes[i], shaderBufferSize);
+        shaderCreationParams.emplace_back(entry.compilerArgs.stage, IShader::E_CONTENT_TYPE::ECT_SPIRV, entry.compilerArgs.preprocessorArgs.sourceIdentifier.data(), sizes[i], shaderBufferSize);
```
devshgraphicsprogramming, line 349: what are  you doing!?

devshgraphicsprogramming, line 349: do we even need to serialize the content type for any reason / keep it in `CPUShaderCreationParams` ?

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -119,7 +189,7 @@ auto IShaderCompiler::CIncludeFinder::getIncludeStandard(const system::path& req
 
     core::blake3_hasher hasher;
     hasher.update((uint8_t*)(retVal.contents.data()), retVal.contents.size() * (sizeof(char) / sizeof(uint8_t)));
-    retVal.hash = { *static_cast<core::blake3_hash_t>(hasher).data };
+    retVal.hash = static_cast<core::blake3_hash_t>(hasher);
```
devshgraphicsprogramming, line 192: instead of putting the code that does header hashing in two places that handle finding includes (which also don't know if you're running with a read or write preprocessing cache at all), is it possible to hash outside/later  in `CCache::find_impl` ?

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -343,13 +412,29 @@ core::smart_refctd_ptr<IShaderCompiler::CCache> IShaderCompiler::CCache::deseria
...
+    size_t srcSize = spirv->getSize() - LZMA_PROPS_SIZE;
+    ELzmaStatus status;
+    ISzAlloc alloc = { SzAlloc, SzFree };
+    SRes res = LzmaDecode(
+        reinterpret_cast<unsigned char*>(uncompressedBuf->getPointer()), &dstSize,
+        reinterpret_cast<const unsigned char*>(spirv->getPointer()) + LZMA_PROPS_SIZE, &srcSize,
+        reinterpret_cast<const unsigned char*>(spirv->getPointer()), LZMA_PROPS_SIZE,
+        LZMA_FINISH_ANY, &status, &alloc);
+    assert(res == SZ_OK);
+    return core::make_smart_refctd_ptr<asset::ICPUShader>(std::move(uncompressedBuf), compilerArgs.stage, IShader::E_CONTENT_TYPE::ECT_SPIRV, compilerArgs.preprocessorArgs.sourceIdentifier.data());
```
devshgraphicsprogramming, line 439: the uncompressed buffer needs its content hash restored (so it needs to be serialized into the cache) so that the `CAssetConverter` works properly on previously cached shaders

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -22,6 +25,73 @@ IShaderCompiler::IShaderCompiler(core::smart_refctd_ptr<system::ISystem>&& syste
...
+
+        assert(propsSize == LZMA_PROPS_SIZE);
+        assert(res == SZ_OK);
+
+        auto compressedSpirvBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(propsSize + destLen);
+        memcpy(compressedSpirvBuffer->getPointer(), compressedSpirv.data(), compressedSpirvBuffer->getSize());
+
+        entry.dependencies = std::move(dependencies);
+        entry.spirv = std::move(compressedSpirvBuffer);
+        entry.uncompressedSize = spirvBuffer->getSize();
```
devshgraphicsprogramming, line 88: maybe `SEntry` needs different constructors:
- for a half-empty dummy lookup element
- for creating a complete entry from deserialization
- for making a complete entry to insert into write cache
and this needs to move into the latter

src/nbl/asset/utils/IShaderCompiler.cpp
alichraghi: nah, we don't

src/nbl/asset/utils/IShaderCompiler.cpp
alichraghi: > for creating a complete entry from deserialization
is it even doable with nlohmann_json?

src/nbl/asset/utils/IShaderCompiler.cpp
alichraghi: debug printing catched

src/nbl/asset/utils/IShaderCompiler.cpp
devshgraphicsprogramming: this needs fixing, do 
```cpp
const auto found = find_impl(mainFile,finder);
if (found==m_container.end())
   return nullptr;
return found->decodeShader();
```

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -83,11 +82,7 @@ inline core::smart_refctd_ptr<ICPUShader> nbl::asset::IShaderCompiler::compileTo
         auto compressedSpirvBuffer = core::make_smart_refctd_ptr<ICPUBuffer>(propsSize + destLen);
         memcpy(compressedSpirvBuffer->getPointer(), compressedSpirv.data(), compressedSpirvBuffer->getSize());
 
-        entry.dependencies = std::move(dependencies);
-        entry.spirv = std::move(compressedSpirvBuffer);
-        entry.uncompressedSize = spirvBuffer->getSize();
-
-        options.writeCache->insert(std::move(entry));
+        options.writeCache->insert(std::move(SEntry(entry, std::move(dependencies), std::move(compressedSpirvBuffer), spirvBuffer->getContentHash(), spirvBuffer->getSize())));
```
devshgraphicsprogramming, line 85: pro C++ tip (I don't actually want you to change anything), there's an `emplace` intead of `insert` for this (constructs an element in-place

src/nbl/asset/utils/IShaderCompiler.cpp
devshgraphicsprogramming: my comment is mostly about not having the LZMA code sit in `IShaderCompiler` methods but rather in `SEntry`

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -354,20 +354,28 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 						lookupHash = std::hash<core::blake3_hash_t>{}(hash);
 					}
 
+					// Making an entry to insert into write cache
+					inline SEntry(const SEntry& other, dependency_container_t dependencies, core::smart_refctd_ptr<asset::ICPUBuffer> spirv,
+						core::blake3_hash_t uncompressedContentHash, size_t uncompressedSize)
+						: mainFileContents(other.mainFileContents), compilerArgs(other.compilerArgs), hash(other.hash),
+						lookupHash(other.lookupHash), dependencies(dependencies), spirv(spirv),
+						uncompressedContentHash(uncompressedContentHash), uncompressedSize(uncompressedSize) {}
```
devshgraphicsprogramming, line 362: but this  kinda does nothing?

its still inline and merely sets the members to values you pass in.

My understanding would be that you no longer make it inline, and MOVE the LZMA compression INTO this constructor

Also if `other` can be overwritten then its better to have a non-const method like `setContent(dependency_container_t&& dependencies, const ICPUShader* spirv)` that will fill out all the stuff

src/nbl/asset/utils/IShaderCompiler.cpp
alichraghi: yeah i forgot this review

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -418,6 +395,39 @@ core::smart_refctd_ptr<IShaderCompiler::CCache> IShaderCompiler::CCache::deseria
...
+void nbl::asset::IShaderCompiler::CCache::SEntry::setContent(const asset::ICPUBuffer* uncompressedSpirvBuffer, dependency_container_t&& dependencies)
+{
+    dependencies = std::move(dependencies);
+    uncompressedContentHash = uncompressedSpirvBuffer->getContentHash();
+    uncompressedSize = uncompressedSpirvBuffer->getSize();
+
+    size_t propsSize = LZMA_PROPS_SIZE;
+    size_t destLen = uncompressedSpirvBuffer->getSize() + uncompressedSpirvBuffer->getSize() / 3 + 128;
+    std::vector<unsigned char> compressedSpirv = {};
+    compressedSpirv.resize(propsSize + destLen);
```
devshgraphicsprogramming, line 410: why are you making a temp vector just to `memcpy` it later into a `spirv` buffer?

just make the buffer now and write into that

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -418,6 +395,39 @@ core::smart_refctd_ptr<IShaderCompiler::CCache> IShaderCompiler::CCache::deseria
...
+
+    ISzAlloc alloc = { SzAlloc, SzFree };
+    int res = LzmaEncode(
+        compressedSpirv.data() + LZMA_PROPS_SIZE, &destLen,
+        reinterpret_cast<const unsigned char*>(uncompressedSpirvBuffer->getPointer()), uncompressedSpirvBuffer->getSize(),
+        &props, compressedSpirv.data(), &propsSize, props.writeEndMark,
+        nullptr, &alloc, &alloc);
+
+    assert(propsSize == LZMA_PROPS_SIZE);
+    assert(res == SZ_OK);
```
devshgraphicsprogramming, line 425: handle failure properly, why would I want my app to crash because LZMA crapped its pants?

return false or something

src/nbl/asset/utils/IShaderCompiler.cpp
alichraghi: the final size after compression is unknown and `ICPUBuffer` can't be resized (unless you ask me to add a `resize` method)?

devshgraphicsprogramming: there's a thing called `CCPUBuffer` with a custom allocator that can also adopt memory

you can `malloc` instead of having a vector, then `realloc` and after realloc is complete you can create that `CCustomAllocatorCPUBuffer` with `core::adopt_memory` tag constructor overload

You just need to provide an "allocator" which will call `free()` on the adopted pointer

devshgraphicsprogramming:
@Erfan-Ahmadi you can merge, this will probably help you a lot with the cache size

src/nbl/asset/utils/IShaderCompiler.cpp
alichraghi: i'll address this in the upcoming *Lib Shader Stage* PR

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -238,7 +286,7 @@ IShaderCompiler::CCache::EntrySet::const_iterator IShaderCompiler::CCache::find_
             else
                 header = finder->getIncludeRelative(dependency.requestingSourceDir, dependency.identifier);
 
-            if (header.hash != dependency.hash || header.contents != dependency.contents)
+            if (header.hash != dependency.hash)
```
Erfan-Ahmadi, line 289: You don't need `m_container.equal_range(mainFile);` and for loop here, since you have a `set` now instead of a `multi_set`
equal_range Is gonna return at most 1 here

src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -22,6 +25,42 @@ IShaderCompiler::IShaderCompiler(core::smart_refctd_ptr<system::ISystem>&& syste
     m_defaultIncludeFinder = core::make_smart_refctd_ptr<CIncludeFinder>(core::smart_refctd_ptr(m_system));
 }
 
+inline core::smart_refctd_ptr<ICPUShader> nbl::asset::IShaderCompiler::compileToSPIRV(const std::string_view code, const SCompilerOptions& options) const
+{
+    CCache::SEntry entry;
+    std::vector<CCache::SEntry::SPreprocessingDependency> dependencies;
+    if (options.readCache || options.writeCache)
+        entry = std::move(CCache::SEntry(code, options));
```
Erfan-Ahmadi, line 33: you don't need the std::move here, `CCache::SEntry(code, options)` is already an r-value

src/nbl/asset/utils/IShaderCompiler.cpp
devshgraphicsprogramming: erm since Lib Shader stage is blocked, copy or cherry pick the fix

alichraghi: huh there's no `CCPUBuffer` in nabla (anymore?)

PR #763, opened 2024-10-29
blake3 encapsulation

devshgraphicsprogramming:


include/nbl/core/hash/blake.h
```diff
@@ -4,12 +4,10 @@
...
 
-
 #include "blake3.h"
 
 #include <span>
 
-
 namespace nbl::core
 {
 struct blake3_hash_t final
```
devshgraphicsprogramming, line 13: missing NBL_API2

src/nbl/core/hash/blake.cpp
```diff
@@ -0,0 +1,29 @@
...
+    ::blake3_hasher_update(&m_state,data,bytes);
+    return *this;
+}
+
+template<typename T>
+core::blake3_hasher& core::blake3_hasher::operator<<(const T& input)
+{
+    update_impl<T>::__call(*this,input);
+    return *this;
+}
```
devshgraphicsprogramming, line 21: Second Commandment of C++, "Thou shalt not define templates outside of Headers"

PR #767, opened 2024-11-03
Improve window creation

alichraghi:
## Testing 
Hello Swapchain

## TODO list:
- [ ] `ECF_FULLSCREEN`
- [ ] `ECF_MAXIMIZED`
- [ ] Multiple monitors
- [ ] Others

### DIFF
--- a/src/nbl/ui/CWindowManagerWin32.cpp
+++ b/src/nbl/ui/CWindowManagerWin32.cpp
@@ -34,41 +34,35 @@ static inline DWORD getWindowStyle(const core::bitflag<IWindow::E_CREATE_FLAGS> 
 	{
 		if (!flags.hasFlags(IWindow::ECF_BORDERLESS))
 		{
-			style |= WS_BORDER;
-			style |= (WS_SYSMENU | WS_CAPTION);
+			style |= WS_BORDER | WS_CAPTION | WS_SYSMENU;
 		}
 		// ? not sure about those below
 		style |= WS_CLIPCHILDREN;
 		style |= WS_CLIPSIBLINGS;
 	}
-	if (flags.hasFlags(IWindow::ECF_MINIMIZED))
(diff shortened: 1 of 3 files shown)


src/nbl/ui/CWindowManagerWin32.cpp
```diff
@@ -29,59 +29,48 @@ IWindowManager::SDisplayInfo CWindowManagerWin32::getPrimaryDisplayInfo() const
...
 {
-	// this could be common to all `createWindow` impl
-	if (creationParams.flags.hasFlags(IWindow::ECF_CAN_RESIZE) || creationParams.flags.hasFlags(IWindow::ECF_CAN_MAXIMIZE))
-		creationParams.flags |= IWindow::ECF_RESIZABLE;
 	// win32 minimize is weird, its a resize to 0,0
-	if (creationParams.flags.hasFlags(IWindow::ECF_CAN_MINIMIZE))
-		creationParams.flags |= IWindow::ECF_CAN_RESIZE;
+	if (creationParams.flags.hasFlags(IWindow::ECF_CAN_MAXIMIZE) || creationParams.flags.hasFlags(IWindow::ECF_CAN_MINIMIZE))
+		creationParams.flags |= IWindow::ECF_UI_RESIZABLE;
+	if (creationParams.flags.hasFlags(IWindow::ECF_FULLSCREEN))
```
Erfan-Ahmadi, line 72: Hmm I wonder why games have full screen vs windowed borderless fullscreen, do we need to consider this as well? Why are you adding borderless flag when you detect fullscreen?

alichraghi: i did some researching and apparently in windows, the only difference between *fullscreen* and *windowed borderless fullscreen* is that the former has the `HWND_TOPMOST` flag. so:
- fullscreen: `borderless + always-on-top + monitor sized`
- windowed borderless fullscreen: `borderless + monitor sized`

PR #771, opened 2024-11-07
Small fix for SPIRV Optimizer + small fix to optimizer pass serialization

Fletterio:
## Description

- There's a disabled optimizer pass so I pushed it to the bottom of the enum (otherwise it fucks up the retrieval later). 
- Update SPIRV version for the optimizer (was at 1.5, now 1.6).
- Change optimizer's message callback so it doesn't crash on nullpointer deref
- Fix a slight overlook when I made serialization of optimizer passes

## Testing 
Added an optimizer pass to my FFT_Bloom example and it worked fine. Caching worked fine as well.

### DIFF
--- a/include/nbl/asset/utils/ISPIRVOptimizer.h
+++ b/include/nbl/asset/utils/ISPIRVOptimizer.h
@@ -39,7 +39,7 @@ class ISPIRVOptimizer final : public core::IReferenceCounted
             EOP_COUNT
         };
 
-        ISPIRVOptimizer(std::span<E_OPTIMIZER_PASS> _passes) : m_passes(_passes.begin(), _passes.end()) {}
+        ISPIRVOptimizer(std::span<const E_OPTIMIZER_PASS> _passes) : m_passes(_passes.begin(), _passes.end()) {}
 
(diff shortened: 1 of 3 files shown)


include/nbl/asset/utils/ISPIRVOptimizer.h
```diff
@@ -34,11 +33,13 @@ class ISPIRVOptimizer final : public core::IReferenceCounted
...
             EOP_STRENGTH_REDUCTION,
             EOP_IF_CONVERSION,
+            EOP_STRIP_DEBUG_INFO,
+            EOP_AGGRESSIVE_DCE,
 
             EOP_COUNT
         };
 
-        ISPIRVOptimizer(std::initializer_list<E_OPTIMIZER_PASS> _passes) : m_passes(std::move(_passes)) {}
+        ISPIRVOptimizer(core::vector<E_OPTIMIZER_PASS>&& _passes) : m_passes(_passes) {}
```
devshgraphicsprogramming, line 42: span<>

Fletterio: done!

PR #777, opened 2024-11-11
Ngfx sdk integration

devshgraphicsprogramming:

### DIFF
--- a/include/nbl/video/IAPIConnection.h
+++ b/include/nbl/video/IAPIConnection.h
@@ -11,7 +11,7 @@
 #include "nbl/video/debug/IDebugCallback.h"
 
 #include "nbl/video/utilities/renderdoc.h"
-
+#include "nbl/video/utilities/ngfx.h"
 
 namespace nbl::video
 {
@@ -61,7 +61,16 @@ class NBL_API2 IAPIConnection : public core::IReferenceCounted
 
         const SFeatures& getEnabledFeatures() const { return m_enabledFeatures; }
 
-        const bool isRunningInRenderdoc() const { return m_rdoc_api; }
+        enum DebuggerType
+        {
+            EDT_NONE,
+            EDT_RENDERDOC,
+            EDT_NGFX
+        };
+        const DebuggerType isRunningInGraphicsDebugger() const {
+            return m_ngfx_api.useNGFX ? EDT_NGFX :      // ngfx takes priority?
+                m_rdoc_api ? EDT_RENDERDOC : EDT_NONE;
+        }
         virtual bool startCapture() = 0;
         virtual bool endCapture() = 0;
 
@@ -70,6 +79,7 @@ class NBL_API2 IAPIConnection : public core::IReferenceCounted
 
         std::vector<std::unique_ptr<IPhysicalDevice>> m_physicalDevices;
         renderdoc_api_t* m_rdoc_api;
+        ngfx_api_t m_ngfx_api;
         SFeatures m_enabledFeatures = {};
 };
 
--- a/include/nbl/video/utilities/ngfx.h
+++ b/include/nbl/video/utilities/ngfx.h
@@ -1,0 +1,79 @@
+#ifndef _NBL_VIDEO_UTILITIES_NGFX_H_INCLUDED_
+#define _NBL_VIDEO_UTILITIES_NGFX_H_INCLUDED_
+
+#include "C:\Program Files\NVIDIA Corporation\Nsight Graphics 2024.1.0\SDKs\NsightGraphicsSDK\0.8.0\include\NGFX_Injection.h"
+
+namespace nbl::video
+{
+    struct SNGFXIntegration
+    {
+        bool useNGFX;
+        NGFX_Injection_InstallationInfo versionInfo;
+    };
+
+    bool injectNGFXToProcess(SNGFXIntegration& api)
+    {
+        uint32_t numInstallations = 0;
+        auto result = NGFX_Injection_EnumerateInstallations(&numInstallations, nullptr);
+        if (numInstallations == 0 || NGFX_INJECTION_RESULT_OK != result)
+        {
+            api.useNGFX = false;
+            return false;
+        }
+
+        std::vector<NGFX_Injection_InstallationInfo> installations(numInstallations);
+        result = NGFX_Injection_EnumerateInstallations(&numInstallations, installations.data());
+        if (numInstallations == 0 || NGFX_INJECTION_RESULT_OK != result)
+        {
+            api.useNGFX = false;
+            return false;
+        }
+
+        // get latest installation
+        api.versionInfo = installations.back();
+
+        uint32_t numActivities = 0;
+        result = NGFX_Injection_EnumerateActivities(&api.versionInfo, &numActivities, nullptr);
+        if (numActivities == 0 || NGFX_INJECTION_RESULT_OK != result)
+        {
+            api.useNGFX = false;
+            return false;
+        }
+
+        std::vector<NGFX_Injection_Activity> activities(numActivities);
+        result = NGFX_Injection_EnumerateActivities(&api.versionInfo, &numActivities, activities.data());
+        if (NGFX_INJECTION_RESULT_OK != result)
+        {
+            api.useNGFX = false;
+            return false;
+        }
+
+        const NGFX_Injection_Activity* pActivityToInject = nullptr;
+        for (const NGFX_Injection_Activity& activity : activities)
+        {
+            if (activity.type == NGFX_INJECTION_ACTIVITY_FRAME_DEBUGGER)    // only want frame debugger
+            {
+                pActivityToInject = &activity;
+                break;
+            }
+        }
+
+        if (!pActivityToInject) {
+            api.useNGFX = false;
+            return false;
+        }
+
+        result = NGFX_Injection_InjectToProcess(&api.versionInfo, pActivityToInject);
+        if (NGFX_INJECTION_RESULT_OK != result)
+        {
+            api.useNGFX = false;
+            return false;
+        }
+
+        return true;
+    }
+
+    using ngfx_api_t = SNGFXIntegration;
+}
+
+#endif //_NBL_VIDEO_UTILITIES_NGFX_H_INCLUDED_
--- a/include/nbl/video/utilities/ngfx.h
+++ b/include/nbl/video/utilities/ngfx.h
@@ -1,6 +1,7 @@
 #ifndef _NBL_VIDEO_UTILITIES_NGFX_H_INCLUDED_
 #define _NBL_VIDEO_UTILITIES_NGFX_H_INCLUDED_
 
+// TODO: hopefully this is temporary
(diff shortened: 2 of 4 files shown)


include/nbl/video/IAPIConnection.h
```diff
@@ -61,7 +61,16 @@ class NBL_API2 IAPIConnection : public core::IReferenceCounted
...
+        enum DebuggerType
+        {
+            EDT_NONE,
+            EDT_RENDERDOC,
+            EDT_NGFX
+        };
+        const DebuggerType isRunningInGraphicsDebugger() const {
+            return m_ngfx_api.useNGFX ? EDT_NGFX :      // ngfx takes priority?
+                m_rdoc_api ? EDT_RENDERDOC : EDT_NONE;
+        }
```
devshgraphicsprogramming, line 73: hide the impl and precompute the enum in the cpp

include/nbl/video/utilities/ngfx.h
```diff
@@ -0,0 +1,85 @@
...
+#include "C:\Program Files\NVIDIA Corporation\Nsight Graphics 2024.1.0\SDKs\NsightGraphicsSDK\0.8.0\include\NGFX_Injection.h"
+
+namespace nbl::video
+{
+    struct SNGFXIntegration
+    {
+        bool useNGFX;
+        NGFX_Injection_InstallationInfo versionInfo;
+    };
+
```
devshgraphicsprogramming, line 14: absolutely should not be public

include/nbl/video/utilities/ngfx.h
```diff
@@ -0,0 +1,85 @@
...
+            api.useNGFX = false;
+            return false;
+        }
+
+        result = NGFX_Injection_InjectToProcess(&api.versionInfo, pActivityToInject);
+        if (NGFX_INJECTION_RESULT_OK != result)
+        {
+            api.useNGFX = false;
+            return false;
+        }
```
devshgraphicsprogramming, line 72: what happens if you are already running through nsight? what do these functions return?

can you detect if you've already been injected into?

src/nbl/video/CVulkanConnection.cpp
```diff
@@ -352,7 +357,9 @@ bool CVulkanConnection::endCapture()
         return false;
     }
 
-    m_rdoc_api->EndFrameCapture(RENDERDOC_DEVICEPOINTER_FROM_VKINSTANCE(m_vkInstance), NULL);
+    if (debugType == EDT_RENDERDOC)
+        m_rdoc_api->EndFrameCapture(RENDERDOC_DEVICEPOINTER_FROM_VKINSTANCE(m_vkInstance), NULL);
+    // no equivalent end frame capture for ngfx, ends captures on next frame delimiter
```
devshgraphicsprogramming, line 362: leave this link in the comments https://www.reddit.com/r/GraphicsProgramming/comments/w0hl9o/graphics_debugger_record_before_first_frame/

include/nbl/video/utilities/ngfx.h
keptsecret: Seems to just crash at the moment because it's already injected. Though, could make it just fail quietly.

keptsecret: Actually, I think this injection happens first. Nsight seems to only begin injection once vulkan/everything (not sure how much) is initialized, at which point it crashes the program.

keptsecret:
Still haven't been able to test because it needs to link dll on compile

src/nbl/video/IAPIConnection.cpp
```diff
@@ -4,13 +4,17 @@
 #include "nbl/video/utilities/renderdoc.h"
 #include "nbl/video/utilities/ngfx.h"
 
+// TODO: temporary hopefully
+#include "C:\Program Files\NVIDIA Corporation\Nsight Graphics 2024.1.0\SDKs\NsightGraphicsSDK\0.8.0\include\NGFX_Injection.h"
+
```
devshgraphicsprogramming, line 9: @AnastaZIuk we need your help for NGFX SDK detection with Cmake and NBL_BUILD_WITH_NGFX or NBL_NGFX_FOUND

include/nbl/video/utilities/ngfx.h
devshgraphicsprogramming: > Seems to just crash at the moment because it's already injected. Though, could make it just fail quietly.

I don't want to inject NGFX when EXE is not launched with NSight.

I want Nsight to do the injection, and us to just know if it happened, and not do it ourselves.

devshgraphicsprogramming:
> Still haven't been able to test because it needs to link dll on compile

You need @AnastaZIuk to figure out the Delay DLL loading for the ngfx dll in a way that it makes sense (probably similar to renderdoc's dll load).

Also the code we have needs to be resilient to DLL not being found, so we need to NOT call NGFX if the DLL was not delay loaded properly.

src/nbl/video/IAPIConnection.cpp
AnastaZIuk: done as part of https://github.com/Devsh-Graphics-Programming/Nabla/pull/789

AnastaZIuk:
> > Still haven't been able to test because it needs to link dll on compile
> 
> You need @AnastaZIuk to figure out the Delay DLL loading for the ngfx dll in a way that it makes sense (probably similar to renderdoc's dll load).
> 
> Also the code we have needs to be resilient to DLL not being found, so we need to NOT call NGFX if the DLL was not delay loaded properly.

done as part of https://github.com/Devsh-Graphics-Programming/Nabla/pull/789

include/nbl/video/IAPIConnection.h
```diff
@@ -61,15 +60,38 @@ class NBL_API2 IAPIConnection : public core::IReferenceCounted
 
         const SFeatures& getEnabledFeatures() const { return m_enabledFeatures; }
 
-        const bool isRunningInRenderdoc() const { return m_rdoc_api; }
+        enum SDebuggerType
+        {
+            EDT_NONE,
+            EDT_RENDERDOC,
+            EDT_NGFX
```
devshgraphicsprogramming, line 67: 1. don't use the `S` prefix for enums, either no prefix or `E`
2. Use `enum class` and always remember to explicitly size the enum (so `: uint8_t`)
3. if you use enum class, then you don't need `EDT_` prefix (because class enum creates a scope for the enum)

include/nbl/video/IAPIConnection.h
```diff
@@ -61,15 +60,38 @@ class NBL_API2 IAPIConnection : public core::IReferenceCounted
...
+        {
+            SNGFXIntegration();
+
+            bool useNGFX = false;
+
+            bool injectNGFXToProcess();
+            bool executeNGFXCommand();
+            inline bool isAPILoaded() { return m_loaded; }
+        private:
+            const bool m_loaded;
```
devshgraphicsprogramming, line 90: wouldn't `m_debuggerType==EDT_NGFX` mean both `useNGFX` and `m_loaded` ?

include/nbl/video/IAPIConnection.h
```diff
@@ -61,15 +60,38 @@ class NBL_API2 IAPIConnection : public core::IReferenceCounted
...
+        };
+        using ngfx_api_t = SNGFXIntegration;
+        ngfx_api_t m_ngfx_api;
```
devshgraphicsprogramming, line 93: typedef is not really necessary, the renderdoc thing is called `_t` cuase it has a C-API and that's what its called

src/nbl/video/IAPIConnection.cpp
```diff
@@ -43,7 +49,145 @@ IAPIConnection::IAPIConnection(const SFeatures& enabledFeatures)
...
+    #endif // NBL_BUILD_WITH_NGFX
+
+    return false;
+}
+
+bool IAPIConnection::SNGFXIntegration::executeNGFXCommand()
+{
+    #ifdef NBL_BUILD_WITH_NGFX
+    if(m_loaded) //! this check is mandatory!
+        return NGFX_Injection_ExecuteActivityCommand() == NGFX_INJECTION_RESULT_OK; // optional TOOD: could log on "else"
```
devshgraphicsprogramming, line 136: yes logging is very useful

src/nbl/video/IAPIConnection.cpp
```diff
@@ -43,7 +49,145 @@ IAPIConnection::IAPIConnection(const SFeatures& enabledFeatures)
         int ret = RENDERDOC_GetAPI(MinRenderdocVersion, (void**)&m_rdoc_api);
         assert(ret == 1);
     #endif
+
+        // probably is platform agnostic, for now
+        m_ngfx_api.injectNGFXToProcess();
+
+        m_debuggerType = m_ngfx_api.useNGFX ? EDT_NGFX :      // ngfx takes priority?
+            m_rdoc_api ? EDT_RENDERDOC : EDT_NONE;
```
devshgraphicsprogramming, line 57: the functions to load the debugger should be moved out of the constructor into a method.

It should work more or less like this
```cpp
if (m_debuggerType!=EDebuggerType::None)
    LOG_ERROR("Graphics Debugger %d Already Attached!,m_debuggerType);
    
 #ifdef NBL_BUILD_WITH_NGFX
 #endif
 
 // if NGFX not loaded, then try Renderdoc
```

include/nbl/video/utilities/ngfx.h
devshgraphicsprogramming: > Actually, I think this injection happens first. Nsight seems to only begin injection once vulkan/everything (not sure how much) is initialized, at which point it crashes the program. Because it should just fail quietly at the moment.

This means that you need to possibly move the Renderdoc and Nsight injection from IAPIConnection constructor to the base class constructor/initialization finish

Idea: have a `init_finish` private method dealign with Renderdoc and Nsight which the derived class needs to call after all the setup is complete (end of most derived ctor).

3rdparty/ngfx/ngfx.cmake
```diff
@@ -0,0 +1,67 @@
+option(NBL_BUILD_WITH_NGFX "Enable NGFX build" OFF)
+
+# NOTE: on windows default installation path is:
+# "C:/Program Files/NVIDIA Corporation/Nsight Graphics <version>/SDKs/NsightGraphicsSDK" <- define as "NGFX_SDK" environment variable
+# then you can pick SDK version with "NGFX_SDK_VERSION" cache variable (CMake GUI list supported)
+
+if(NBL_BUILD_WITH_NGFX)
+    if(NOT DEFINED ENV{NGFX_SDK})
+        message(FATAL_ERROR "\"NGFX_SDK\" environment variable must be defined to build with NBL_BUILD_WITH_NGFX enabled!")
+    endif()
```
devshgraphicsprogramming, line 10: image

What am I doing wrong @AnastaZIuk ?

devshgraphicsprogramming: Wtf I need to define an envvar!? https://stackoverflow.com/questions/14343643/can-i-add-an-environment-variable-in-cmake-gui-wizard

PR #789, opened 2024-11-20
NGFX CMake Nabla build system integration

AnastaZIuk:


devshgraphicsprogramming:
@keptsecret merge this into your branch

src/nbl/CMakeLists.txt
```diff
@@ -659,6 +659,21 @@ write_source_definitions("${_NBL_DEFINE_FILE_WRAPPER_}" "${_NBL_SOURCE_DEFINITIO
...
+	else()
+		target_link_libraries(Nabla PRIVATE ngfx)
+	endif()
+
+	target_include_directories(Nabla PRIVATE $<TARGET_PROPERTY:ngfx,INTERFACE_INCLUDE_DIRECTORIES>)
+	target_compile_definitions(Nabla 
+		PRIVATE NBL_BUILD_WITH_NGFX
+		PRIVATE $<TARGET_PROPERTY:ngfx,INTERFACE_COMPILE_DEFINITIONS>
+	)
+endif()
```
devshgraphicsprogramming, line 675: @AnastaZIuk why is there an ngfx target cmake knows about?

we dont do this for Vulkan, CUDA, Optix, we delay load from DLL and are not supposed to crash if not found

AnastaZIuk: > @AnastaZIuk why is there an ngfx target cmake knows about?

because I introduce modern cmake there as I do on `newBuildSystem` branch, [interface](<https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries>) targets don't have build rules but you can interact with them

> we dont do this for Vulkan, CUDA, Optix

it's actually not quite true, for example when you call `find_package(Vulkan)` you get imported targets if [found](<https://cmake.org/cmake/help/latest/module/FindVulkan.html>) which are similar to interface targets and because there is no existing `ngfx` module which I can request I wrote a simple one myself just to embed whole logic into single logical target (it could also be imported target just like vulkan targets - doesn't matter too much). 

If the target is available it means:

- someone enabled `NBL_BUILD_WITH_NGFX`
- the SDK with the injection header has been found and passed a simple validation

> we delay load from DLL and are not supposed to crash if not found

and what makes you think it will crash? We do delay load from DLL [here](<https://github.com/Devsh-Graphics-Programming/Nabla/pull/789/files#diff-b3edcb9f53793e7e0e01f9fe5ebca8fc3cacd7bd9033b9c88a506d7e2ccaed91R64>) as well & we have [`m_loaded`](<https://github.com/Devsh-Graphics-Programming/Nabla/pull/789/files#diff-067cf624533e28b72cf0261aa15e54fa579284e9f8843537b3bb87f4430e515fR90>) member to make sure we can call the API only if we first loaded the DLL

devshgraphicsprogramming: ok thanks for explaining it to me

devshgraphicsprogramming: but whats the deal with `target_link_libraries` and the ngfx being an INTERFACE instead of PRIVATE?

AnastaZIuk: > but whats the deal with `target_link_libraries` and the ngfx being an INTERFACE instead of PRIVATE?

the reason for

```cmake
if(NBL_STATIC_BUILD)
  target_link_libraries(Nabla INTERFACE ngfx)
else()
  target_link_libraries(Nabla PRIVATE ngfx)
endif()
```

is that in static build mode each executable must link against the import library - using the `INTERFACE` qualifier ensures this linkage is propagated.

On the other hand when building Nabla as a DLL the `PRIVATE` qualifier is used to prevent any inheritance of ngfx linkage for targets linking Nabla since the import library is already linked directly to the DLL.

PR #791, opened 2024-11-21
build: Add ClangCL profiles

alichraghi:
Also fix some compilation errors catched by clang


CMakeLists.txt
```diff
@@ -35,10 +35,10 @@ if(NBL_STATIC_BUILD)
 	message(STATUS "Static Nabla build enabled!")
 else()
 	if(MSVC)
```
AnastaZIuk, line 37: remove the `MSVC` check here - for `NOT NBL_STATIC_BUILD` case we should only check now if `NBL_COMPILER_DYNAMIC_RUNTIME` is enabled otherwise throw an error to make sure we have dynamic runtime for dynamic builds

cmake/adjust/flags.cmake
```diff
@@ -173,7 +213,7 @@ function(nbl_adjust_flags)
 			
 			set(MAPPED_CONFIG $<TARGET_GENEX_EVAL:${NBL_TARGET_ITEM},$<TARGET_PROPERTY:${NBL_TARGET_ITEM},NBL_CONFIGURATION_MAP>>)
 			
-			if(MSVC)
+			if(MSVC AND CMAKE_CXX_COMPILER_ID STREQUAL MSVC)
```
AnastaZIuk, line 216: `if(CMAKE_CXX_COMPILER_ID STREQUAL MSVC)`

alichraghi: ah was pretty sure i changed this..

3rdparty/CMakeLists.txt
```diff
@@ -496,11 +496,7 @@ if (NBL_BUILD_IMGUI)
 endif()
 
 foreach(trgt IN LISTS NBL_3RDPARTY_TARGETS)
-		if(NBL_DYNAMIC_MSVC_RUNTIME)
-			set_property(TARGET ${trgt} PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
-		else()
-			set_property(TARGET ${trgt} PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
-		endif()
+		set_property(TARGET ${trgt} PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>$<$<BOOL:${NBL_COMPILER_DYNAMIC_RUNTIME}>:DLL>")
```
devshgraphicsprogramming, line 499: AFAIK in Clang and GCC there are options about static linking of libstd-c++

CMakeLists.txt
```diff
@@ -35,14 +35,8 @@ endif()
...
-		if(NBL_DYNAMIC_MSVC_RUNTIME)
-			message(STATUS "Shared Nabla build enabled!")
-		else()
-			message(FATAL_ERROR "Turn NBL_DYNAMIC_MSVC_RUNTIME on! For dynamic Nabla builds dynamic MSVC runtime is mandatory!")
-		endif()
-	else()
-		message(FATAL_ERROR "Nabla can't be built with shared libraries! Please make sure you are targetting Windows OS and MSVC compiler!")
+	if(NOT NBL_COMPILER_DYNAMIC_RUNTIME)
+		message(FATAL_ERROR "Turn NBL_COMPILER_DYNAMIC_RUNTIME on! For dynamic Nabla builds dynamic runtime is mandatory!")
 	endif()
```
devshgraphicsprogramming, line 40: aah ok so we want to ban compiling with `static-libstdc++`

cmake/adjust/template/vendor/CXX_Clang.cmake
```diff
@@ -0,0 +1,51 @@
...
+)
+
+# RelWithDebInfo
+set(NBL_CXX_RELWITHDEBINFO_COMPILE_OPTIONS "")
+
+# Global
+list(APPEND NBL_CXX_COMPILE_OPTIONS
+	-Wextra
+	-fno-strict-aliasing
+	-msse4.2
```
devshgraphicsprogramming, line 20: mavx

cmake/adjust/template/vendor/CXX_Clang.cmake
```diff
@@ -0,0 +1,51 @@
...
+	-Wno-c++98-compat-pedantic
+	-Wno-padded
+	-Wno-unsafe-buffer-usage
+	-Wno-switch-enum
+	-Wno-error=ignored-attributes
+	-Wno-error=unused-function
+	-Wno-error=unused-variable
+	-Wno-error=unused-parameter
+	-Wno-error=ignored-attributes
+	-Wno-error=non-pod-varargs
```
devshgraphicsprogramming, line 37: I take it that we'll remove this over time?

cmake/adjust/template/vendor/CXX_Clang.cmake
```diff
@@ -0,0 +1,51 @@
...
+
+# Debug
+set(NBL_CXX_DEBUG_COMPILE_OPTIONS
+	-ggdb3 -Wall -fno-omit-frame-pointer -fstack-protector-strong
+)
+
+# Release
+set(NBL_CXX_RELEASE_COMPILE_OPTIONS
+	-fexpensive-optimizations
+)
```
devshgraphicsprogramming, line 11: I think O3 enables this by default ?

cmake/adjust/template/vendor/CXX_Clang.cmake
```diff
@@ -0,0 +1,51 @@
...
+
+# RelWithDebInfo
+set(NBL_CXX_RELWITHDEBINFO_COMPILE_OPTIONS "")
```
devshgraphicsprogramming, line 14: what about O1 or something ? Alos don't you need `ggdb3` ?

cmake/adjust/template/vendor/CXX_Clang.cmake
```diff
@@ -0,0 +1,51 @@
+include_guard(GLOBAL)
+
+# Debug
+set(NBL_CXX_DEBUG_COMPILE_OPTIONS
+	-ggdb3 -Wall -fno-omit-frame-pointer -fstack-protector-strong
```
devshgraphicsprogramming, line 5: is `ggdb3` sufficient?

devshgraphicsprogramming: shouldn't `-g` also be used?

devshgraphicsprogramming: btw GCC/Clang have an equivalent of ffast math that we use in MSVC

3rdparty/CMakeLists.txt
AnastaZIuk: yes there [are](<https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-static-libstdc>) and this [property](<https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html>) has effect only if MSVC ABI is the target

CMakeLists.txt
AnastaZIuk: yes we agreed it must match

cmake/adjust/template/vendor/CXX_Clang.cmake
AnastaZIuk: yes, @YasInvolved is going to kill all warnings

cmake/adjust/template/vendor/CXX_Clang.cmake
```diff
@@ -0,0 +1,51 @@
+include_guard(GLOBAL)
+
+# Debug
```
devshgraphicsprogramming, line 3: what about macro definitions like `NDEBUG` ?

P.S. should MSVC even define should macros from commandline instead of configuring theminto the header?

cmake/adjust/template/vendor/CXX_Clang.cmake
devshgraphicsprogramming: I think you might also need a flag for stripping debug symbols

cmake/adjust/template/vendor/C_Clang.cmake
```diff
@@ -0,0 +1,51 @@
...
+	list(APPEND NBL_C_COMPILE_OPTIONS -fsanitize=address)
+endif()
+
+if(NBL_SANITIZE_THREAD)
+	list(APPEND NBL_C_COMPILE_OPTIONS -fsanitize=thread)
+endif()
+
+# our pervious flags-set function called this, does not affect flags nor configs so I will keep it here temporary
+# TODO: move it out from the profile
+link_libraries(-fuse-ld=gold)
```
devshgraphicsprogramming, line 51: same comments as for C++, maybe unify a bit with a common cmake include ?

cmake/adjust/template/vendor/CXX_Clang.cmake
AnastaZIuk: actually `ggdb3` doesn't exist on clang and it throws a warning about unknown arg

include/nbl/asset/IDescriptorSetLayout.h
```diff
@@ -330,7 +330,7 @@ class IDescriptorSetLayout : public IDescriptorSetLayoutBase
 				bindings[i].binding = i;
 				bindings[i].type = type;
 				bindings[i].createFlags = SBinding::E_CREATE_FLAGS::ECF_NONE;
-				bindings[i].stageFlags = stageAccessFlags ? stageAccessFlags[i]:asset::IShader::ESS_ALL_OR_LIBRARY;
+				bindings[i].stageFlags = stageAccessFlags ? stageAccessFlags[i]:asset::IShader::E_SHADER_STAGE::ESS_ALL_OR_LIBRARY;
```
devshgraphicsprogramming, line 333: hlsl::ShaderStage`

include/nbl/macros.h
```diff
@@ -81,7 +81,7 @@
 
 //! Workarounds for compiler specific bugs
 // MSVC 2019 is a special snowflake
-#if defined(_MSC_VER) && _MSC_VER>=1920
+#if defined(_MSC_VER) && !defined(__clang__) && _MSC_VER>=1920
```
devshgraphicsprogramming, line 84: can we just ban old MSC_VER and old GCC ? and not do weird macro tricks ( or less)

include/nbl/video/ISwapchain.h
```diff
@@ -21,6 +21,8 @@ class ISwapchain : public IBackendObject
 
         struct SSharedCreationParams
         {
+            SSharedCreationParams() {}
+
```
devshgraphicsprogramming, line 25: we'll lose designated initializers with a default ctor

src/nbl/video/IGPUAccelerationStructure.cpp
```diff
@@ -140,11 +140,11 @@ uint32_t IGPUBottomLevelAccelerationStructure::BuildInfo<BufferType>::valid(cons
...
-template uint32_t IGPUBottomLevelAccelerationStructure::BuildInfo<asset::ICPUBuffer>::template valid<uint32_t>(const uint32_t* const) const;
+template uint32_t IGPUBottomLevelAccelerationStructure::BuildInfo<IGPUBuffer>::valid<uint32_t>(const uint32_t* const) const;
+template uint32_t IGPUBottomLevelAccelerationStructure::BuildInfo<asset::ICPUBuffer>::valid<uint32_t>(const uint32_t* const) const;
 using BuildRangeInfo = hlsl::acceleration_structures::bottom_level::BuildRangeInfo;
-template uint32_t IGPUBottomLevelAccelerationStructure::BuildInfo<IGPUBuffer>::template valid<BuildRangeInfo>(const BuildRangeInfo* const) const;
-template uint32_t IGPUBottomLevelAccelerationStructure::BuildInfo<asset::ICPUBuffer>::template valid<BuildRangeInfo>(const BuildRangeInfo* const) const;
+template uint32_t IGPUBottomLevelAccelerationStructure::BuildInfo<IGPUBuffer>::valid<BuildRangeInfo>(const BuildRangeInfo* const) const;
+template uint32_t IGPUBottomLevelAccelerationStructure::BuildInfo<asset::ICPUBuffer>::valid<BuildRangeInfo>(const BuildRangeInfo* const) const;
```
devshgraphicsprogramming, line 147: and this change doesn't break msvc ?

AnastaZIuk: it doesn't

cmake/adjust/template/vendor/CXX_Clang.cmake
devshgraphicsprogramming: also ggdb3 is for GDB, I think on Clang one is supposed to use LLDB

cmake/adjust/template/vendor/impl/Clang.cmake
```diff
@@ -0,0 +1,89 @@
...
+endif()
+
+NBL_REQUEST_COMPILE_OPTION_SUPPORT(LANG ${LANG} OPTIONS
+	# latest Clang(CL) 19.1.1 shipped with VS seems to require explicitly features to be listed (simdjson)
+	# TODO: Yas, use with REQUEST_VAR, if the request fail then do not promote simdjson to build with 
+	# HASWELL implementation because those flags + avx2 compose subset it wants in this case
+
+	# also instead of enabling single options maybe we could consider requesting an
+	# instruction implementation set instead, eg -march=haswel, though this approach
+	# could add a few more flags then we actually need while building - to rethink
```
AnastaZIuk, line 28: @devshgraphicsprogramming

AnastaZIuk: https://github.com/llvm/llvm-project/blob/b46f16c8d7f4a9e37271cdc4717dd858a4c813dc/llvm/lib/Target/X86/X86.td#L937

devshgraphicsprogramming: `-march` is a bad idea unless you're deploying on a particular CPU (i.e. renderfarm or amazon AWS)

include/nbl/macros.h
AnastaZIuk: we can ban old compilers at header level, there are a few headers checking for compiler version and doing weird tricks so imo we should do it in another PR and delegate @YasInvolved

cmake/adjust/template/vendor/CXX_Clang.cmake
AnastaZIuk: > mavx

resolved

cmake/adjust/template/vendor/CXX_Clang.cmake
```diff
@@ -0,0 +1,51 @@
...
+	-Wno-padded
+	-Wno-unsafe-buffer-usage
+	-Wno-switch-enum
+	-Wno-error=ignored-attributes
+	-Wno-error=unused-function
+	-Wno-error=unused-variable
+	-Wno-error=unused-parameter
+	-Wno-error=ignored-attributes
+	-Wno-error=non-pod-varargs
+	-fno-exceptions
```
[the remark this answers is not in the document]
AnastaZIuk: > I dont think we can compile without exceptions, there are some try-catch block in our code IIRC ?

resolved, `-fno-exceptions` should not be there + its possible to compile with this flag but it changes runtime behaviour which we don't want (asserts)

cmake/adjust/template/vendor/CXX_Clang.cmake
AnastaZIuk: yes however [`-fexpensive-optimizations`](<https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-fexpensive-optimizations>) is not Clang's option but GCC's

cmake/adjust/template/vendor/CXX_Clang.cmake
AnastaZIuk: > what about O1 or something ? Alos don't you need `ggdb3` ?

https://github.com/Devsh-Graphics-Programming/Nabla/pull/791#discussion_r2033336186 + resolved

cmake/adjust/template/vendor/CXX_Clang.cmake
AnastaZIuk: on windows `ClangCL` which is Clang frontend compatible with MSVC CLI uses `lld-link` linker

> C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\bin>lld.exe
lld is a generic driver.
Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead

cmake/adjust/template/vendor/CXX_Clang.cmake
```diff
@@ -0,0 +1,51 @@
...
+	list(APPEND NBL_CXX_COMPILE_OPTIONS -fsanitize=address)
+endif()
+
+if(NBL_SANITIZE_THREAD)
+	list(APPEND NBL_CXX_COMPILE_OPTIONS -fsanitize=thread)
+endif()
+
+# our pervious flags-set function called this, does not affect flags nor configs so I will keep it here temporary
+# TODO: move it out from the profile
+link_libraries(-fuse-ld=gold)
```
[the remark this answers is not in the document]
AnastaZIuk: resolved, we use incremental linking

cmake/adjust/template/vendor/CXX_Clang.cmake
AnastaZIuk: > what about macro definitions like `NDEBUG` ?

in theory we set them in MSVC frontend profile but we could skip doing so both for this profile & Clang since CMake sets `NDEBUG` at the very beginning with `CMAKE_<LANG>_FLAGS_<CONFIG>_INIT` within some defaults (ofc, it can be overwritten) - so its not MSVC which sets it but CMake

- <https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows-MSVC.cmake#L504>
- <https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Windows-Clang.cmake#L117>

> instead of configuring theminto the header?

to be discussed but we already generate a set of defines which is included in nabla.h iirc

cmake/adjust/template/vendor/C_Clang.cmake
AnastaZIuk: resolved

cmake/adjust/template/vendor/impl/Clang.cmake
AnastaZIuk: > `-march` is a bad idea unless you're deploying on a particular CPU (i.e. renderfarm or amazon AWS)

understood, also found it too much - resolved

cmake/adjust/template/vendor/CXX_Clang.cmake
AnastaZIuk: https://github.com/Devsh-Graphics-Programming/Nabla/commit/a2a7e72f42501fe71b6f4f141b92bac4d2b8cd93 I stay at defining it myself though

AnastaZIuk:
@YasInvolved we will target warnings fixes into another PR

PR #760, opened 2024-10-25
Remove Matrices

devshgraphicsprogramming:

PR #795, opened 2024-11-23
More fft utils

Fletterio:
## Description
Still adding more changes on this side

## Testing 
All thesting done on the FFT_Bloom example on its respective branch


include/nbl/builtin/hlsl/complex.hlsl
```diff
@@ -379,6 +379,19 @@ complex_t<Scalar> rotateRight(NBL_CONST_REF_ARG(complex_t<Scalar>) value)
     return retVal;
 }
 
+// Annoyed at having to write a lot of boilerplate to do a select
+// Essentially returns what you'd expect from doing `condition ? a : b`
+template<typename Scalar>
+complex_t<Scalar> ternaryOperator(bool condition, NBL_CONST_REF_ARG(complex_t<Scalar>) a, NBL_CONST_REF_ARG(complex_t<Scalar>) b)
```
devshgraphicsprogramming, line 385: make a functor similar to `plus_assign`, `plus` etc. thats called `ternary_operator`

I think following might work for `!is_fundamental_v<T>` types
```cpp
if (condition)
   return a;
else
   return b;
```

Fletterio: Checked godbolt, using `?` with vector types compiles down to a single select, doing `if (condition) else` compiles down to two selects :/

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -1,9 +1,47 @@
...
+    for (auto i : axes)
+    {
+        dimensions[i] = core::roundUpToPoT(dimensions[i]);
+        if (realFFT && !axisCount++)
+            dimensions[i] /= 2;
+    }
+    return dimensions;
+}
+
+static inline uint64_t getOutputBufferSize(const uint32_t3& inputDimensions, uint32_t numChannels, std::span<uint16_t> axes, bool realFFT = false, bool halfFloats = false)
```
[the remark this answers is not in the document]
Fletterio: `roundUpToPoT` is in `core/math/intutil.h`. I can create a `builtin/math/intutil.hlsl`, copy most functions over and refactor every usage of the functions in that file

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -1,13 +1,45 @@
...
+    uint16_t elementPerInvocationLog2 = 1u + hlsl::findMSB(core::max((inputArrayLength >> workgroupSizeLog2) - 1u, 1u));
+    return { elementPerInvocationLog2, workgroupSizeLog2 };
+}
+
+}
+}
+}
+}
+
+#else
```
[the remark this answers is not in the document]
Fletterio: HLSL has no findMSB, I implemented the SPIR-V intrinsics in the std.450 instruction set and the GLSL equivalents in `glsl_compat/core.hlsl` but not sure how to unify them since one in in `hlsl` namespace and the other is in `hlsl::glsl`

include/nbl/builtin/hlsl/complex.hlsl
devshgraphicsprogramming: btw soon `lerp`/`mix` with a boolean mixer can be used to work around the ternary shit on structs

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -135,6 +136,61 @@ float32_t2 unpackSnorm2x16(uint32_t p)
...
+        }
+    };
+} // namespace impl
+
+template<typename T>
+vector<int32_t, uint32_t(sizeof(T) / sizeof(scalar_type_t<T>))> findMSB(T value)
+{
+    return impl::findMSB<T, is_signed_v<T>, is_integral_v<T> >::__call(value);
+}
+
```
devshgraphicsprogramming, line 193: @Przemog1 I'll merge this PR before #801 , when resolving conflicts after merging `master` to 801, make sure to throw all this away.

include/nbl/builtin/hlsl/workgroup/fft.hlsl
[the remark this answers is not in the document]
devshgraphicsprogramming: > HLSL has no findMSB, I implemented the SPIR-V intrinsics in the std.450 instruction set and the GLSL equivalents in `glsl_compat/core.hlsl` but not sure how to unify them since one in in `hlsl` namespace and the other is in `hlsl::glsl`

@Przemog1 will fix in #801

include/nbl/builtin/hlsl/spirv_intrinsics/core.hlsl
```diff
@@ -234,13 +248,13 @@ template<class T, class U>
...
+template<typename Integral>
 [[vk::ext_instruction( spv::OpBitFieldUExtract )]]
-enable_if_t<is_unsigned_v<Unsigned>, Unsigned> bitFieldUExtract( Unsigned val, uint32_t offsetBits, uint32_t numBits );
+enable_if_t<is_integral_v<Integral>, Integral> bitFieldUExtract( Integral val, uint32_t offsetBits, uint32_t numBits );
 
-template<typename Signed>
+template<typename Integral>
 [[vk::ext_instruction( spv::OpBitFieldSExtract )]]
-enable_if_t<is_signed_v<Signed>, Signed> bitFieldSExtract( Signed val, uint32_t offsetBits, uint32_t numBits );
+enable_if_t<is_integral_v<Integral>, Integral> bitFieldSExtract( Integral val, uint32_t offsetBits, uint32_t numBits );
```
devshgraphicsprogramming, line 257: @Przemog1 see if you have this stuff in #801

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -1,13 +1,89 @@
...
+    const uint16_t elementsPerInvocationLog2 = 1u + findMSB(max((inputArrayLength >> workgroupSizeLog2) - 1u, 1u));
+    const OptimalFFTParameters retVal = { elementsPerInvocationLog2, workgroupSizeLog2 };
+    return retVal;
+}
+
+}
+}
+}
+}
+// ------------------------------- END CPP ONLY -------------------------------------------
```
devshgraphicsprogramming, line 73: why C++ only, you have literally nothing left that will only compile in C++

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -1,13 +1,89 @@
...
+{
+namespace workgroup
+{
+namespace fft
+{
+
+struct OptimalFFTParameters
+{
+    uint16_t elementsPerInvocationLog2;
+    uint16_t workgroupSizeLog2;
```
devshgraphicsprogramming, line 57: bitfields for extra memory sparing points (2x8 in single `uint16_t`)

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -1,13 +1,89 @@
...
+};
+
+inline OptimalFFTParameters optimalFFTParameters(const uint32_t maxWorkgroupSize, uint32_t inputArrayLength)
+{
```
devshgraphicsprogramming, line 61: you need to return {0,0} (invalid value) if the workgroupsize is not at least 4 or the input array is less than 2

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -1,13 +1,89 @@
...
+    // This is the logic found in core::roundUpToPoT to get the log2
+    const uint16_t workgroupSizeLog2 = 1u + findMSB(min(inputArrayLength / 2, maxWorkgroupSize) - 1u);
+    const uint16_t elementsPerInvocationLog2 = 1u + findMSB(max((inputArrayLength >> workgroupSizeLog2) - 1u, 1u));
```
devshgraphicsprogramming, line 64: There's an issue `inputArrayLength==2` leads to `findMSB(0)`, which I'm not sure if both our C++ and HLSL impl return `-1` in that case (c.f. with @Przemog1 )

You should also write an assert before computation of `elementsPerInvocationLog2` that `(inputArrayLength>>workgroupSizeLog2)>1` (obviously commented out for HLSL)

There should be no `max(,1)` necessary, as workgroup size may never be larger than input array length.

TBH I'd first **round up** the inputArrayLength to PoT and **round down** the maxWorkgroupSize to PoT while working out their logarithms, then proceed with the rest of the computations

include/nbl/builtin/hlsl/fft/common.hlsl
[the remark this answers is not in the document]
devshgraphicsprogramming: > `roundUpToPoT` is in `core/math/intutil.h`. I can create a `builtin/math/intutil.hlsl`, copy most functions over and refactor every usage of the functions in that file

ask @Przemog1 about the preferred location so it fits in with the spirit of #801 

You don't actually need to do big refactor if you provide a "legacy" alias (make the old `nbl::core::roundUpToPoT` call `nbl::hlsl::roundUpToPoT`) and add the `[[deprecated]]` attribute

include/nbl/builtin/hlsl/concepts/accessors/fft.hlsl
```diff
@@ -0,0 +1,87 @@
...
+#define NBL_CONCEPT_PARAM_0 (accessor, T)
+#define NBL_CONCEPT_PARAM_1 (index, uint32_t)
+#define NBL_CONCEPT_PARAM_2 (val, uint32_t)
+NBL_CONCEPT_BEGIN(3)
+#define accessor NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_0
+#define index NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_1
+#define val NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+NBL_CONCEPT_END(
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((accessor.set(index, val)), is_same_v, void))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((accessor.get(index, val)), is_same_v, void))
```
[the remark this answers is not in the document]
Fletterio: For that I need to touch the stuff in `memory_accessors.hlsl`

include/nbl/builtin/hlsl/fft/common.hlsl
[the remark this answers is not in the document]
Fletterio: I'm creating `builtin/hlsl/math/intutil.hlsl`, @Przemog1 lmk if you'd rather have it be named different or placed somewhere else

include/nbl/builtin/hlsl/workgroup/fft.hlsl
Fletterio: Actually the inputArray needs to be at least twice the subgroupsize of the device, since we require full subgroups

Fletterio: Not sure how to assert that from cpp side unless I make the methods divergent between host and device

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
devshgraphicsprogramming: hehe turns out 801 got merged before this.

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -1,60 +1,128 @@
...
+namespace fft 
+{
+
+// ------------------------------------------------- Utils ---------------------------------------------------------
+
+// Bit-reverses T as a binary string of length given by Bits
+template<typename T, uint16_t Bits NBL_FUNC_REQUIRES(is_integral_v<T> && Bits <= sizeof(T) * 8)
+T bitReverse(T value)
+{
+    return spirv::bitReverse<uint32_t>(value) >> (sizeof(T) * 8 - Bits);
```
devshgraphicsprogramming, line 119: `return spirv::bitReverse<T>(value) >> promote<T>(sizeof(T) * 8 - Bits);`

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -1,60 +1,128 @@
...
+    return newDimensions;
+}
+
+// template parameter N controls the number of dimensions of the input
+// template parameter M controls the number of dimensions we run an FFT along AND store the result
+// "axes" indicates which dimensions we run an FFT along AND store the result
+template <uint16_t N, uint16_t M NBL_FUNC_REQUIRES(M <= N)
+inline uint64_t getOutputBufferSize(NBL_CONST_REF_ARG(vector<uint32_t, N>) inputDimensions, uint32_t numChannels, NBL_CONST_REF_ARG(vector<uint16_t, M>) axes, bool realFFT = false, bool halfFloats = false)
+{
+    const vector<uint64_t, 3> paddedDims = padDimensions<N, M>(inputDimensions, axes);
```
devshgraphicsprogramming, line 41: the axes stuff is mightly confusing, see https://github.com/Devsh-Graphics-Programming/Nabla/pull/795#discussion_r1888150303

include/nbl/builtin/hlsl/fft/common.hlsl
devshgraphicsprogramming: also I don't get why you made it C++ only

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -219,10 +226,17 @@ T bitfieldInsert(T base, T insert, uint32_t offset, uint32_t bits)
...
 {
-    return spirv::bitFieldReverse<T>(value);
+    using unsigned_t = typename make_unsigned<T>::type;
+    // Bit-cast to uint to use the uint version of MSB
+    unsigned_t unsignedCasted = spirv::bitcast<unsigned_t, T>(value);
+    unsigned_t bits = findMSB(unsignedCasted);
+    NBL_CONSTEXPR_STATIC_INLINE unsigned_t width = promote<unsigned_t, scalar_type_t<unsigned_t> >(scalar_type_t<unsigned_t>(sizeof(scalar_type_t<unsigned_t>)));
+    // Do the shift in unsigned to ensure we do not get an arithmetic shift if signed, then cast back to original type
+    return spirv::bitcast<T, unsigned_t>(spirv::bitReverse<unsigned_t>(unsignedCasted) >> (width - bits));
 }
```
devshgraphicsprogramming, line 240: what are you doing? glm and SPIR-V bitfield reverse work exactly the same way!

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -206,6 +206,17 @@ inline FloatingPoint rsqrt(FloatingPoint x)
...
 
+template<typename Integer>
+inline Integer bitReverse(Integer base)
+{
+#ifdef __HLSL_VERSION
+	return spirv::bitReverse(base);
+#else
+	return glm::bitfieldReverse(base);
+#endif
+}
```
devshgraphicsprogramming, line 217: can stay like this until @Przemog1 institutes the amazing convention and rewrite.

devshgraphicsprogramming: actually don't call `spirv` directly, call `glsl`

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -30,6 +31,12 @@ genIUType bitfieldInsert(genIUType const& Base, genIUType const& Insert, int Off
 	return glm::bitfieldInsert<genIUType>(Base, Insert, Offset, Bits);
 }
 
+template<typename genIUType>
+genIUType bitfieldReverse(genIUType const& Value)
+{
+    return glm::bitfieldReverse<genIUType>(Value);
+}
+
```
devshgraphicsprogramming, line 39: remove all the C++ cases, make the whole header `__HLSL_VERSION` only

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -222,7 +229,7 @@ T bitfieldInsert(T base, T insert, uint32_t offset, uint32_t bits)
 template<typename T>
 T bitfieldReverse(T value)
 {
-    return spirv::bitFieldReverse<T>(value);
+    return spirv::bitReverse(value);
```
devshgraphicsprogramming, line 232: I know that GLM is crap and can't have the template parameter passed directly, but the spirv intrinsic can

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
devshgraphicsprogramming: that way you can also make sure you're passing the template parameter explictly.

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
Fletterio: I'm sure I had rolled this back a while ago wtf

include/nbl/builtin/hlsl/math/intutil.hlsl
```diff
@@ -0,0 +1,88 @@
...
+    uint64_t retval{};
+    for (const auto& it : initializer)
+        retval |= (1ull << it);
+    return retval;
+}
+
+#endif
+
+} // end namespace hlsl
+} // end namespace nbl
```
devshgraphicsprogramming, line 84: very nice, I'll let @Przemog1 review, because he's also getting rid of the old `round` and `isPoT` in another PR #760 (or the one after)

include/nbl/builtin/hlsl/fft/common.hlsl
Fletterio: The thing is you only want to pad along the dimensions you're going to run an FFT and store the dimensions afterward. For example for 2D FFT running y-axis first you only want to pad the y-axis since you do FFT along y -> store -> FFT along x -> NO store -> conv + IFFT along x -> only then store back

Fletterio: If you want the full result of a 2D FFT only then you pad both dimensions

include/nbl/builtin/hlsl/fft/common.hlsl
Fletterio: works for unsigned vectors only, should I require T be unsigned and be done with that?

include/nbl/core/math/intutil.h
```diff
@@ -83,13 +64,10 @@ NBL_FORCE_INLINE constexpr INT_TYPE align(INT_TYPE alignment, INT_TYPE size, INT
...
+[[deprecated("Use the nbl::hlsl version in builtin/hlsl/math/intutil.hlsl")]] 
 NBL_FORCE_INLINE constexpr uint64_t createBitmask(std::initializer_list<BITMASK_TYPE> initializer)
 {
-    static_assert(std::is_integral<BITMASK_TYPE>::value || std::is_enum<BITMASK_TYPE>::value, "Integral or enum required.");
-    uint64_t retval {};
-    for (const auto& it : initializer)
-        retval |= (1ull << it);
-    return retval;
+    return hlsl::createBitmask<BITMASK_TYPE>(initializer);
 }
```
devshgraphicsprogramming, line 71: don't try to port this one to HLSL, there's no initializer list type in HLSL

devshgraphicsprogramming (commented):
I hate GH reviews

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
devshgraphicsprogramming: ah  yes you did.

include/nbl/builtin/hlsl/fft/common.hlsl
devshgraphicsprogramming: > works for unsigned vectors only, should I require T be unsigned and be done with that?

yes require unsigned, it makes sense that way

include/nbl/builtin/hlsl/fft/common.hlsl
devshgraphicsprogramming: hmm, ok I see your point, problem was that your `axes` variable was unused in `padDimensions`

I'd always have `padDimensions` return the full final FFT size (for this only matters if the FFT is of a real signal and if it is what axis goes first)

Then `outputBufferSize` can return the buffer size after a certain pass.

For the stuff to make sense semantically, instead of `vector<uint16_t,N> axes` you need to call it `passAxisOrder` (component 0 tells you along which basis we perform the FFT), and default would be (0,1,2) (a.k.a. `iota`)

It makes no sense to have `N` and `M`, just `N` would do.

You'd also need a `const uint16_t pass`. so 
```cpp
uint64_t getOutputBufferSize(
   const uint32_t numChannels
   const vector<uint32_t,N> inputDimensions,
   const uint16_t passIx,
   const vector<uint16_t,N> axisPassOrder = truncate<vector<uint16_t,N> >(uint16_t4(0,1,2,3)),
   bool realFFT = false,
   bool halfFloats = false
)
```

Alos the thing has so many arguments, it should probably be a struct with `operator()` instead XD

Although then you get no default params.

include/nbl/builtin/hlsl/workgroup/fft.hlsl
devshgraphicsprogramming: > Actually the inputArray needs to be at least twice the subgroupsize of the device, since we require full subgroups

The thing is you can't run a workgroup smaller than a subgroup, so whenever FFT input array is less than that, the input signal should be padded with periodically repeated copies of itself (it makes the FFT output contain 0s in the "padding" frequencies)

devshgraphicsprogramming: > Not sure how to assert that from cpp side unless I make the methods divergent between host and device

take a ~~minWorkgroupSize~~ minSubgroupSize and a maxWorkgroupSize.

The min should be user provided.

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -1,9 +1,57 @@
...
+    vector<uint32_t, N> newDimensions = dimensions;
+    uint16_t axisCount = 0;
+    for (uint16_t i = 0u; i < M; i++)
+    {
+        newDimensions[i] = core::roundUpToPoT(newDimensions[i]);
+        if (realFFT && !axisCount++)
+            newDimensions[i] /= 2;
+    }
+    return newDimensions;
+}
```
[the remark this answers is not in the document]
devshgraphicsprogramming: pseudocode superseded in another conversion btw, due to remarks about the dimension depending on axis done, pass etc.

include/nbl/builtin/hlsl/fft/common.hlsl
devshgraphicsprogramming: > I'd always have padDimensions return the full final FFT size (for this only matters if the FFT is of a real signal and if it is what axis goes first)

Btw its easy and makes more sense to later just use `mix/lerp` to get the actualy output size:
```cpp
vector<bool,3> axesDone;
uint32_t3 passOutputDimension = mix(inputDimensions,paddedDimensions,axesDone);
```

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
devshgraphicsprogramming: @Przemog1 will sort it out

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
devshgraphicsprogramming: @Przemog1's job now

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
devshgraphicsprogramming: passed to @Przemog1

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -7,7 +7,11 @@
 #include "nbl/builtin/hlsl/cpp_compat/basic.h"
 #include "nbl/builtin/hlsl/spirv_intrinsics/core.hlsl"
 #include "nbl/builtin/hlsl/type_traits.hlsl"
+<<<<<<< HEAD
+#include "nbl/builtin/hlsl/bit.hlsl"
+=======
 #include "nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl"
+>>>>>>> master
```
devshgraphicsprogramming, line 14: you comitted a conflict

Fletterio: Yep, fixed but it seems I forgot to push

include/nbl/core/math/intutil.h
devshgraphicsprogramming: yes leave depr warning

Fletterio: oh I meant in the sense that the user should include the .hlsl file not the .h

include/nbl/builtin/hlsl/fft/common.hlsl
Fletterio: Cannot move this one to `bit.hlsl`, it causes a dependency cycle: I would need `cpp_compat.hlsl` defined in `bit.hlsl`, but `cpp_compat.hlsl` includes `iee754.hlsl` which includes `bit.hlsl`

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -38,24 +39,122 @@ struct OptimalFFTParameters
...
 
-inline OptimalFFTParameters optimalFFTParameters(const uint32_t maxWorkgroupSize, uint32_t inputArrayLength)
+/**
+* @brief Returns the best parameters (according to our metric) to run an FFT
+*
+* @param [in] maxWorkgroupSize The max number of threads that can be launched in a single workgroup
+* @param [in] inputArrayLength The length of the array to run an FFT on
+* @param [in] minSubgroupSize The smallest possible number of threads that can run in a single subgroup. 32 by default.
+*/
+inline OptimalFFTParameters optimalFFTParameters(uint32_t maxWorkgroupSize, uint32_t inputArrayLength, uint32_t minSubgroupSize = 32u)
```
devshgraphicsprogramming, line 57: don't default the subgroup size parameter

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -38,24 +39,122 @@ struct OptimalFFTParameters
...
 {
+    NBL_CONSTEXPR_STATIC OptimalFFTParameters invalidParameters = { 0 , 0 };
+
     // Round inputArrayLength to PoT
-    uint32_t FFTLength = 1u << (1u + findMSB(_static_cast<uint32_t>(inputArrayLength - 1u)));
+    const uint32_t FFTLength = 1u << (1u + findMSB(_static_cast<uint32_t>(inputArrayLength - 1u)));
     // Round maxWorkgroupSize down to PoT
-    uint32_t actualMaxWorkgroupSize = 1u << (findMSB(maxWorkgroupSize));
+    const uint32_t actualMaxWorkgroupSize = 1u << (findMSB(maxWorkgroupSize));
```
devshgraphicsprogramming, line 64: why are `roundUpToPoT` and `roundDownToPoT` not used?

include/nbl/builtin/hlsl/workgroup/fft.hlsl
```diff
@@ -38,24 +39,122 @@ struct OptimalFFTParameters
...
+* @brief Returns the best parameters (according to our metric) to run an FFT
+*
+* @param [in] maxWorkgroupSize The max number of threads that can be launched in a single workgroup
+* @param [in] inputArrayLength The length of the array to run an FFT on
+* @param [in] minSubgroupSize The smallest possible number of threads that can run in a single subgroup. 32 by default.
+*/
+inline OptimalFFTParameters optimalFFTParameters(uint32_t maxWorkgroupSize, uint32_t inputArrayLength, uint32_t minSubgroupSize = 32u)
 {
+    NBL_CONSTEXPR_STATIC OptimalFFTParameters invalidParameters = { 0 , 0 };
+
```
devshgraphicsprogramming, line 60: sanity checks need to be performed first:
```
minSubgroupSize>=4
maxWorkgroupSize>=minSubgroupSize
inputArrayLength>minSubgroupSize
```

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -14,34 +14,63 @@ namespace hlsl
...
+/**
+* @brief Returns the size of the full FFT computed, in terms of number of complex elements.
+*
+* @tparam N Number of dimensions of the signal to perform FFT on.
+*
+* @param [in] dimensions Size of the signal.
+* @param [in] realFFT Indicates whether the signal is real. False by default.
+* @param [in] firstAxis Indicates which axis the FFT is performed on first. Only relevant for real-valued signals. Must be less than N. 0 by default.
+*/
+inline vector<uint64_t, N> padDimensions(NBL_CONST_REF_ARG(vector<uint32_t, N>) dimensions, bool realFFT = false, uint16_t firstAxis = 0u)
```
devshgraphicsprogramming, line 28: there's no need for both a bool and `firstAxis` a `firstAxis` set to >=N tells you its not a realFFt (invalid value for a parameter that only makes sense if it is)

include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -14,34 +14,63 @@ namespace hlsl
...
+* @param [in] passIx Which pass the size is being computed for.
+* @param [in] axisPassOrder Order of the axis in which the FFT is computed in. Default is xyzw.
+* @param [in] realFFT True if the signal is real. False by default.
+* @param [in] halfFloats True if using half-precision floats. False by default.
+*/
+inline uint64_t getOutputBufferSize(
+    uint32_t numChannels,
+    NBL_CONST_REF_ARG(vector<uint32_t, N>) inputDimensions,
+    uint16_t passIx,
+    NBL_CONST_REF_ARG(vector<uint16_t, N>) axisPassOrder = _static_cast<vector<uint16_t, N> >(uint16_t4(0, 1, 2, 3)),
```
devshgraphicsprogramming, line 57: pro tip, both for C++ and HLSL it makes no sense/perf benefit to use constant reference (or `in`) for native vector types... the macro just impedes readability, you can just use `const` the `inline` takes care of everything

devshgraphicsprogramming: @Przemog1 does `_sttatic_cast` truncase now properly ?

include/nbl/builtin/hlsl/fft/common.hlsl
Fletterio: I don't get it

include/nbl/builtin/hlsl/fft/common.hlsl
Fletterio: wait there's no reference to vector types in Cpp?

include/nbl/builtin/hlsl/fft/common.hlsl
devshgraphicsprogramming: `realFFT==false` implies you don't care about the value of `firstAxis` and if you cared the only valid parameter range for `firstAxis` is [0,N)

hence why not derive `realFFT` from `firstAxis<N` and declare just `firstAxis=N`

include/nbl/builtin/hlsl/workgroup/fft.hlsl
Fletterio: They were in the .h and I had written this before I made the hlsl version, then forgot to update

PR #797, opened 2024-12-04
Hlsl bxdfs

devshgraphicsprogramming:


keptsecret:
TODO: all `LightSample::create` calls need to be templated. Forgot about that

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -15,30 +17,74 @@ namespace hlsl
 namespace bxdf
 {
 
+// returns unnormalized vector
+// TODO: template these?
+float computeUnnormalizedMicrofacetNormal(bool _refract, float3 V, float3 L, float orientedEta)
```
devshgraphicsprogramming, line 22: yes you could, on the scalar type, for good measure

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -15,30 +17,74 @@ namespace hlsl
...
+{
+    const float etaFactor = (_refract ? orientedEta : 1.0);
+    const float3 tmpH = V + L * etaFactor;
+    return _refract ? (-tmpH) : tmpH;
```
devshgraphicsprogramming, line 26: btw you could make a function somewhere (to be moved into the @Przemog1 ieee754 namespace) that conditionally flips the sign of a float

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -15,30 +17,74 @@ namespace hlsl
...
+}
+// returns normalized vector, but NaN when
```
devshgraphicsprogramming, line 28: finish comment "NaN when result is 0 length"

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -15,30 +17,74 @@ namespace hlsl
...
+#define NBL_CONCEPT_PARAM_2 (dirDotN,U)
+NBL_CONCEPT_BEGIN(4)
+#define rdirinfo NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_0
+#define N NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_1
+#define dirDotN NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+NBL_CONCEPT_END(
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((rdirinfo.direction), ::nbl::hlsl::is_same_v, vector<U, 3>))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((rdirinfo.getDirection()), ::nbl::hlsl::is_same_v, vector<U, 3>))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((rdirinfo.transmit()), ::nbl::hlsl::is_same_v, T))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((rdirinfo.reflect(N, dirDotN)), ::nbl::hlsl::is_same_v, T))
```
devshgraphicsprogramming, line 59: maybe lets require a refract as well ?

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -48,358 +94,883 @@ struct Basic
...
+#define normT NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+#define normB NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
+NBL_CONCEPT_END(
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((aniso.T), ::nbl::hlsl::is_same_v, vector<U,3>))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((aniso.B), ::nbl::hlsl::is_same_v, vector<U,3>))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((aniso.TdotV), ::nbl::hlsl::is_same_v, U))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((aniso.BdotV), ::nbl::hlsl::is_same_v, U))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((T::create(iso,normT,normB)), ::nbl::hlsl::is_same_v, T))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((T::create(iso,normT)), ::nbl::hlsl::is_same_v, T))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((T::create(iso)), ::nbl::hlsl::is_same_v, T))
```
devshgraphicsprogramming, line 161: the tangent-frame-less constructor shouldn't be required for the concept

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -48,358 +94,883 @@ struct Basic
...
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((aniso.getTangentFrame()), ::nbl::hlsl::is_same_v, matrix<U,3,3>))
+) && Isotropic<I, B, U> && ray_dir_info::Basic<B, U>;
+#undef normB
+#undef normT
+#undef iso
+#undef aniso
+#include <nbl/builtin/hlsl/concepts/__end.hlsl>
+
+template<class RayDirInfo, typename U NBL_PRIMARY_REQUIRES(ray_dir_info::Basic<RayDirInfo, U>)
+struct SAnisotropic : SIsotropic<RayDirInfo, U>
```
devshgraphicsprogramming, line 172: we may need to debate compositon vs. inheritance for this

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -15,30 +17,74 @@ namespace hlsl
...
-  
-  Basic reflect(const float3 N, const float directionDotN)
-  {
-    Basic retval;
-    retval.direction = nbl::hlsl::reflect(direction,N,directionDotN);
-    return retval;
-  }
-
-  float3 direction;
+    using vector_t = vector<T, 3>;
```
devshgraphicsprogramming, line 70: I'd call it `vector_type` to keep with stdlib conventions for names of nested typedefs

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: you probably want to require some nested typedefs exist like `vector_type`

devshgraphicsprogramming: then if you require `scalar_type` exists in `T` you don't need to take a `U`

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: and add a `scalar_type`

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -48,358 +94,883 @@ struct Basic
 namespace surface_interactions
 {
 
-template<class RayDirInfo>
-struct Isotropic
+#define NBL_CONCEPT_NAME Isotropic
+#define NBL_CONCEPT_TPLT_PRM_KINDS (typename)(typename)(typename)
+#define NBL_CONCEPT_TPLT_PRM_NAMES (T)(B)(U)                        // B is type Basic<T>
```
devshgraphicsprogramming, line 99: if you require `Basic<B>` to have a `vector_type` and `scalar_type` you don't need the B and U parameters

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -48,358 +94,883 @@ struct Basic
...
+#define NBL_CONCEPT_PARAM_0 (iso,T)
+#define NBL_CONCEPT_PARAM_1 (normV,B)
+#define NBL_CONCEPT_PARAM_2 (normN,vector<U, 3>)
```
devshgraphicsprogramming, line 102: I haven't checked, but `(normV,typename T::ray_dir_info_type)` and `(normN,typename T::vector_type)` should work

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -48,358 +94,883 @@ struct Basic
...
 };
 
 }
 
 
-template<class RayDirInfo>
-struct LightSample
+#define NBL_CONCEPT_NAME Sample
+#define NBL_CONCEPT_TPLT_PRM_KINDS (typename)(typename)(typename)(typename)(typename)
+#define NBL_CONCEPT_TPLT_PRM_NAMES (T)(I)(A)(B)(U)      // I type Isotropic, A type Aniso, B type Basic (getting clunky)
```
devshgraphicsprogramming, line 220: I'll review this later cause its pretty meaty and you need the `_type` aliases first so there's less template params.

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -48,358 +94,883 @@ struct Basic
...
+#define NBL_CONCEPT_PARAM_0 (sample_,T)
+#define NBL_CONCEPT_PARAM_1 (iso,I)
+#define NBL_CONCEPT_PARAM_2 (aniso,A)
+#define NBL_CONCEPT_PARAM_3 (rdirinfo,B)
+#define NBL_CONCEPT_PARAM_4 (pV,vector<U, 3>)
+#define NBL_CONCEPT_PARAM_5 (frame,matrix<U, 3, 3>)
+#define NBL_CONCEPT_PARAM_6 (pT,vector<U, 3>)
+#define NBL_CONCEPT_PARAM_7 (pB,vector<U, 3>)
+#define NBL_CONCEPT_PARAM_8 (pN,vector<U, 3>)
```
devshgraphicsprogramming, line 229: btw you don't need to make separate (new) arguments of the same type for a concept constraint, its enough you have one `normDir,vector<U,3>` and feed it in place of every T,B,N,V,L

include/nbl/builtin/hlsl/bxdf/ndf.hlsl
```diff
@@ -0,0 +1,144 @@
...
+	return w2 * w2 * atab * numbers::inv_pi<T>;
+}
+
+template<typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+T ggx_aniso(T TdotH2, T BdotH2, T NdotH2, T ax, T ay, T ax2, T ay2)
+{
+	T a2 = ax * ay;
+	T denom = TdotH2 / ax2 + BdotH2 / ay2 + NdotH2;
+	return numbers::inv_pi<T> / (a2 * denom * denom);
+}
```
devshgraphicsprogramming, line 137: every ndf should be a struct in CamelCase with argumentless `operator()`, and the arguments should be members

devshgraphicsprogramming: then maybe you can have some common argument struct like `Isotropic` and `Anisotropic` (and Squared non Squared types) which has the `NdotH` or NdotH2` which will make the interop with the MicrofacetCache introp nicer

include/nbl/builtin/hlsl/bxdf/ndf.hlsl
```diff
@@ -0,0 +1,144 @@
...
+template<typename T, bool ggx NBL_FUNC_REQUIRES(is_scalar_v<T>)
+T microfacet_to_light_measure_transform(T NDFcos, T maxNdotV)
+{
+    return impl::microfacet_to_light_measure_transform<T,ggx>::__call(NDFcos, maxNdotV);
+}
+
+
+// blinn-phong
+template<typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+T blinn_phong(T NdotH, T n)
```
devshgraphicsprogramming, line 81: leave a comment that `n` is the "shininess exponent" in the original paper

include/nbl/builtin/hlsl/bxdf/ndf.hlsl
devshgraphicsprogramming: also remember `using scalar_type = T`

include/nbl/builtin/hlsl/bxdf/ndf.hlsl
```diff
@@ -0,0 +1,144 @@
...
+{
+namespace bxdf
+{
+namespace ndf
+{
+
+// common
+namespace impl
+{
+template<typename T, bool ggx = false>
```
devshgraphicsprogramming, line 22: instead of having a bool, you can define a trait `ndf::is_ggx_v<NDF>` and take the `NDF` instead of the `T`

This way you get a more "rock solid" partial spec

devshgraphicsprogramming: that will require you to move the `microfaced_to_light_measure_transform` to the bottom of the file

include/nbl/builtin/hlsl/bxdf/ndf.hlsl
```diff
@@ -0,0 +1,144 @@
...
+T microfacet_to_light_measure_transform(T NDFcos, T absNdotV, bool transmitted, T VdotH, T LdotH, T VdotHLdotH, T orientedEta)
+{
+    return impl::microfacet_to_light_measure_transform<T,ggx>::__call(NDFcos, absNdotV, transmitted, VdotH, LdotH, VdotHLdotH, orientedEta);
+}
+
+template<typename T, bool ggx NBL_FUNC_REQUIRES(is_scalar_v<T>)
+T microfacet_to_light_measure_transform(T NDFcos, T maxNdotV)
+{
+    return impl::microfacet_to_light_measure_transform<T,ggx>::__call(NDFcos, maxNdotV);
+}
```
devshgraphicsprogramming, line 76: just have the thing as a struct with partial spec, no need to have it a function.

Also same thing with the arguments, turn into members

devshgraphicsprogramming: Also imho it would be best to split this into 6 cases:
- GGX vs non GGX
- BxDF, refract only, reflect only (brdf doesn't transmit/refract, so no eta needed)

you can template on an enum
```cpp
REFLECT_BIT = 0b01,
REFRACT_BIT = 0b10,
REFLECT_REFRACT_BIT = 0b11
```

include/nbl/builtin/hlsl/bxdf/common.hlsl
Fletterio: Minor nit but I was copying this code to figure out how to set up concepts on the fft branch and I think the `N` in `NBL_CONCEPT_BEGIN(N)` refers not to the number of requirements you put in `NBL_CONCEPT_END` but the number of `#define`s you do between `NBL_CONCEPT_BEGIN` and `NBL_CONCEPT_END`

devshgraphicsprogramming: > Minor nit but I was copying this code to figure out how to set up concepts on the fft branch and I think the `N` in `NBL_CONCEPT_BEGIN(N)` refers not to the number of requirements you put in `NBL_CONCEPT_END` but the number of `#define`s you do between `NBL_CONCEPT_BEGIN` and `NBL_CONCEPT_END`

yes exactly, numver of the implicit local parameters

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -50,7 +50,11 @@ T computeMicrofacetNormal(bool _refract, vector<T,3> V, vector<T,3> L, T oriente
 // if V and L are on different sides of the surface normal, then their dot product sign bits will differ, hence XOR will yield 1 at last bit
 bool isTransmissionPath(float NdotV, float NdotL)
 {
+#ifdef __HLSL_VERSION
     return bool((asuint(NdotV) ^ asuint(NdotL)) & 0x80000000u);
+#else
+    return bool((reinterpret_cast<uint32_t &>(NdotV) ^ reinterpret_cast<uint32_t &>(NdotL)) & 0x80000000u);
+#endif
```
devshgraphicsprogramming: use `bit_cast` then you can cover both HLSL and C++

devshgraphicsprogramming:
Comments are as they are, opening a new PR

devshgraphicsprogramming:
Everything resolved in #899

PR #801, opened 2024-12-11
Cpp compat intrinsics refactor

Przemog1:


include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
```diff
@@ -0,0 +1,46 @@
...
+[[vk::ext_instruction(GLSLstd450::GLSLstd450FindILsb, "GLSL.std.450")]]
+enable_if_t<is_integral_v<Integral> && (sizeof(scalar_type_t<Integral>) == 4), Integral> findILsb(Integral value);
+
+template<typename Integral>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450FindSMsb, "GLSL.std.450")]]
+enable_if_t<is_integral_v<Integral> && (sizeof(scalar_type_t<Integral>) == 4), Integral> findSMsb(Integral value);
+
+template<typename Integral>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450FindUMsb, "GLSL.std.450")]]
+enable_if_t<is_integral_v<Integral> && (sizeof(scalar_type_t<Integral>) == 4), Integral> findUMsb(Integral value);
```
devshgraphicsprogramming, line 27: I think these you can just not template because `Integral` can only be `uint32_t` or `int32_t` for the who functions

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
```diff
@@ -0,0 +1,46 @@
...
+[[vk::ext_instruction(GLSLstd450::GLSLstd450Exp2, "GLSL.std.450")]]
+enable_if_t<is_floating_point<FloatingPoint>::value, FloatingPoint> Exp2(FloatingPoint val);
+
+template<typename FloatingPoint>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450InverseSqrt, "GLSL.std.450")]]
+enable_if_t<is_floating_point_v<FloatingPoint>, FloatingPoint> InverseSqrt(FloatingPoint val);
+
+template<typename FloatingPoint>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450Floor, "GLSL.std.450")]]
+enable_if_t<is_floating_point_v<FloatingPoint>, FloatingPoint> Floor(FloatingPoint val);
```
devshgraphicsprogramming, line 39: consistency, start with lower case, so `exp2`, `inverseSqrt`, etc.

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
Przemog1: hence `(sizeof(scalar_type_t<Integral>) == 4)`

devshgraphicsprogramming: I mean that `S` in the intrinsic means `int` (signed) only and `U` means `uint` (unsigned) only so there really isn't any room for any other type

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
```diff
@@ -38,6 +38,18 @@ template<typename FloatingPoint>
 [[vk::ext_instruction(GLSLstd450::GLSLstd450Floor, "GLSL.std.450")]]
 enable_if_t<is_floating_point_v<FloatingPoint>, FloatingPoint> floor(FloatingPoint val);
 
+template<typename FloatingPoint>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450Cross, "GLSL.std.450")]]
+enable_if_t<is_floating_point_v<FloatingPoint>, vector<FloatingPoint, 3> > cross(in vector<FloatingPoint, 3> lhs, in vector<FloatingPoint, 3> rhs);
+
+template<typename FloatingPoint>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450FMix, "GLSL.std.450")]]
+enable_if_t<is_floating_point_v<FloatingPoint>, FloatingPoint> fMix(FloatingPoint val);
```
devshgraphicsprogramming, line 47: takes 3 arguments

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
```diff
@@ -0,0 +1,58 @@
...
+[[vk::ext_instruction(GLSLstd450::GLSLstd450Exp2, "GLSL.std.450")]]
+enable_if_t<is_floating_point<FloatingPoint>::value, FloatingPoint> exp2(FloatingPoint val);
+
+template<typename FloatingPoint>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450InverseSqrt, "GLSL.std.450")]]
+enable_if_t<is_floating_point_v<FloatingPoint>, FloatingPoint> inverseSqrt(FloatingPoint val);
+
+template<typename FloatingPoint>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450Floor, "GLSL.std.450")]]
+enable_if_t<is_floating_point_v<FloatingPoint>, FloatingPoint> floor(FloatingPoint val);
```
devshgraphicsprogramming, line 39: btw if you read the spec, some of these take not only floating points but vectors of floating point

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -0,0 +1,302 @@
...
+DEFINE_EXP2_SPECIALIZATION(uint32_t)
+DEFINE_EXP2_SPECIALIZATION(uint64_t)
+
+template<typename FloatingPoint>
+inline FloatingPoint rsqrt(FloatingPoint x)
+{
+#ifdef __HLSL_VERSION
+	return spirv::inverseSqrt(x);
+#else
+	return 1.0f / std::sqrt(x);
```
devshgraphicsprogramming, line 295: leave a comment `TODO: https://stackoverflow.com/a/62239778`

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -0,0 +1,302 @@
...
+	min(a, b);
+#else
+	return std::min(a, b);
+#endif
+}
+
+template<typename T>
+inline T max(NBL_CONST_REF_ARG(T) a, NBL_CONST_REF_ARG(T) b)
+{
+	return lerp<T>(a, b, b > a);
```
devshgraphicsprogramming, line 242: do it same way as min

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -0,0 +1,302 @@
...
+#else
+	return reinterpret_cast<matrix<T, N, M>&>(glm::inverse(reinterpret_cast<typename matrix<T, N, M>::Base const&>(m)));
+#endif
+}
+
+template<typename T, typename U>
+inline T lerp(NBL_CONST_REF_ARG(T) x, NBL_CONST_REF_ARG(T) y, NBL_CONST_REF_ARG(U) a)
+{
+#ifdef __HLSL_VERSION
+	return spirv::fMix(x, y, a);
```
devshgraphicsprogramming, line 198: `fMix` only works on floats, you really need an impl-struct and partial specs

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -0,0 +1,302 @@
...
+	return dot_product_impl::dot_helper<T>::dot(lhs, rhs);
+}
+
+// determinant not defined cause its implemented via hidden friend
+// https://stackoverflow.com/questions/67459950/why-is-a-friend-function-not-treated-as-a-member-of-a-namespace-of-a-class-it-wa
+template<typename T, uint16_t N, uint16_t M>
+inline T determinant(NBL_CONST_REF_ARG(matrix<T, N, M>) m)
+{
+#ifdef __HLSL_VERSION
+
```
devshgraphicsprogramming, line 112: HLSL/SPIR-V version?

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -0,0 +1,302 @@
...
+namespace nbl
+{
+namespace hlsl
+{
+
+template<typename Integer>
+int bitCount(NBL_CONST_REF_ARG(Integer) val)
+{
+#ifdef __HLSL_VERSION
+	return countbits(val);
```
devshgraphicsprogramming, line 26: HLSL bitcount only works on 32bit values, GLM probably too, need care here

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -0,0 +1,302 @@
...
+{
+#ifdef __HLSL_VERSION
+
+#else
+	return glm::determinant(reinterpret_cast<typename matrix<T, N, M>::Base const&>(m));
+#endif
+}
+
+template<typename Integer>
+int findLSB(NBL_CONST_REF_ARG(Integer) val)
```
devshgraphicsprogramming, line 119: clamp, floor, and everything below this line should have overloads for vectors.

Some of these like clamp, min, max, floor, etc. actually work on vectors too in SPIR-V

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
devshgraphicsprogramming: whenever an operation in SPIR-V does vectors and scalars (or in general two very different types), you can actually declare an overload (the intrinsic twice) as long as the `enable_if_t` clauses are mutually exclusive (no ambiguity).

will works just fine with SFINAE.

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
Przemog1: makes sense

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
Przemog1: no need for overloads, it will work

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -20,10 +20,19 @@ namespace hlsl
...
 
 template<typename Integer>
-int bitCount(NBL_CONST_REF_ARG(Integer) val)
+inline int bitCount(NBL_CONST_REF_ARG(Integer) val)
 {
 #ifdef __HLSL_VERSION
+	if (sizeof(Integer) == 8u)
+	{
+		uint32_t lowBits = val;
+		uint32_t highBits = val >> 32u;
```
devshgraphicsprogramming, line 29: I think you need to bitcast `val` to unsigned int first, right shifting a signed int has weird behaviours w.r.t the sign bit

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -191,28 +201,52 @@ inline matrix<T, N, M> inverse(NBL_CONST_REF_ARG(matrix<T, N, M>) m)
...
 
+namespace cpp_compat_intrinsics_impl
+{
+
+	// TODO: concept requiring T to be a float
 template<typename T, typename U>
-inline T lerp(NBL_CONST_REF_ARG(T) x, NBL_CONST_REF_ARG(T) y, NBL_CONST_REF_ARG(U) a)
+struct lerp_helper
 {
+	static inline T lerp(NBL_CONST_REF_ARG(T) x, NBL_CONST_REF_ARG(T) y, NBL_CONST_REF_ARG(U) a)
```
devshgraphicsprogramming: I'd constrain that either `is_same_v<T,U> || is_same_v<scalar_type_t<T>,U>` or something similar, you probably don't want a weird `U`  on this spec

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
```diff
@@ -18,13 +18,11 @@ template<typename Integral>
...
 
-template<typename Integral>
 [[vk::ext_instruction(GLSLstd450::GLSLstd450FindSMsb, "GLSL.std.450")]]
-enable_if_t<is_integral_v<Integral> && (sizeof(scalar_type_t<Integral>) == 4), Integral> findSMsb(Integral value);
+int32_t findSMsb(int32_t value);
 
-template<typename Integral>
 [[vk::ext_instruction(GLSLstd450::GLSLstd450FindUMsb, "GLSL.std.450")]]
-enable_if_t<is_integral_v<Integral> && (sizeof(scalar_type_t<Integral>) == 4), Integral> findUMsb(Integral value);
+uint32_t findUMsb(uint32_t value);
```
devshgraphicsprogramming, line 24: check if `findUMsb` doesn't return -1 for 0, if so then it needs to return `int32_t` despite taking in `uint32_t`

devshgraphicsprogramming: also check `findILsb` while you're at it, i think it might return `int32_t` only as well

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
```diff
@@ -44,11 +42,11 @@ enable_if_t<is_floating_point_v<FloatingPoint>, vector<FloatingPoint, 3> > cross
 
 template<typename FloatingPoint>
 [[vk::ext_instruction(GLSLstd450::GLSLstd450FMix, "GLSL.std.450")]]
-enable_if_t<is_floating_point_v<FloatingPoint>, FloatingPoint> fMix(FloatingPoint val);
+enable_if_t<is_floating_point_v<FloatingPoint>, FloatingPoint> fMix(FloatingPoint val, FloatingPoint min, FloatingPoint max);
```
devshgraphicsprogramming: are you  sure its `interpolant,min,max` and not `min,max,interpolant`

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -190,21 +190,21 @@ inline T floor(NBL_CONST_REF_ARG(T) val)
...
 #endif
 }
 
 namespace cpp_compat_intrinsics_impl
 {
-
-	// TODO: concept requiring T to be a float
+	// TODO: concept requiring T to be a float when U is not bool
 template<typename T, typename U>
 struct lerp_helper
```
devshgraphicsprogramming, line 209: btw you don't need a default spec, its okay to foward declare, and only provide partials (useful if you want the "default" to be constrained with a concept)

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
```diff
@@ -0,0 +1,301 @@
...
+	using ScalarType = typename vector_traits<VectorType>::scalar_type;\
+\
+	static inline ScalarType dot(NBL_CONST_REF_ARG(VectorType) lhs, NBL_CONST_REF_ARG(VectorType) rhs)\
+	{\
+		return RETURN_VALUE;\
+	}\
+};\
+
+#ifdef __HLSL_VERSION
+#define BUILTIN_VECTOR_SPECIALIZATION_RET_VAL dot(lhs, rhs)
```
devshgraphicsprogramming, line 42: https://godbolt.org/z/7EPsWvPe7

```
DXC (trunk)
1
2
3
DXC (trunk) - 325ms

<source>:7:5: error: recursive functions are not allowed: function 'main' calls recursive function 'Functor<vector<float, 3> >::dot'
    static inline float dot(const in T lhs, const in T rhs)
    ^
```

only happens upon instantiation, but `::` before the HLSL intrinsics doesn't work, so you need to call your static function `__call` not `dot`

devshgraphicsprogramming: I;ve filed a bug btw, https://github.com/microsoft/DirectXShaderCompiler/issues/7035

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
```diff
@@ -0,0 +1,301 @@
...
+			const as_int valAsInt = reinterpret_cast<const as_int&>(val);
+			return glm::findMSB(valAsInt);
+		}
+		else
+		{
+			// GLM accepts only integer types, so idea is to cast input to integer type
+			using as_uint = typename unsigned_integer_of_size<sizeof(Integer)>::type;
+			const as_uint valAsUnsignedInt = reinterpret_cast<const as_uint&>(val);
+			return glm::findMSB(valAsUnsignedInt);
+		}
```
devshgraphicsprogramming, line 77: when were we using `findMSB` on floats or or something similar?

devshgraphicsprogramming: also you have a bunch of specializations later which call glm, what is this default spec for? other intger sizes than 32 ?

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
```diff
@@ -0,0 +1,301 @@
...
+};
+template<typename Integer, int N>
+struct find_msb_return_type<vector<Integer, N> >
+{
+	using type = vector<int32_t, N>;
+};
+template<typename Integer>
+using find_lsb_return_type = find_msb_return_type<Integer>;
+
+template<typename T, typename U>
```
devshgraphicsprogramming, line 231: add `NBL_STRUCT_CONSTRAINABLE` instead of closing brace, so that you can have constrained partial specs later

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
```diff
@@ -0,0 +1,301 @@
...
+#undef MIX_FUNCTION
+
+template<typename T>
+struct lerp_helper<T, bool>
+{
+	static inline T lerp(NBL_CONST_REF_ARG(T) x, NBL_CONST_REF_ARG(T) y, NBL_CONST_REF_ARG(bool) a)
+	{
+		return a ? y : x;
+	}
+};
```
devshgraphicsprogramming, line 281: you either need to use `if-else` here or do a partial spec with a constraint (`NBL_PARTIAL_REQ_TOP` and `NBL_PARTIAL_REQ_BOT`) on `is_fundamental_v` because structs don't support ternary op in HLSL

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
Przemog1: > when were we using `findMSB` on floats or or something similar?

mostly `core::bitflag`, which you can't static_cast to underlying type

include/nbl/builtin/hlsl/cpp_compat/intrinsics.hlsl
```diff
@@ -0,0 +1,341 @@
...
+T clamp(NBL_CONST_REF_ARG(T) val, NBL_CONST_REF_ARG(T) min, NBL_CONST_REF_ARG(T) max)
+{
+#ifdef __HLSL_VERSION
+	return clamp(val, min, max);
+#else
+	return glm::clamp(val, min, max);
+#endif
+}
+
+namespace cpp_compat_intrinsics_impl
```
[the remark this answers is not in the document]
Przemog1: done

include/nbl/asset/ICPUGraphicsPipeline.h
```diff
@@ -93,7 +93,7 @@ class ICPUGraphicsPipeline final : public ICPUPipeline<IGraphicsPipeline<ICPUPip
 
 		inline int8_t stageToIndex(const ICPUShader::E_SHADER_STAGE stage) const override
 		{
-			const auto stageIx = hlsl::findLSB(stage);
+			const auto stageIx = hlsl::findLSB(static_cast<std::underlying_type_t<ICPUShader::E_SHADER_STAGE>>(stage));
```
devshgraphicsprogramming, line 96: either partial spec for C++ only with `is_enum` constraint...

or use `core::bitflag` instead of `static_cast<std::underlying_type_t`

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
```diff
@@ -276,7 +313,10 @@ struct lerp_helper<T, bool>
 {
 	static inline T lerp(NBL_CONST_REF_ARG(T) x, NBL_CONST_REF_ARG(T) y, NBL_CONST_REF_ARG(bool) a)
 	{
-		return a ? y : x;
+		if (a)
+			return y;
+		else
+			return x;
 	}
 };
```
devshgraphicsprogramming: @Fletterio you can add a partial spec that uses `?` for `is_fundamental_v` if it means that much to you

devshgraphicsprogramming:
Is there a companion Examples-Tests PR or does this leave all examples unaffected ?

Przemog1:
> Is there a companion Examples-Tests PR or does this leave all examples unaffected ?

here: https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/162

include/nbl/asset/ICPUGraphicsPipeline.h
Przemog1: done

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
Fletterio: If there's no perf hit I'll just use this one

PR #803, opened 2024-12-16
Maths for HLSL BxDFs (template cmath, tgmath)

keptsecret:


include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -120,6 +121,26 @@ inline FloatingPoint isinf(NBL_CONST_REF_ARG(FloatingPoint) val)
 #endif
 }
 
+template<typename  T>
+inline T pow(NBL_CONST_REF_ARG(T) x, NBL_CONST_REF_ARG(T) y)
+{
+#ifdef __HLSL_VERSION
+    return spirv::pow(x, y);
+#else
+    return std::pow(x, y);
```
devshgraphicsprogramming, line 130: pass the template argument explicitly always so you don't fall victim to implicit conversions

Przemog1:
@keptsecret merge [cpp_intrinsics_improvements](https://github.com/Devsh-Graphics-Programming/Nabla/tree/cpp_intrinsics_improvements) please

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -290,8 +290,8 @@ bool partitionRandVariable(float leftProb, NBL_REF_ARG(float) xi, NBL_REF_ARG(fl
 #ifdef __HLSL_VERSION
     NBL_CONSTEXPR float NEXT_ULP_AFTER_UNITY = asfloat(0x3f800001u);
 #else
-    NBL_CONSTEXPR uint32_t val = 0x3f800001u;
-    NBL_CONSTEXPR float32_t NEXT_ULP_AFTER_UNITY = reinterpret_cast<float32_t &>( val );
+    uint32_t val = 0x3f800001u;
+    float32_t NEXT_ULP_AFTER_UNITY = reinterpret_cast<float32_t &>( val );
```
devshgraphicsprogramming, line 294: `bit_cast<float32_t>(0x3f800001u)` to make C++20 happy

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -366,7 +366,14 @@ struct bitFields    // need template?
...
     uint32_t __overwrite()
     {
+#ifdef __HLSL_VERSION
         return spirv::bitFieldInsert<uint32_t>(base, value, offset, count);
+#else
+        // TODO: double check implementation
+        const uint32_t shifted_masked_value = ~(0xffffffffu << count) << offset;
+        base &= ~shifted_masked_value;
+        return base | (value << offset);
+#endif
```
devshgraphicsprogramming, line 376: use `hlsl::bifieldInsert` should work both for C++ and HLSL AFAIK

include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -88,7 +88,7 @@ template<typename T>
 inline T floor(NBL_CONST_REF_ARG(T) val)
 {
 #ifdef __HLSL_VERSION
-    return spirv::floor(val);
+    return spirv::floor<T>(val);
 #else
     return glm::floor(val);
```
devshgraphicsprogramming, line 93: aren't the GLM functions templated as well?

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,580 @@
...
+{
+    impl::bitFields b = impl::bitFields::create(base, value, offset, count);
+    return b.__overwrite();
+}
+
+uint32_t bitFieldInsert(uint32_t base, uint32_t value, uint32_t offset, uint32_t count)
+{
+    impl::bitFields b = impl::bitFields::create(base, value, offset, count);
+    return b.__insert();
+}
```
devshgraphicsprogramming, line 396: but we have glm and spirv for this ?

include/nbl/builtin/hlsl/tgmath.hlsl
devshgraphicsprogramming: @Przemog1 will take ownership

include/nbl/builtin/hlsl/math/functions.hlsl
keptsecret: The situation with `bitfieldInsert` is confusing to me. This mostly reflects the GLSL version we have: https://github.com/Devsh-Graphics-Programming/Nabla/blob/c7bdd2a3ec4a5a83c485533b2d68e7c9c5b8994b/include/nbl/builtin/glsl/math/functions.glsl#L263-L277

keptsecret: But there's also these ones in `glsl_compat/core.hlsl`: https://github.com/Devsh-Graphics-Programming/Nabla/blob/c7bdd2a3ec4a5a83c485533b2d68e7c9c5b8994b/include/nbl/builtin/hlsl/glsl_compat/core.hlsl#L210-L226

So maybe I can remove this one? But then there's no `bitfieldOverwrite` and I don't know if they do the same thing as the GLSL version.

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,580 @@
...
+vector<T, M> applyChainRule(vector<T, M> dFdG, T dGdR)
+{
+    return impl::applyChainRule2D<T,M>::__call(dFdG, dGdR);
+}
+
+template<typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+T applyChainRule(T dFdG, T dGdR)
+{
+    return impl::applyChainRule1D<T>::__call(dFdG, dGdR);
+}
```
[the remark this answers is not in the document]
keptsecret: GLM uses `*` for matrix multiplications, doesn't it?  
Could change to use `mul` in #804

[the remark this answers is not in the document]
keptsecret: Also don't need partial spec?

include/nbl/builtin/hlsl/tgmath.hlsl
keptsecret: The compiler doesn't seem to like using GLM templated for some reason.  
Gives a `no matching overloaded function found`.

include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -0,0 +1,231 @@
...
+{
+#ifdef __HLSL_VERSION
+    return spirv::log<T>(val);
+#else
+    return std::log(val);
+#endif
+}
+
+template<typename  T>
+inline T abs(NBL_CONST_REF_ARG(T) val)
```
keptsecret, line 179: Getting an `ambiguous call` when using this function. Possible overlap with this one?
https://github.com/Devsh-Graphics-Programming/Nabla/blob/c7bdd2a3ec4a5a83c485533b2d68e7c9c5b8994b/include/nbl/builtin/hlsl/complex.hlsl#L331

devshgraphicsprogramming: could be, to be sure you can "hide"/comment out the `complex` one

Anyway the compiler usually tells you what are the options.

Example by przemek
image

include/nbl/builtin/hlsl/math/functions.hlsl
devshgraphicsprogramming: @Przemog1 is responsible here, but in general:

There's should be the SPIR-V intrinsic in `spirv` namespace which provides this GLSL builtin to HLSL.

Then there should be an implementation in C++ which does the same thing, its very likely that `glm` already has one.

And finally a `nbl::hlsl::` namespace function which calls the SPIR-V or C++ depending on the language `__HLSL_VERSION`

do not remove anything, coordinate with Przemek.

include/nbl/builtin/hlsl/tgmath.hlsl
devshgraphicsprogramming: > The compiler doesn't seem to like using GLM templated for some reason. Gives a `no matching overloaded function found`.

what were you trying?

It might be because `floor` in `glm` has two overloads and is unfortunately templated differently for vectors
https://github.com/g-truc/glm/blob/33b4a621a697a305bc3a7610d290677b96beb181/glm/common.hpp#L65

Its probably better to use `std::floor` and forward per channel because GLM will refuse to work on any other vector type.

devshgraphicsprogramming: leave as is, @Przemog1 will improve later.

include/nbl/builtin/hlsl/tgmath.hlsl
Przemog1: it can't be because of `abs` from `complex.hlsl`. `abs` from `complex.hlsl` only alows `complex_t` on input, and since it is a templated function, no implicit conversion will be allowed

include/nbl/builtin/hlsl/math/functions.hlsl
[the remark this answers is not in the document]
devshgraphicsprogramming: > GLM uses `*` for matrix multiplications, doesn't it? Could change to use `mul` in #804

we have our own `mul` now, for both C++ and HLSL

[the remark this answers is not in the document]
devshgraphicsprogramming: > Also don't need partial spec?

There's no custom/specialized behaviour to enable, the only thing we could do is maybe allow a different `Matrix` type.

But that can be noted as a TODO in a comment.

include/nbl/builtin/hlsl/tgmath.hlsl
devshgraphicsprogramming: anyway I'll let @Przemog1 fix

include/nbl/builtin/hlsl/ieee754.hlsl
```diff
@@ -0,0 +1,137 @@
...
+	return (ieee754::impl::bitCastToUintType(x) & traits<AsFloat>::signMask) >> ((sizeof(T) * 8) - 1);
+}
+
+template <typename T>
+NBL_CONSTEXPR_INLINE_FUNC typename unsigned_integer_of_size<sizeof(T)>::type extractSignPreserveBitPattern(T x)
+{
+	using AsFloat = typename float_of_size<sizeof(T)>::type;
+	return ieee754::impl::bitCastToUintType(x) & traits<AsFloat>::signMask;
+}
+
```
devshgraphicsprogramming, line 132: @Przemog1 we should have a `signFlip(x)` and `copySign(x,y)` (sign of y gets copied into x)

Przemog1: in `ieee754.hlsl`? also HLSL have `negate` intrinsic i can implement

devshgraphicsprogramming: > in `ieee754.hlsl`? also HLSL have `negate` intrinsic i can implement

ok negate then instead of signFlip :D

Przemog1: done

Przemog1: > ~in `ieee754.hlsl`? also HLSL have `negate` intrinsic i can implement~

it doesn't :V

Przemog1: actually will implement `flipSign` in `ieee754.hlsl`

src/nbl/builtin/CMakeLists.txt
```diff
@@ -227,6 +227,10 @@ LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/tgmath.hlsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/array_accessors.hlsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/matrix_utils/matrix_traits.hlsl")
 LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/vector_utils/vector_traits.hlsl")
+LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/matrix_utils/matrix_traits.hlsl")
+LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/matrix_utils/mul_output_t.hlsl")
```
keptsecret, line 231: duplicated `matrix_traits.hlsl`  
one needs removing

src/nbl/builtin/CMakeLists.txt
```diff
@@ -227,6 +227,10 @@ LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/tgmath.hlsl")
...
+LIST_BUILTIN_RESOURCE(NBL_RESOURCES_TO_EMBED "hlsl/tgmath.hlsl")
```
keptsecret, line 232: also duplicated `tgmath.hlsl` with line 226

include/nbl/builtin/hlsl/concepts/core.hlsl
```diff
@@ -0,0 +1,110 @@
...
+NBL_BOOL_CONCEPT integral = nbl::hlsl::is_integral_v<T> && nbl::hlsl::is_scalar_v<T>;
+
+template<typename T>
+NBL_BOOL_CONCEPT signed_integral = nbl::hlsl::is_signed_v<T> && nbl::hlsl::is_integral_v<T> && nbl::hlsl::is_scalar_v<T>;
+
+template<typename T>
+NBL_BOOL_CONCEPT unsigned_integral = !nbl::hlsl::is_signed_v<T> && ::nbl::hlsl::is_integral_v<T> && nbl::hlsl::is_scalar_v<T>;
+
+template<typename T>
+NBL_BOOL_CONCEPT floating_point = nbl::hlsl::is_floating_point_v<T> && nbl::hlsl::is_scalar_v<T>;
```
devshgraphicsprogramming, line 90: its weird that the type trait says vectors are true, but concept then says they are not.

I'd define these identically for HLSL and C++ and remove the `is_scalar_v` requirement

include/nbl/builtin/hlsl/concepts/core.hlsl
```diff
@@ -0,0 +1,110 @@
...
+template <typename T>
+concept floating_point = std::floating_point<T>;
+
+// Some other useful concepts.
+
+template<typename T, typename... Ts>
+concept any_of = (same_as<T, Ts> || ...);
+
+template <typename T>
+concept scalar = floating_point<T> || integral<T>;
```
devshgraphicsprogramming, line 58: I'd want `is_scalar_v && ` added because of https://github.com/Devsh-Graphics-Programming/Nabla/pull/803/files#r1904299686

include/nbl/builtin/hlsl/concepts/core.hlsl
```diff
@@ -0,0 +1,110 @@
...
+template<typename T>
+struct IsEmulatingFloatingPointType
+{
+	static const bool value = false;
+};
+}
+
+//! Floating point types are native floating point types or types that imitate native floating point types (for example emulated_float64_t)
+template<typename T>
+NBL_BOOL_CONCEPT FloatingPointLike = (nbl::hlsl::is_floating_point_v<T> && nbl::hlsl::is_scalar_v<T>) || impl::IsEmulatingFloatingPointType<T>::value;
```
devshgraphicsprogramming, line 105: I'd remove the `is_scalar_v` in line with https://github.com/Devsh-Graphics-Programming/Nabla/pull/803/files#r1904299686

and fully implement in terms of `impl`, let the default specialization of the impl struct be `value = is_floating_point_v<T>;` then it will feel less specifically hardcoded for our emulated float

include/nbl/builtin/hlsl/shapes/beziers.hlsl
```diff
@@ -412,7 +412,7 @@ struct Quadratic
         // p'(1/2) = 2(C-A)
             
         // exponent so large it would wipe the mantissa on any relative operation
-        const float_t PARAMETER_THRESHOLD = exp2(24);
+        const float_t PARAMETER_THRESHOLD = exp2(24.0f);
```
devshgraphicsprogramming, line 415: leave a comment that it should be `exp2<float_t>(numeric_limits<float_t>::digits)` ater `tgmath` has an `exp2`

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+{
+
+namespace impl
+{
+template<typename T, uint32_t LP, bool Odd=(LP&0x1)>
+struct lp_norm;
+
+// infinity case
+template<typename T>
+struct lp_norm<T,0,false>
```
devshgraphicsprogramming, line 25: you'll never hit it because 0 is not Odd :)

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+    static scalar_type_t<T> __call(const T v)
+    {
+        scalar_type_t<T> retval = abs<T>(v[0]);
+        for (int i = 1; i < rank<T>::value; i++)
+            retval = max<T>(abs<T>(v[i]),retval);
+        return retval;
+    }
+};
+
+// TOOD: is this doing what it should be?
```
devshgraphicsprogramming, line 36: yes ofc

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+        return __sum(v);
+    }
+};
+
+template<typename T>
+struct lp_norm<T,2,false>
+{
+    static scalar_type_t<T> __sum(const T v)
+    {
+        return dot<T>(v, v);   // TODO: wait for overloaded dot?
```
devshgraphicsprogramming, line 59: use the nabla templated dot and we're good

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+    }
+
+    static scalar_type_t<T> __call(const T v)
+    {
+        return sqrt<T>(__sum(v));
```
devshgraphicsprogramming, line 64: wrong template param to `sqrt` should be the scalar_type instead of `T`

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+{
+namespace hlsl
+{
+namespace math
+{
+
+namespace impl
+{
+template<typename T, uint32_t LP, bool Odd=(LP&0x1)>
+struct lp_norm;
```
devshgraphicsprogramming, line 21: to general nitpicks:
1. use the `functional.hlsl` array getters instead of `operator[]`
2. use the `vector_traits` to establish the `scalar_type_t` because the one from `type_traits` is only a trait for native vectors

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+template <typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+vector<T, 3> reflect(vector<T, 3> I, vector<T, 3> N, T NdotI)
+{
+    return N * 2.0f * NdotI - I;
+}
+
+template <typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+vector<T, 3> reflect(vector<T, 3> I, vector<T, 3> N)
+{
+    T NdotI = dot<T>(N, I);
```
devshgraphicsprogramming, line 93: 1. your `dot` wont compile, because its `dot<VectorT>`
2. I'd probably require `/*Vectorial<T> && */vector_traits<T>::Dimension==3` instead of taking `T` and slapping it in a `vector` (perf note, vector_traits existing is same as checking Vectorial but will make Clang/DXC choke a lot less)

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+        return backside;
+    }
+};
+}
+
+template<typename T NBL_FUNC_REQUIRES(is_scalar_v<T> || is_vector_v<T>)
+bool getOrientedEtas(NBL_REF_ARG(T) orientedEta, NBL_REF_ARG(T) rcpOrientedEta, scalar_type_t<T> NdotI, T eta)
+{
+    return impl::orientedEtas<T>::__call(orientedEta, rcpOrientedEta, NdotI, eta);
+}
```
devshgraphicsprogramming, line 134: this stuff belongs in `bxdf/common` or even `bxdf/fresnel`

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+    impl::refract<T> r = impl::refract<T>::create(I, N, backside, NdotI, NdotI2, rcpOrientedEta, rcpOrientedEta2);
+    return r.doReflectRefract(_refract);
+}
+
+template<typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+vector<T,3> reflectRefract(bool _refract, vector<T,3> I, vector<T,3> N, T NdotI, T eta)
+{
+    impl::refract<T> r = impl::refract<T>::create(I, N, NdotI, eta);
+    return r.doReflectRefract(_refract);
+}
```
devshgraphicsprogramming, line 266: remove them all, the only reason (as I've explained to @keptsecret ) so many overloads existed is because we didn't have structs & methods in HLSL.

Now that we have, we will oonly use structs for complex things like this.

Also this is fresnel related stuff and should be somewhere in `bxdf/fresnel` not regular common math

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+        return doReflectRefract(r, I, N, NdotI, NdotTorR, rcpOrientedEta);
+    }
+
+    vector_type I;
+    vector_type N;
+    bool backside;
+    T NdotI;
+    T NdotI2;
+    T rcpOrientedEta;
+    T rcpOrientedEta2;
```
devshgraphicsprogramming, line 213: stuff that deals with refractive indices definitely belongs in `bxdf/fresnel`

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+}
+
+
+// valid only for `theta` in [-PI,PI]
+template <typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+void sincos(T theta, NBL_REF_ARG(T) s, NBL_REF_ARG(T) c)
+{
+    c = cos<T>(theta);
+    s = sqrt<T>(1.0-c*c);
+    s = (theta < 0.0) ? -s : s; // TODO: test with XOR
```
devshgraphicsprogramming, line 275: make that `signflip` or `negate` funuction that takes a bool, and then use it here

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+template<typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+vector<T,3> reflectRefract(bool _refract, vector<T,3> I, vector<T,3> N, T NdotI, T eta)
+{
+    impl::refract<T> r = impl::refract<T>::create(I, N, NdotI, eta);
+    return r.doReflectRefract(_refract);
+}
+
+
+// valid only for `theta` in [-PI,PI]
+template <typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
```
devshgraphicsprogramming, line 270: you should change most `is_scalar_v` into a bool concept that matches float scalars and also allows `emulated_float64_t`

Also `scalar` matches integers and booleans, on which sin and cos don't make sense

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+void sincos(T theta, NBL_REF_ARG(T) s, NBL_REF_ARG(T) c)
+{
+    c = cos<T>(theta);
+    s = sqrt<T>(1.0-c*c);
+    s = (theta < 0.0) ? -s : s; // TODO: test with XOR
+}
+
+template <typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
+matrix<T, 3, 2> frisvad(vector<T, 3> n) // TODO: confirm dimensions of matrix
```
devshgraphicsprogramming, line 279: frankly I'd make it `void frisvad(const T normal, NBL_REF_ARG(T) tangent, NBL_REF_ARG(T) bitangent)` requiring that `vector_traits<T>::Dimension==3`

the reason is that this way its more semantically sound, instead of trying to scratch your head what row/column of the matrix is what (in the original GLSL, the first column - the first vector you get with `operator[]` or provide with the constructor was the tangent)

We had this conversation with @keptsecret on his thread and PR

Also https://github.com/Devsh-Graphics-Programming/Nabla/pull/803/files#r1904982026

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+{
+	const T a = 1.0 / (1.0 + n.z);
+	const T b = -n.x * n.y * a;
+	return (n.z < -0.9999999) ? matrix<T, 2, 3>(vector<T, 3>(0.0,-1.0,0.0), vector<T, 3>(-1.0,0.0,0.0)) : 
+        matrix<T, 2, 3>(vector<T, 3>(1.0-n.x*n.x*a, b, -n.x), vector<T, 3>(b, 1.0-n.y*n.y*a, -n.y));
```
devshgraphicsprogramming, line 284: also use an if instead of `?` we get no perf diff, but much better legibility (also the `n.z` condition can be moved up)

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+{
+    const float32_t3 condAbs = asfloat(asuint(x) & select(cond, (uint32_t3)0x7fFFffFFu, (uint32_t3)0xffFFffFFu));
+    return max(condAbs,limit);
+}
+
+template <>
+float32_t4 conditionalAbsOrMax<float32_t4>(bool cond, float32_t4 x, float32_t4 limit)
+{
+    const float32_t4 condAbs = asfloat(asuint(x) & select(cond, (uint32_t4)0x7fFFffFFu, (uint32_t4)0xffFFffFFu));
+    return max(condAbs,limit);
```
devshgraphicsprogramming, line 337: Require that `T` is floating point (matching emulated_float too), leave signed integer impl as a TODO

because the `condAbs` is not constexpr, you can use our `nbl::hlsl::bit_cast`

you can use `nbl::hlsl::promote` to turn a single scalar into a vector.

Also use `mix` instead of `select` then it should all work cross C++ and HLSL

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -0,0 +1,453 @@
...
+// TODO: even/odd cases
+}
+
+template<typename T, uint32_t LP NBL_FUNC_REQUIRES(LP>0)
+scalar_type_t<T> lpNormPreroot(NBL_CONST_REF_ARG(T) v)
+{
+    return impl::lp_norm<T,LP>::__sum(v);
+}
+
+template<typename T, uint32_t LP>
```
devshgraphicsprogramming, line 77: you probably want to require that `vector_traits<T>::scalar_type` is a float (including emulated float)

include/nbl/builtin/hlsl/math/functions.hlsl
Przemog1: there is `flipSign` function in the `ieee754.hlsl` file

include/nbl/builtin/hlsl/concepts/core.hlsl
Przemog1: done

include/nbl/builtin/hlsl/concepts/core.hlsl
Przemog1: done

include/nbl/builtin/hlsl/concepts/core.hlsl
Przemog1: done

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -276,7 +276,7 @@ void sincos(T theta, NBL_REF_ARG(T) s, NBL_REF_ARG(T) c)
 }
 
 template <typename T NBL_FUNC_REQUIRES(is_scalar_v<T>)
-matrix<T, 3, 2> frisvad(vector<T, 3> n) // TODO: confirm dimensions of matrix
+matrix<T, 2, 3> frisvad(vector<T, 3> n)
```
devshgraphicsprogramming, line 279: dont use `matrix` its semantically confusing

include/nbl/builtin/hlsl/concepts/core.hlsl
devshgraphicsprogramming: bump, the new code still only has `FloatingPoint<T> || Integral<T>`

include/nbl/builtin/hlsl/ieee754.hlsl
```diff
@@ -130,6 +130,27 @@ NBL_CONSTEXPR_INLINE_FUNC typename unsigned_integer_of_size<sizeof(T)>::type ext
...
 
+template <typename FloatingPoint NBL_FUNC_REQUIRES(hlsl::is_floating_point_v<FloatingPoint> && hlsl::is_scalar_v<FloatingPoint>)
+NBL_CONSTEXPR_INLINE_FUNC FloatingPoint copySign(FloatingPoint to, FloatingPoint from)
+{
+	using AsUint = typename unsigned_integer_of_size<sizeof(FloatingPoint)>::type;
+
+	const AsUint toAsUint = ieee754::impl::bitCastToUintType(to);
+	const AsUint fromAsUint = ieee754::impl::bitCastToUintType(from);
+
+	return bit_cast<FloatingPoint>(toAsUint | extractSignPreserveBitPattern(from));
```
devshgraphicsprogramming, line 141: this does not copy sign, because you need to clear the sign bit on `toAsUint` first

right now it will make `to` negative if any of the inputs are negative

btw are you sure you want to force that `FloatingPoint` is non-emulated and scalar? It could work on any float even emulated 64

include/nbl/builtin/hlsl/ieee754.hlsl
```diff
@@ -130,6 +130,27 @@ NBL_CONSTEXPR_INLINE_FUNC typename unsigned_integer_of_size<sizeof(T)>::type ext
...
+}
+
+template <typename FloatingPoint NBL_FUNC_REQUIRES(hlsl::is_floating_point_v<FloatingPoint>&& hlsl::is_scalar_v<FloatingPoint>)
+NBL_CONSTEXPR_INLINE_FUNC FloatingPoint flipSign(FloatingPoint val)
```
devshgraphicsprogramming, line 145: bool indicating whether to flip would be super useful actually and it would always get optimized away if const

> btw are you sure you want to force that FloatingPoint is non-emulated and scalar? It could work on any float even emulated 64

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
```diff
@@ -3,9 +3,17 @@
...
 #include <nbl/builtin/hlsl/concepts.hlsl>
 #include <nbl/builtin/hlsl/spirv_intrinsics/core.hlsl>
 #include <nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl>
+#include <nbl/builtin/hlsl/ieee754.hlsl>
+#include <nbl/builtin/hlsl/concepts.hlsl>
+#include <nbl/builtin/hlsl/concepts/vector.hlsl>
+
+#ifndef __HLSL_VERSION
+#include <bitset>
+#endif
```
devshgraphicsprogramming, line 16: is this needed here?

include/nbl/builtin/hlsl/spirv_intrinsics/glsl.std.450.hlsl
```diff
@@ -41,10 +41,26 @@ template<int N>
...
+[[vk::ext_instruction(GLSLstd450::GLSLstd450Exp, "GLSL.std.450")]]
+enable_if_t<is_floating_point<FloatingPoint>::value && !is_matrix_v<FloatingPoint>, FloatingPoint> exp(FloatingPoint val);
+
 template<typename FloatingPoint>
 [[vk::ext_instruction(GLSLstd450::GLSLstd450Exp2, "GLSL.std.450")]]
 enable_if_t<is_floating_point<FloatingPoint>::value && !is_matrix_v<FloatingPoint>, FloatingPoint> exp2(FloatingPoint val);
 
+template<typename FloatingPoint>
+[[vk::ext_instruction(GLSLstd450::GLSLstd450Log, "GLSL.std.450")]]
+enable_if_t<is_floating_point<FloatingPoint>::value && !is_matrix_v<FloatingPoint>, FloatingPoint> log(FloatingPoint val);
```
devshgraphicsprogramming, line 58: pow, exp and log don't take float64 without an extension

also as per our discord conversation, we need 5 named concepts in `spirv` namespace so the `enable_if_t` are clearer and easier to compare with SPIR-V spec

include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -4,70 +4,130 @@
...
+    return tgmath_impl::exp2_helper<T>::__call(x);
+}
+
+template<typename T NBL_FUNC_REQUIRES(concepts::FloatingPointLikeScalar<T> || concepts::FloatingPointLikeVectorial<T>)
+inline T log(NBL_CONST_REF_ARG(T) x)
+{
+    return tgmath_impl::log_helper<T>::__call(x);
+}
+
+template<typename T NBL_FUNC_REQUIRES(concepts::FloatingPointLikeScalar<T> || concepts::SignedIntegral<T> || concepts::FloatingPointLikeVectorial<T> || concepts::SignedIntVectorial<T>)
```
devshgraphicsprogramming, line 100: maybe don't put any constraints on the functions, and let the helpers do all the error reporting (no point doing it twice) as long as each `_helper` declares `NBL_STRUCT_CONSTRAINABLE` in the EMPTY forward declaration of the primary template

include/nbl/builtin/hlsl/math/functions.hlsl
devshgraphicsprogramming: > there is `flipSign` function in the `hlsl/ieee754.hlsl` file

the `flipSign` needs a bool arg to tell whether to flip or not

include/nbl/builtin/hlsl/tgmath.hlsl
```diff
@@ -4,70 +4,130 @@
...
+inline FloatingPoint isinf(NBL_CONST_REF_ARG(FloatingPoint) val)
+{
+    return tgmath_impl::isinf_helper<FloatingPoint>::__call(val);
+}
+
+template<typename V NBL_FUNC_REQUIRES(concepts::FloatingPointLikeVectorial<V>)
+inline vector<bool, vector_traits<V>::Dimension> isinf(NBL_CONST_REF_ARG(V) val)
+{
+    return tgmath_impl::isinf_helper<V>::__call(val);
+}
```
devshgraphicsprogramming, line 73: instead of overload, maybe you just want to require that this `_helper` declares a `using return_t = ` ?

include/nbl/builtin/hlsl/impl/tgmath_impl.hlsl
```diff
@@ -0,0 +1,767 @@
...
+{
+
+template<typename T, typename U NBL_STRUCT_CONSTRAINABLE>
+struct lerp_helper;
+
+#ifdef __HLSL_VERSION
+#define MIX_FUNCTION spirv::fMix
+#else
+#define MIX_FUNCTION glm::mix
+#endif
```
devshgraphicsprogramming, line 25: as per our discord conversation, every helper should have a FIRST partial specialization done with `is_same_v<decltype(spirv::...),expected_Return_type>` constraint in a `#ifdef __HLSL_VERSION` block

then the C++ versions and cross-language adaptors for emulated should come

include/nbl/builtin/hlsl/impl/tgmath_impl.hlsl
```diff
@@ -0,0 +1,767 @@
...
+	}
+};
+
+// ISNAN
+
+template<typename UnsignedInteger NBL_FUNC_REQUIRES(hlsl::is_integral_v<UnsignedInteger> && hlsl::is_unsigned_v<UnsignedInteger>)
+inline bool isnan_uint_impl(UnsignedInteger val)
+{
+	using AsFloat = typename float_of_size<sizeof(UnsignedInteger)>::type;
+	return bool((ieee754::extractBiasedExponent<UnsignedInteger>(val) == ieee754::traits<AsFloat>::specialValueExp) && (val & ieee754::traits<AsFloat>::mantissaMask));
```
devshgraphicsprogramming, line 95: there's a faster check:
1. AND with `(numeric_limits<UnsignedInteger>::max>>1)` - clear the sign mask
2. check for `> (specialValueExp<<mantissaBits)`

include/nbl/builtin/hlsl/impl/tgmath_impl.hlsl
```diff
@@ -0,0 +1,767 @@
...
+		return tgmath_impl::isnan_uint_impl(reinterpret_cast<const AsUint&>(x));
+#endif
+	}
+};
+
+template<typename V>
+NBL_PARTIAL_REQ_TOP(concepts::Vectorial<V>)
+struct isnan_helper<V NBL_PARTIAL_REQ_BOT(concepts::Vectorial<V>) >
+{
+	using output_t = vector<bool, hlsl::vector_traits<V>::Dimension>;
```
devshgraphicsprogramming, line 122: as per the discord convom, its best if there's a whole separate SPIR-V specialization which is not predicated on a concept but if the SPIR-V intrinsic is callable/compiles with that type

and the C++ specialization can be just for a scalar.

So you have 2 specializations in two different parts of `__HLSL_VERSION` block, instead of one that tries to reconcile incompatible concepts

include/nbl/builtin/hlsl/impl/tgmath_impl.hlsl
```diff
@@ -0,0 +1,767 @@
...
+};
+
+// ERF
+
+template<typename T NBL_STRUCT_CONSTRAINABLE>
+struct erf_helper;
+
+template<typename FloatingPoint>
+NBL_PARTIAL_REQ_TOP(concepts::FloatingPointScalar<FloatingPoint>)
+struct erf_helper<FloatingPoint NBL_PARTIAL_REQ_BOT(concepts::FloatingPointScalar<FloatingPoint>) >
```
devshgraphicsprogramming, line 231: the non-spirv forwarding functions can be done however you like, most sane is probably to do the per-channel broadcast specs as macros for partial spec

include/nbl/builtin/hlsl/impl/tgmath_impl.hlsl
```diff
@@ -0,0 +1,767 @@
...
+{
+	static FloatingPoint __call(NBL_CONST_REF_ARG(FloatingPoint) _x)
+	{
+#ifdef __HLSL_VERSION
+		const FloatingPoint a1 = 0.254829592;
+		const FloatingPoint a2 = -0.284496736;
+		const FloatingPoint a3 = 1.421413741;
+		const FloatingPoint a4 = -1.453152027;
+		const FloatingPoint a5 = 1.061405429;
+		const FloatingPoint p = 0.3275911;
```
devshgraphicsprogramming, line 241: pay attention to 2 things:
1. whether the constant literal has all the digits needed (float32 and float64 often have different digits
2. to get true FP64 constants in HLSL you need to use a stupid suffix, there's actually an `NBL_` macro you need to place around them cause DXC is holding back on implementing explicit sized fp16, fp32, fp64 suffices on float literals


Also, usually an implementation of a complex math function is different for float16, float32 and float64 (different levels of polynomial approximation/coefficients and number of steps/iterations) so I'd have a separate partial spec in preprocessor blocks of `__HLSL_VERSION

include/nbl/builtin/hlsl/impl/tgmath_impl.hlsl
devshgraphicsprogramming: Ideally I'd keep the functions without SPIR-V intrinsic equivalents in a separate file or near the bottom of this one

devshgraphicsprogramming:
`intrinsics.hlsl` and `functions.hlsl` I'll review some other day

include/nbl/builtin/hlsl/concepts/core.hlsl
Przemog1: now its just `concept scalar = nbl::hlsl::is_scalar_v<T>;`

include/nbl/builtin/hlsl/impl/tgmath_impl.hlsl
Przemog1: done here:
https://github.com/Devsh-Graphics-Programming/Nabla/pull/817

include/nbl/builtin/hlsl/tgmath.hlsl
Przemog1: leaving as it is

Przemog1: actually i will do it

Przemog1: done here: https://github.com/Devsh-Graphics-Programming/Nabla/pull/817

include/nbl/builtin/hlsl/tgmath.hlsl
Przemog1: done here: https://github.com/Devsh-Graphics-Programming/Nabla/pull/817

include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl
Przemog1: yes, i use the `bitset::count` function

include/nbl/builtin/hlsl/ieee754.hlsl
Przemog1: I don't get it, you want an additional `bool` function parameter indicating if function will do what it is supposed to do?

Przemog1: allowed emulated float as an input to this function here: https://github.com/Devsh-Graphics-Programming/Nabla/pull/817

include/nbl/builtin/hlsl/ieee754.hlsl
Przemog1: done here: https://github.com/Devsh-Graphics-Programming/Nabla/pull/817

devshgraphicsprogramming:
@keptsecret merge `master` into this and I review again

include/nbl/builtin/hlsl/math/functions.hlsl
keptsecret: `0 & 0x1` doesn't evaluate to `false`?

include/nbl/builtin/hlsl/math/functions.hlsl
keptsecret: I'll move `reflect` into  `bxdf/fresnel` as well then

include/nbl/builtin/hlsl/math/functions.hlsl
Przemog1: it does

devshgraphicsprogramming: my bad, excuse my braindeadness.

Przemog1: happens

include/nbl/builtin/hlsl/concepts/core.hlsl
```diff
@@ -29,13 +29,13 @@ template<typename T>
...
 template<typename T>
-NBL_BOOL_CONCEPT FloatingPoint = nbl::hlsl::is_floating_point_v<T>;
+NBL_BOOL_CONCEPT FloatingPoint = nbl::hlsl::is_floating_point_v<T> || nbl::hlsl::is_same_v<T, float16_t>;
 
 template<typename T>
 NBL_BOOL_CONCEPT Boolean = nbl::hlsl::is_same_v<T, bool> || (nbl::hlsl::is_vector_v<T> && nbl::hlsl::is_same_v<typename vector_traits<T>::scalar_type, bool>);
 
 template <typename T>
-NBL_BOOL_CONCEPT Scalar = nbl::hlsl::is_scalar_v<T>;
+NBL_BOOL_CONCEPT Scalar = nbl::hlsl::is_scalar_v<T> || nbl::hlsl::is_same_v<T, float16_t>;
```
devshgraphicsprogramming, line 38: merge `master`, this was the "wrong" fix

include/nbl/builtin/hlsl/tgmath/impl.hlsl
```diff
@@ -429,6 +451,91 @@ struct erfInv_helper<FloatingPoint NBL_PARTIAL_REQ_BOT(concepts::FloatingPointSc
...
 };
 
+template<>
+struct erfInv_helper<float64_t>
+{
+	static float64_t __call(NBL_CONST_REF_ARG(float64_t) _x)
+	{
+		float64_t x = clamp<float64_t>(_x, NBL_FP64_LITERAL(-0.99999), NBL_FP64_LITERAL(0.99999));
+
+		float64_t w = float64_t(-log_helper<float32_t>::__call((float32_t(NBL_FP64_LITERAL(1.0)) - x) * float32_t(NBL_FP64_LITERAL(1.0)) + x));
```
devshgraphicsprogramming, line 461: use `float64_t` in the exp, and then comment out (or otherwise disable) the entire partial spec

devshgraphicsprogramming: also check the license of the other implementations of our `erf` and `erfInv` functions and make sure they don't come from GLIBC or other GPL thing, and if they do find different library which has them under a different license AND MAKE SURE TO WRITE A COMMENT STATING WHERE THE CODE CAME FROM

keptsecret: The current ones don't seem to have a license attached. The `erf` approximations came from the formula in A&S "Handbook of Mathematical Functions". `erfInv` is from GPU Computing Gems.

I could re-add an `erf<float64_t>` implementation from MUSL `libm` which uses apache license like us. But it's still for HLSL only so not that useful.

devshgraphicsprogramming: then write down for the float64_t

devshgraphicsprogramming: also note the book, usually books are sane with their code snippets and don't license them weirdly (thats the whole point of selling them)

include/nbl/builtin/hlsl/math/functions.hlsl
```diff
@@ -292,13 +100,21 @@ void sincos(T theta, NBL_REF_ARG(T) s, NBL_REF_ARG(T) c)
...
+	const typename vector_traits<T>::scalar_type b = -normal.x * normal.y * a;
+    if (normal.z < -NBL_FP64_LITERAL(0.9999999))
+    {
+        tangent = T(0.0,-1.0,0.0);
+        bitangent = T(-1.0,0.0,0.0);
+    }
+    else
+    {
+        tangent = T(NBL_FP64_LITERAL(1.0)-normal.x*normal.x*a, b, -normal.x);
+        bitangent = T(b, NBL_FP64_LITERAL(1.0)-normal.y*normal.y*a, -normal.y);
```
devshgraphicsprogramming, line 116: why are you using fp64 literals?

just do 
```cpp
using scalar_t = typename vector_traits<T>::scalar_type;
const scalar_t unit = _static_cast<scalar_t>(1);
```

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -0,0 +1,148 @@
...
+        return backside;
+    }
+};
+}
+
+template<typename T NBL_FUNC_REQUIRES(is_scalar_v<T> || is_vector_v<T>)
+bool getOrientedEtas(NBL_REF_ARG(T) orientedEta, NBL_REF_ARG(T) rcpOrientedEta, scalar_type_t<T> NdotI, T eta)
+{
+    return impl::orientedEtas<T>::__call(orientedEta, rcpOrientedEta, NdotI, eta);
+}
```
devshgraphicsprogramming, line 56: we still have the construction of the old GLSL, you're then passing the 3 values around to other functions instead of taking a whole struct

what I need is
```cpp
namespace nbl::hlsl::bxdf::fresnel
{
template<typename T>
struct OrientedEtas
{
   T value;
   T rcp;
   bool backside;
};

template<typename T>
struct OrientedEtaRcps
{
   T value;
   T value2;
   bool backside;
};
}
```

and just different `create` methods

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -0,0 +1,148 @@
...
+
+
+template <typename T NBL_FUNC_REQUIRES(concepts::Vectorial<T>)
+T reflect(T I, T N, typename vector_traits<T>::scalar_type NdotI)
+{
+    return N * 2.0f * NdotI - I;
+}
```
devshgraphicsprogramming, line 63: should be a struct with a `computeNdotI` like the `refract`

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    vector_type doRefract()
+    {
+        return N * (NdotI * rcpOrientedEta + computeNdotT(backside, NdotI2, rcpOrientedEta2)) - rcpOrientedEta * I;
+    }
+
+    static vector_type doReflectRefract(bool _refract, NBL_CONST_REF_ARG(vector_type) _I, NBL_CONST_REF_ARG(vector_type) _N, scalar_type _NdotI, scalar_type _NdotTorR, scalar_type _rcpOrientedEta)
+    {
+        return _N * (_NdotI * (_refract ? _rcpOrientedEta : 1.0f) + _NdotTorR) - _I * (_refract ? _rcpOrientedEta : 1.0f);
```
devshgraphicsprogramming, line 125: btw I learnt recently `?` never emits a Select in SPIR-V we need to use `mix(T,T,bool)` unfortunately.

include/nbl/builtin/hlsl/tgmath/impl.hlsl
devshgraphicsprogramming: @keptsecret can you please put in the comments where the code comes from!?

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
keptsecret: From checking the intrinsics, `mix(T,T,bool)` also uses `?`
https://github.com/Devsh-Graphics-Programming/Nabla/blob/23124c582fde6496af21ea2cf85535ab2ecc0922/include/nbl/builtin/hlsl/cpp_compat/impl/intrinsics_impl.hlsl#L243-L251

Only `mix(T,T,T)` will use `spirv::fMix`

devshgraphicsprogramming: it needs to use `OpSelect` if its callable with those operands

PR #812, opened 2025-01-06
NBL_WITH_DOCKER cmake option

YasInvolved:
## Description
Installs docker on the machine if it's not present

## Testing 
Script tested on a clean Win10 VM without docker

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,6 +216,29 @@ if (NBL_WITH_DOCKER)
 			message(FATAL_ERROR "Docker installer exited with non-zero return code")
 		endif()
 	endif()
+
+	macro(wait5sec)
+		execute_process(COMMAND powershell -Command "Start-Sleep -Seconds 5")
+	endmacro()
+	
+	# DDT stands for Docker Daemon Test
+	set(DDT_MAX_ATTEMPTS 10)
+	set(DDT_DOCKER_RESPONDING False)
+	foreach(ATTEMPT RANGE 1 ${DDT_MAX_ATTEMPTS})
+		message(STATUS "Attempt ${ATTEMPT} of ${DDT_MAX_ATTEMPTS}: Checking Docker Endpoint")
+		execute_process(COMMAND ${DOCKER_EXECUTABLE} info RESULT_VARIABLE DOCKER_EXIT_CODE OUTPUT_QUIET)
+
+		if (DOCKER_EXIT_CODE EQUAL 0)
+			set(DDT_DOCKER_RESPONDING True)
+			break()
+		endif()
+
+		wait5sec()
+	endforeach()
+
+	if (NOT DDT_DOCKER_RESPONDING)
+		message(FATAL_ERROR "Docker Daemon is not responding. Please make sure it's running in the background")
+	endif()
 endif()
 
 set(THIRD_PARTY_SOURCE_DIR "${PROJECT_SOURCE_DIR}/3rdparty")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,8 +223,8 @@ if (NBL_WITH_DOCKER)
 	endmacro()
 
 	# helper macro for calling docker, takes args as a string
-	macro(calldocker args)
-		execute_process(COMMAND ${DOCKER_EXECUTABLE} ${args} RESULT_VARIABLE DOCKER_EXIT_CODE OUTPUT_QUIET)
+	macro(calldocker)
+		execute_process(COMMAND ${DOCKER_EXECUTABLE} ${ARGN} RESULT_VARIABLE DOCKER_EXIT_CODE OUTPUT_QUIET)
 	endmacro()
 	
 	# check if the docker daemon is running
@@ -233,7 +233,7 @@ if (NBL_WITH_DOCKER)
 	set(DDT_DOCKER_RESPONDING False)
 	foreach(ATTEMPT RANGE 1 ${DDT_MAX_ATTEMPTS})
 		message(STATUS "Attempt ${ATTEMPT} of ${DDT_MAX_ATTEMPTS}: Checking Docker Endpoint")
-		calldocker("info")
+		calldocker(info)
 
 		if (DOCKER_EXIT_CODE EQUAL 0)
 			set(DDT_DOCKER_RESPONDING True)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -193,6 +193,20 @@ option(NBL_BUILD_BULLET "Enable Bullet Physics building and integration?" OFF)
 option(NBL_BUILD_DOCS "Enable building documentation?" OFF) # No one has doxygen installed, plus we dont know when was the last time we generated working doxy and we'll use SphinX in the future
 option(NBL_ENABLE_PROJECT_JSON_CONFIG_VALIDATION "" ON)
 option(NBL_EMBED_BUILTIN_RESOURCES "Embed built-in resources?" ON)
+option(NBL_WITH_DOCKER "Detects and installs docker if not present in the system" OFF)
+
+if (NBL_WITH_DOCKER)
+	find_program(DOCKER_EXECUTABLE docker)
```
AnastaZIuk, line 199: fail if host is not windows

CMakeLists.txt
```diff
@@ -193,6 +193,20 @@ option(NBL_BUILD_BULLET "Enable Bullet Physics building and integration?" OFF)
 option(NBL_BUILD_DOCS "Enable building documentation?" OFF) # No one has doxygen installed, plus we dont know when was the last time we generated working doxy and we'll use SphinX in the future
 option(NBL_ENABLE_PROJECT_JSON_CONFIG_VALIDATION "" ON)
 option(NBL_EMBED_BUILTIN_RESOURCES "Embed built-in resources?" ON)
+option(NBL_WITH_DOCKER "Detects and installs docker if not present in the system" OFF)
```
AnastaZIuk, line 196: "Enables docker integration, if client is not found Docker Desktop will be installed"

CMakeLists.txt
```diff
@@ -193,6 +193,20 @@ option(NBL_BUILD_BULLET "Enable Bullet Physics building and integration?" OFF)
...
+
+if (NBL_WITH_DOCKER)
+	find_program(DOCKER_EXECUTABLE docker)
+
+	if (DOCKER_EXECUTABLE)
+		message(STATUS "Found docker executable: ${DOCKER_EXECUTABLE}")
+	else()
+		execute_process(COMMAND powershell -Command "& {
+			Invoke-WebRequest -Uri https://desktop.docker.com/win/main/amd64/Docker%20Desktop%20Installer.exe -OutFile DockerDesktopInstaller.exe;
+			Start-Process -FilePath .\\DockerDesktopInstaller.exe -ArgumentList 'install','--quiet','--accept-license','--backend=wsl-2'
```
AnastaZIuk, line 206: don't forget to remove the installer after success

CMakeLists.txt
```diff
@@ -193,6 +193,25 @@ option(NBL_BUILD_BULLET "Enable Bullet Physics building and integration?" OFF)
...
+	if (NOT CMAKE_HOST_WIN32)
+		message(FATAL_ERROR "NBL_WITH_DOCKER works only with Windows machines. Please disable this option")
+	endif()
+
+	find_program(DOCKER_EXECUTABLE docker)
+
+	if (DOCKER_EXECUTABLE)
+		message(STATUS "Found docker executable: ${DOCKER_EXECUTABLE}")
+	else()
+		execute_process(COMMAND powershell -Command "& {
```
AnastaZIuk, line 208: handle errors, terminate the configuration on fail

CMakeLists.txt
```diff
@@ -193,6 +193,30 @@ option(NBL_BUILD_BULLET "Enable Bullet Physics building and integration?" OFF)
...
+			Invoke-WebRequest -Uri https://desktop.docker.com/win/main/amd64/Docker%20Desktop%20Installer.exe -OutFile DockerDesktopInstaller.exe;
+			$exitCode = Start-Process -Wait -FilePath .\\DockerDesktopInstaller.exe -ArgumentList 'install','--quiet','--accept-license','--backend=wsl-2','--always-run-service';
+			Remove-Item -Force DockerDesktopInstaller.exe;
+			exit $exitCode;
+		}" RESULT_VARIABLE DOCKER_INSTALLER_EXIT_CODE)
+
+		if (DOCKER_INSTALLER_EXIT_CODE NOT EQUAL 0)
+			message(FATAL_ERROR "Docker installer exited with non-zero return code")
+		endif()
+	endif()
```
AnastaZIuk, line 218: now you need to check we see the endpoint

so execute process with `docker info` and check exit status in a loop, do as long as we detect 0 code which means we see client, daemon and an endpoint. Do eg 10 iterations with 5 seconds wait interval (`Attempt i/10: Checking Docker endpoint...`) - last iteration is your timeout on which on fail you display stderr (on success also display std out). On fail add "Could not detect docker endpoint, run the daemon!" msg at the end.

AnastaZIuk: another important thing to do is to detect we have `docker_default` network

execute process with `docker network inspect docker_default` and check exit error, if there is no such network let cmake create one with `docker network create -d nat docker_default`, if it could not create the network fail the configuration

YasInvolved: > find_program(DOCKER_EXECUTABLE docker REQUIRED)

we can't make the docker required because it will fail the configuration early

AnastaZIuk: there is a reason the comment is on line 218 - we can

CMakeLists.txt
```diff
@@ -193,6 +193,68 @@ option(NBL_BUILD_BULLET "Enable Bullet Physics building and integration?" OFF)
...
+	if (NOT DDT_DOCKER_RESPONDING)
+		message(FATAL_ERROR "Docker Daemon is not responding. Please make sure it's running in the background")
+	endif()
+
+	set(DNT_NETWORK_NAME "docker_default")
+	message(STATUS "Checking wether \"${DNT_NETWORK_NAME}\" docker network exist.")
+	calldocker(network inspect ${DNT_NETWORK_NAME})
+	if (NOT DOCKER_EXIT_CODE EQUAL 0)
+		message(STATUS "Docker network \"${DNT_NETWORK_NAME}\" doesn't exist. Creating one.'")
+		calldocker(network create -d nat ${DNT_NETWORK_NAME})
```
AnastaZIuk, line 255: you must check exit code and fail configure if non zero

PR #813, opened 2025-01-08
png loader: fix grayscale/alpha images bug

alichraghi:

### DIFF
--- a/src/nbl/asset/interchange/CImageLoaderPNG.cpp
+++ b/src/nbl/asset/interchange/CImageLoaderPNG.cpp
@@ -178,6 +178,8 @@ asset::SAssetBundle CImageLoaderPng::loadAsset(system::IFile* _file, const asset
 	
 	if (ColorType == PNG_COLOR_TYPE_PALETTE)
 		png_set_palette_to_rgb(png_ptr);
+	else if (ColorType == PNG_COLOR_TYPE_GRAY || ColorType == PNG_COLOR_TYPE_GRAY_ALPHA)
+		png_set_gray_to_rgb(png_ptr);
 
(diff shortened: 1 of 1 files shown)


src/nbl/asset/interchange/CImageLoaderPNG.cpp
```diff
@@ -178,6 +178,8 @@ asset::SAssetBundle CImageLoaderPng::loadAsset(system::IFile* _file, const asset
 	
 	if (ColorType == PNG_COLOR_TYPE_PALETTE)
 		png_set_palette_to_rgb(png_ptr);
+	else if (ColorType == PNG_COLOR_TYPE_GRAY || ColorType == PNG_COLOR_TYPE_GRAY_ALPHA)
```
devshgraphicsprogramming, line 181: GRAY without alpha you DONT want to promote to R8G8B8A8_SRGB, because we have the R8_SRGB format

devshgraphicsprogramming: they all do the same because it took till OpenGL 4.6 to introduce an R8_SRGB format, and its still optional.

We have that format (and handle promotion), so I don't want you promoting an 8 bit image to 32bit or 24bit without need.

I want `ColorType == PNG_COLOR_TYPE_GRAY` to load as `R8_SRGB`

devshgraphicsprogramming:
works?

alichraghi:
tested it with pngs you sent + some other from png test suite. not tested on scene yet because no VM available

devshgraphicsprogramming:
> tested it with pngs you sent + some other from png test suite. not tested on scene yet because no VM available

this is `master` branch so the paht tracer is not here

PR #814, opened 2025-01-08
ditt branch: png loader: fix grayscale/alpha images bug

alichraghi:


devshgraphicsprogramming:
cool this is sufficient for this branch (`ditt` doesn't have format promotion on upload)

PR #818, opened 2025-01-17
FFT Fixes

devshgraphicsprogramming:

### DIFF
--- a/include/nbl/builtin/hlsl/fft/common.hlsl
+++ b/include/nbl/builtin/hlsl/fft/common.hlsl
@@ -56,10 +56,47 @@ inline uint64_t getOutputBufferSize(
     uint16_t passIx,
     NBL_CONST_REF_ARG(vector<uint16_t, N>) axisPassOrder = _static_cast<vector<uint16_t, N> >(uint16_t4(0, 1, 2, 3)),
     bool realFFT = false,
+
     bool halfFloats = false
 )
 {
     const vector<uint32_t, N> paddedDimensions = padDimensions<N>(inputDimensions, realFFT, axisPassOrder[0]);
+    vector<bool, N> axesDone = promote<vector<bool, N>, bool>(false);
+    for (uint16_t i = 0; i <= passIx; i++)
+        axesDone[axisPassOrder[i]] = true;
+    const vector<uint32_t, N> passOutputDimension = lerp(inputDimensions, paddedDimensions, axesDone);
+    uint64_t numberOfComplexElements = uint64_t(numChannels);
+    for (uint16_t i = 0; i < N; i++)
+        numberOfComplexElements *= uint64_t(passOutputDimension[i]);
+    return numberOfComplexElements * (halfFloats ? sizeof(complex_t<float16_t>) : sizeof(complex_t<float32_t>));
+}
+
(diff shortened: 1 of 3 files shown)


include/nbl/builtin/hlsl/fft/common.hlsl
```diff
@@ -70,6 +71,42 @@ inline uint64_t getOutputBufferSize(
...
+* @param [in] axisPassOrder Order of the axis in which the FFT is computed in. Default is xyzw.
+* @param [in] realFFT True if the signal is real. False by default.
+* @param [in] halfFloats True if using half-precision floats. False by default.
+*/
+inline uint64_t getOutputBufferSizeConvolution(
+    uint32_t numChannels,
+    NBL_CONST_REF_ARG(vector<uint32_t, N>) inputDimensions,
+    NBL_CONST_REF_ARG(vector<uint32_t, N>) kernelDimensions,
+    uint16_t passIx,
+    NBL_CONST_REF_ARG(vector<uint16_t, N>) axisPassOrder = _static_cast<vector<uint16_t, N> >(uint16_t4(0, 1, 2, 3)),
```
devshgraphicsprogramming, line 93: no renferences needed

devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/795#discussion_r1917440037

First, a pointer is 8 bytes, anything smaller or equal to 8 bytes is always better off being passed by value (unless you explicitly want to observe changes to it).

You may actually get screwed over by `fno-strict-aliasing` (or havintg two references of the same type - where compiler assumes aliasing) performance wise, because compiler is not free to assume that the references don't alias, and as long as you have at least one mutable reference (strict aliasing -> of the same type), it can't eliminate loads of the const references of the same type.

If your function gets inlined (you've used inline keyword, and SPIR-V gets aggressively inlined by default) and you no longer have a true function call, the Load/Store ops get eliminated during the most basic optimization. So it doesn't matter how big your struct is, 

If the compiler suspects aliasing, they won't get eliminated. Arguments passed by value (copied) never alias.

Ok so when you actually have a function call, and the caller is supposedly supposed to take a copy of the arguments by pushing them on the stack and the callee is supposed to pop them.

Most CPU architectures have a fast calling convention allowing the return value to be passed (or more accurately left) by the called function in the register, and the first few arguments that fit be passed (left) by the callee in a few registers to the called function.

X86-64 has a new calling convention compared to x86 stdcall and ccall, basically `fastcall`/`vectorcall` is now default. This means:
- integers get passed in registers
- structs smaller than a 64bit int get bytecasted and passed by registers
- floats get passed in AVX `xmm` registers , also groups of floats (up to SIMD width)

Obviously only the arguments that fit the above and on a first come first served basis get passed in this way.

The only time this sort of optimization doesn't happen is for functions labelled ccall or stdcall, which are cross-DLL extern C API functions (for C libraries), those are not even generated by `NBL_API2`, but by utilites such as SWIG for C++ - other-lang interop.

P.S. Leaf subroutines (functions that don't call other functions / no recursion possible) are prime candidates for inlining, and even if they are not inlined they behave slightly differently w.r.t the usage of the stack.

Fletterio:
Addressed that + leftover comments from last merge

PR #819, opened 2025-01-18
Prefix sum box blur

alichraghi:

### DIFF
--- a/include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
+++ b/include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
@@ -1,0 +1,262 @@
+#include "nbl/builtin/hlsl/cpp_compat.hlsl"
+#include "nbl/builtin/hlsl/glsl_compat/core.hlsl"
+#include "nbl/builtin/hlsl/workgroup/basic.hlsl"
+#include "nbl/builtin/hlsl/workgroup/arithmetic.hlsl"
+#include "nbl/builtin/hlsl/workgroup/scratch_size.hlsl"
+#include "nbl/builtin/hlsl/device_capabilities_traits.hlsl"
+#include "nbl/builtin/hlsl/enums.hlsl"
+
+namespace nbl
+{
+namespace hlsl
+{
+namespace box_blur
+{
+
+template<
+    typename DataAccessor,
+    typename SharedAccessor,
+    typename ScanSharedAccessor,
+    typename Sampler,
+    uint16_t WorkgroupSize,
+    class device_capabilities=void> // TODO: define concepts for the Box1D and apply constraints
+struct Box1D
+{
+    // TODO: Generalize later on when Francesco enforces accessor-concepts in `workgroup` and adds a `SharedMemoryAccessor` concept
+    struct ScanSharedAccessorWrapper
+    {
+        void get(const uint16_t ix, NBL_REF_ARG(float32_t) val)
+        {
+            val = base.template get<float32_t, uint16_t>(ix);
+        }
+
+        void set(const uint16_t ix, const float32_t val)
+        {
+            base.template set<float32_t, uint16_t>(ix, val);
+        }
+
+        void workgroupExecutionAndMemoryBarrier()
+        {
+            base.workgroupExecutionAndMemoryBarrier();
+        }
+
+        ScanSharedAccessor base;
+    };
+
+    void operator()(
+        NBL_REF_ARG(DataAccessor) data,
+        NBL_REF_ARG(SharedAccessor) scratch,
+        NBL_REF_ARG(ScanSharedAccessor) scanScratch,
+        NBL_REF_ARG(Sampler) boxSampler,
+        const uint16_t channel)
+    {
+        const uint16_t end = data.linearSize();
+        const uint16_t localInvocationIndex = workgroup::SubgroupContiguousIndex();
+
+        // prefix sum
+        // note the dynamically uniform loop condition 
+        for (uint16_t baseIx = 0; baseIx < end;)
+        {
+            const uint16_t ix = localInvocationIndex + baseIx;
+            float32_t input = data.template get<float32_t>(channel, ix);
+            // dynamically uniform condition
+            if (baseIx != 0)
+            {
+                // take result of previous prefix sum and add it to first element here
+                if (localInvocationIndex == 0)
+                    input += scratch.template get<float32_t>(baseIx - 1);
+            }
+            // need to copy-in / copy-out the accessor cause no references in HLSL - yay!
+            ScanSharedAccessorWrapper scanScratchWrapper;
+            scanScratchWrapper.base = scanScratch;
+            const float32_t sum = workgroup::inclusive_scan<plus<float32_t>, WorkgroupSize, device_capabilities>::template __call(input, scanScratchWrapper);
+            scanScratch = scanScratchWrapper.base;
+            // loop increment
+            baseIx += WorkgroupSize;
+            // if doing the last prefix sum, we need to barrier to stop aliasing of temporary scratch for `inclusive_scan` and our scanline
+            // TODO: might be worth adding a non-aliased mode as NSight says nr 1 hotspot is barrier waiting in this code
+            if (end + ScanSharedAccessor::Size > SharedAccessor::Size)
+                scratch.workgroupExecutionAndMemoryBarrier();
+            // save prefix sum results
+            if (ix < end)
+                scratch.template set<float32_t>(ix, sum);
+            // previous prefix sum must have finished before we ask for results
+            scratch.workgroupExecutionAndMemoryBarrier();
+        }
+
+        const float32_t last = end - 1;
+        const float32_t normalizationFactor = 1.f / (2.f * radius + 1.f);
+
+        for (float32_t ix = localInvocationIndex; ix < end; ix += WorkgroupSize)
+        {
+            const float32_t result = boxSampler(scratch, ix, radius, borderColor[channel]);
+            data.template set<float32_t>(channel, uint16_t(ix), result * normalizationFactor);
+        }
+    }
+
+    vector<float32_t, DataAccessor::Channels> borderColor;
+    float32_t radius;
+};
+
+template<typename PrefixSumAccessor, typename T>
+struct BoxSampler
+{
+    uint16_t wrapMode;
+    uint16_t linearSize;
+
+    T operator()(NBL_REF_ARG(PrefixSumAccessor) prefixSumAccessor, float32_t ix, float32_t radius, float32_t borderColor)
+    {
+        const float32_t alpha = radius - floor(radius);
+        const float32_t lastIdx = linearSize - 1;
+        const float32_t rightIdx = float32_t(ix) + radius;
+        const float32_t leftIdx = float32_t(ix) - radius;
+        const int32_t rightFlIdx = (int32_t)floor(rightIdx);
+        const int32_t rightClIdx = (int32_t)ceil(rightIdx);
+        const int32_t leftFlIdx = (int32_t)floor(leftIdx);
+        const int32_t leftClIdx = (int32_t)ceil(leftIdx);
+
+        T result = 0;
+        if (rightFlIdx < linearSize)
+        {
+            result += lerp(prefixSumAccessor.template get<T, uint32_t>(rightFlIdx), prefixSumAccessor.template get<T, uint32_t>(rightClIdx), alpha);
+        }
+        else
+        {
+            switch (wrapMode) {
+            case ETC_REPEAT:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T floored = prefixSumAccessor.template get<T, uint32_t>(rightFlIdx % linearSize) + ceil(float32_t(rightFlIdx % lastIdx) / linearSize) * last;
+                const T ceiled = prefixSumAccessor.template get<T, uint32_t>(rightClIdx % linearSize) + ceil(float32_t(rightClIdx % lastIdx) / linearSize) * last;
+                result += lerp(floored, ceiled, alpha);
+                break;
+            }
+            case ETC_CLAMP_TO_BORDER:
+            {
+                result += prefixSumAccessor.template get<T, uint32_t>(lastIdx) + (rightIdx - lastIdx) * borderColor;
+                break;
+            }
+            case ETC_CLAMP_TO_EDGE:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T lastMinusOne = prefixSumAccessor.template get<T, uint32_t>(lastIdx - 1);
+                result += (rightIdx - lastIdx) * (last - lastMinusOne) + last;
+                break;
+            }
+            case ETC_MIRROR:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                T floored, ceiled;
+                int32_t d = rightFlIdx - lastIdx;
+
+                if (d % (2 * linearSize) == linearSize)
+                    floored = ((d + linearSize) / linearSize) * last;
+                else
+                {
+                    const uint32_t period = uint32_t(ceil(float32_t(d) / linearSize));
+                    if ((period & 0x1u) == 1)
+                        floored = period * last + last - prefixSumAccessor.template get<T, uint32_t>(lastIdx - uint32_t(d % linearSize));
+                    else
+                        floored = period * last + prefixSumAccessor.template get<T, uint32_t>((d - 1) % linearSize);
+                }
+                
+                d = rightClIdx - lastIdx;
+                if (d % (2 * linearSize) == linearSize)
+                    ceiled = ((d + linearSize) / linearSize) * last;
+                else
+                {
+                    const uint32_t period = uint32_t(ceil(float32_t(d) / linearSize));
+                    if ((period & 0x1u) == 1)
+                        ceiled = period * last + last - prefixSumAccessor.template get<T, uint32_t>(lastIdx - uint32_t(d % linearSize));
+                    else
+                        ceiled = period * last + prefixSumAccessor.template get<T, uint32_t>((d - 1) % linearSize);
+                }
+
+                result += lerp(floored, ceiled, alpha);
+                break;
+            }
+            case ETC_MIRROR_CLAMP_TO_EDGE:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T first = prefixSumAccessor.template get<T, uint32_t>(0);
+                const T firstPlusOne = prefixSumAccessor.template get<T, uint32_t>(1);
+                result += (rightIdx - lastIdx) * (firstPlusOne - first) + last;
+                break;
+            }
+            }
+        }
+
+        if (leftFlIdx >= 0)
+        {
+            result -= lerp(prefixSumAccessor.template get<T, uint32_t>(leftFlIdx), prefixSumAccessor.template get<T, uint32_t>(leftClIdx), alpha);
+        }
+        else
+        {
+            switch (wrapMode) {
+            case ETC_REPEAT:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T floored = prefixSumAccessor.template get<T, uint32_t>(abs(leftFlIdx) % linearSize) + ceil(T(leftFlIdx) / linearSize) * last;
+                const T ceiled = prefixSumAccessor.template get<T, uint32_t>(abs(leftClIdx) % linearSize) + ceil(float32_t(leftClIdx) / linearSize) * last;
+                result -= lerp(floored, ceiled, alpha);
+                break;
+            }
+            case ETC_CLAMP_TO_BORDER:
+            {
+                result -= prefixSumAccessor.template get<T, uint32_t>(0) + leftIdx * borderColor;
+                break;
+            }
+            case ETC_CLAMP_TO_EDGE:
+            {
+                result -= leftIdx * prefixSumAccessor.template get<T, uint32_t>(0);
+                break;
+            }
+            case ETC_MIRROR:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                T floored, ceiled;
+
+                if (abs(leftFlIdx + 1) % (2 * linearSize) == 0)
+                    floored = -(abs(leftFlIdx + 1) / linearSize) * last;
+                else
+                {
+                    const uint32_t period = uint32_t(ceil(float32_t(abs(leftFlIdx + 1)) / linearSize));
+                    if ((period & 0x1u) == 1)
+                        floored = -(period - 1) * last - prefixSumAccessor.template get<T, uint32_t>((abs(leftFlIdx + 1) - 1) % linearSize);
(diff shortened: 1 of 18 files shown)


src/nbl/asset/IAssetManager.cpp
```diff
@@ -307,9 +307,9 @@ void IAssetManager::insertBuiltinAssets()
 	// samplers
 	{
 		asset::ISampler::SParams params;
-		params.TextureWrapU = asset::ISampler::ETC_REPEAT;
-		params.TextureWrapV = asset::ISampler::ETC_REPEAT;
-		params.TextureWrapW = asset::ISampler::ETC_REPEAT;
+		params.TextureWrapU = asset::ISampler::E_TEXTURE_CLAMP::ETC_REPEAT;
+        params.TextureWrapV = asset::ISampler::E_TEXTURE_CLAMP::ETC_REPEAT;
+        params.TextureWrapW = asset::ISampler::E_TEXTURE_CLAMP::ETC_REPEAT;
```
devshgraphicsprogramming, line 312: indentation got messed up

also at the time you have to type out `asset::ISampler::E_TEXTURE_CLAMP` I'd use a local `using` alias on line 309

devshgraphicsprogramming:
Did you test the FFT example, 24 Colorspace test and ImGUI examples don't break from your changes?

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+
+template<
+    typename DataAccessor,
+    typename SharedAccessor,
+    typename ScanSharedAccessor,
+    typename Sampler,
+    uint16_t WorkgroupSize,
+    class device_capabilities=void> // TODO: define concepts for the Box1D and apply constraints
+struct Box1D
+{
```
devshgraphicsprogramming, line 24: why do we have:
1. a file that contains every struct for different things? (other things use a folder with the same name as the general include)
2. a namespace with a different name to the header

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+    T operator()(NBL_REF_ARG(PrefixSumAccessor) prefixSumAccessor, float32_t ix, float32_t radius, float32_t borderColor)
+    {
+        const float32_t alpha = radius - floor(radius);
+        const float32_t lastIdx = linearSize - 1;
+        const float32_t rightIdx = float32_t(ix) + radius;
+        const float32_t leftIdx = float32_t(ix) - radius;
+        const int32_t rightFlIdx = (int32_t)floor(rightIdx);
+        const int32_t rightClIdx = (int32_t)ceil(rightIdx);
+        const int32_t leftFlIdx = (int32_t)floor(leftIdx);
+        const int32_t leftClIdx = (int32_t)ceil(leftIdx);
```
devshgraphicsprogramming, line 116: what happens if you get your `ClIdx` as `FlIdx+1` instead? I think `alpha` should be 0 anyway whenever ceil==floor and never use the invalid value of `ClIdx` anyway

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+
+        T result = 0;
+        if (rightFlIdx < linearSize)
+        {
+            result += lerp(prefixSumAccessor.template get<T, uint32_t>(rightFlIdx), prefixSumAccessor.template get<T, uint32_t>(rightClIdx), alpha);
```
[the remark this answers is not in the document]
devshgraphicsprogramming: so shouldn't you be checking `rightClIdx` instead?

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+        }
+        else
+        {
+            switch (wrapMode) {
+            case ETC_REPEAT:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T floored = prefixSumAccessor.template get<T, uint32_t>(rightFlIdx % linearSize) + ceil(float32_t(rightFlIdx % lastIdx) / linearSize) * last;
+                const T ceiled = prefixSumAccessor.template get<T, uint32_t>(rightClIdx % linearSize) + ceil(float32_t(rightClIdx % lastIdx) / linearSize) * last;
```
devshgraphicsprogramming, line 130: something modulo `N-1` and then divided by `N` and then ceiled will ALWAYS return 1, unless the modulo was 0

this calculation must be wrong, also the overlap counting part is common to both inputs to `lerp` so can be hoisted out of it.

its only when floored is N-1 and ceiled is 0 after modulo, you have a wraparound, and need to add a `last` to the `ceiled` to compensate

and independently of that you add `(rightFlrIdx/N)*last` to the overall interpolated thing

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+        }
+        else
+        {
+            switch (wrapMode) {
+            case ETC_REPEAT:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T floored = prefixSumAccessor.template get<T, uint32_t>(abs(leftFlIdx) % linearSize) + ceil(T(leftFlIdx) / linearSize) * last;
+                const T ceiled = prefixSumAccessor.template get<T, uint32_t>(abs(leftClIdx) % linearSize) + ceil(float32_t(leftClIdx) / linearSize) * last;
+                result -= lerp(floored, ceiled, alpha);
```
devshgraphicsprogramming, line 201: symmetrically broken to https://github.com/Devsh-Graphics-Programming/Nabla/pull/819/files?diff=unified&w=0#r1923818897

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+                break;
+            }
+            case ETC_MIRROR_CLAMP_TO_EDGE:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T first = prefixSumAccessor.template get<T, uint32_t>(0);
+                const T firstPlusOne = prefixSumAccessor.template get<T, uint32_t>(1);
+                result += (rightIdx - lastIdx) * (firstPlusOne - first) + last;
+                break;
+            }
```
devshgraphicsprogramming, line 185: I'm gonna accept on blind faith these work, don't have time to review, please double check or get someone else to play devil's advocate

devshgraphicsprogramming: @Przemog1 could be appropriate person to check the math here

I've spotted one wrong thing already, since your prefix sum is inclusive, `first` is already the value you want, not `firstPlusOne-first` for clamp to edge. Also you don't handle the first mirror correctly.

Btw now that this is separate from the shader, you can simple-test this box sampler on the CPU side with a simple array of `{1,2,3,4,5}` and its prefix sum

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T floored = prefixSumAccessor.template get<T, uint32_t>(abs(leftFlIdx) % linearSize) + ceil(T(leftFlIdx) / linearSize) * last;
+                const T ceiled = prefixSumAccessor.template get<T, uint32_t>(abs(leftClIdx) % linearSize) + ceil(float32_t(leftClIdx) / linearSize) * last;
+                result -= lerp(floored, ceiled, alpha);
+                break;
+            }
+            case ETC_CLAMP_TO_BORDER:
+            {
+                result -= prefixSumAccessor.template get<T, uint32_t>(0) + leftIdx * borderColor;
```
devshgraphicsprogramming, line 206: I think your code requires an inclusive prefix sum? should be documented.

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+                break;
+            }
+            case ETC_CLAMP_TO_EDGE:
+            {
+                result -= leftIdx * prefixSumAccessor.template get<T, uint32_t>(0);
```
devshgraphicsprogramming, line 211: nitpick, `leftIdx=-1` maps to 0 in an inclusive prefix sum because `leftIdx=0` maps to first value, you need a +1 here

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+                result -= lerp(floored, ceiled, alpha);
+                break;
+            }
+            case ETC_MIRROR_CLAMP_TO_EDGE:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T lastMinusOne = prefixSumAccessor.template get<T, uint32_t>(lastIdx - 1);
+                result -= leftIdx * (last - lastMinusOne);
+                break;
+            }
```
devshgraphicsprogramming, line 252: might be suffering to similar things as https://github.com/Devsh-Graphics-Programming/Nabla/pull/819/files?diff=unified&w=0#r1923852734

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+namespace box_blur
+{
+
+template<
+    typename DataAccessor,
+    typename SharedAccessor,
+    typename ScanSharedAccessor,
+    typename Sampler,
+    uint16_t WorkgroupSize,
+    class device_capabilities=void> // TODO: define concepts for the Box1D and apply constraints
```
devshgraphicsprogramming, line 22: okay, so in C++ you'd have the option of a `template<class> class` parameter for the `Sampler` but its of limited usefulness anyway.

Instead what you can require is that `Sampler::prefix_sum_accessor_t` exists, and then you don't need to ask for the type of `SharedAccessor` in the template at all.

Even better yet, you could add a `prefixSumAccessor` member to `Sampler` and then you don't need to pass `scratch` to `Box1D::operator()` at all as an argument (they agree).

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        ScanSharedAccessor base;
+    };
+
+    void operator()(
+        NBL_REF_ARG(DataAccessor) data,
+        NBL_REF_ARG(SharedAccessor) scratch,
+        NBL_REF_ARG(ScanSharedAccessor) scanScratch,
+        NBL_REF_ARG(Sampler) boxSampler,
```
devshgraphicsprogramming, line 50: 1. Why are we calling our struct `Box1D` should be `Blur1D` because we can use any sampler, not just box!
2. same reason for calling out argument `boxSampler`, could be just `_sampler`

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+                scratch.workgroupExecutionAndMemoryBarrier();
+            // save prefix sum results
+            if (ix < end)
+                scratch.template set<float32_t>(ix, sum);
+            // previous prefix sum must have finished before we ask for results
+            scratch.workgroupExecutionAndMemoryBarrier();
+        }
+
+        const float32_t last = end - 1;
+        const float32_t normalizationFactor = 1.f / (2.f * radius + 1.f);
```
devshgraphicsprogramming, line 88: appplying the normalization should be part of the `BoxSampler`'s `operator()`

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
```diff
@@ -0,0 +1,262 @@
...
+            if (end + ScanSharedAccessor::Size > SharedAccessor::Size)
+                scratch.workgroupExecutionAndMemoryBarrier();
+            // save prefix sum results
+            if (ix < end)
+                scratch.template set<float32_t>(ix, sum);
+            // previous prefix sum must have finished before we ask for results
+            scratch.workgroupExecutionAndMemoryBarrier();
+        }
+
+        const float32_t last = end - 1;
```
devshgraphicsprogramming, line 87: I'd try splitting the Box1D (or what I'd call `Blur1D`) into two separate functors:
1. multi-wg-wide prefix sum
2. the SAT sampling

don't do it but leave a comment about it

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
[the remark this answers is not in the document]
alichraghi: right. typo

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
alichraghi: yes `floor`/`ceil`/`alpha` thing is for correct sub-pixels calculation so when radius is a natural number calculating both floor and ceil is pointless because lerp would result the same value since `floor=ceil` but branching for calculating one of them when `alpha=0` isn't any better.

alichraghi: > you may want to assert that linearSize>1

HLSL/SPIR-V has asserts?

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
```diff
@@ -0,0 +1,183 @@
...
+        }
+        else
+        {
+            switch (wrapMode) {
+            case ETC_REPEAT:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T floored = prefixSumAccessor.template get<T, uint32_t>(rightFlIdx % linearSize) + last;
+                const T ceiled = prefixSumAccessor.template get<T, uint32_t>(rightClIdx % linearSize) + last;
+                result += lerp(floored, ceiled, alpha);
```
devshgraphicsprogramming, line 51: this is surely broken, see https://github.com/Devsh-Graphics-Programming/Nabla/pull/819#discussion_r1923818897

your last isn't even multiplied by how many multiples of linearSize you're outside the range 🤦

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
```diff
@@ -0,0 +1,183 @@
...
+        }
+        else
+        {
+            switch (wrapMode) {
+            case ETC_REPEAT:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T floored = prefixSumAccessor.template get<T, uint32_t>((lastIdx + leftFlIdx) % linearSize) + floor(T(leftFlIdx) / linearSize) * last;
+                const T ceiled = prefixSumAccessor.template get<T, uint32_t>((lastIdx + leftClIdx) % linearSize) + floor(T(leftClIdx) / linearSize) * last;
+                result -= lerp(floored, ceiled, alpha);
```
devshgraphicsprogramming, line 120: its interesting how you've allowed this to compute `last` multiple properly but not the right side case

doesn't the non symmetry of your cases not disturb you immediately?

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
```diff
@@ -0,0 +1,183 @@
...
+                }
+
+                result -= lerp(floored, ceiled, alpha);
+                break;
+            }
+            case ETC_MIRROR_CLAMP_TO_EDGE:
+            {
+                const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
+                const T lastMinusOne = prefixSumAccessor.template get<T, uint32_t>(lastIdx - 1);
+                result -= (1 - abs(leftIdx)) * (last - lastMinusOne);
```
devshgraphicsprogramming, line 169: you know left index is already negative, so make it `1+leftIdx` no need for the `abs`

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
devshgraphicsprogramming: where's the documeentation

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
```diff
@@ -0,0 +1,183 @@
...
+        const float32_t alpha = frac(radius);
+        const float32_t rightIdx = float32_t(ix) + radius;
+        const float32_t leftIdx = float32_t(ix) - radius - 1;
+        const int32_t lastIdx = linearSize - 1;
+        const int32_t rightFlIdx = (int32_t)floor(rightIdx);
+        const int32_t rightClIdx = (int32_t)ceil(rightIdx);
+        const int32_t leftFlIdx = (int32_t)floor(leftIdx);
+        const int32_t leftClIdx = (int32_t)ceil(leftIdx);
+
+        assert(linearSize > 1);
```
devshgraphicsprogramming, line 36: may wanna also assert that `radius>numeric_limits::min` or something

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
```diff
@@ -0,0 +1,183 @@
...
+
+template<typename PrefixSumAccessor, typename T>
+struct BoxSampler
+{
+    using prefix_sum_accessor_t = PrefixSumAccessor;
+
+    PrefixSumAccessor prefixSumAccessor;
+    uint16_t wrapMode;
+    uint16_t linearSize;
+    T normalizationFactor;
```
devshgraphicsprogramming, line 23: its always tied and derived from radius, don't even make it a parameter, work it out inside `operator()~`

include/nbl/builtin/hlsl/prefix_sum_blur/blur.hlsl
```diff
@@ -0,0 +1,107 @@
...
+            ScanSharedAccessorWrapper scanScratchWrapper;
+            scanScratchWrapper.base = scanScratch;
+            const float32_t sum = workgroup::inclusive_scan<plus<float32_t>, WorkgroupSize, device_capabilities>::template __call(input, scanScratchWrapper);
+            scanScratch = scanScratchWrapper.base;
+            // loop increment
+            baseIx += WorkgroupSize;
+            // if doing the last prefix sum, we need to barrier to stop aliasing of temporary scratch for `inclusive_scan` and our scanline
+            // TODO: might be worth adding a non-aliased mode as NSight says nr 1 hotspot is barrier waiting in this code
+            if (end + ScanSharedAccessor::Size > Sampler::prefix_sum_accessor_t::Size)
+                _sampler.prefixSumAccessor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming, line 80: do it, also why were you asking for a `scanScratch` when  it was assumed to alias `_sampler.prefixSumAccessor` ?

in this light should be just as easy as deleting `ScanSharedAccessorWrapper` struct and this if-statement then just documenting that this function requires `scanScract` and `_sampler.prefixSumAccessor`  not to alias.

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
devshgraphicsprogramming: you want to specifically assert that `raidus` is bigger than a denormalized float (hence the `bit_cast<float32_t>(numeric_limits<float32_t>::min)`

denormalized numbers can flush to 0, which means expression like `1+2*radius` can equal  0 which in turn will make your normalization factor inf

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
```diff
@@ -127,36 +138,36 @@ struct BoxSampler
...
             }
             case ETC_MIRROR:
             {
                 const T last = prefixSumAccessor.template get<T, uint32_t>(lastIdx);
                 T floored, ceiled;
 
-                if (abs(leftFlIdx + 1) % (2 * linearSize) == 0)
-                    floored = -(abs(leftFlIdx + 1) / linearSize) * last;
+                if (abs(leftFlIdx) % (2 * linearSize) == 0)
+                    floored = -(abs(leftFlIdx) / linearSize) * last;
```
devshgraphicsprogramming, line 150: you already know `leftFlIdx` is negative, can use `-leftFlIdx` instead of the abs

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
```diff
@@ -127,36 +138,36 @@ struct BoxSampler
...
                 else
                 {
-                    const uint32_t period = uint32_t(ceil(float32_t(abs(leftClIdx + 1)) / linearSize));
+                    const uint32_t period = uint32_t(ceil(float32_t(abs(leftClIdx)) / linearSize));
                     if ((period & 0x1u) == 1)
-                        ceiled = -(period - 1) * last - prefixSumAccessor.template get<T, uint32_t>((abs(leftClIdx + 1) - 1) % linearSize);
+                        ceiled = -(period - 1) * last - prefixSumAccessor.template get<T, uint32_t>((abs(leftClIdx) - 1) % linearSize);
                     else
-                        ceiled = -(period - 1) * last - (last - prefixSumAccessor.template get<T, uint32_t>((leftClIdx + 1) % linearSize - 1));
+                        ceiled = -(period - 1) * last - (last - prefixSumAccessor.template get<T, uint32_t>(leftClIdx % linearSize - 1));
```
devshgraphicsprogramming, line 170: all your `abs(leftClIdx)` and `abs(leftFlIdx)` could be `-leftClIfx` and `leftFlIdx` respectively because `leftIdx<0` and the highest you can ceil to is 0

include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl
```diff
@@ -166,13 +177,13 @@ struct BoxSampler
...
                 const T lastMinusOne = prefixSumAccessor.template get<T, uint32_t>(lastIdx - 1);
-                result -= (1 - abs(leftIdx)) * (last - lastMinusOne);
+                result -= (leftIdx + 1) * (last - lastMinusOne);
                 break;
             }
             }
         }
 
-        return result * normalizationFactor;
+        return result / (2 * radius + 1);
```
devshgraphicsprogramming, line 186: use floating point literals, use typed literals in general

include/nbl/builtin/hlsl/prefix_sum_blur.hlsl
alichraghi: https://github.com/Devsh-Graphics-Programming/Nabla/blob/b65dc7f9173b5e4a34fd5e783db99754718a17d8/include/nbl/builtin/hlsl/prefix_sum_blur/box_sampler.hlsl#L15

devshgraphicsprogramming:
still works?, if yes, then I'll merge

alichraghi:
yes it works

PR #823, opened 2025-01-29
Adds a resizable LRUCache

Fletterio:
https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/181


include/nbl/core/containers/GrowableDoublyLinkedList.h
```diff
@@ -0,0 +1,82 @@
...
+		void* newReservedSpace = _NBL_ALIGNED_MALLOC(firstPart + newCapacity * sizeof(node_t), alignof(node_t));
+
+		// Malloc failed, not possible to grow
+		if (!newReservedSpace)
+			return false;
+
+		node_t* newArray = reinterpret_cast<node_t*>(reinterpret_cast<uint8_t*>(newReservedSpace) + firstPart);
+
+		// Copy memory over to new buffer, then free old one
+		memcpy(reinterpret_cast<void*>(newArray), reinterpret_cast<void*>(this->m_array), this->m_cap * sizeof(node_t));
```
Erfan-Ahmadi, line 64: It's uncommon practice to reinterpret cast to a `void*` in a function (memcpy) getting void*.
 just use the raw pointer, better readability

include/nbl/core/containers/GrowableDoublyLinkedList.h
```diff
@@ -0,0 +1,82 @@
...
+		_NBL_ALIGNED_FREE(this->m_reservedSpace);
+
+		// Finally, create new address allocator from old one
+		this->m_addressAllocator = std::unique_ptr<address_allocator_t>(new address_allocator_t(newCapacity, std::move(*(this->m_addressAllocator)), newReservedSpace));
```
Erfan-Ahmadi, line 68: ah, is this the special PoolAddressAllocator constructor with copyState?

Fletterio: yep, that's right

include/nbl/core/containers/LRUCache.h
```diff
@@ -16,206 +18,399 @@ namespace core
...
-		unordered_set<uint32_t,WrapHash,WrapEquals> m_shortcut_map;
-};
+	/**
+	* @brief Resizes the cache by extending its capacity so it can hold more elements. Returns a bool indicating if capacity was indeed increased.
+	*
+	* @param [in] newCapacity New number of elements to hold. MUST be greater than current cache capacity.
+	*/
+	inline bool grow(uint32_t newCapacity)
+	{
+		m_shortcut_map.reserve(newCapacity);
```
Erfan-Ahmadi, line 393: let's check for newCapacity being more than previous one here as well.
we don't know what these containers will be or how they'll behave in the future

include/nbl/core/alloc/AddressAllocatorBase.h
```diff
@@ -123,14 +121,24 @@ namespace core
...
             }
 
+            void invalidate()
+            {
+                reservedSpace = nullptr;
+                addressOffset = invalid_address;
+                alignOffset = invalid_address;
+                maxRequestableAlignment = invalid_address;
+                combinedOffset = invalid_address;
+            }
```
devshgraphicsprogramming, line 140: for style points you can make `AddressAllocatorBase()` call `invalidate()`

include/nbl/core/containers/DoublyLinkedList.h
```diff
@@ -0,0 +1,330 @@
...
+
+	inline void common_detach(node_t* node)
+	{
+		if (node->next != invalid_iterator)
+			get(node->next)->prev = node->prev;
+		if (node->prev != invalid_iterator)
+			get(node->prev)->next = node->next;
+	}
+
+	std::unique_ptr<address_allocator_t> m_addressAllocator;
```
devshgraphicsprogramming, line 315: newing this on the heap is horribly meh, just saying.

devshgraphicsprogramming: the good place to use a `unique_ptr` would have been the arrays below

include/nbl/core/containers/DoublyLinkedList.h
```diff
@@ -0,0 +1,330 @@
...
+			}
+			m_dispose_f(back->data);
+		}
+	}
+
+	inline void common_delete(uint32_t address)
+	{
+		if (m_dispose_f)
+			m_dispose_f(get(address)->data);
+		get(address)->~node_t();
```
devshgraphicsprogramming, line 303: style hint, use `std::uninitialized_construct` and destruct instead of placement new and explicit destructor invoke, see the commandbuffer/pool stuff

include/nbl/core/containers/DoublyLinkedList.h
```diff
@@ -0,0 +1,330 @@
...
+	* @param [in] newCapacity New number of elements to hold. MUST be greater than current list capacity.
+	*/
+	inline bool grow(uint32_t newCapacity)
+	{
+		// Must at least make list grow
+		if (newCapacity <= m_cap)
+			return false;
+		// Same as code found in ContiguousMemoryLinkedListBase to create aligned space
+		const auto firstPart = core::alignUp(address_allocator_t::reserved_size(1u, newCapacity, 1u), alignof(node_t));
+		void* newReservedSpace = _NBL_ALIGNED_MALLOC(firstPart + newCapacity * sizeof(node_t), alignof(node_t));
```
devshgraphicsprogramming, line 191: like every good container take an `allocator` template param, and use it instead of malloc

include/nbl/core/containers/DoublyLinkedList.h
devshgraphicsprogramming: or use a true C++11 style `allocator_traits` and its construct/destruct methods https://github.com/Devsh-Graphics-Programming/Nabla/pull/823/files#r1979485663

include/nbl/core/containers/DoublyLinkedList.h
Fletterio: I don't get where I'd use a `unique_ptr` here

devshgraphicsprogramming: the allocator reserved space (state)

include/nbl/core/containers/DoublyLinkedList.h
```diff
@@ -279,20 +296,18 @@ class DoublyLinkedList
...
-			while (begin != back)
-			{
-				m_dispose_f(begin->data);
-				begin = get(begin->next);
-			}
-			m_dispose_f(back->data);
+			node_t* currentNode = get(currentAddress);
+			uint32_t nextAddress = currentNode->next;
+			if (m_dispose_f) m_dispose_f(currentNode->data);
+			currentNode->~node_t();
```
devshgraphicsprogramming, line 309: https://en.cppreference.com/w/cpp/memory/allocator_traits/destroy

include/nbl/core/containers/DoublyLinkedList.h
```diff
@@ -0,0 +1,348 @@
...
+	DoublyLinkedList(const uint32_t capacity, disposal_func_t&& dispose_f = disposal_func_t(), const allocator_t& _allocator = allocator_t()) 
+		: m_dispose_f(std::move(dispose_f)), m_allocator(_allocator)
+	{
+		// Have to consider allocating enough space for list AND state of the address allocator
+		const auto firstPart = core::alignUp(address_allocator_t::reserved_size(1u, capacity, 1u), alignof(node_t));
+		// Allocator can only allocate in terms of nodes
+		const size_t firstPartNodes = (firstPart + sizeof(node_t) - 1) / sizeof(node_t);
+		m_currentAllocationSize = firstPartNodes + capacity;
+		m_reservedSpace = reinterpret_cast<void*>(allocator_traits_t::allocate(m_allocator, m_currentAllocationSize));
+		m_array = reinterpret_cast<node_t*>(reinterpret_cast<uint8_t*>(m_reservedSpace) + firstPart);
```
Erfan-Ahmadi, line 230: Ok, from my understanding this first part is for the address allocator, and the second part is for the actual nodes.
and I see the allocations now need to be aligned by `alignof(node_t)` and the allocation doesn't get the size in bytes anymore, it gets count of "nodes"

I might be micromanaging a little but this little piece of code caught my eye and caused me to think more about what it exactly does.

I have 2 thing here I want to say:

1. Isn't `(firstPart + sizeof(node_t) - 1) / sizeof(node_t);` implicitly another `alignUp` and a division but this time with sizeof(node) instead of alignof(node)?

So may I suggest just do the alignUp with `sizeof(node)` instead of `alignof(node)` and then do the simple division by sizeof(node)

2. I suggest renaming the variables in such a way that it's understood the first part is not actually for the nodes, but it's for the address allocator.

Fletterio: makes sense

Fletterio: issue is that alignUp takes PoT arguments for alignment

Fletterio: nvm, that doesn't matter I can do the hand-alignment

Erfan-Ahmadi: > issue is that alignUp takes PoT arguments for alignment
you're right, I had forgotten about that.
> nvm, that doesn't matter I can do the hand-alignment
exactly, still something redundant is being done here, no need for alignUp since you're aligning by sizeof instead.

Erfan-Ahmadi: let's rename `firstPart` to `addressAllocatorMemSize`  and `firstPartNodes` to `addressAllocatorMemAsNodes`.

PR #829, opened 2025-02-03
video: adjust device_limits/features to the updated profiles

alichraghi:


src/nbl/video/device_capabilities/device_features.json
```diff
@@ -8,12 +8,12 @@
         {
             "type": "bool",
             "name": "robustBufferAccess",
-            "value": false,
+            "value": true,
```
devshgraphicsprogramming, line 11: look, there's no "expose" (which means it defaults to optional), the value NEEDS to be false becvause thats the default/initializer value (you don't want to enable `robustBufferAccess` unless the user explicitly asks for it

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -49,7 +49,7 @@
                 {
                     "type": "bool",
                     "name": "tessellationShader",
-                    "value": false
+                    "value": true
```
devshgraphicsprogramming, line 52: look, there's no "expose" (which means it defaults to optional), the value NEEDS to be false becvause thats the default/initializer value (you don't want to enable `robustBufferAccess` unless the user explicitly asks for it

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -203,7 +203,7 @@
                 {
                     "type": "bool",
                     "name": "vertexPipelineStoresAndAtomics",
-                    "value": false,
+                    "value": true,
                     "expose": "MOVE_TO_LIMIT",
```
devshgraphicsprogramming, line 207: lets not mandate it yet, the androids may not have that feature for a looong time

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -212,16 +212,16 @@
                 {
                     "type": "bool",
                     "name": "fragmentStoresAndAtomics",
-                    "value": false,
+                    "value": true,
                     "expose": "MOVE_TO_LIMIT",
                     "comment": [
-                        "ROADMAP 2022 no supporton iOS GPUs"
+                        "ROADMAP 2024 no supporton iOS GPUs"
```
devshgraphicsprogramming, line 218: delete the part of the comment about iOS GPUs

src/nbl/video/device_capabilities/device_features.json
devshgraphicsprogramming: also value needs to be `false`

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -544,10 +544,10 @@
         {
             "type": "bool",
             "name": "samplerYcbcrConversion",
-            "value": false,
+            "value": true,
             "expose": "DISABLE",
```
devshgraphicsprogramming, line 548: look at the `expose` its disabled, means we don't even care about it, so leave the value as was

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -212,16 +212,16 @@
...
-                        "ROADMAP 2022 no supporton iOS GPUs"
+                        "ROADMAP 2024 no supporton iOS GPUs"
                     ]
                 },
                 {
                     "type": "bool",
                     "name": "shaderTessellationAndGeometryPointSize",
-                    "value": false,
+                    "value": true,
                     "expose": "MOVE_TO_LIMIT",
```
devshgraphicsprogramming, line 225: this is a limit, leave the default value as is (the default values are "worst case" basically)

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -698,9 +698,9 @@
                 {
                     "type": "bool",
                     "name": "shaderUniformBufferArrayNonUniformIndexing",
-                    "value": false,
+                    "value": true,
                     "expose": "MOVE_TO_LIMIT",
-                    "comment": ["[EXPOSE AS A LIMIT] ROADMAP 2022 mandates but poor device support"]
+                    "comment": ["[EXPOSE AS A LIMIT] ROADMAP 2024 mandates but poor device support"]
```
devshgraphicsprogramming, line 703: you need to update the comment to match and do `expose` `REQUIRE`

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -829,14 +829,14 @@
...
+                "Roadmap 2024 requires support for these we always enable and they're unlikely to harm performance",
                 "or VK_EXT_scalar_block_layout"
             ]
         },
         {
             "type": "bool",
             "name": "imagelessFramebuffer",
-            "value": false,
+            "value": true,
             "expose": "DISABLE",
```
devshgraphicsprogramming, line 840: look its disabled, we decided not to expose them

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -950,14 +950,14 @@
...
                     "expose": "MOVE_TO_LIMIT",
                     "comment": ["ALIAS: VK_EXT_shader_viewport_index_layer"]
                 },
                 {
                     "type": "bool",
                     "name": "shaderOutputLayer",
-                    "value": false,
+                    "value": true,
                     "expose": "MOVE_TO_LIMIT",
                     "comment": ["ALIAS: VK_EXT_shader_viewport_index_layer"]
```
devshgraphicsprogramming, line 962: leave as is, I have a feeling Metal and Android may have issues with this

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -980,7 +980,7 @@
         {
             "type": "bool",
             "name": "robustImageAccess",
-            "value": false,
+            "value": true,
```
devshgraphicsprogramming, line 983: do not ever init the robustness features to default on, even if everyone has them

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -999,7 +999,7 @@
                 {
                     "type": "bool",
                     "name": "inlineUniformBlock",
-                    "value": false,
+                    "value": true,
                     "expose": "DISABLE",
```
devshgraphicsprogramming, line 1003: look, its disabled

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -1106,7 +1106,7 @@
         {
             "type": "bool",
             "name": "dynamicRendering",
-            "value": false,
+            "value": true,
             "expose": "DISABLE",
```
devshgraphicsprogramming, line 1110: you need to REQUIRE it, but don't do it yet because we're not implementing dynamic rendering yet

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -135,7 +135,7 @@
                 {
                     "type": "uint32_t",
                     "name": "maxBoundDescriptorSets",
-                    "value": 4,
+                    "value": 8,
```
devshgraphicsprogramming, line 138: I saw Metal only having 7 and roadmap only mandating 7, also Nabla is hardcoded to 4 so leave as 4

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -915,69 +915,67 @@
...
                     "name": "shaderRoundingModeRTZFloat32",
-                    "value": false
+                    "value": true
                 },
                 {
                     "type": "bool",
                     "name": "shaderRoundingModeRTZFloat64",
-                    "value": false
+                    "value": true
                 }
```
devshgraphicsprogramming, line 979: coordinate with @nipunG314 if these need to be expose DISABLE (like the above) or expose REQUIRE

devshgraphicsprogramming: I'd REQUIRE, it also means you need to always enable the shader float control vk1.2 feature (and fail if its not present) during device list and init

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -992,45 +990,45 @@
                 {
                     "type": "bool",
                     "name": "shaderUniformBufferArrayNonUniformIndexingNative",
-                    "value": false
+                    "value": true
```
devshgraphicsprogramming, line 993: native indexing of UBOs is a bit stinky, don't require

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -992,45 +990,45 @@
...
                     "name": "shaderInputAttachmentArrayNonUniformIndexingNative",
-                    "value": false,
+                    "value": true,
                     "comment": ["promotion candidate"]
                 },
                 {
                     "type": "bool",
                     "name": "robustBufferAccessUpdateAfterBind",
-                    "value": false
+                    "value": true
```
devshgraphicsprogramming, line 1021: if you want to require, then you need the `expose` tag

Also the `robustBufferAccessUpdateAfterBind`  will need special treatment, one should only require/check for it in CVulkanPhysicalDevice.cpp if the robust buffer access feature is enabled

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -992,45 +990,45 @@
...
                 },
                 {
                     "type": "bool",
                     "name": "quadDivergentImplicitLod",
-                    "value": false
+                    "value": true
```
devshgraphicsprogramming, line 1026: I dont like this feature, don't require

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -1120,12 +1118,12 @@
...
                     "type": "bool",
                     "name": "independentResolveNone",
-                    "value": false
+                    "value": true
                 },
                 {
                     "type": "bool",
                     "name": "independentResolve",
-                    "value": false
+                    "value": true
```
devshgraphicsprogramming, line 1126: expose and require tags needed

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -1176,18 +1174,18 @@
...
                     "type": "uint8_t",
                     "name": "maxSubgroupSize",
-                    "value": "4"
+                    "value": "32"
                 },
                 {
                     "type": "uint32_t",
                     "name": "maxComputeWorkgroupSubgroups",
-                    "value": "16"
+                    "value": "32"
```
devshgraphicsprogramming, line 1188: this you must leave as was before

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -1245,152 +1243,152 @@
...
                     "type": "bool",
                     "name": "integerDotProductAccumulatingSaturating64BitSignedAccelerated",
-                    "value": false
+                    "value": true
                 },
                 {
                     "type": "bool",
                     "name": "integerDotProductAccumulatingSaturating64BitMixedSignednessAccelerated",
-                    "value": false
+                    "value": true
```
devshgraphicsprogramming, line 1391: if you were to promote them, then you'd need `expose` tag, I'd suggest to leave as is right now, its difficult to roll back requirements

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -1575,7 +1573,7 @@
         {
             "type": "uint32_t",
             "name": "maxPushDescriptors",
-            "value": 0,
+            "value": 32,
             "expose": "DISABLE",
```
devshgraphicsprogramming, line 1577: look its disabled

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -1442,7 +1440,7 @@
         {
             "type": "uint32_t",
             "name": "minImportedHostPointerAlignment",
-            "value": "0x1u << 31",
+            "value": "0x1u << 14u",
             "compare": "REVERSE",
             "comment": [
```
devshgraphicsprogramming, line 1445: do we require external memory host extension?

Also we probably shouldn't in-case we do WebGPU in the future.

so thats the reason the alignment is max uint32_t Power of two, so you can't fail limit intersection when extension/feature not enabled

Also its a good worst case value to report

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -1914,7 +1912,7 @@
         {
             "type": "uint32_t",
             "name": "lineSubPixelPrecisionBits",
-            "value": 0,
+            "value": 4,
```
devshgraphicsprogramming, line 1915: do we require line drawing feature though?

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -2078,13 +2076,13 @@
...
-                    "value": false,
+                    "value": true,
                     "expose": "DISABLE"
                 },
                 {
                     "type": "bool",
                     "name": "graphicsPipelineLibraryIndependentInterpolationDecoration",
-                    "value": false,
+                    "value": true,
                     "expose": "DISABLE"
```
devshgraphicsprogramming, line 2086: we don't require the graphics pipeline feature (yet) and we've disabled, no need to set it to true

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -2141,29 +2139,29 @@
...
                     "type": "bool",
                     "name": "workgroupMemoryExplicitLayout8BitAccess",
-                    "value": false
+                    "value": true
                 },
                 {
                     "type": "bool",
                     "name": "workgroupMemoryExplicitLayout16BitAccess",
-                    "value": false
+                    "value": true
```
devshgraphicsprogramming, line 2157: are you sure about this, requiring 8  and 16 bit explicit layout access would mean we require 8 and 16 bit regular access.

First off, HLSL doesn't let us use 8 bit types, so lets not require anything 8 bit.

I'm not sure all GPUs we care about can do 16 bit access to smem.

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -2173,7 +2171,7 @@
         {
            "type": "core::bitflag<asset::IShader::E_SHADER_STAGE>",
             "name": "cooperativeMatrixSupportedStages",
-            "value": "asset::IShader::E_SHADER_STAGE::ESS_UNKNOWN",
+            "value": "asset::IShader::E_SHADER_STAGE::ESS_COMPUTE | asset::IShader::E_SHADER_STAGE::ESS_ALL",
```
devshgraphicsprogramming, line 2174: **we DONT require that extension!**

src/nbl/video/device_capabilities/device_limits.json
devshgraphicsprogramming: **But wait `VK_KHR_workgroup_memory_explicit_layout` IS NOT in NablaCore.json**

So why are you requiring limits from it!?

src/nbl/video/device_capabilities/device_limits.json
alichraghi: only when renderdoc is used

src/nbl/video/device_capabilities/device_limits.json
alichraghi: not required
```cpp
        if (isExtensionSupported(VK_EXT_LINE_RASTERIZATION_EXTENSION_NAME))
            properties.limits.lineSubPixelPrecisionBits = lineRasterizationProperties.lineSubPixelPrecisionBits;
```

src/nbl/video/device_capabilities/device_limits.json
alichraghi: reverted as apple and android doesn't support these

src/nbl/video/device_capabilities/device_limits.json
devshgraphicsprogramming: so leave as 0

src/nbl/video/device_capabilities/device_limits.json
devshgraphicsprogramming: ok then don't change

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -304,7 +304,9 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
         //VkPhysicalDeviceGraphicsPipelineLibraryPropertiesEXT    graphicsPipelineLibraryProperties = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GRAPHICS_PIPELINE_LIBRARY_PROPERTIES_EXT };
         VkPhysicalDeviceFragmentDensityMap2PropertiesEXT        fragmentDensityMap2Properties = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FRAGMENT_DENSITY_MAP_2_PROPERTIES_EXT };
         VkPhysicalDeviceRayTracingPipelinePropertiesKHR         rayTracingPipelineProperties = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_RAY_TRACING_PIPELINE_PROPERTIES_KHR };
+#if 0 // TODO
         VkPhysicalDeviceCooperativeMatrixPropertiesKHR          cooperativeMatrixProperties = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_COOPERATIVE_MATRIX_PROPERTIES_KHR };
+#endif
```
devshgraphicsprogramming, line 309: why did you undo  !? the Vulkan headers now have the extension, what you did to remove the TOODO was ok

alichraghi: > **we DONT require that extension!**

i thought you want me to revert here? or you meant to add REQUIRE tag?

devshgraphicsprogramming: that was a comment for your JSON

see how all the other non required optional features are handled (AFAIK the code is already done for that) the `if 0` was only because we were using old Vulkan headers that didn't have the published enums and structure definitions yet

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -904,80 +904,91 @@
...
-                    "value": false
+                    "value": true,
+                    "expose": "REQUIRE"
                 },
                 {
                     "type": "bool",
                     "name": "shaderRoundingModeRTZFloat64",
-                    "value": false
+                    "value": true,
+                    "expose": "REQUIRE"
```
devshgraphicsprogramming, line 991: do not require limits that are contingent on features (in this case `shaderFloat16` and `shaderFloat32`)

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -2163,20 +2176,34 @@
...
+                {
+                    "type": "bool",
+                    "name": "cooperativeMatrix",
+                    "value": false
+                },
+                {
+                    "type": "bool",
+                    "name": "cooperativeMatrixRobustBufferAccess",
+                    "value": false
+                },
```
devshgraphicsprogramming, line 2201: but you already declared these as features, they can't also be in limits!

src/nbl/video/device_capabilities/device_features.json
```diff
@@ -4138,13 +4138,20 @@
...
+            "entries": [
+                {
+                    "type": "bool",
+                    "name": "cooperativeMatrix",
+                    "value": false
+                },
+                {
+                    "type": "bool",
+                    "name": "cooperativeMatrixRobustBufferAccess",
+                    "value": false
```
devshgraphicsprogramming, line 4153: you want to declare the `cooperativeMatrix` feature as `MOVE_TO_LIMIT` (add the `expose` tag)

while `cooperativeMatrixRobustBufferAccess` remains a feature like it is right now

src/nbl/video/device_capabilities/device_limits.json
devshgraphicsprogramming: only the supported stages shall be a limit

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -1128,10 +1134,9 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
             features.rasterizationOrderDepthAttachmentAccess = rasterizationOrderAttachmentAccessFeatures.rasterizationOrderDepthAttachmentAccess;
             features.rasterizationOrderStencilAttachmentAccess = rasterizationOrderAttachmentAccessFeatures.rasterizationOrderStencilAttachmentAccess;
         }
-#if 0
+
         if (isExtensionSupported(VK_KHR_COOPERATIVE_MATRIX_EXTENSION_NAME))
             features.cooperativeMatrixRobustBufferAccess = cooperativeMatrixFeatures.cooperativeMatrixRobustBufferAccess;
-#endif
```
devshgraphicsprogramming, line 1140: special request from yours truly, can you check for `maintenance5` extension or `graphics_pipline_library` extension and return nullptr if neither is available.

If they're available get both their feature and property structs into pNext:
https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_maintenance5.html
https://registry.khronos.org/vulkan/specs/latest/man/html/VK_EXT_graphics_pipeline_library.html#_new_structures

But later when you create Logical Device, enable maintenance5 if available, and only enable graphics pipeline libraries as backup if available.

src/nbl/video/device_capabilities/device_limits.json
alichraghi: These are actually not supported on all profiles yet (even on my GPU). the profile merger script probably got confused because they exist on both `VkPhysicalDeviceDescriptorIndexingProperties` and `VkPhysicalDeviceVulkan12Properties`

src/nbl/video/CVulkanPhysicalDevice.cpp
alichraghi: hmm `VkPhysicalDeviceMaintenance5Properties` is declared in `vk1.4`. are we updating vulkan-headers now?

src/nbl/video/device_capabilities/device_limits.json
devshgraphicsprogramming: ouch

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -398,19 +394,32 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
+        if (!vulkan12Properties.shaderRoundingModeRTEFloat32)
+            return nullptr;
+        if (!vulkan12Properties.shaderRoundingModeRTEFloat64)
+            return nullptr;
+        if (!vulkan12Properties.shaderRoundingModeRTZFloat16)
+            return nullptr;
+        if (!vulkan12Properties.shaderRoundingModeRTZFloat32)
+            return nullptr;
+        if (!vulkan12Properties.shaderRoundingModeRTZFloat64)
+            return nullptr;
```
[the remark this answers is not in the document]
devshgraphicsprogramming: you want to move the checks of the limits into where the features are already known, yes

src/nbl/video/CVulkanPhysicalDevice.cpp
devshgraphicsprogramming: you can update Vulkan headers, but for anything that is core in vulkan 1.4 you don't use any vulkan14 struct you need to use the extension struct (we're still running 1.3 and all 1.4 features we get as extensions)

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -903,81 +903,77 @@
...
                     "type": "bool",
                     "name": "shaderRoundingModeRTZFloat32",
-                    "value": false
+                    "value": true
                 },
                 {
                     "type": "bool",
                     "name": "shaderRoundingModeRTZFloat64",
-                    "value": false
+                    "value": true
```
devshgraphicsprogramming, line 976: this is a mess and needs to get fixed

devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/829#discussion_r1944463467

devshgraphicsprogramming: you can only DISABLE and `value=true` the 32bit ones, rest you need to DISABLE and `value=false`

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -394,31 +397,25 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
         properties.limits.shaderRoundingModeRTZFloat64 = vulkan12Properties.shaderRoundingModeRTZFloat64;
-            
+
+
         // descriptor indexing
         properties.limits.maxUpdateAfterBindDescriptorsInAllPools                 = vulkan12Properties.maxUpdateAfterBindDescriptorsInAllPools;
-        properties.limits.shaderUniformBufferArrayNonUniformIndexingNative        = vulkan12Properties.shaderUniformBufferArrayNonUniformIndexingNative;
-        properties.limits.shaderSampledImageArrayNonUniformIndexingNative         = vulkan12Properties.shaderSampledImageArrayNonUniformIndexingNative;
-        properties.limits.shaderStorageBufferArrayNonUniformIndexingNative        = vulkan12Properties.shaderStorageBufferArrayNonUniformIndexingNative;
-        properties.limits.shaderStorageImageArrayNonUniformIndexingNative         = vulkan12Properties.shaderStorageImageArrayNonUniformIndexingNative;
```
devshgraphicsprogramming, line 420: where do you check they're not false?

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -895,6 +897,43 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
+                return nullptr;
+            if (!properties.limits.shaderDenormPreserveFloat64)
+                return nullptr;
+            if (!properties.limits.shaderDenormFlushToZeroFloat64)
+                return nullptr;
+            if (!properties.limits.shaderRoundingModeRTEFloat64)
+                return nullptr;
+            if (!properties.limits.shaderRoundingModeRTZFloat64)
+                return nullptr;
+        }
```
devshgraphicsprogramming, line 936: if you don't expose them, then you need to move up the Vk Structs and check from the VkStructs, not our own limits struct

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -1577,10 +1611,19 @@ core::smart_refctd_ptr<ILogicalDevice> CVulkanPhysicalDevice::createLogicalDevic
...
         REQUIRE_EXTENSION_IF(enabledFeatures.cooperativeMatrixRobustBufferAccess,VK_KHR_COOPERATIVE_MATRIX_EXTENSION_NAME,&cooperativeMatrixFeatures);
-#endif
+        
+        VkPhysicalDeviceMaintenance5FeaturesKHR maintenance5Features = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MAINTENANCE_5_FEATURES_KHR, nullptr };
+        VkPhysicalDeviceGraphicsPipelineLibraryFeaturesEXT graphicsPipelineLibraryFeatures = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_GRAPHICS_PIPELINE_LIBRARY_FEATURES_EXT, nullptr };
+
+        // Enable maintenance5 and graphics pipeline libraries as backup if available
+        if (!enableExtensionIfAvailable(VK_KHR_MAINTENANCE_5_EXTENSION_NAME, &maintenance5Features))
+        {
+            enableExtensionIfAvailable(VK_KHR_PIPELINE_LIBRARY_EXTENSION_NAME, nullptr);
```
devshgraphicsprogramming, line 1624: why do you need to enable the pipeline library extension? does the Graphics Pipeline Lib Extension require it?

btw I think its best to actually pass a struct pointer, not sure how well passing nullptr works on the pNext chains :uneasy_smile:

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -1823,10 +1866,13 @@ core::smart_refctd_ptr<ILogicalDevice> CVulkanPhysicalDevice::createLogicalDevic
...
-        cooperativeMatrixFeatures.cooperativeMatrix = true;
-        cooperativeMatrixFeatures.cooperativeMatrixRobustBufferAccess = enabledFeatures.cooperativeMatrixRobustBufferAccess;
-#endif
+
+        if (limits.cooperativeMatrix) {
+            cooperativeMatrixFeatures.cooperativeMatrixRobustBufferAccess = enabledFeatures.cooperativeMatrixRobustBufferAccess;
+        }
+
+        maintenance5Features.maintenance5 = true;
+        graphicsPipelineLibraryFeatures.graphicsPipelineLibrary = maintenance5Features.maintenance5 == 0;
```
devshgraphicsprogramming, line 1875: erm but you just set maintenance5 to true, maybe you should incorporate the feature struct bool setting into the ig statements at https://github.com/Devsh-Graphics-Programming/Nabla/pull/829/files#r1947966841

src/nbl/video/CVulkanPhysicalDevice.cpp
alichraghi: so query vk features in `Get physical device's limits/properties` scope?

src/nbl/video/CVulkanPhysicalDevice.cpp
alichraghi: i saw it being added in [Vulkan-Samples](https://github.com/KhronosGroup/Vulkan-Samples/blob/40e3b440ef0f1c5e5abcf7c85f9de332e1f13503/samples/extensions/graphics_pipeline_library/graphics_pipeline_library.cpp#L58). also passing `nullptr` here is fine. it won't be added to pNext chains but the extension will be inserted.

devshgraphicsprogramming: > i saw it being added in [Vulkan-Samples](https://github.com/KhronosGroup/Vulkan-Samples/blob/40e3b440ef0f1c5e5abcf7c85f9de332e1f13503/samples/extensions/graphics_pipeline_library/graphics_pipeline_library.cpp#L58).

image

its better to quote straight from the spec, instead of relying on someone else not messing up 

P.S. unrelated note, this is why I dislike devs trusting ChatGPT and asking it to distil the WebGPU spec for them

devshgraphicsprogramming: Ok in this case I think its best to enable the pipeline_library extension independently if its there and pass the struct, because at some point someone will want to add that extensions and add it a second time 500 lines away and we'll have a debugging nightmare why the feature struct isn't in pNext (we have a sentinel that prevents double add)

src/nbl/video/CVulkanPhysicalDevice.cpp
devshgraphicsprogramming: for a limit thats required pending a feature, don't do `Vulkan Limits -> Nabla Limit` to then check `Nabla Feature Predicate && not Nabla  Limit`  do `Nabla Feature && ! Vulkan Limit`

because keeping it as is, would require you to expose a limit to Nabla which has no business being exposed (i.e. we can assume the limit is fullfilled if we have the feature).

Example in plain english, if you have 16 bit floats, you can acount on having float controls for float16 type ops. If you don't then whether you would have had controls is irrelevant because you can't use 16bit floats.

src/nbl/video/CVulkanPhysicalDevice.cpp
alichraghi: `VK_KHR_pipeline_library` has no property/feature structs btw
image

src/nbl/video/CVulkanPhysicalDevice.cpp
alichraghi: by "Nabla feature", you mean features defined in `device_features.json`? `shaderFloat16/64` are not defined there but in `device_limit.json` so they are a limit

src/nbl/video/CVulkanPhysicalDevice.cpp
devshgraphicsprogramming: `maintenance5Features.maintenance5 == 0` will never kick in because you're setting it to true on the line immediately above

src/nbl/video/CVulkanPhysicalDevice.cpp
devshgraphicsprogramming: sorry I forgot `shaderFloat` and `shaderInt` are nabla limits, but still you'd want to check
```cpp
if (properties.limits.shaderFloat16 && !vkShaderFloatControlProperties.shaderRoundingModeRTZFloat16)
   return nullptr;
```

so that `properties.limits.shaderRoundingModeRTZFloat16` need not exist

src/nbl/video/CVulkanPhysicalDevice.cpp
alichraghi: removing now as they aren't available on my gpu

devshgraphicsprogramming: so the code you removed, should be back? right...?

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -992,29 +988,34 @@
...
+                    "expose": "REQUIRE",
                     "comment": ["promotion candidate"]
                 },
                 {
                     "type": "bool",
                     "name": "shaderInputAttachmentArrayNonUniformIndexingNative",
-                    "value": false,
+                    "value": true,
+                    "expose": "REQUIRE",
                     "comment": ["promotion candidate"]
```
devshgraphicsprogramming, line 1019: if your GPU doesn't support those, then you need to revert this part of the JSON, right?

src/nbl/video/CVulkanPhysicalDevice.cpp
alichraghi: absolutely

src/nbl/video/device_capabilities/device_limits.json
alichraghi: yes done

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -397,25 +399,51 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
+                return nullptr;
+            if (!vulkan12Properties.shaderDenormFlushToZeroFloat16)
+                return nullptr;
+            if (!vulkan12Properties.shaderRoundingModeRTEFloat16)
+                return nullptr;
+            if (!vulkan12Properties.shaderRoundingModeRTZFloat16)
+                return nullptr;
+        }
+
+        if (properties.limits.shaderFloat64) {
```
devshgraphicsprogramming, line 427: `properties.limits.shaderFloat` are not initialized until line 925

Please be diligent in doing your work and follow my advice https://github.com/Devsh-Graphics-Programming/Nabla/pull/829#discussion_r1945494266

devshgraphicsprogramming: Either you initialize your `properties.limits` first or you need to use `deviceProperties2.`

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -1618,10 +1609,16 @@ core::smart_refctd_ptr<ILogicalDevice> CVulkanPhysicalDevice::createLogicalDevic
...
-        if (!enableExtensionIfAvailable(VK_KHR_MAINTENANCE_5_EXTENSION_NAME, &maintenance5Features))
+        if (enableExtensionIfAvailable(VK_KHR_MAINTENANCE_5_EXTENSION_NAME, &maintenance5Features))
         {
-            enableExtensionIfAvailable(VK_KHR_PIPELINE_LIBRARY_EXTENSION_NAME, nullptr);
+            maintenance5Features.maintenance5 = true;
+        }
+        else
+        {
+            graphicsPipelineLibraryFeatures.graphicsPipelineLibrary = true;
             enableExtensionIfAvailable(VK_EXT_GRAPHICS_PIPELINE_LIBRARY_EXTENSION_NAME, &graphicsPipelineLibraryFeatures);
```
devshgraphicsprogramming, line 1622: you need to `return nullptr` if this extension enable of last resort fails

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -903,77 +903,92 @@
...
-                    "value": true
+                    "value": false,
+                    "expose": "DISABLE"
                 },
                 {
                     "type": "bool",
                     "name": "shaderSignedZeroInfNanPreserveFloat32",
-                    "value": true
+                    "value": true,
+                    "expose": "DISABLE"
```
devshgraphicsprogramming, line 913: the `value=true` (the 32bit ones) should still be REQUIRE

src/nbl/video/device_capabilities/device_limits.json
devshgraphicsprogramming: sorry if its `true` then it needs to be `REQUIRE` my bad

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -1176,7 +1189,7 @@
                 {
                     "type": "uint8_t",
                     "name": "minSubgroupSize",
-                    "value": "64",
+                    "value": "32",
```
devshgraphicsprogramming, line 1192: no there's Android HW that has uber wide subgroups, actually set it to 128

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -1677,11 +1717,11 @@ core::smart_refctd_ptr<ILogicalDevice> CVulkanPhysicalDevice::createLogicalDevic
         vulkan12Features.shaderInputAttachmentArrayDynamicIndexing = limits.shaderInputAttachmentArrayDynamicIndexing;
         vulkan12Features.shaderUniformTexelBufferArrayDynamicIndexing = true; // implied by `descriptorIndexing`
         vulkan12Features.shaderStorageTexelBufferArrayDynamicIndexing = true; // implied by `descriptorIndexing`
-        vulkan12Features.shaderUniformBufferArrayNonUniformIndexing = limits.shaderUniformBufferArrayNonUniformIndexing;
+        vulkan12Features.shaderUniformBufferArrayNonUniformIndexing = true; // implied by `descriptorIndexing`
         vulkan12Features.shaderSampledImageArrayNonUniformIndexing = true; // implied by `descriptorIndexing`
         vulkan12Features.shaderStorageBufferArrayNonUniformIndexing = true; // implied by `descriptorIndexing`
         vulkan12Features.shaderStorageImageArrayNonUniformIndexing = true; // require
-        vulkan12Features.shaderInputAttachmentArrayNonUniformIndexing = limits.shaderInputAttachmentArrayNonUniformIndexing;
+        vulkan12Features.shaderInputAttachmentArrayNonUniformIndexing = true; // implied by `descriptorIndexing`
```
devshgraphicsprogramming, line 1724: you messed around with the JSONs regarding the descriptor indexing then reverted the changes, I think you're not done undoing the C++ device report and init

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -903,11 +933,9 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
             return nullptr;
         // not uniform at all
-        properties.limits.shaderUniformBufferArrayNonUniformIndexing = vulkan12Features.shaderUniformBufferArrayNonUniformIndexing;
-        if (!vulkan12Features.shaderSampledImageArrayNonUniformIndexing || !vulkan12Features.shaderStorageBufferArrayNonUniformIndexing || !vulkan12Features.shaderStorageImageArrayNonUniformIndexing)
+        if (!vulkan12Features.shaderUniformBufferArrayNonUniformIndexing || !vulkan12Features.shaderSampledImageArrayNonUniformIndexing || !vulkan12Features.shaderStorageBufferArrayNonUniformIndexing || !vulkan12Features.shaderStorageImageArrayNonUniformIndexing)
             return nullptr;
-        properties.limits.shaderInputAttachmentArrayNonUniformIndexing = vulkan12Features.shaderInputAttachmentArrayNonUniformIndexing;
-        if (!vulkan12Features.shaderUniformTexelBufferArrayNonUniformIndexing || !vulkan12Features.shaderStorageTexelBufferArrayNonUniformIndexing)
+        if (!vulkan12Features.shaderInputAttachmentArrayNonUniformIndexing || !vulkan12Features.shaderUniformTexelBufferArrayNonUniformIndexing || !vulkan12Features.shaderStorageTexelBufferArrayNonUniformIndexing)
```
devshgraphicsprogramming, line 938: you messed around with the JSONs regarding the descriptor indexing then reverted the changes, I think you're not done undoing the C++ device report and init?

src/nbl/video/CVulkanPhysicalDevice.cpp
alichraghi: `vkShaderFloatControlProperties` isn't a thing and we can't access `vulkan12Properties` after limits are initialized if that's what you meant. because it's in a different scope

alichraghi: we also can't *initialize `properties.limits` first* because `shaderFloat16/64` are features not properties

devshgraphicsprogramming: > `vkShaderFloatControlProperties` isn't a thing and we can't access `vulkan12Properties` after limits are initialized if that's what you meant. because it's in a different scope

then hoist the variable above the scope...

devshgraphicsprogramming: next time suggest solutions while listing problems

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -1618,8 +1619,11 @@ core::smart_refctd_ptr<ILogicalDevice> CVulkanPhysicalDevice::createLogicalDevic
         }
         else
         {
+            if (!enableExtensionIfAvailable(VK_EXT_GRAPHICS_PIPELINE_LIBRARY_EXTENSION_NAME, &graphicsPipelineLibraryFeatures)) {
+                return nullptr;
+            }
+            assert(pipelineLibraryEnabled);
```
devshgraphicsprogramming, line 1625: dont forget to leave comments next to asserts

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -903,10 +934,8 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
         if (!vulkan12Features.shaderUniformTexelBufferArrayDynamicIndexing || !vulkan12Features.shaderStorageTexelBufferArrayDynamicIndexing)
             return nullptr;
         // not uniform at all
-        properties.limits.shaderUniformBufferArrayNonUniformIndexing = vulkan12Features.shaderUniformBufferArrayNonUniformIndexing;
         if (!vulkan12Features.shaderSampledImageArrayNonUniformIndexing || !vulkan12Features.shaderStorageBufferArrayNonUniformIndexing || !vulkan12Features.shaderStorageImageArrayNonUniformIndexing)
             return nullptr;
-        properties.limits.shaderInputAttachmentArrayNonUniformIndexing = vulkan12Features.shaderInputAttachmentArrayNonUniformIndexing;
```
devshgraphicsprogramming, line 909: related to https://github.com/Devsh-Graphics-Programming/Nabla/pull/829#discussion_r1976697029

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -895,6 +900,32 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
+            if (!vulkan12Properties.shaderSignedZeroInfNanPreserveFloat64)
+                return nullptr;
+            if (!vulkan12Properties.shaderDenormPreserveFloat64)
+                return nullptr;
+            if (!vulkan12Properties.shaderDenormFlushToZeroFloat64)
+                return nullptr;
+            if (!vulkan12Properties.shaderRoundingModeRTEFloat64)
+                return nullptr;
+            if (!vulkan12Properties.shaderRoundingModeRTZFloat64)
+                return nullptr;
```
devshgraphicsprogramming, line 927: Denorm and RTZ are not all true on the RTX 3060
image

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -903,81 +903,92 @@
...
-                    "value": false
+                    "value": true,
+                    "expose": "REQUIRE"
                 },
                 {
                     "type": "bool",
                     "name": "shaderRoundingModeRTZFloat64",
-                    "value": false
+                    "value": false,
+                    "expose": "DISABLE"
```
devshgraphicsprogramming, line 991: something went wrong here
image

PR #817, opened 2025-01-16
New tgmath

Przemog1:

PR #832, opened 2025-02-07
Rm core matrix

Przemog1:


devshgraphicsprogramming:
Depends on #817 so will review after that is merged

PR #835, opened 2025-02-12
nbl::system::isDebuggerAttached

YasInvolved:


include/nbl/system/CSystemLinux.h
```diff
@@ -16,6 +16,9 @@ class CSystemLinux final : public ISystemPOSIX
 
 		NBL_API2 SystemInfo getSystemInfo() const override;
 };
+
+bool isDebuggerAttached();
```
devshgraphicsprogramming, line 20: NBL_API2

src/nbl/system/CSystemLinux.cpp
```diff
@@ -26,4 +29,37 @@ ISystem::SystemInfo CSystemLinux::getSystemInfo() const
...
+   if (status == -1)
+      return false;
+
+   const size_t numRead = read(status, static_cast<void*>(buf), sizeof(buf) - 1);
+   close(status);
+
+   buf[numRead] = '\0';
+   const auto offset = strstr(buf, tracerPidStr);
+   if (not offset)
+      return false;
```
devshgraphicsprogramming, line 48: there's probably some more errors to handle here

include/nbl/system/CSystemLinux.h
devshgraphicsprogramming: Also why is this a random global function? We don't do those.

add it as `static` method to `ISystem` and implement in `ISystem.cpp`

devshgraphicsprogramming:
@AnastaZIuk it looks good, probably mergable

PR #838, opened 2025-02-21
added -fconstexpr-ops-limit for gcc

YasInvolved:


YasInvolved:
Will be tested on debian as soon as my toolchain completes setup

PR #839, opened 2025-02-21
IShader, no more conversions, stage is now part of spec info.

devshgraphicsprogramming:
## Description

Remove unnecessary complexity of ICPUShader and IGPUShader.

## Testing 

All examples using shaders need to be tested.

## TODO list:

- [x] require maintenance5 or graphics_pipeline_library
- [x] pass SPIR-V directly to pipeline creation without creating a `VkShaderModule` using `maintenance5`/`graphics_pipeline_library` `pNext` new API
- [ ] set debug object name on created graphics pipeline libraries
- [x] use SPIR-V Tools & Opt to create a copy of the SPIR-V, remove entry points not used by spec info, then run DCE and capability stripping pass


include/nbl/asset/IPipeline.h
```diff
@@ -223,7 +223,7 @@ class IPipelineBase
 
 			const IShader* shader = nullptr;
 			// A name of the function where the entry point of an shader executable begins. It's often "main" function.
-			std::string_view entryPoint = {};
+			std::string entryPoint = {};
```
devshgraphicsprogramming: I want this struct to remain lightweird an "allocator less" so keep it as a `string_view`

note that the spec constant map is a pointer.

Maybe we need the `SShaderSpecInfo` to be templated/behave differently for IGPU and ICPU where the ICPU version doesn't point at the spec_constant_map_t and entrypoint but those are mutable things instead.

include/nbl/video/utilities/CAssetConverter.h
```diff
@@ -440,6 +440,7 @@ class CAssetConverter : public core::IReferenceCounted
 				{
 					public:
 						virtual const patch_t<asset::ICPUSampler>* operator()(const lookup_t<asset::ICPUSampler>&) const = 0;
+						virtual const patch_t<asset::IShader>* operator()(const lookup_t<asset::IShader>&) const = 0;
```
devshgraphicsprogramming, line 443: you don't have a patch, so don't make the method pure or virtual at all, see the methods below with comment
```cpp
// certain items are not patchable, so there's no `patch_t` with non zero size
```

include/nbl/asset/metadata/CHLSLMetadata.h
```diff
@@ -0,0 +1,30 @@
...
+
+class CHLSLMetadata final : public IAssetMetadata
+{
+    public:
+        explicit CHLSLMetadata(hlsl::ShaderStage shaderStage): shaderStage(shaderStage) {}
+
+        _NBL_STATIC_INLINE_CONSTEXPR const char* LoaderName = "CHLSLLoader";
+        const char* getLoaderName() const override { return LoaderName; }
+        
+        hlsl::ShaderStage shaderStage;
```
devshgraphicsprogramming, line 24: there's only one metadata per asset bundle, in a bundle they may be multiple assets of the same type, so this needs to ne an array (smart refctd dynamic or similar) matching in length with the bundle

src/nbl/video/utilities/CAssetConverter.cpp
```diff
@@ -851,10 +852,13 @@ class HashVisit : public CAssetConverter::CHashCache::hash_impl_base
 						default:
 							break;
 					}
-					for (const auto& specConstant : *arg0.entries)
+					if (arg0.entries)
 					{
-						hasher << specConstant.first;
-						hasher.update(specConstant.second.data,specConstant.second.size);
+					  for (const auto& specConstant : *arg0.entries)
```
devshgraphicsprogramming, line 857: nice catch

include/nbl/asset/utils/ISPIRVDebloater.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include "nbl/system/ILogger.h"
+
+namespace nbl::asset
+{
+
+class ISPIRVDebloater final : public core::IReferenceCounted
+{
+    public:
+        ISPIRVDebloater();
```
devshgraphicsprogramming, line 16: maybe call it EntryPointRemover or Trimmer

include/nbl/asset/utils/ISPIRVDebloater.h
```diff
@@ -0,0 +1,40 @@
...
+
+        struct Result
+        {
+            bool isAllEntryPointsFound;
```
devshgraphicsprogramming, line 20: slight grammar nitpick, multiple is either `are` or `were`

Also another nitpick, make sure to order your member fields from largest to smallest (best packing with alignment padding). Right now you'll get same size either way but its a good habit to get into

include/nbl/asset/utils/ISPIRVDebloater.h
```diff
@@ -0,0 +1,40 @@
...
+        struct EntryPoint
+        {
+          std::string_view name;
+          hlsl::ShaderStage shaderStage;
+
+          bool operator==(const EntryPoint& rhs) const = default;
+        };
+
+        Result debloat(const ICPUBuffer* spirvBuffer, std::span<const EntryPoint> entryPoints, system::logger_opt_ptr logger) const;
+
```
devshgraphicsprogramming, line 33: add an overload which returns `core::smart_refctd_ptr<const ICPUBuffer>` such that the `spirvBuffer` is returned if the `Result::spirv` is null but `wereAllEntryPointsFound==true`

This will probably be the most ergonomic and preferred way to call it

include/nbl/asset/utils/ISPIRVDebloater.h
```diff
@@ -0,0 +1,40 @@
...
+        struct EntryPoint
+        {
+          std::string_view name;
+          hlsl::ShaderStage shaderStage;
+
+          bool operator==(const EntryPoint& rhs) const = default;
+        };
+
+        Result debloat(const ICPUBuffer* spirvBuffer, std::span<const EntryPoint> entryPoints, system::logger_opt_ptr logger) const;
```
devshgraphicsprogramming, line 32: when you use the `opt` logger types, it makes sense to default them to `nullptr` so they're actually optional when you call the function

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "nbl/system/ILogger.h"
+
+
+using namespace nbl::asset;
+
+ISPIRVDebloater::ISPIRVDebloater()
+{
+    constexpr auto optimizationPasses = std::array{
+        ISPIRVOptimizer::EOP_ELIM_DEAD_FUNCTIONS,
+        ISPIRVOptimizer::EOP_ELIM_DEAD_VARIABLES,
```
devshgraphicsprogramming, line 15: you may need EOP_REMOVE_UNUSED_INTERFACE_VARIABLE_PASS and others because dead variables may only remove globals, descriptors, push constants and groupshared (read the source/docs)

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+        ISPIRVOptimizer::EOP_ELIM_DEAD_CONSTANTS,
```
devshgraphicsprogramming, line 16: does it also remove spec constants?

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+        // TODO: remove unused type. no dedicated flag for type. maybe need agressive dce to remove type.
```
devshgraphicsprogramming, line 17: DeadMember optimization may remove the members from structs, and then you have empty structs so maybe empty `OpStruct` will get removed?

Then the only things left behind would be `OpType` for SPIR-V types, I wonder what pass can remove those...

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+            case  spv::ExecutionModelMeshEXT: return hlsl::ESS_MESH;
+            case  spv::ExecutionModelRayGenerationKHR : return hlsl::ESS_RAYGEN;
+            case  spv::ExecutionModelAnyHitKHR : return hlsl::ESS_ANY_HIT;
+            case  spv::ExecutionModelClosestHitKHR : return hlsl::ESS_CLOSEST_HIT;
+            case  spv::ExecutionModelMissKHR : return hlsl::ESS_MISS;
+            case  spv::ExecutionModelIntersectionKHR : return hlsl::ESS_INTERSECTION;
+            case  spv::ExecutionModelCallableKHR : return hlsl::ESS_CALLABLE;
+            default:
+            {
+                assert(false);
```
devshgraphicsprogramming, line 48: NEVER assert to handle a regular user error, we dont want to crash in debug mode because someone misused the engine

Use the logger instead.

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+                return hlsl::ESS_UNKNOWN;
+            }
+        }
+      };
+    static constexpr auto HEADER_SIZE = 5;
+
+    core::vector<bool> foundEntryPoints(entryPoints.size(), false);
```
devshgraphicsprogramming, line 55: use an `core::set` and remove from it as you find (will create less O(f(entryPointCount)) work for you)

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+    std::vector<uint32_t> minimizedSpirv;
+    bool needDebloat = false;
+
+    auto offset = HEADER_SIZE;
```
devshgraphicsprogramming, line 59: please validate that  the header is present and SPIR-V version is at least 1.4

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+        const auto opcode = instruction & 0x0ffffu;
+
+        offset += length;
+
+        if (opcode == spv::OpEntryPoint)
+        {
+            const auto curExecutionModel = static_cast<spv::ExecutionModel>(spirv[curOffset + 1]);
+
+            // TODO: check whether this reinterpret_cast is UB
+            const auto curEntryPointName = std::string_view(reinterpret_cast<const char*>(spirv + curOffset + 3)); // entryPoint name is the third parameter
```
devshgraphicsprogramming, line 75: no its not UB, but will the string be null terminated as per SPIR-V ?

Also you need to protect against someone shoving garbage into the debloater, so please protect against non-null terminated strings. You need to actually find the null terminator with a limit of looking up to `COUNT-curOffset-4` away.

also `while` checks that `offset<COUNT` but it doesn't check `offset+3<COUNT`

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+        const auto instruction = spirv[curOffset];
+
+        const auto length = instruction >> 16;
+        const auto opcode = instruction & 0x0ffffu;
+
+        offset += length;
+
+        if (opcode == spv::OpEntryPoint)
+        {
+            const auto curExecutionModel = static_cast<spv::ExecutionModel>(spirv[curOffset + 1]);
```
devshgraphicsprogramming, line 72: no protection against reading past the end of buffer

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+                    minimizedSpirv.reserve(spirvDwordCount);
+                    minimizedSpirv.insert(minimizedSpirv.end(), spirv, spirv + curOffset);
+                    needDebloat = true;
+                }
+                continue;
+            }
+        }
+        
+        if (!needDebloat) continue;
+        minimizedSpirv.insert(minimizedSpirv.end(), spirv + curOffset, spirv + offset);
```
devshgraphicsprogramming, line 98: you're not checking that `offset` is still within the buffer

src/nbl/asset/utils/ISPIRVDebloater.cpp
devshgraphicsprogramming: then your `isAllEntryPointsFound` check just becomes `entryPointsToFind.empty()`

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+        ISPIRVOptimizer::EOP_ELIM_DEAD_CONSTANTS,
+        // TODO: remove unused type. no dedicated flag for type. maybe need agressive dce to remove type.
+        ISPIRVOptimizer::EOP_TRIM_CAPABILITIES,
+    };
+    m_optimizer = core::make_smart_refctd_ptr<ISPIRVOptimizer>(std::span(optimizationPasses));
+}
+
+ISPIRVDebloater::Result ISPIRVDebloater::debloat(const  ICPUBuffer* spirvBuffer, std::span<const EntryPoint> entryPoints, system::logger_opt_ptr logger) const
+{
+    const auto* spirv = static_cast<const uint32_t*>(spirvBuffer->getPointer());
```
devshgraphicsprogramming, line 25: basic validation missing:
- non null buffer of sufficient size to store the header and as many entry point opcodes as you have `entryPoints`
- the entry point list is not empty

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    const auto isAllEntryPointsFound = std::all_of(foundEntryPoints.begin(), foundEntryPoints.end(), std::identity{});
+
+    if (!isAllEntryPointsFound || !needDebloat) {
+        return Result {
+            .isAllEntryPointsFound = isAllEntryPointsFound,
+            .spirv = nullptr,
+        };
+    }
```
devshgraphicsprogramming, line 108: handle the no-debloat-needed success case separately from failure due to missing entry points

include/nbl/asset/utils/ISPIRVDebloater.h
```diff
@@ -0,0 +1,40 @@
...
+
+    private:
+        core::smart_refctd_ptr<ISPIRVOptimizer> m_optimizer;
```
devshgraphicsprogramming, line 35: one CANNOT have globals or any other inherent hidden state in a class if you want the methods to be thread-safe (they need to be pure functional to be such all context provided as arguments)

You will call the debloater from the `ILogicalDevice` and that class is _mostly_ thread safe.

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -11,11 +11,10 @@
 using namespace nbl;
 using namespace nbl::video;
 
-
-
 CVulkanLogicalDevice::CVulkanLogicalDevice(core::smart_refctd_ptr<const IAPIConnection>&& api, renderdoc_api_t* const rdoc, const IPhysicalDevice* const physicalDevice, const VkDevice vkdev, const SCreationParams& params)
 : ILogicalDevice(std::move(api),physicalDevice,params,rdoc),
-m_vkdev(vkdev), m_devf(vkdev), m_deferred_op_mempool(NODES_PER_BLOCK_DEFERRED_OP*sizeof(CVulkanDeferredOperation),1u,MAX_BLOCK_COUNT_DEFERRED_OP,static_cast<uint32_t>(sizeof(CVulkanDeferredOperation)))
+m_vkdev(vkdev), m_devf(vkdev), m_deferred_op_mempool(NODES_PER_BLOCK_DEFERRED_OP*sizeof(CVulkanDeferredOperation),1u,MAX_BLOCK_COUNT_DEFERRED_OP,static_cast<uint32_t>(sizeof(CVulkanDeferredOperation))),
+m_spirvDebloater(core::make_smart_refctd_ptr<asset::ISPIRVDebloater>())
```
devshgraphicsprogramming, line 17: make the debloater live in `ILogicalDevice` because I want you to give the same treatment to SPIR-V modules which are inputs to pipeline creation for any potential future backend

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -1164,7 +1168,15 @@ void CVulkanLogicalDevice::createComputePipelines_impl(
...
         initPipelineCreateInfo(outCreateInfo,info);
-        outCreateInfo->stage = getVkShaderStageCreateInfoFrom(info.shader,outShaderModule,outEntryPoints, outRequiredSubgroupSize,outSpecInfo,outSpecMapEntry,outSpecData);
+        const auto& spec = info.shader;
+        const auto debloatEntryPoints = std::array{ asset::ISPIRVDebloater::EntryPoint{spec.entryPoint , spec.stage} };
+        const auto debloatResult = m_spirvDebloater->debloat(spec.shader->getContent(), debloatEntryPoints, m_logger);
+        if (!debloatResult.isAllEntryPointsFound)
+        {
+            m_logger.log("Entry point %s is not found!", system::ILogger::ELL_ERROR, spec.entryPoint);
+            return;
+        }
```
devshgraphicsprogramming, line 1178: entry point removal and capability trimming need to be done before we hit `create...Pipelines_impl` 
https://github.com/Devsh-Graphics-Programming/Nabla/pull/839/files/7c548b7c80ff07d588a49571c7566b857ae9b215..31724ef75dc4e25499d95d318cd1024bce97b914#r2026648971

include/nbl/asset/utils/ISPIRVDebloater.h
```diff
@@ -17,19 +17,31 @@ class ISPIRVDebloater final : public core::IReferenceCounted
...
+            {
+                if (auto cmp = shaderStage <=> other.shaderStage; cmp != 0)
+                    return cmp;
+                return name <=> other.name;
+            }
         };
 
-        Result debloat(const ICPUBuffer* spirvBuffer, std::span<const EntryPoint> entryPoints, system::logger_opt_ptr logger) const;
+        Result debloat(const ICPUBuffer* spirvBuffer, std::span<const EntryPoint> entryPoints, system::logger_opt_ptr logger = nullptr) const;
+        core::smart_refctd_ptr<ICPUBuffer> debloat(core::smart_refctd_ptr<ICPUBuffer>&& code, std::span<const EntryPoint> entryPoints, system::logger_opt_ptr logger = nullptr) const;
```
devshgraphicsprogramming, line 44: take a `const ICPUBuffer` and return a `const ICPUBuffer`

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -8,23 +8,92 @@
...
 
+static constexpr spv_target_env SPIRV_VERSION = spv_target_env::SPV_ENV_UNIVERSAL_1_6;
+
 ISPIRVDebloater::ISPIRVDebloater()
 {
     constexpr auto optimizationPasses = std::array{
+        ISPIRVOptimizer::EOP_DEAD_BRANCH_ELIM,
+        ISPIRVOptimizer::EOP_ELIM_DEAD_FUNCTIONS,
+        ISPIRVOptimizer::EOP_DEAD_BRANCH_ELIM,
         ISPIRVOptimizer::EOP_ELIM_DEAD_FUNCTIONS,
```
devshgraphicsprogramming, line 19: why are you doing two rounds of dead branch elimination followed by dead function elimination ?

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -8,23 +8,92 @@
...
+          nullptr,
+          false
+       };
+    }
+
+    // We will remove found entry point one by one. We set all entry points as unfound initially
+    core::set<EntryPoint> unfoundEntryPoints;
+    for (const auto& entryPoint : entryPoints)
+    {
+       if (unfoundEntryPoints.find(entryPoint) != unfoundEntryPoints.end())
```
devshgraphicsprogramming, line 76: this should be fine, as long as the data match 100% (its just a duplicate entry in a set), only problem is if name matches but stage doesn't

src/nbl/asset/utils/ISPIRVOptimizer.cpp
```diff
@@ -46,8 +46,9 @@ nbl::core::smart_refctd_ptr<ICPUBuffer> ISPIRVOptimizer::optimize(const uint32_t
         &spvtools::CreateIfConversionPass,
         &spvtools::CreateStripDebugInfoPass,
         &spvtools::CreateTrimCapabilitiesPass,
-        &spvtools::CreateAggressiveDCEPass
+        &spvtools::CreateAggressiveDCEPass,
         &spvtools::CreateRemoveUnusedInterfaceVariablesPass,
+        &spvtools::CreateEliminateDeadInputComponentsSafePass,
     };
```
devshgraphicsprogramming, line 52: I want you to have a think if there'a less fragile way to initialize this array (maybe via a consteval function) so its not susceptible to getting the ordering wrong between the nable enum and SPV tools optimizer pass token. Maybe something like a list of
```cpp
array[NABLA_ENUM] = &Token;
```

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -1169,14 +1169,7 @@ void CVulkanLogicalDevice::createComputePipelines_impl(
...
         const auto& spec = info.shader;
-        const auto debloatEntryPoints = std::array{ asset::ISPIRVDebloater::EntryPoint{spec.entryPoint , spec.stage} };
-        const auto debloatResult = m_spirvDebloater->debloat(spec.shader->getContent(), debloatEntryPoints, m_logger);
-        if (!debloatResult.isAllEntryPointsFound)
-        {
-            m_logger.log("Entry point %s is not found!", system::ILogger::ELL_ERROR, spec.entryPoint);
-            return;
-        }                 
-        outCreateInfo->stage = getVkShaderStageCreateInfoFrom(spec, debloatResult.spirv.get(), outShaderModule, outEntryPoints, outRequiredSubgroupSize, outSpecInfo, outSpecMapEntry, outSpecData);
+        outCreateInfo->stage = getVkShaderStageCreateInfoFrom(spec, nullptr, outShaderModule, outEntryPoints, outRequiredSubgroupSize, outSpecInfo, outSpecMapEntry, outSpecData);
```
devshgraphicsprogramming, line 1172: there's suddenly a nullptr where the spirv bytecode used to be ?

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -45,71 +114,131 @@ ISPIRVDebloater::Result ISPIRVDebloater::debloat(const  ICPUBuffer* spirvBuffer,
...
+        if (opcode == spv::OpEntryPoint) break;
+        offset += length;
+    }
+
+    // handle entry points removal
     while (offset < spirvDwordCount) {
         const auto curOffset = offset;
         const auto instruction = spirv[curOffset];
+        const auto [length, opcode] = parse_instruction(instruction);
+        if (opcode != spv::OpEntryPoint) break;
```
devshgraphicsprogramming, line 149: what's the logic behind breaking out on the very first instruction that's not an entry point? shouldn't this be a `continue` ?

devshgraphicsprogramming: also done after advancing `offset += length`

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -45,71 +114,131 @@ ISPIRVDebloater::Result ISPIRVDebloater::debloat(const  ICPUBuffer* spirvBuffer,
...
+                minimizedSpirv.reserve(spirvDwordCount);
+                minimizedSpirv.insert(minimizedSpirv.end(), spirv, spirv + curOffset);
+                needDebloat = true;
             }
+            removedEntryPointIds.insert(curEntryPointId);
+            continue;
         }
-        
         if (!needDebloat) continue;
         minimizedSpirv.insert(minimizedSpirv.end(), spirv + curOffset, spirv + offset);
```
devshgraphicsprogramming, line 186: because of the `break` on the first instruction that's not an entry point, your `minimizedSpirv` will only contain entry point opcodes !

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -45,71 +114,131 @@ ISPIRVDebloater::Result ISPIRVDebloater::debloat(const  ICPUBuffer* spirvBuffer,
...
+    if (!needDebloat)
+    {
+        return {
             .spirv = nullptr,
+            .isSuccess = true,
         };
     }
 
+    // handle execution model removal
+    while (offset < spirvDwordCount)
```
devshgraphicsprogramming, line 208: quote https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#LogicalLayout in the comment so not resetting `offset` is not taken to be a bug

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -8,23 +8,92 @@
...
+        std::string location;
+        if (src)
+            location = src + ":"s + std::to_string(pos.line) + ":" + std::to_string(pos.column);
+        else
+            location = "";
+
+        logger.log(location, lvl, msg);
+    };
+    spvtools::SpirvTools core(SPIRV_VERSION);
+    core.SetMessageConsumer(msgConsumer);
```
devshgraphicsprogramming, line 54: are there any options about how strong it will validate?

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -739,6 +765,55 @@ asset::ICPUPipelineCache::SCacheKey ILogicalDevice::getPipelineCacheKey() const
...
+        NBL_LOG_ERROR("Invalid parameters were given");
+        return false;
+    }
+
+    core::vector<IGPUComputePipeline::SCreationParams> newParams(params.begin(), params.end());
+    const auto shaderCount = std::accumulate(params.begin(), params.end(), 0, [](uint32_t sum, auto& param)
+    {
+        return sum + param.getShaders().size();
+    });
+    core::vector<core::smart_refctd_ptr<asset::IShader>> debloatedShaders(shaderCount); // vector to hold all the debloated shaders, so the pointer from the new ShaderSpecInfo is not dangling
```
devshgraphicsprogramming, line 792: you only want to `reserve` that many

include/nbl/asset/utils/ISPIRVDebloater.h
devshgraphicsprogramming: the `m_spirvDebloater` is still a global value (member of ILogicalDevice)

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -7,10 +7,36 @@
...
+        } else
+        {
+          *outShaders = nullptr;
+        }
+        auto debloatedShaderSpec = shaderSpec;
+        debloatedShaderSpec.shader = outShaders->get();
+        *outShaderSpecInfos = debloatedShaderSpec;
+
+        outShaders++;
+        outShaderSpecInfos++;
```
devshgraphicsprogramming, line 31: this would have crashed your creation code, you're taking a pointer to shader spec infos by reference from the creation params and scrolling by one

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -7,10 +7,36 @@
...
+            .name = shaderSpec.entryPoint,
+            .shaderStage = shaderSpec.stage,
+        };
+        if (shaderSpec.shader)
+        {
+          const auto* shader = shaderSpec.shader;
+          *outShaders = core::make_smart_refctd_ptr<asset::IShader>(debloater.debloat(shader->getCode(), std::span(&entryPoint, 1)), shader->getContentType(), std::string(shader->getFilepathHint()));
+        } else
+        {
+          *outShaders = nullptr;
```
devshgraphicsprogramming, line 24: reserve in the `outShaders` and only push_back if a debloat was necessary

include/nbl/asset/utils/ISPIRVDebloater.h
devshgraphicsprogramming: also it seems that it now makes most sense to take a `const core::set<EntryPoint>&` instead of a span

src/nbl/asset/utils/ISPIRVDebloater.cpp
devshgraphicsprogramming: ok I asked you to use a set or map instead of span, so there will be no duplicates I guess

src/nbl/asset/utils/ISPIRVDebloater.cpp
```diff
@@ -55,11 +77,14 @@ static bool validate(const uint32_t* binary, uint32_t binarySize, nbl::system::l
...
 }
 
-ISPIRVDebloater::Result ISPIRVDebloater::debloat(const  ICPUBuffer* spirvBuffer, std::span<const EntryPoint> entryPoints, system::logger_opt_ptr logger) const
+ISPIRVDebloater::Result ISPIRVDebloater::debloat(const  ICPUBuffer* spirvBuffer, const core::set<EntryPoint>& entryPoints, system::logger_opt_ptr logger) const
 {
     const auto* spirv = static_cast<const uint32_t*>(spirvBuffer->getPointer());
     const auto spirvDwordCount = spirvBuffer->getSize() / 4;
 
+    std::cout << "Before strip capabilities: " << std::endl;
+    printCapabilities(spirv, spirvDwordCount);
```
devshgraphicsprogramming: use the logger please

src/nbl/asset/utils/ISPIRVOptimizer.cpp
```diff
@@ -76,8 +112,15 @@ nbl::core::smart_refctd_ptr<ICPUBuffer> ISPIRVOptimizer::optimize(const uint32_t
...
-    for (E_OPTIMIZER_PASS pass : m_passes)
-        opt.RegisterPass(create_pass_f[pass]());
+    for (E_OPTIMIZER_PASS pass : m_passes) {
+        if (getSpirvOptimizerPass(pass) != nullptr)
+        {
+            opt.RegisterPass(create_pass_f[pass]());
+        } else
+        {
+            logger.log("Optimizer pass is unknown or not supported!", system::ILogger::ELL_WARNING);
+        }
```
devshgraphicsprogramming, line 122: whats going on here, why not use `getSpirvOptoimizerPass(pass)` directly?

devshgraphicsprogramming: so that we can get rid of `create_pass_f` ? https://github.com/Devsh-Graphics-Programming/Nabla/pull/839#discussion_r2035368381

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -7,27 +7,25 @@
...
-          const auto* shader = shaderSpec.shader;
-          *outShaders = core::make_smart_refctd_ptr<asset::IShader>(debloater.debloat(shader->getCode(), std::span(&entryPoint, 1)), shader->getContentType(), std::string(shader->getFilepathHint()));
-        } else
+        const auto* shader = shaderSpec.shader;
+        auto debloatedShaderSpec = shaderSpec;
+        if (shader != nullptr)
         {
-          *outShaders = nullptr;
+            outShaders.push_back(debloater.debloat(shader, entryPoints));
+            debloatedShaderSpec.shader = outShaders.back().get();
```
devshgraphicsprogramming: its somewhat dangerous to write code like this, I get that the reserve is supposed to prevent reallocation, but you better assert that `data()` before the push_back is exactly the same as `data()` after

include/nbl/asset/utils/ISPIRVDebloater.h
```diff
@@ -0,0 +1,66 @@
...
+                    return cmp;
+                return name <=> other.name;
+            }
+        };
+
+        Result debloat(const ICPUBuffer* spirvBuffer, const core::set<EntryPoint>& entryPoints, system::logger_opt_ptr logger = nullptr) const;
+
+        inline core::smart_refctd_ptr<const IShader> debloat(const IShader* shader, const core::set<EntryPoint>& entryPoints, system::logger_opt_ptr logger = nullptr) const
+        {
+            const auto buffer = shader->getContent();
```
devshgraphicsprogramming, line 51: you need to check the `shader->getContentType()` is SPIR-V and error out before doing anything

include/nbl/asset/utils/ISPIRVDebloater.h
```diff
@@ -0,0 +1,66 @@
...
+            const auto result = debloat(buffer, entryPoints, logger);
+            if (result && result.spirv.get() == nullptr)
+            {
+                return core::smart_refctd_ptr<const IShader>(shader);
+            }
+            return core::make_smart_refctd_ptr<IShader>(core::smart_refctd_ptr(result.spirv), shader->getContentType(), std::string(shader->getFilepathHint()));
```
devshgraphicsprogramming, line 57: creating a shader from a null buffer is probably a bad idea (when you have a fail), also the content will always be SPIR-V

include/nbl/video/IGPURayTracingPipeline.h
```diff
@@ -10,9 +10,9 @@
 namespace nbl::video
 {
 
-class IGPURayTracingPipeline : public IBackendObject, public asset::IRayTracingPipeline<const IGPUPipelineLayout, const IGPUShader>
+class IGPURayTracingPipeline : public IBackendObject, public asset::IRayTracingPipeline<const IGPUPipelineLayout, const asset::IShader>
 {
-        using pipeline_t = asset::IRayTracingPipeline<const IGPUPipelineLayout,const IGPUShader>;
+        using pipeline_t = asset::IRayTracingPipeline<const IGPUPipelineLayout,const asset::IShader>;
```
devshgraphicsprogramming, line 15: the `I....Pipeline` no longer need to be templated on the shader type (there's only one now)

src/nbl/asset/utils/ISPIRVDebloater.cpp
devshgraphicsprogramming: leave a comment in the code if you still think that is corect

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -7,10 +7,48 @@
...
+static void debloatShaders(const asset::ISPIRVDebloater& debloater, std::span<const asset::IPipelineBase::SShaderSpecInfo> shaderSpecs, core::vector<core::smart_refctd_ptr<const asset::IShader>>& outShaders, asset::IPipelineBase::SShaderSpecInfo* outShaderSpecInfos, system::logger_opt_ptr logger = nullptr)
+{
+    using EntryPoints = core::set<asset::ISPIRVDebloater::EntryPoint>;
+    core::map<const asset::IShader*, EntryPoints> entryPointsMap;
+
+    // collect all entry points first before we debloat
+    for (const auto& shaderSpec : shaderSpecs) {
+        const auto* shader = shaderSpec.shader;
+        if (entryPointsMap.contains(shader))
+            entryPointsMap.emplace(shader, EntryPoints());
```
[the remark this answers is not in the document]
devshgraphicsprogramming: also perf nitpick, I think you can write
```cpp
auto it = entryPointsMap.find(shader);
if (it==entryPointsMap.end())
  it = entryPointsMap.emplace_hint(it);|
it.second->insert({ .name = shaderSpec.entryPoint, .stage = shaderSpec.stage });
```

include/nbl/asset/utils/ISPIRVDebloater.h
devshgraphicsprogramming: ok looked the innards of `ISPIRVOptimizer` and what it uses and it seems to be thread safe

devshgraphicsprogramming:
Everything resolved, work continues in #887

PR #844, opened 2025-03-03
Fixed float16_t concepts

Przemog1:


include/nbl/builtin/hlsl/concepts/core.hlsl
```diff
@@ -29,13 +29,13 @@ template<typename T>
...
 template<typename T>
-NBL_BOOL_CONCEPT FloatingPoint = nbl::hlsl::is_floating_point_v<T>;
+NBL_BOOL_CONCEPT FloatingPoint = nbl::hlsl::is_floating_point_v<T> || nbl::hlsl::is_same_v<T, float16_t>;
 
 template<typename T>
 NBL_BOOL_CONCEPT Boolean = nbl::hlsl::is_same_v<T, bool> || (nbl::hlsl::is_vector_v<T> && nbl::hlsl::is_same_v<typename vector_traits<T>::scalar_type, bool>);
 
 template <typename T>
-NBL_BOOL_CONCEPT Scalar = nbl::hlsl::is_scalar_v<T>;
+NBL_BOOL_CONCEPT Scalar = nbl::hlsl::is_scalar_v<T> || nbl::hlsl::is_same_v<T, float16_t>;
```
devshgraphicsprogramming, line 38: you don't want to "fix" like this, get the `nbl::hlsl::is_scalar` and `nbl::hlsl::is_floating_point` to return true for `float16_t` otherwise we'll end up with some really horrible behaviour

Btw if you need something to tell one not to use `float16_t` on CPU and `emulated_float64_t` on GPU, add an `is_emulated_arithmetic_v = is_arithmetic_v && is_fundamental_v`

PR #845, opened 2025-03-05
Fixed Demote-Promote-Writer-Readers Lock

devshgraphicsprogramming:

### DIFF
--- a/include/nbl/system/demote_promote_writer_readers_lock.h
+++ b/include/nbl/system/demote_promote_writer_readers_lock.h
@@ -71,7 +71,7 @@ public:
 				return oldState & preemptedMask;
 			};
 
-		perform_under_locked_state(std::move(success), preemptionCheck);
+		perform_under_locked_state(success, preemptionCheck);
 	}
 
 	/**
@@ -92,7 +92,7 @@ public:
 				return (oldState & currentReadersMask) && !(oldState & writingMask);
 			};
 
-		perform_under_locked_state(std::move(success), defaultPreemptionCheck, sanityChecks);
+		perform_under_locked_state(success, DefaultPreemptionCheck(), sanityChecks);
 	}
 
 	/**
@@ -133,7 +133,7 @@ public:
 				return preemptedState;
 			};
 
-		perform_under_locked_state(std::move(success), preemptionCheck, defaultSanityChecks, preempted);
+		perform_under_locked_state(success, preemptionCheck, defaultSanityChecks, preempted);
 	}
 
 	/**
@@ -154,7 +154,7 @@ public:
 				return (oldState & writingMask) && !(oldState & currentReadersMask);
 			};
 
-		perform_under_locked_state(std::move(success), defaultPreemptionCheck, sanityChecks);
+		perform_under_locked_state(success, DefaultPreemptionCheck(), sanityChecks);
 	}
 
 	/**
@@ -206,7 +206,7 @@ public:
 				return preemptedState;
 			};
 
-		perform_under_locked_state(std::move(success), preemptionCheck, sanityChecks, preempted);
+		perform_under_locked_state(success, preemptionCheck, sanityChecks, preempted);
 	}
 
 	/**
@@ -229,19 +229,26 @@ public:
 				return (oldState & writingMask) && !(oldState & currentReadersMask);
 			};
 
-		perform_under_locked_state(std::move(success), defaultPreemptionCheck, sanityChecks);
+		perform_under_locked_state(success, DefaultPreemptionCheck(), sanityChecks);
 	}
 
 private:
 
-	constexpr static auto defaultPreemptionCheck = [](const state_lock_value_t oldState) -> bool {return false; };
+	struct DefaultPreemptionCheck
+	{
+		bool operator()(state_lock_value_t oldState)
+		{
+			return false;
+		}
+	};
+
 	constexpr static auto defaultPreempted = [](const state_lock_value_t oldState)->state_lock_value_t {return oldState; };
 	constexpr static auto defaultSanityChecks = [](const state_lock_value_t oldState)->bool {return true; };
 
-	template<typename Success, typename PreemptionCheck = decltype(defaultPreemptionCheck), typename SanityChecks = decltype(defaultSanityChecks), typename Preempted = decltype(defaultPreempted)>
+	template<typename Success, typename PreemptionCheck = DefaultPreemptionCheck, typename SanityChecks = decltype(defaultSanityChecks), typename Preempted = decltype(defaultPreempted)>
 	inline void perform_under_locked_state(
 		Success&& success,
-		PreemptionCheck& preemptionCheck = defaultPreemptionCheck,
+		PreemptionCheck&& preemptionCheck = DefaultPreemptionCheck(),
 		SanityChecks& sanityChecks = defaultSanityChecks,
 		Preempted& preempted = defaultPreempted
 		)
--- a/include/nbl/system/demote_promote_writer_readers_lock.h
+++ b/include/nbl/system/demote_promote_writer_readers_lock.h
@@ -13,34 +13,34 @@ namespace nbl::system
 // Ugly workaround, structs with constevals can't be nested
 namespace impl
 {
-	template<typename U>
-	struct DPWRLStateLock
-	{
-		using type = std::atomic<U>;
-	};
 
-	// will fail if `atomic_unsiged_lock_free` doesn't exist or its value type has wrong size
-	template<typename U> requires (sizeof(std::atomic_unsigned_lock_free::value_type) >= sizeof(U))
-	struct DPWRLStateLock<U>
-	{
-		using type = std::atomic_unsigned_lock_free;
-	};
+template<typename U>
+struct DPWRLStateLock
+{
+	using type = std::atomic<U>;
+};
 
-	struct DPWRLStateSemantics
-	{
-		using state_lock_value_t = typename DPWRLStateLock<uint32_t>::type::value_type;
-		uint32_t currentReaders : 10 = 0;
-		uint32_t pendingWriters : 10 = 0;
-		uint32_t pendingUpgrades : 10 = 0;
-		uint32_t writing : 1 = 0;
-		uint32_t stateLocked : 1 = 0;
+// will fail if `atomic_unsiged_lock_free` doesn't exist or its value type has wrong size
+template<typename U> requires (sizeof(std::atomic_unsigned_lock_free::value_type) >= sizeof(U))
+struct DPWRLStateLock<U>
+{
+	using type = std::atomic_unsigned_lock_free;
+};
 
-		consteval inline operator state_lock_value_t() const
-		{ 
-			return static_cast<state_lock_value_t>((currentReaders << 22) | (pendingWriters << 12) | (pendingUpgrades << 2) | (writing << 1) | stateLocked);
-		}
-	};
-} //namespace impl
+struct DPWRLStateSemantics
+{
+	using state_lock_value_t = typename DPWRLStateLock<uint32_t>::type::value_type;
+	uint32_t currentReaders : 10 = 0;
+	uint32_t pendingWriters : 10 = 0;
+	uint32_t pendingUpgrades : 10 = 0;
+	uint32_t writing : 1 = 0;
+	uint32_t stateLocked : 1 = 0;
+
+	consteval inline operator state_lock_value_t() const
+	{ 
+		return static_cast<state_lock_value_t>((currentReaders << 22) | (pendingWriters << 12) | (pendingUpgrades << 2) | (writing << 1) | stateLocked);
+	}
+};
 
 enum DPWR_LOCK_DEBUG_STAGE
 {
@@ -51,19 +51,21 @@ enum DPWR_LOCK_DEBUG_STAGE
 	PREEMPTED
 };
 
-namespace impl
+struct DPWRLVoidDebugCallback
 {
-	template<typename T>
-	concept DPWRLDebugCallback = requires(T t, DPWR_LOCK_DEBUG_STAGE stage) {
-		{ t(stage) } -> std::same_as<void>;
-	};
+};
 
-	class DPWRLVoidDebugCallback
-	{
-	public:
-		void operator()(DPWR_LOCK_DEBUG_STAGE) {};
-	};
-}
+template<typename T>
+/**
+* @brief A valid debug callback overrides `void operator()(DPWR_LOCK_DEBUG_STAGE stage)`
+*/
+concept DPWRLDebugCallback = requires(T t, DPWR_LOCK_DEBUG_STAGE stage) {
+	{ t(stage) } -> std::same_as<void>;
+} || std::is_same_v<T, DPWRLVoidDebugCallback>;
+
+} //namespace impl
+
+
 
 template <impl::DPWRLDebugCallback DebugCallback = impl::DPWRLVoidDebugCallback>
 /**
@@ -265,7 +267,7 @@ public:
 
 private:
 
-	constexpr static inline bool usingDebugCallback = std::is_same_v<DebugCallback, impl::DPWRLVoidDebugCallback>;
+	constexpr static inline bool usingDebugCallback = ! std::is_same_v<DebugCallback, impl::DPWRLVoidDebugCallback>;
 
 	struct DefaultPreemptionCheck
 	{
@@ -286,6 +288,7 @@ private:
 		Preempted& preempted = defaultPreempted
 		)
 	{
+		DebugCallback callback = {};
 		for (uint32_t spinCount = 0; true; spinCount++)
 		{
 			const state_lock_value_t oldState = state.fetch_or(flipLock);
@@ -301,14 +304,14 @@ private:
 			const state_lock_value_t newState = wasPreempted ? preempted(oldState) : success(oldState);
 			// new state must unlock the state lock
 			assert(!(newState & flipLock));
-			if (usingDebugCallback) DebugCallback(DPWR_LOCK_DEBUG_STAGE::BEFORE_STATE_UPDATE);
+			if constexpr (usingDebugCallback) callback(impl::DPWR_LOCK_DEBUG_STAGE::BEFORE_STATE_UPDATE);
 			state.store(newState);
-			if (usingDebugCallback) DebugCallback(DPWR_LOCK_DEBUG_STAGE::AFTER_STATE_UPDATE);
+			if constexpr (usingDebugCallback) callback(impl::DPWR_LOCK_DEBUG_STAGE::AFTER_STATE_UPDATE);
 			if (wasPreempted)
 			{
-				if (usingDebugCallback)
+				if constexpr (usingDebugCallback)
 				{
-					DebugCallback(DPWR_LOCK_DEBUG_STAGE::PREEMPTED);
+					callback(impl::DPWR_LOCK_DEBUG_STAGE::PREEMPTED);
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
+#ifndef __NBL_DEMOTE_PROMOTE_WRITER_READERS_LOCK_H_INCLUDED__
+#define __NBL_DEMOTE_PROMOTE_WRITER_READERS_LOCK_H_INCLUDED__
+
+#include <boost/atomic/atomic.hpp>
+#include <thread>
+#include <mutex> // for std::adopt_lock_t
+
+// TODO: Reset spins if have to yield?
```
devshgraphicsprogramming, line 8: no

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+// TODO atomic_unsigned_lock_free? since C++20 (from SReadWriteSpinLock)
```
devshgraphicsprogramming, line 9: yes, just assert the size

devshgraphicsprogramming: but then you need to be a bit more clever with the `StateSemantics` left over bits (if any)

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+	{
+		explicit inline operator uint32_t() const { return std::bit_cast<uint32_t>(*this); }
+
+		uint32_t currentReaders : 10 = 0;
+		uint32_t pendingWriters : 10 = 0;
+		uint32_t pendingUpgrades : 10 = 0;
+		uint32_t writing : 1 = 0;
+		uint32_t stateLocked : 1 = 0; 
+	};
+	boost::atomic_unsigned_lock_free state;
```
devshgraphicsprogramming, line 28: don't use boost in top level headers, you'll pollute everyone's project with it, do not want

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+
+// TODO: Reset spins if have to yield?
+// TODO atomic_unsigned_lock_free? since C++20 (from SReadWriteSpinLock)
+
+namespace nbl::system
+{
+
+class demote_promote_writer_readers_lock
+{
+	constexpr static inline uint32_t MaxActors = 1023;
```
devshgraphicsprogramming, line 16: add a comment about what this is, so people don't get confused like you did

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+
+	/**
+	* @brief Acquires lock for reading. This is the operation with the lowest priority, thread will be blocked until there are no writers and no readers waiting for a writer upgrade
+	*/
+	void read_lock()
+	{
+		constexpr uint32_t preemptedMask = StateSemantics{ .currentReaders = 0,.pendingWriters = MaxActors,.pendingUpgrades = MaxActors,.writing = true,.stateLocked = false };
+		for (uint32_t spinCount = 0; true; spinCount++)
+		{
+			uint32_t actual = state.fetch_or(flipLock);
```
devshgraphicsprogramming, line 41: its not the actual state, its the old state, change variable name

include/nbl/system/demote_promote_writer_readers_lock.h
devshgraphicsprogramming: actually if you want to use `atomic_unsigned_lock_free` then you need `atomic_unsigned_lock_free::value_type` used throughout instead of `uint32_t` for the conversion op, bitfield, etc.

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+			{
+				state.store(actual);
+				if (spinCount > SpinsBeforeYield)
+					std::this_thread::yield();
+				continue;
+			}
+			actual += static_cast<uint32_t>(StateSemantics{ .currentReaders = 1 });
+			// release the flipLock, increment currentReaders 
+			state.store(actual);
+			reading = true;
```
devshgraphicsprogramming, line 61: what's `reading` it doesn't exist as a variable anywhere

include/nbl/system/demote_promote_writer_readers_lock.h
devshgraphicsprogramming: also make it a `const`

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+			// Must release flipLock and yield for other threads to progress
+			if (actual & preemptedMask)
+			{
+				state.store(actual);
+				if (spinCount > SpinsBeforeYield)
+					std::this_thread::yield();
+				continue;
+			}
+			actual += static_cast<uint32_t>(StateSemantics{ .currentReaders = 1 });
+			// release the flipLock, increment currentReaders
```
devshgraphicsprogramming, line 59: you might want to pepper a few sanity check asserts here and there

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+			}
+			actual -= static_cast<uint32_t>(StateSemantics{ .currentReaders = 1 });
+			// release the flipLock, decrement currentReaders
+			state.store(actual);
+			break;
+		}
+	}
+
+	/**
+	* @brief Acquires lock for writing. Is superseded by reader upgrades: if any thread with a reader's lock is trying to upgrade, then this thread will be blocked until that reader becomes a writer
```
devshgraphicsprogramming, line 89: wrong doc/bad grammar this thread will be blocked until all previous readers and writers release their locks, but it will preempt any further readers from acquiring lock

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+			}
+			// If there are any pending upgrades or a thread is currently reading or writing, cannot acquire lock (`preemptedMask`), so release flipLock and yield for other threads to progress
+			// Also registers this thread as a pending writer if not done so already
+			if (actual & preemptedMask)
+			{
+				if (!registeredPending)
+				{
+					registeredPending = true;
+					actual += static_cast<uint32_t>(StateSemantics{ .pendingWriters = 1 });
+				}
```
devshgraphicsprogramming, line 113: you want to register as a pending writer ALWAYS no matter if you're getting preempted or not (how else will everyone know that you're pending), its just the state lock you need to acquire

include/nbl/system/demote_promote_writer_readers_lock.h
devshgraphicsprogramming: same stlistic thing in all functions

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+		for (uint32_t spinCount = 0; true; spinCount++)
+		{
+			uint32_t actual = state.fetch_or(flipLock);
+			// stateLocked: some other thread is doing work and expects to update state
+			if (actual & flipLock)
+			{
+				if (spinCount > SpinsBeforeYield)
+					std::this_thread::yield();
+				continue;
+			}
```
devshgraphicsprogramming, line 206: hmm I see this 6 times completely identical, maybe wrap it up into a private method `lock_state`?

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+			{
+				actual |= writingMask;
+				// If had been registered earlier
+				if (registeredPending)
+				{
+					actual -= static_cast<uint32_t>(StateSemantics{ .pendingUpgrades = 1 });
+				}
+				state.store(actual);
+				break;
+			}
```
devshgraphicsprogramming, line 177: you're not checking for current readers, follow discord more https://discord.com/channels/593902898015109131/1344268285566517279/1346491987272011807

Also all your other operations are written, all failure cases first and `continue` with a final break

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -0,0 +1,217 @@
...
+			{
+				if (spinCount > SpinsBeforeYield)
+					std::this_thread::yield();
+				continue;
+			}
+			actual &= ~writingMask;
+			actual += static_cast<uint32_t>(StateSemantics{ .currentReaders = 1 });
+			// Release flipLock, declare that we're no longer writing and that we're reading
+			state.store(actual);
+		}
```
devshgraphicsprogramming, line 211: where's the `break` at the end?

Fletterio: It's the break on 176

include/nbl/system/demote_promote_writer_readers_lock.h
Fletterio: Oh yeah I think I forgot about the preemption mask

include/nbl/system/demote_promote_writer_readers_lock.h
Fletterio: Forgot 💀

include/nbl/system/demote_promote_writer_readers_lock.h
Fletterio: Is that def anywhere else? My system couldn't find it, I'll see about compiling when I'm writing the tests

include/nbl/system/demote_promote_writer_readers_lock.h
Fletterio: not sure how to pull it out, because it's integrated into the loop

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -1,64 +1,88 @@
...
+	template<typename U> requires (sizeof(std::atomic_unsigned_lock_free::value_type) >= sizeof(U))
+	struct DPWRLstate_lock<U>
 	{
-		explicit inline operator uint32_t() const { return std::bit_cast<uint32_t>(*this); }
+		using type = std::atomic_unsigned_lock_free;
+	};
+
+	constexpr inline DPWRLstate_lock<uint32_t>::type::value_type DPWRLMaxActors = 1023;
 
+	struct DPWRLStateSemantics
```
devshgraphicsprogramming, line 29: why are we mixing snake case with camel case

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -1,64 +1,88 @@
...
 		uint32_t currentReaders : 10 = 0;
 		uint32_t pendingWriters : 10 = 0;
 		uint32_t pendingUpgrades : 10 = 0;
 		uint32_t writing : 1 = 0;
-		uint32_t stateLocked : 1 = 0; 
+		uint32_t stateLocked : 1 = 0;
+
+		consteval inline operator state_lock_value_t() const
+		{ 
+			assert(currentReaders <= DPWRLMaxActors && pendingWriters <= DPWRLMaxActors && pendingUpgrades <= DPWRLMaxActors);
```
devshgraphicsprogramming, line 40: assert is pointless, it will always be true

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -1,64 +1,88 @@
...
+
+	// will fail if `atomic_unsiged_lock_free` doesn't exist or its value type has wrong size
+	template<typename U> requires (sizeof(std::atomic_unsigned_lock_free::value_type) >= sizeof(U))
+	struct DPWRLstate_lock<U>
 	{
-		explicit inline operator uint32_t() const { return std::bit_cast<uint32_t>(*this); }
+		using type = std::atomic_unsigned_lock_free;
+	};
+
+	constexpr inline DPWRLstate_lock<uint32_t>::type::value_type DPWRLMaxActors = 1023;
```
devshgraphicsprogramming, line 27: this needs to be public in the lock class, not an implementation detail

devshgraphicsprogramming: I see it already is aliased, but it has no business being here

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -1,64 +1,88 @@
...
+} //namespace impl
 
-	constexpr uint32_t flipLock = StateSemantics{ .currentReaders = 0,.pendingWriters = 0,.pendingUpgrades = 0,.writing = false,.stateLocked = true };
-	constexpr uint32_t writingMask = StateSemantics{ .currentReaders = 0,.pendingWriters = 0,.pendingUpgrades = 0,.writing = true,.stateLocked = false };
+class demote_promote_writer_readers_lock
+{
+public:
+	using state_lock_t = impl::DPWRLstate_lock<uint32_t>::type;
+	using state_lock_value_t = impl::DPWRLStateSemantics::state_lock_value_t;
+	// Limit on how many threads can be launched concurrently that try to read/write from/to the same file
```
devshgraphicsprogramming, line 51: "file?"

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -1,64 +1,88 @@
...
+	constexpr static inline state_lock_value_t MaxActors = impl::DPWRLMaxActors;
+	
+	friend class dp_read_lock_guard;
+	friend class dp_write_lock_guard;
 
 	/**
-	* @brief Acquires lock for reading. This is the operation with the lowest priority, thread will be blocked until there are no writers and no readers waiting for a writer upgrade
+	* @brief Acquires lock for reading. This thread will be blocked until there are no writers (writing or pending) and no readers waiting for a writer upgrade
```
devshgraphicsprogramming, line 58: > but it will preempt any further readers from acquiring lock between first state lock acquire and releasing the lock it will acquire when this method returns

important thing to note

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -1,64 +1,88 @@
...
-	constexpr static inline uint32_t MaxActors = 1023;
-	constexpr static inline uint32_t SpinsBeforeYield = 5000;
-	struct StateSemantics
+	template<typename U>
+	struct DPWRLstate_lock
+	{
+		using type = std::atomic<U>;
+	};
+
+	// will fail if `atomic_unsiged_lock_free` doesn't exist or its value type has wrong size
```
devshgraphicsprogramming, line 20: btw another solution is to have `constexpr` function like `flipLockMaks` instead of the ``flipLock` `constexpr static inline` variable

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -1,64 +1,88 @@
...
 	/**
-	* @brief Acquires lock for reading. This is the operation with the lowest priority, thread will be blocked until there are no writers and no readers waiting for a writer upgrade
+	* @brief Acquires lock for reading. This thread will be blocked until there are no writers (writing or pending) and no readers waiting for a writer upgrade
 	*/
 	void read_lock()
 	{
-		constexpr uint32_t preemptedMask = StateSemantics{ .currentReaders = 0,.pendingWriters = MaxActors,.pendingUpgrades = MaxActors,.writing = true,.stateLocked = false };
-		for (uint32_t spinCount = 0; true; spinCount++)
+		constexpr state_lock_value_t preemptedMask = pendingWritersMask | pendingUpgradesMask | writingMask;
+		for (state_lock_value_t spinCount = 0; true; spinCount++)
```
devshgraphicsprogramming, line 63: you don't want to use `state_lock_value_t` for this, it has nothing in common/todo with it

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -130,62 +163,75 @@ class demote_promote_writer_readers_lock
...
-			// Release flipLock
-			state.store(actual);
-			if (spinCount > SpinsBeforeYield)
-				std::this_thread::yield();
+				// Release flipLock
+				state.store(nextState);
+				if (spinCount > SpinsBeforeYield)
+					std::this_thread::yield();
 			}
+			// Release flipLock, declare that this thread is writing, mark that this thread is no longer pending IF had been registered as such
```
devshgraphicsprogramming, line 229: you're missing a `continue` after the yield/spin

This is a great argument for keeping your code DRY and unifying the control flow 
https://github.com/Devsh-Graphics-Programming/Nabla/pull/845#discussion_r1986890925

include/nbl/system/demote_promote_writer_readers_lock.h
Fletterio: Does it preempt further readers from acquiring? It can release the state lock without returning if it has to yield

devshgraphicsprogramming: damn I left the comment in the wrong place, this is the doc for read_lock, nvm this review comment

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -193,86 +164,122 @@ class demote_promote_writer_readers_lock
...
+private:
+
+	constexpr static auto defaultPreemptionCheck = [](const state_lock_value_t oldState) -> bool {return false; };
+	constexpr static auto defaultPreempted = [](const state_lock_value_t oldState)->state_lock_value_t {return oldState; };
+	constexpr static auto defaultSanityChecks = [](const state_lock_value_t oldState)->bool {return true; };
+
+	template<typename Success, typename PreemptionCheck = decltype(defaultPreemptionCheck), typename SanityChecks = decltype(defaultSanityChecks), typename Preempted = decltype(defaultPreempted)>
+	inline void perform_under_locked_state(
+		Success&& success,
+		PreemptionCheck& preemptionCheck = defaultPreemptionCheck,
```
devshgraphicsprogramming, line 244: don't make `preemptionCheck` stateful then, just take it as a `typename` and call `operator()` on a fresh object

devshgraphicsprogramming:
looks good

devshgraphicsprogramming:
Ok lets merge

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -41,16 +42,46 @@ namespace impl
...
 
-class demote_promote_writer_readers_lock
+enum DPWR_LOCK_DEBUG_STAGE
+{
+	// To ensure correctness during debug, the callback at `BEFORE_STATE_UPDATE` MUST make the virtual thread that called it run again. This is so that during debug
+	// only the virtual thread holding the flipLock progresses
+	BEFORE_STATE_UPDATE,
+	AFTER_STATE_UPDATE,
+	PREEMPTED
+};
```
devshgraphicsprogramming, line 52: hide in impl namespace

include/nbl/system/demote_promote_writer_readers_lock.h
```diff
@@ -261,18 +301,27 @@ class demote_promote_writer_readers_lock
...
 			// new state must unlock the state lock
 			assert(!(newState & flipLock));
+			if (usingDebugCallback) DebugCallback(DPWR_LOCK_DEBUG_STAGE::BEFORE_STATE_UPDATE);
 			state.store(newState);
+			if (usingDebugCallback) DebugCallback(DPWR_LOCK_DEBUG_STAGE::AFTER_STATE_UPDATE);
 			if (wasPreempted)
 			{
-				if (spinCount > SpinsBeforeYield)
-					std::this_thread::yield();
+				if (usingDebugCallback)
```
devshgraphicsprogramming, line 309: `if constexpr`

devshgraphicsprogramming:
@Fletterio get the GCC/Clang fixes in, and we merge.

PR #858, opened 2025-03-25
CAD DTM Rendering

Erfan-Ahmadi:
CAD DTM Rendering Example


include/nbl/video/alloc/SubAllocatedDescriptorSet.h
```diff
@@ -99,6 +108,7 @@ class SubAllocatedDescriptorSet : public core::IReferenceCounted
 				}
 			protected:
 				core::smart_refctd_dynamic_array<value_type> m_addresses;
+				core::smart_refctd_dynamic_array<ref_t> m_objectsToHold;
```
devshgraphicsprogramming, line 111: ideally these two should share an allocation to be faster.

PR #859, opened 2025-03-27
Build info cmake target

YasInvolved:


cmake/buildinfo/buildinfo.cmake
```diff
@@ -0,0 +1,29 @@
+cmake_host_system_information(RESULT CPU_NAME QUERY PROCESSOR_NAME)
+cmake_host_system_information(RESULT CPU_DESC QUERY PROCESSOR_DESCRIPTION)
+
+add_custom_command(
+	OUTPUT cmake_info.json
+	COMMAND ${CMAKE_COMMAND} -E capabilities >> cmake_info.json
+	COMMENT "Generating ${NBL_ROOT_PATH_BINARY}/cmake_info.json"
+)
+
+set(VKSDK_INFO_COMMAND "$ENV{VULKAN_SDK}/bin/vulkaninfoSDK")
```
AnastaZIuk, line 10: don't assume the ENV [exists](<https://cmake.org/cmake/help/v3.31/module/FindVulkan.html>)

```cmake
cmake_path(GET Vulkan_INCLUDE_DIR PARENT_PATH VULKAN_SDK_INSTALL_DIRECTORY)
set(VKSDK_INFO_COMMAND "${VULKAN_SDK_INSTALL_DIRECTORY}/bin/vulkaninfoSDK")
```

you should also check if vulkan driver is enabled at all _first_ (`if (Vulkan_FOUND)`), if not then pack empty "vulkan-driver-disabled.txt" file (no idea if it compiles without currently, but do it please just for our sanity) - at some point when we finally have separate DLLs it makes sense to compile without the driver when only minimal namespace is considered which doesn't require it

cmake/buildinfo/buildinfo.cmake
```diff
@@ -0,0 +1,29 @@
+cmake_host_system_information(RESULT CPU_NAME QUERY PROCESSOR_NAME)
+cmake_host_system_information(RESULT CPU_DESC QUERY PROCESSOR_DESCRIPTION)
+
+add_custom_command(
+	OUTPUT cmake_info.json
+	COMMAND ${CMAKE_COMMAND} -E capabilities >> cmake_info.json
+	COMMENT "Generating ${NBL_ROOT_PATH_BINARY}/cmake_info.json"
```
AnastaZIuk, line 7: since it's a script, you must check if `NBL_ROOT_PATH_BINARY` is defined otherwise error

cmake/buildinfo/buildinfo.cmake
devshgraphicsprogramming: why are you calling it "vulkdan driver" this is about the SDK, the whole point of the Vulkan Loader and ICD is that there's a separate vulkan loader lib that handles loading different drivers for different gpus.

hence you should be able to compile a vulkan app even if no Vulkan GPU or driver installed on the system

AnastaZIuk: indeed and I think I just blindly copy pasted your words https://github.com/Devsh-Graphics-Programming/Nabla/commit/7e00fd639581416363f5c970eff568b229f79fa2 but yes, its all about SDK so maybe lets call the file "vulkan-sdk-not-found.txt" or sth instead

cmake/buildinfo/buildinfo.cmake
```diff
@@ -1,18 +1,28 @@
...
+	add_custom_command(
+		OUTPUT vulkan_info.json
+		COMMAND ${VKSDK_INFO_COMMAND} -j -o ${NBL_ROOT_PATH_BINARY}/vulkan_info.json
+		COMMENT "Generating ${NBL_ROOT_PATH_BINARY}/vulkan_info.json"
+	)
+else()
+	add_custom_command(
+		OUTPUT vulkan_driver_disabled.txt
+		COMMAND ${CMAKE_COMMAND} -E touch ${NBL_ROOT_PATH_BINARY}/vulkan_driver_disabled.txt
+		COMMENT "Generating ${NBL_ROOT_PATH_BINARY}/vulkan_driver_disabled.txt"
```
AnastaZIuk, line 24: https://github.com/Devsh-Graphics-Programming/Nabla/pull/859#discussion_r2019926984 + https://github.com/Devsh-Graphics-Programming/Nabla/commit/7e00fd639581416363f5c970eff568b229f79fa2#r154593064

cmake/buildinfo/buildinfo.cmake
devshgraphicsprogramming: mea maxima culpa, yes I followed the Irrlicht tradition of misnaming things in back in 2020

cmake/buildinfo/buildinfo.cmake
```diff
@@ -0,0 +1,39 @@
+cmake_host_system_information(RESULT CPU_NAME QUERY PROCESSOR_NAME)
+cmake_host_system_information(RESULT CPU_DESC QUERY PROCESSOR_DESCRIPTION)
+
+if (NOT DEFINED NBL_ROOT_PATH_BINARY)
+	message(FATAL_ERROR "NBL_ROOT_PATH_BINARY is not defined or it's empty")
+endif
```
AnastaZIuk, line 6: typo + how did you test it when it never configured?

cmake/buildinfo/buildinfo.cmake
AnastaZIuk: typo + how did you test it when it never configured?

cmake/buildinfo/buildinfo.cmake
AnastaZIuk: done by me

cmake/buildinfo/buildinfo.cmake
AnastaZIuk: done by me

PR #867, opened 2025-04-11
Yas fix clang warnings

YasInvolved:


src/nbl/video/utilities/CAssetConverter.cpp
```diff
@@ -110,7 +109,7 @@ bool CAssetConverter::acceleration_structure_patch_base::valid(const ILogicalDev
 	if (!features.accelerationStructure)
 		return false;
 	// just make the flags agree/canonicalize
-	allowCompaction = allowCompaction || compactAfterBuild;
+	allowCompaction = allowCompaction | compactAfterBuild;
```
devshgraphicsprogramming, line 112: i fixed this already, you'll conflict. I actually did a lot edits that will cause conflicts

PR #870, opened 2025-04-25
New workgroup reduce + scan using subgroup2

keptsecret:


devshgraphicsprogramming:
GH still highlighting wrong

PR #881, opened 2025-05-24
Dtm lru cache upgrades

Fletterio:
Cache Upgrades discussed in Discord: Iteration (both mutable and const), Eviction callback for whole pair, deep copy constructor

### DIFF
--- a/include/nbl/core/containers/DoublyLinkedList.h
+++ b/include/nbl/core/containers/DoublyLinkedList.h
@@ -74,12 +74,17 @@ public:
 	template <bool Mutable>
 	friend class Iterator;
 
+	using iterator = Iterator<true>;
+	using const_iterator = Iterator<false>;
+
 	using allocator_t = allocator;
 	using allocator_traits_t = std::allocator_traits<allocator_t>;
 	using address_allocator_t = PoolAddressAllocator<uint32_t>;
 	using node_t = SDoublyLinkedNode<Value>;
 	using value_t = Value;
 	using disposal_func_t = std::function<void(value_t&)>;
+
+	static constexpr bool IsTriviallyCopyable = std::is_trivially_copyable_v<Value>;
 
 	_NBL_STATIC_INLINE_CONSTEXPR uint32_t invalid_iterator = node_t::invalid_iterator;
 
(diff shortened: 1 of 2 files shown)


include/nbl/core/containers/DoublyLinkedList.h
```diff
@@ -277,14 +287,14 @@ class DoublyLinkedList
...
 	// Iterator stuff
-	Iterator begin();
-	Iterator end();
-	Iterator cbegin() const;
-	Iterator cend() const;
-	std::reverse_iterator<Iterator> rbegin();
-	std::reverse_iterator<Iterator> rend();
-	std::reverse_iterator<Iterator> crbegin() const;
-	std::reverse_iterator<Iterator> crend() const;
+	Iterator<true> begin();
```
Erfan-Ahmadi, line 290: Instead of <true> <false> use "Iterator" and "Const Iterator" names
make naming convention consistent with all other typdefs in the LRUCache and DoublyLinkedList, use `iterator` and `const_iterator` just like stl
your class name could still be `Iterator` templated.

include/nbl/core/containers/LRUCache.h
```diff
@@ -38,6 +38,13 @@ class LRUCacheBase
 		LRUCacheBase(const uint32_t capacity, MapHash&& _hash, MapEquals&& _equals, disposal_func_t&& df) : m_list(capacity, std::move(df)), m_hash(std::move(_hash)), m_equals(std::move(_equals)), searchedKey(nullptr)
 		{ }
 
+		// CAREFUL: searchedKey was left as nullptr because for ResizableLRUCache, it's always set by some method before being used
+		// If you implement another LRUCache where this is not the case and the state of the copy has to match exactly, it needs to be consistent.
+		// What I would do in such a case (either in this same class or dividing this class again into two, one that leaves the searchedKey null and 
+		// another that sets it to a consistent value) would be to iterate over every Key-Value pair in the copy, and set searchedKey to point
+		// to the Key such that `MapEquals(Key, *other.searchedKey)`
+		LRUCacheBase(const LRUCacheBase& other) : m_list(other.m_list), m_hash(other.m_hash), m_equals(other.m_equals), searchedKey(nullptr) {}
```
Erfan-Ahmadi, line 46: searchedKey is fine to be left as nullptr, it's just a helper when `m_shortcut_map.find(invalid_iterator);` is used which leads to getReference here.
I think the whole wrapHash stuff is to avoid using `unordered_map` of Key->nodeAddr and avoid storing Key in another place.
instead we use an unordered_set, and when we want to get key associated with a `nodeAddr` we look up in our doublyLinkedList

PR #884, opened 2025-06-03
Block Chain Scan

devshgraphicsprogramming:
## Description

Implementation of Block Chain Scan 

## Testing 

Example XXX

## TODO list:

We'll let @kpentaris review.

### DIFF
--- a/include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
+++ b/include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
@@ -107,27 +107,6 @@ struct reduce<Config, multiplies<T>, ForwardProgressGuarantees, device_capabilit
     }
 };
 
-// template<class Config, class BinOp, bool ForwardProgressGuarantees, class device_capabilities>
-// struct reduce;
-// {
-//     using scalar_t = typename BinOp::type_t;
-//     using arith_config_t = typename Config::arith_config_t;
-//     using workgroup_reduce_t = workgroup2::reduction<arith_config_t, BinOp, device_capabilities>;
-
-//     template<class DataAccessor, class OutputAccessor, class StatusAccessor, class ScratchAccessor>
-//     void __call(NBL_REF_ARG(DataAccessor) dataAccessor, NBL_REF_ARG(OutputAccessor) outputAccessor, NBL_REF_ARG(StatusAccessor) statusAccessor, NBL_REF_ARG(ScratchAccessor) sharedMemScratchAccessor)
-//     {
-//         const scalar_t localReduction = workgroup_reduce_t::template __call<DataAccessor, ScratchAccessor>(dataAccessor, sharedMemScratchAccessor);
-
-//         const bool lastInvocation = (workgroup::SubgroupContiguousIndex() == Config::WorkgroupSize-1);
-//         if (lastInvocation)
-//         {
-//             spirv::atomicIAdd(outputAccessor.getPtr().deref().__get_spv_ptr(), spv::ScopeDevice, spv::MemorySemanticsReleaseMask/*|spv::MemorySemanticsMakeAvailableMask*/, localReduction);
-//             spirv::atomicIAdd(statusAccessor.getPtr().deref().__get_spv_ptr(), spv::ScopeDevice, spv::MemorySemanticsReleaseMask/*|spv::MemorySemanticsMakeAvailableMask*/, 1u);
-//         }
-//     }
-// };
-
 // TODO: change this to scan, it totally won't work for reduce anyways
 template<class Config, class BinOp, bool ForwardProgressGuarantees, class device_capabilities>
 struct scan


include/nbl/builtin/hlsl/scan/arithmetic.hlsl
```diff
@@ -0,0 +1,34 @@
...
+#include "nbl/builtin/hlsl/scan/arithmetic_impl.hlsl"
+
+namespace nbl
+{
+namespace hlsl
+{
+namespace scan
+{
+
+template<class Config, class BinOp, bool ForwardProgressGuarantees, class device_capabilities=void>
```
devshgraphicsprogramming, line 16: we should make a fake device feature called `forwardProgressGuarantees` which is basically always false

include/nbl/builtin/hlsl/scan/arithmetic.hlsl
```diff
@@ -0,0 +1,34 @@
...
+struct reduction
+{
+    using scalar_t = typename BinOp::type_t;
+
+    template<class ReadOnlyDataAccessor, class ScratchAccessor NBL_FUNC_REQUIRES(ArithmeticDataAccessor<ReadOnlyDataAccessor> && ArithmeticSharedMemoryAccessor<ScratchAccessor>)
+    static scalar_t __call(NBL_REF_ARG(ReadOnlyDataAccessor) dataAccessor, NBL_REF_ARG(ScratchAccessor) sharedMemScratchAccessor)   // scratch bda?
```
devshgraphicsprogramming: leave as is (remove comment), we may want to prefix sum over an image for example

include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
```diff
@@ -0,0 +1,120 @@
...
+{
+
+template<typename T>    // only uint32_t or uint64_t for now?
+struct Constants
+{
+    NBL_CONSTEXPR_STATIC_INLINE T NOT_READY = 0;
+    NBL_CONSTEXPR_STATIC_INLINE T LOCAL_COUNT = T(0x1u) << (sizeof(T)*8-2);
+    NBL_CONSTEXPR_STATIC_INLINE T GLOBAL_COUNT = T(0x1u) << (sizeof(T)*8-1);
+    NBL_CONSTEXPR_STATIC_INLINE T STATUS_MASK = LOCAL_COUNT | GLOBAL_COUNT;
+};
```
devshgraphicsprogramming, line 40: you can use `enum class` if you update DXC btw also with `: uint16_t` or `: uint64_t`

not everything needs to be a crazy template

include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
```diff
@@ -0,0 +1,120 @@
...
+    using config_t = Config;
+    using arith_config_t = typename Config::arith_config_t;
+    using workgroup_reduce_t = workgroup2::reduction<arith_config_t, BinOp, device_capabilities>;
+    using binop_t = BinOp;
+
+    template<class DataAccessor, class ScratchAccessor>
+    scalar_t __call(NBL_REF_ARG(DataAccessor) dataAccessor, NBL_REF_ARG(ScratchAccessor) sharedMemScratchAccessor)
+    {
+        const scalar_t localReduction = workgroup_reduce_t::__call<DataAccessor, ScratchAccessor>(dataAccessor, sharedMemScratchAccessor);
+        bda::__ptr<T> scratch = dataAccessor.getScratchPtr();   // scratch data should be at least T[NumWorkgroups]
```
devshgraphicsprogramming: ask for a separate accessor for the workgroup flags

Also 32bit atomics are always present and probably the cheapest, there's no harm  in packing 16 workgroup flags into the same `uint32_t` but please benchmark if its a net gain (more likely to sit in GPU's L2 cache) or loss (contention)

devshgraphicsprogramming: aaah that might not be compatible with using `UMax` though

include/nbl/builtin/hlsl/scan/arithmetic.hlsl
```diff
@@ -0,0 +1,34 @@
...
+namespace scan
+{
+
+template<class Config, class BinOp, bool ForwardProgressGuarantees, class device_capabilities=void>
+struct reduction
+{
+    using scalar_t = typename BinOp::type_t;
+
+    template<class ReadOnlyDataAccessor, class ScratchAccessor NBL_FUNC_REQUIRES(workgroup2::ArithmeticReadOnlyDataAccessor<ReadOnlyDataAccessor,scalar_t> && workgroup2::ArithmeticSharedMemoryAccessor<ScratchAccessor,scalar_t>)
+    static scalar_t __call(NBL_REF_ARG(ReadOnlyDataAccessor) dataAccessor, NBL_REF_ARG(ScratchAccessor) sharedMemScratchAccessor)
```
devshgraphicsprogramming: you need a 3rd accessor which is an atomic accessor (to both accumulate the result and figure out when everyone is done)

include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
```diff
@@ -0,0 +1,177 @@
...
+        {
+            // wait until last workgroup does reduction
+            while (!(value & constants_t::GLOBAL_COUNT))
+            {
+                // value = spirv::atomicLoad(scratchLast.__get_spv_ptr(), spv::ScopeWorkgroup, spv::MemorySemanticsAcquireMask);
+                value = spirv::atomicIAdd(scratchLast.__get_spv_ptr(), spv::ScopeWorkgroup, spv::MemorySemanticsAcquireMask, 0u);
+            }
+        }
+        value = workgroup::Broadcast(value, sharedMemScratchAccessor, Config::WorkgroupSize-1);
+        return value & (~constants_t::STATUS_MASK);
```
devshgraphicsprogramming: This won't work even with forward progress guarantees, you just need to let the workgroup quit

include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
```diff
@@ -0,0 +1,177 @@
...
+{
+    using scalar_t = typename BinOp::type_t;
+    using constants_t = Constants<scalar_t>;
+    using config_t = Config;
+    using arith_config_t = typename Config::arith_config_t;
+    using workgroup_reduce_t = workgroup2::reduction<arith_config_t, BinOp, device_capabilities>;
+    using binop_t = BinOp;
+
+    template<class DataAccessor, class ScratchAccessor>
+    scalar_t __call(NBL_REF_ARG(DataAccessor) dataAccessor, NBL_REF_ARG(ScratchAccessor) sharedMemScratchAccessor)
```
devshgraphicsprogramming, line 122: you have the readonly accessor to get your element, and you have the scratch memory accessor (for the workgroup scans/reductions) but you don't have:
- accessor for the Device-Scope scratch
- accessor for where to store the reduction result (reduction is special compared to a scan, you can't get the result right away)

include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
```diff
@@ -0,0 +1,177 @@
...
+    {
+        const scalar_t localReduction = workgroup_reduce_t::template __call<DataAccessor, ScratchAccessor>(dataAccessor, sharedMemScratchAccessor);
+        bda::__ptr<scalar_t> scratch = dataAccessor.getScratchPtr();   // scratch data should be at least T[NumWorkgroups]
+
+        const bool lastInvocation = (workgroup::SubgroupContiguousIndex() == Config::WorkgroupSize-1);
+        if (lastInvocation)
+        {
+            bda::__ref<scalar_t> scratchId = (scratch + glsl::gl_WorkGroupID().x).deref();
+            spirv::atomicUMax(scratchId.__get_spv_ptr(), spv::ScopeWorkgroup, spv::MemorySemanticsReleaseMask, localReduction|constants_t::LOCAL_COUNT);
```
devshgraphicsprogramming: you want to separate the storage of the reduction from the flags I think.

include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
```diff
@@ -0,0 +1,177 @@
...
+        scalar_t prefix = scalar_t(0);
+        // decoupled lookback
+        if (ForwardProgressGuarantees)
+        {
+            if (lastInvocation) // don't make whole block work and do busy stuff
+            {
+                // for (uint32_t prevID = glsl::gl_WorkGroupID().x-1; prevID >= 0u; prevID--)   // won't run properly this way for some reason, results in device lost
+                for (uint32_t i = 1; i <= glsl::gl_WorkGroupID().x; i++)
+                {
+                    const uint32_t prevID = glsl::gl_WorkGroupID().x-i;
```
devshgraphicsprogramming, line 144: don't use `gl_WorkGroupID` ask for the `virtualWorkgroupIndex` in the function call

include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
```diff
@@ -0,0 +1,177 @@
...
+                    scalar_t value = constants_t::NOT_READY;
+                    {
+                        // spin until something is ready
+                        while (value == constants_t::NOT_READY)
+                        {
+                            bda::__ref<scalar_t> scratchPrev = (scratch + prevID).deref();
+                            // value = spirv::atomicLoad(scratchPrev.__get_spv_ptr(), spv::ScopeWorkgroup, spv::MemorySemanticsAcquireMask);
+                            value = spirv::atomicIAdd(scratchPrev.__get_spv_ptr(), spv::ScopeWorkgroup, spv::MemorySemanticsAcquireMask, 0u);
```
devshgraphicsprogramming: you'll have multiple workgroups doing this, you'll mess up the results, you want to accumulate those locally here in a register

so while you're walking backwards, you only do `prefix += atomicLoad(scratchPrev.__get_spv_ptr(), spv::ScopeDevice, MakeVisible);` 

Also this requires that you have two different scratch store locations:
- one for local reduction
- one for global reduction

If you keep the global and local on the same address, you get nasty data races because the status is a flag and not a mutex, so you can overwrite a local result with a global while another workgroup reads and it thinks that the value it reads is a local result because flag is not updated yet.

P.S. You probably don't want to be writing out the GLOBAL results and updating status flags here even though you can, because other workgroups before you are obviously "just about" to write out their results, and this way you'll just introduce more uncached memory traffic.

include/nbl/builtin/hlsl/scan/arithmetic_impl.hlsl
```diff
@@ -0,0 +1,177 @@
...
+
+        binop_t binop;
+        scalar_t prefix = scalar_t(0);
+        // decoupled lookback
+        if (ForwardProgressGuarantees)
+        {
+            if (lastInvocation) // don't make whole block work and do busy stuff
+            {
+                // for (uint32_t prevID = glsl::gl_WorkGroupID().x-1; prevID >= 0u; prevID--)   // won't run properly this way for some reason, results in device lost
+                for (uint32_t i = 1; i <= glsl::gl_WorkGroupID().x; i++)
```
devshgraphicsprogramming, line 142: actually using the whole workgroup or at least subgroup (benchmark it) would be much faster here, so each invocation checks a workgroup and you can use `workgroup2::reduce` with a Max binop to find the first preceeding workgroup with a ready GLOBAL scan value

You'd also be able to accumulate the `prefix` faster over the ones which have LOCAL ready

PR #892, opened 2025-06-21
Restore numbers in example names

YasInvolved:


cmake/scripts/nbl/projectTargetName.cmake
```diff
@@ -21,9 +21,7 @@ if(NOT DEFINED _NBL_PROJECT_DIRECTORY_)
 endif()
 
 get_filename_component(EXECUTABLE_NAME ${_NBL_PROJECT_DIRECTORY_} NAME)
-string(REGEX REPLACE "^[0-9]+\." "" EXECUTABLE_NAME ${EXECUTABLE_NAME})
 string(TOLOWER ${EXECUTABLE_NAME} EXECUTABLE_NAME)
-string(MAKE_C_IDENTIFIER ${EXECUTABLE_NAME} EXECUTABLE_NAME)
```
AnastaZIuk, line 26: the code is back again in the macro, edit it right [there](<https://github.com/Devsh-Graphics-Programming/Nabla/blob/mesh_loaders/cmake/common.cmake#L25>) + remove this script (I forgot to do so), it was only for Jenkins purpose which we don't use anymore to build Nabla

Krzysztof made it like 7 years now, I believe it could be related to CMake restrictions in target name though it is not clear from the docs, [CMP0037](<https://cmake.org/cmake/help/v3.30/policy/CMP0037.html>)

> Target names may contain upper and lower case letters, numbers, the underscore character (_), dot(.), plus(+) and minus(-). As a special case, ALIAS and IMPORTED targets may contain two consecutive colons.

hence keep the 24 line removed but I would be careful with 26, remove and see if CI fails anywhere due to the line

AnastaZIuk: > remove and see if CI fails anywhere due to the line

ok I tried myself on different branch and it will work ok

AnastaZIuk:
done in https://github.com/Devsh-Graphics-Programming/Nabla/commit/d6fc177c80bbf8cba609c94fab90f8871d1ddd4d

PR #894, opened 2025-06-25
Mesh loaders kevin

kevyuu:
Implement extra geometries

### DIFF
--- a/src/nbl/asset/utils/CGeometryCreator.cpp
+++ b/src/nbl/asset/utils/CGeometryCreator.cpp
@@ -15,6 +15,11 @@
 namespace nbl::asset
 {
 
+static uint8_t packSnorm(float val)
+{
+	return round(hlsl::clamp(val, -1.0f, 1.0f) * 127);
+}
+
 core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createCube(const hlsl::float32_t3 size) const
 {
 	using namespace hlsl;
@@ -156,7 +161,7 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createCube(const h
 	{
 		const hlsl::vector<int8_t, 3> norm[6] =
 		{
-			hlsl::vector<int8_t,3>(0, 0, 127),
+			hlsl::vector<int8_t,3>(0, 0, 1),
 			hlsl::vector<int8_t,3>(127, 0, 0),
 			hlsl::vector<int8_t,3>(0, 0,-127),
 			hlsl::vector<int8_t,3>(-127, 0, 0),
@@ -186,121 +191,12 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createCube(const h
 	return retval;
 }
 
-#if 0
-
-/*
-	a cylinder, a cone and a cross
-	point up on (0,1.f, 0.f )
-*/
-core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createArrow(
-	const uint32_t tesselationCylinder,
-	const uint32_t tesselationCone,
-	const float height,
-	const float cylinderHeight,
-	const float width0,
-	const float width1,
-	const video::SColor vtxColor0,
-	const video::SColor vtxColor1
-) const
+core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float radius,
+				uint32_t polyCountX, uint32_t polyCountY, CQuantNormalCache* const quantNormalCacheOverride) const
 {
-    assert(height > cylinderHeight);
+	using namespace hlsl;
 
-    auto cylinder = createCylinderMesh(width0, cylinderHeight, tesselationCylinder, vtxColor0);
-    auto cone = createConeMesh(width1, height-cylinderHeight, tesselationCone, vtxColor1, vtxColor1);
-
-	auto cylinderVertices = reinterpret_cast<CylinderVertex*>(cylinder.bindings[0].buffer->getPointer());
-	auto coneVertices = reinterpret_cast<ConeVertex*>(cone.bindings[0].buffer->getPointer());
-
-	auto cylinderIndecies = reinterpret_cast<uint16_t*>(cylinder.indexBuffer.buffer->getPointer());
-	auto coneIndecies = reinterpret_cast<uint16_t*>(cone.indexBuffer.buffer->getPointer());
-
-	const auto cylinderVertexCount = cylinder.bindings[0].buffer->getSize() / sizeof(CylinderVertex);
-	const auto coneVertexCount = cone.bindings[0].buffer->getSize() / sizeof(ConeVertex);
-	const auto newArrowVertexCount = cylinderVertexCount + coneVertexCount;
-
-	const auto cylinderIndexCount = cylinder.indexBuffer.buffer->getSize() / sizeof(uint16_t);
-	const auto coneIndexCount = cone.indexBuffer.buffer->getSize() / sizeof(uint16_t);
-	const auto newArrowIndexCount = cylinderIndexCount + coneIndexCount;
-
-	for (auto i = 0ull; i < coneVertexCount; ++i)
-	{
-		core::vector3df_SIMD newPos = coneVertices[i].pos;
-		newPos.rotateYZByRAD(-1.5707963268);
-
-		for (auto c = 0; c < 3; ++c)
-			coneVertices[i].pos[c] = newPos[c];
-	}
-
-	auto newArrowVertexBuffer = asset::ICPUBuffer::create({ newArrowVertexCount * sizeof(ArrowVertex) });
-	newArrowVertexBuffer->setUsageFlags(newArrowVertexBuffer->getUsageFlags() | asset::IBuffer::EUF_VERTEX_BUFFER_BIT);
-	auto newArrowIndexBuffer = asset::ICPUBuffer::create({ newArrowIndexCount * sizeof(uint16_t) });
-	newArrowIndexBuffer->setUsageFlags(newArrowIndexBuffer->getUsageFlags() | asset::IBuffer::EUF_INDEX_BUFFER_BIT);
-
-	for (auto z = 0ull; z < newArrowVertexCount; ++z)
-	{
-		auto arrowVertex = reinterpret_cast<ArrowVertex*>(newArrowVertexBuffer->getPointer()) + z;
-
-		if (z < cylinderVertexCount)
-		{
-			auto cylinderVertex = (cylinderVertices + z);
-			memcpy(arrowVertex, cylinderVertex, sizeof(ArrowVertex));
-		}
-		else
-		{
-			auto coneVertex = (coneVertices + z - cylinderVertexCount);
-			memcpy(arrowVertex, coneVertex, offsetof(ConeVertex, normal)); // copy position and color
-			arrowVertex->uv[0] = 0;
-			arrowVertex->uv[1] = 0;
-			arrowVertex->normal = coneVertex->normal;
-		}
-	}
-
-	{
-		auto ArrowIndices = reinterpret_cast<uint16_t*>(newArrowIndexBuffer->getPointer());
-		auto newConeIndices = (ArrowIndices + cylinderIndexCount);
-
-		memcpy(ArrowIndices, cylinderIndecies, sizeof(uint16_t) * cylinderIndexCount);
-		memcpy(newConeIndices, coneIndecies, sizeof(uint16_t) * coneIndexCount);
-
-		for (auto i = 0ull; i < coneIndexCount; ++i)
-			*(newConeIndices + i) += cylinderVertexCount;
-	}
-
-	return_type arrow;
-
-	constexpr size_t vertexSize = sizeof(ArrowVertex);
-	arrow.inputParams = 
-	{ 0b1111u,0b1u,
-		{
-			{0u,EF_R32G32B32_SFLOAT,offsetof(ArrowVertex,pos)},
-			{0u,EF_R8G8B8A8_UNORM,offsetof(ArrowVertex,color)},
-			{0u,EF_R32G32_SFLOAT,offsetof(ArrowVertex,uv)},
-			{0u,EF_A2B10G10R10_SNORM_PACK32,offsetof(ArrowVertex,normal)}
-		},
-		{vertexSize,SVertexInputBindingParams::EVIR_PER_VERTEX} 
-	};
-
-	arrow.bindings[0] = { 0, std::move(newArrowVertexBuffer) }; 
-	arrow.indexBuffer = { 0, std::move(newArrowIndexBuffer) };
-	arrow.indexCount = newArrowIndexCount;
-	arrow.indexType = EIT_16BIT;
-
-    return arrow;
-}
-
-/* A sphere with proper normals and texture coords */
-core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float radius, uint32_t polyCountX, uint32_t polyCountY, IMeshManipulator* const meshManipulatorOverride) const
-{
-	// we are creating the sphere mesh here.
-	return_type retval;
-	constexpr size_t vertexSize = sizeof(CGeometryCreator::SphereVertex);
-	CQuantNormalCache* const quantNormalCache = (meshManipulatorOverride == nullptr) ? defaultMeshManipulator->getQuantNormalCache() : meshManipulatorOverride->getQuantNormalCache();
-	retval.inputParams = { 0b1111u,0b1u,{
-											{0u,EF_R32G32B32_SFLOAT,offsetof(SphereVertex,pos)},
-											{0u,EF_R8G8B8A8_UNORM,offsetof(SphereVertex,color)},
-											{0u,EF_R32G32_SFLOAT,offsetof(SphereVertex,uv)},
-											{0u,EF_A2B10G10R10_SNORM_PACK32,offsetof(SphereVertex,normal)}
-										},{vertexSize,SVertexInputBindingParams::EVIR_PER_VERTEX} };
+	CQuantNormalCache* const quantNormalCache = quantNormalCacheOverride == nullptr ? m_params.normalCache.get() : quantNormalCacheOverride;
 
 	if (polyCountX < 2)
 		polyCountX = 2;
@@ -308,15 +204,20 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
 		polyCountY = 2;
 
 	const uint32_t polyCountXPitch = polyCountX + 1; // get to same vertex on next level
+  const size_t vertexCount = (polyCountXPitch * polyCountY) + 2;
 
-	retval.indexCount = (polyCountX * polyCountY) * 6;
-	auto indices = asset::ICPUBuffer::create({ sizeof(uint32_t) * retval.indexCount });
+	auto retval = core::make_smart_refctd_ptr<ICPUPolygonGeometry>();
+	retval->setIndexing(IPolygonGeometryBase::TriangleList());
 
 	// Create indices
+	using index_t = uint32_t;
 	{
-		uint32_t level = 0;
+    const auto indexCount = (polyCountX * polyCountY) * 6;
+		const auto bytesize = sizeof(index_t) * indexCount;
+		auto indices = ICPUBuffer::create({bytesize,IBuffer::EUF_INDEX_BUFFER_BIT});
+		auto indexPtr = reinterpret_cast<index_t*>(indices->getPointer());
+    uint32_t level = 0;
 		size_t indexAddIx = 0;
-		uint32_t* indexPtr = (uint32_t*)indices->getPointer();
 		for (uint32_t p1 = 0; p1 < polyCountY - 1; ++p1)
 		{
 			//main quads, top to bottom
@@ -372,23 +273,123 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
 		indexPtr[indexAddIx++] = polyCountSqM1 + polyCountX - 1;
 		indexPtr[indexAddIx++] = polyCountSqM1;
 		indexPtr[indexAddIx++] = polyCountSq1;
-	}
-	indices->setUsageFlags(indices->getUsageFlags() | asset::IBuffer::EUF_INDEX_BUFFER_BIT);
-	retval.indexBuffer = {0ull, std::move(indices)};
 
-	// handle vertices
+		shapes::AABB<4,index_t> aabb;
+		aabb.minVx[0] = 0;
+		aabb.maxVx[0] = vertexCount - 1;
+		retval->setIndexView({
+			.composed = {
+				.encodedDataRange = {.u32=aabb},
+				.stride = sizeof(index_t),
+				.format = EF_R16_UINT,
+				.rangeFormat = IGeometryBase::EAABBFormat::U16
+			},
+			.src = {.offset=0,.size=bytesize,.buffer=std::move(indices)}
+		});
+	}
+
+	constexpr auto NormalCacheFormat = EF_R8G8B8_SNORM;
+	constexpr auto NormalFormat = EF_R8G8B8A8_SNORM;
+
+	// Create vertex attributes with NONE usage because we have no clue how they'll be used
+	hlsl::float32_t3* positions;
+  hlsl::vector<uint8_t, 4>* normals;
+	hlsl::vector<uint8_t, 2>* uvs;
+	hlsl::vector<uint8_t, 4>* colors;
 	{
-		size_t vertexSize = 3 * 4 + 4 + 2 * 4 + 4;
-		size_t vertexCount = (polyCountXPitch * polyCountY) + 2;
-		auto vtxBuf = asset::ICPUBuffer::create({ vertexCount * vertexSize });
-		auto* tmpMem = reinterpret_cast<uint8_t*>(vtxBuf->getPointer());
+		{
+			constexpr auto AttrSize = sizeof(decltype(*positions));
+			auto buff = ICPUBuffer::create({{AttrSize * vertexCount,IBuffer::EUF_NONE}});
+			positions = reinterpret_cast<decltype(positions)>(buff->getPointer());
+			shapes::AABB<4, float32_t> aabb;
+			aabb.maxVx = float32_t4(radius, radius, radius, 0.0f);
+			aabb.minVx = float32_t4(-radius, -radius, -radius, 0.0f);
+			retval->setPositionView({
+				.composed = {
+					.encodedDataRange = {.f32 = aabb},
+					.stride = AttrSize,
+					.format = EF_R32G32B32_SFLOAT,
+					.rangeFormat = IGeometryBase::EAABBFormat::F32
+				},
+				.src = {
+				  .offset=0,
+				  .size = buff->getSize(),
+				  .buffer = std::move(buff),
+				}
+			});
+		}
+		{
+			constexpr auto AttrSize = sizeof(decltype(*normals));
+			auto buff = ICPUBuffer::create({{AttrSize * vertexCount,IBuffer::EUF_NONE}});
+			normals = reinterpret_cast<decltype(normals)>(buff->getPointer());
+			shapes::AABB<4, int8_t> aabb;
+			aabb.maxVx = hlsl::vector<int8_t,4>(127,127,127,0);
+			aabb.minVx = -aabb.maxVx;
+			retval->setNormalView({
+				.composed = {
+					.encodedDataRange = {.s8=aabb},
+					.stride = AttrSize,
+					.format = NormalFormat,
+					.rangeFormat = IGeometryBase::EAABBFormat::S8_NORM
+				},
+				.src = {
+				  .offset = 0,
+				  .size = buff->getSize(),
+				  .buffer = std::move(buff)
+				}
+			});
+		}
+		{
+			constexpr auto AttrSize = sizeof(decltype(*uvs));
+			auto buff = ICPUBuffer::create({{AttrSize * vertexCount,IBuffer::EUF_NONE}});
+			uvs = reinterpret_cast<decltype(uvs)>(buff->getPointer());
+			shapes::AABB<4, uint8_t> aabb;
+			aabb.minVx = hlsl::vector<uint8_t, 4>(0,0,0,0);
+			aabb.maxVx = hlsl::vector<uint8_t, 4>(255,255,0,0);
+			retval->getAuxAttributeViews()->push_back({
+				.composed = {
+					.encodedDataRange = {.u8=aabb},
+					.stride = AttrSize,
+					.format = EF_R8G8_UNORM,
+					.rangeFormat = IGeometryBase::EAABBFormat::U8_NORM
+				},
+				.src = {
+				  .offset = 0,
+				  .size = buff->getSize(),
+				  .buffer = std::move(buff),
+				}
+			});
+		}
+		{
+			constexpr auto AttrSize = sizeof(decltype(*colors));
+			auto buff = ICPUBuffer::create({{AttrSize * vertexCount,IBuffer::EUF_NONE}});
+			colors = reinterpret_cast<decltype(colors)>(buff->getPointer());
+			shapes::AABB<4, uint8_t> aabb;
+			aabb.minVx = hlsl::vector<uint8_t, 4>(0,0,0,0);
+			aabb.maxVx = hlsl::vector<uint8_t, 4>(255,255,0,0);
+			retval->getAuxAttributeViews()->push_back({
+				.composed = {
+					.encodedDataRange = {.u8=aabb},
+					.stride = AttrSize,
+					.format = EF_R8G8B8A8_UNORM,
+					.rangeFormat = IGeometryBase::EAABBFormat::U8_NORM
+				},
+				.src = {
+				  .offset = 0,
+				  .size = buff->getSize(),
+				  .buffer = std::move(buff),
+				}
+			});
+		}
+	}
+
+	// fill vertices
+	{
 		for (size_t i = 0; i < vertexCount; i++)
 		{
-			tmpMem[i * vertexSize + 3 * 4 + 0] = 255;
-			tmpMem[i * vertexSize + 3 * 4 + 1] = 255;
-			tmpMem[i * vertexSize + 3 * 4 + 2] = 255;
-			tmpMem[i * vertexSize + 3 * 4 + 3] = 255;
+			colors[i] = { 255,255,255,255 };
 		}
+
 		// calculate the angle which separates all points in a circle
 		const float AngleX = 2 * core::PI<float>() / polyCountX;
 		const float AngleY = core::PI<float>() / polyCountY;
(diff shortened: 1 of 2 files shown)


src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -15,6 +15,11 @@
 namespace nbl::asset
 {
 
+static uint8_t packSnorm(float val)
+{
+	return round(hlsl::clamp(val, -1.0f, 1.0f) * 127);
+}
```
devshgraphicsprogramming: `cpp_compat/intrinsics.hlsl` has those

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -156,7 +161,7 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createCube(const h
 	{
 		const hlsl::vector<int8_t, 3> norm[6] =
 		{
-			hlsl::vector<int8_t,3>(0, 0, 127),
+			hlsl::vector<int8_t,3>(0, 0, 1),
```
devshgraphicsprogramming: huh ? you've made a bug

kevyuu: ah sorry

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -372,23 +273,123 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+
+		shapes::AABB<4,index_t> aabb;
+		aabb.minVx[0] = 0;
+		aabb.maxVx[0] = vertexCount - 1;
+		retval->setIndexView({
+			.composed = {
+				.encodedDataRange = {.u32=aabb},
+				.stride = sizeof(index_t),
+				.format = EF_R16_UINT,
+				.rangeFormat = IGeometryBase::EAABBFormat::U16
```
devshgraphicsprogramming: you defined `index_t = uint32_t` ... then format should be ~~R16 and U16 respectively~~

Edit: I meant R32 and U32 wote wrong thing by accident

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -372,23 +273,123 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
-	indices->setUsageFlags(indices->getUsageFlags() | asset::IBuffer::EUF_INDEX_BUFFER_BIT);
-	retval.indexBuffer = {0ull, std::move(indices)};
 
-	// handle vertices
+	constexpr auto NormalCacheFormat = EF_R8G8B8_SNORM;
+	constexpr auto NormalFormat = EF_R8G8B8A8_SNORM;
+
+	// Create vertex attributes with NONE usage because we have no clue how they'll be used
+	hlsl::float32_t3* positions;
+  hlsl::vector<uint8_t, 4>* normals;
```
devshgraphicsprogramming: `int8_t`

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -372,23 +273,123 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+	hlsl::vector<uint8_t, 2>* uvs;
```
devshgraphicsprogramming: thats not enough precision, you need uint16_t, see the disk code

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -398,258 +399,615 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+      {
+        // calculate points position
+
+        float32_t3 pos(static_cast<float>(cos(axz) * sinay),
+          static_cast<float>(cos(ay)),
+          static_cast<float>(sin(axz) * sinay));
+        // for spheres the normal is the position
+        core::vectorSIMDf normal(&pos.x);
+        normal.makeSafe3D();
+        const auto quantizedNormal = quantNormalCache->quantize<NormalCacheFormat>(normal);
```
devshgraphicsprogramming: can you make the quantization cache work with `hlsl::` vector types instead of `core::vectorSIMD` (I'm trying to deprecate them)

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -398,258 +399,615 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
-        vertices[i+halfIx] = vertices[i];
-        vertices[i+halfIx].pos[2] = length;
-        vertices[i+halfIx].uv[1] = 1.f;
-    }
+	using namespace hlsl;
+
+	CQuantNormalCache* const quantNormalCache = quantNormalCacheOverride == nullptr ? m_params.normalCache.get() : quantNormalCacheOverride;
+
+  const uint16_t halfIx = static_cast<uint16_t>(tesselation);
+  const uint16_t vertexCount = 2 * static_cast<uint16_t>(tesselation);
```
devshgraphicsprogramming: you may want to protect against overflow and return nullptr when uint16 is not enough for given tessellation (also take the tesselation as uint16)

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -398,258 +399,615 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+      auto buff = ICPUBuffer::create({{AttrSize * vertexCount,IBuffer::EUF_NONE}});
+      colors = reinterpret_cast<decltype(colors)>(buff->getPointer());
+      shapes::AABB<4, uint8_t> aabb;
+      aabb.minVx = hlsl::vector<uint8_t, 4>(0,0,0,0);
+      aabb.maxVx = hlsl::vector<uint8_t, 4>(255,255,0,0);
+      retval->getAuxAttributeViews()->push_back({
+        .composed = {
+          .encodedDataRange = {.u8=aabb},
+          .stride = AttrSize,
+          .format = EF_R8G8B8A8_UNORM,
```
devshgraphicsprogramming: colors should be SRGB

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -398,258 +399,615 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+        }
+      });
+    }
+    {
+      constexpr auto AttrSize = sizeof(decltype(*colors));
+      auto buff = ICPUBuffer::create({{AttrSize * vertexCount,IBuffer::EUF_NONE}});
+      colors = reinterpret_cast<decltype(colors)>(buff->getPointer());
+      shapes::AABB<4, uint8_t> aabb;
+      aabb.minVx = hlsl::vector<uint8_t, 4>(0,0,0,0);
+      aabb.maxVx = hlsl::vector<uint8_t, 4>(255,255,0,0);
```
devshgraphicsprogramming: correct this everywhere, the min and max are supposed to be a single value

Actually, since the colors of the geometries are constant, there's no point having them as attributes (we win one less place in the codebase with `video::SColor`)

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -398,258 +399,615 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+	const auto cylinderIndices = cylinder->getIndexView().src.buffer->getPointer();
+	const auto coneIndices = cone->getIndexView().src.buffer->getPointer();
+
+	const auto cylinderVertexCount = cylinder->getPositionView().getElementCount();
+	const auto coneVertexCount = cone->getPositionView().getElementCount();
+	const auto newArrowVertexCount = cylinderVertexCount + coneVertexCount;
+
+	const auto cylinderIndexCount = cylinder->getVertexReferenceCount();
+	const auto coneIndexCount = cone->getVertexReferenceCount();
+	const auto newArrowIndexCount = cylinderIndexCount + coneIndexCount;
```
devshgraphicsprogramming: Actually. make this function return a geometry collection!

Its a perfect use case! (untouched geometries, just transforms applied)

You can skip rendering the arrow in ex09 (only test with BLAS stuff in 67)

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: So I remove all video::SColor argument from create<Mesh>?

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Remove all colors from createMesh. Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: no single packSnorm. only packSnorm2x16 and packSnorm4x8

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

include/nbl/asset/utils/CGeometryCreator.h
```diff
@@ -58,11 +58,10 @@ class NBL_API2 CGeometryCreator final : public core::IReferenceCounted
 		\param colorCone color of the cone
 		\return Generated mesh.
 		*/
-		core::smart_refctd_ptr<ICPUPolygonGeometry> createArrow(const uint32_t tesselationCylinder = 4,
+		core::vector<core::smart_refctd_ptr<ICPUPolygonGeometry>> createArrow(const uint32_t tesselationCylinder = 4,
```
devshgraphicsprogramming: no, return a `ICPUGeometryCollection`

P.S. never use containers such as `vector` for cross-DLL calls (anything thats not inline)

src/nbl/asset/utils/CGeometryCreator.cpp
devshgraphicsprogramming: its still 1 on the latest commit

src/nbl/asset/utils/CGeometryCreator.cpp
devshgraphicsprogramming: still wrong, you've got 32bit index but you're saying the index buffer view fromat is 16 bit which will cause the sphere to render wrong

src/nbl/asset/utils/CGeometryCreator.cpp
devshgraphicsprogramming: ex 67 does not test the collection

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -378,23 +279,96 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+				}
+			});
+		}
+		{
+			constexpr auto AttrSize = sizeof(decltype(*uvs));
+			auto buff = ICPUBuffer::create({{AttrSize * vertexCount,IBuffer::EUF_NONE}});
+			uvs = reinterpret_cast<decltype(uvs)>(buff->getPointer());
+			shapes::AABB<4, uint16_t> aabb;
+			aabb.minVx = hlsl::vector<uint8_t, 4>(0,0,0,0);
+			aabb.maxVx = hlsl::vector<uint8_t, 4>(255,255,0,0);
```
devshgraphicsprogramming: wrong max value, might want to use `numeric_limits` or something to work that out/keep in sync

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -378,23 +279,96 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+			retval->getAuxAttributeViews()->push_back({
+				.composed = {
+					.encodedDataRange = {.u16=aabb},
+					.stride = AttrSize,
+					.format = EF_R8G8_UNORM,
+					.rangeFormat = IGeometryBase::EAABBFormat::U8_NORM
```
devshgraphicsprogramming: wrong formats, should be 16bit

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
 				// calculate texture coordinates via sphere mapping
 				// tu is the same on each level, so only calculate once
 				float tu = 0.5f;
 				//if (y==0)
 				//{
-				if (normal.Y != -1.0f && normal.Y != 1.0f)
-					tu = static_cast<float>(acos(core::clamp(normal.X / sinay, -1.0, 1.0)) * 0.5 * core::RECIPROCAL_PI<double>());
-				if (normal.Z < 0.0f)
+				if (normal.y != -1.0f && normal.y != 1.0f)
+					tu = static_cast<float>(acos(core::clamp(normal.x / sinay, -1.0, 1.0)) * 0.5 * core::RECIPROCAL_PI<double>());
```
devshgraphicsprogramming: get rid of `core::RECIPROCAL_PI` we have some `numbers.hlsl` header that gives that

do everything in `hlsl::` tgmath, also `double` overkill here, use float in the templates

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
 
-			((float*)tmpMemPtr)[0] = ((float*)oldTmpMemPtr)[0];
-			((float*)tmpMemPtr)[1] = ((float*)oldTmpMemPtr)[1];
-			((float*)tmpMemPtr)[2] = ((float*)oldTmpMemPtr)[2];
-			((float*)tmpMemPtr)[4] = 1.f;
-			((float*)tmpMemPtr)[5] = ((float*)oldTmpMemPtr)[5];
-			((uint32_t*)tmpMemPtr)[6] = ((uint32_t*)oldTmpMemPtr)[6];
-			tmpMemPtr += vertexSize;
+			positions[vertex_i] = positions[old_vertex_i];
+			uvs[vertex_i] = { 127, uvs[old_vertex_i].y };
```
devshgraphicsprogramming: again, stop using constant literals, you're making the code exteremely bug prone (you're supposed to have `uint16_t` max value here, not `int8_t` max)

src/nbl/asset/utils/CGeometryCreator.cpp
devshgraphicsprogramming: erm you're using this to encode sphere UV coords which should be `unorm16`, just use `encodePixel` or make this templated so one can't make mistakes like this with implicit conversions

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
-		BoundingBox.addInternalPoint(-radius, -radius, -radius);
-
-		// set vertex buffer
-		vtxBuf->setUsageFlags(vtxBuf->getUsageFlags() | asset::IBuffer::EUF_VERTEX_BUFFER_BIT);
-		retval.bindings[0] = { 0ull,std::move(vtxBuf) };
-		retval.indexType = asset::EIT_32BIT;
-		retval.bbox = BoundingBox;
+		vertex_i++;
+		positions[vertex_i] = { 0.f, -radius, 0.f };
+		uvs[vertex_i] = { 63, 127};
```
devshgraphicsprogramming: again, wrong values

src/nbl/asset/utils/CGeometryCreator.cpp
devshgraphicsprogramming: would actually be best to use `encodePixel` with 1.f for X or `pack<uint16_t>(1.0)` to make sure hardcoding doesn't hurt you

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
 
 		// the vertex at the top of the sphere
-		((float*)tmpMemPtr)[0] = 0.f;
-		((float*)tmpMemPtr)[1] = radius;
-		((float*)tmpMemPtr)[2] = 0.f;
-		((float*)tmpMemPtr)[4] = 0.5f;
-		((float*)tmpMemPtr)[5] = 0.f;
-		((quant_normal_t*)tmpMemPtr)[6] = quantNormalCache->quantize<EF_A2B10G10R10_SNORM_PACK32>(core::vectorSIMDf(0.f, 1.f, 0.f));
+		positions[vertex_i] = { 0.f, radius, 0.f };
+		uvs[vertex_i] = { 0, 63};
```
devshgraphicsprogramming: again wrong constant literals

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
-	retval.indexBuffer = { 0ull, std::move(idxBuf) };
-	vtxBuf->setUsageFlags(vtxBuf->getUsageFlags() | asset::IBuffer::EUF_VERTEX_BUFFER_BIT);
-	retval.bindings[0] = { 0ull, std::move(vtxBuf) };
-	retval.indexType = asset::EIT_16BIT;
-	//retval.bbox = ?;
+	using namespace hlsl;
+
+	CQuantNormalCache* const quantNormalCache = quantNormalCacheOverride == nullptr ? m_params.normalCache.get() : quantNormalCacheOverride;
+
+	const auto halfIx = static_cast<uint16_t>(tesselation);
```
devshgraphicsprogramming: take `tessellation` as a `uint16_t` then

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+					.format = EF_R8G8_UNORM,
+					.rangeFormat = IGeometryBase::EAABBFormat::U8_NORM
+				},
+				.src = {
+					.offset = 0,
+					.size = buff->getSize(),
+					.buffer = std::move(buff),
+				}
+			});
+		}
```
devshgraphicsprogramming: again format mismatch, bad constant literals, wrong UVs, make some functions to DRY your code for attribute view creation

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+	}
+
+	const float tesselationRec = core::reciprocal_approxim<float>(static_cast<float>(tesselation));
```
devshgraphicsprogramming: I want to deprecate `reciprocal_approxim` use a regular `1.f/static_cast<float>(tessellation)`

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+	const float step = 2.f * core::PI<float>() * tesselationRec;
```
devshgraphicsprogramming: again use our hlsl numbers header

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+	for (uint32_t i = 0u; i < tesselation; ++i)
+	{
+		const auto f_i = static_cast<float>(i);
+		hlsl::float32_t3 p(std::cos(f_i * step), std::sin(f_i * step), 0.f);
+		p *= radius;
+		const auto n = quantNormalCache->quantize<NormalCacheFormat>(hlsl::normalize(p));
```
devshgraphicsprogramming: you can also `p *= radius` after you quantize the normal, this way you can skip the normalization

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+
+		positions[i] = { p.x, p.y, p.z };
+		memcpy(normals + i, &n, sizeof(n));
+		uvs[i] = { packSnorm(f_i * tesselationRec), packSnorm(0.0) };
```
devshgraphicsprogramming: again, encoding to `int8_t` signed fromat for UVs

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+
+		positions[i + halfIx] = { p.x, p.y, length };
+		normals[i + halfIx] = normals[i];
+		uvs[i + halfIx] = { packSnorm(1.0f), packSnorm(0.0f) };
```
devshgraphicsprogramming: huh? why on earth are all the UVs the same at the top of the cylinder?

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
-	//vertex positions
-	for (uint32_t i = 0u; i < tesselation; i++)
+	// Create indices
+	using index_t = uint16_t;
 	{
-		core::vectorSIMDf v(std::cos(i * step), 0.0f, std::sin(i * step), 0.0f);
-		v *= radius;
+		constexpr uint32_t RowCount = 2u;
+		const auto IndexCount = 3 * tesselation;
+		const auto bytesize = sizeof(index_t) * IndexCount;
```
devshgraphicsprogramming: this is wrong, sizing is like for a cylinder, but this is a cone (only 1 row)

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
+		u2 = hlsl::normalize(hlsl::cross(u2, v0ToApex)) * angleWeight;
+
+
+		const auto baseNormal = quantNormalCache->quantize<NormalCacheFormat>(hlsl::normalize(u1 + u2));
+		memcpy(normals + i, &baseNormal, sizeof(baseNormal));
 
-		baseVertices[i].normal = quantNormalCache->quantize<EF_A2B10G10R10_SNORM_PACK32>(core::normalize(u1 + u2));
-		apexVertices[i].normal = quantNormalCache->quantize<EF_A2B10G10R10_SNORM_PACK32>(core::normalize(u1));
+		const auto apexNormal = quantNormalCache->quantize<NormalCacheFormat>(hlsl::normalize(u1));
+		memcpy(normals + apexVertexBase_i + i, &apexNormal, sizeof(apexNormal));
```
devshgraphicsprogramming: a cone can't have smooth normals that make sense, don't generate them.

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -423,239 +395,373 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createSphere(float
...
 	{
-		indices[i * 3] = firstIndexOfApexVertices + i;
-		indices[(i * 3) + 1] = firstIndexOfBaseVertices + i;
-		indices[(i * 3) + 2] = i == (tesselation - 1) ? firstIndexOfBaseVertices : firstIndexOfBaseVertices + i + 1;
+		auto& conePosition = conePositions[i];
+		core::vector3df_SIMD newPos(conePosition.x, conePosition.y, conePosition.z);
+		newPos.rotateYZByRAD(-1.5707963268);
+
+		conePosition = {newPos.x, newPos.y, newPos.z};
 	}
```
devshgraphicsprogramming: no, a geometry collection has transform matrices for each geometry, you shouldn't need to transform vertices in the actual vertex buffer

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -871,19 +977,20 @@ core::smart_refctd_ptr<ICPUPolygonGeometry> CGeometryCreator::createDisk(const f
...
-    The icosphere with N=1 (default) has 80 triangles by subdividing a triangle
-    of icosahedron into 4 triangles. If N=0, it is identical to icosahedron.
+		The icosphere with N=1 (default) has 80 triangles by subdividing a triangle
+		of icosahedron into 4 triangles. If N=0, it is identical to icosahedron.
 */
 
 class Icosphere
 {
 public:
+	using index_t = unsigned int;
```
devshgraphicsprogramming: please use sized types like `uint32_t`

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -1838,38 +1945,109 @@ class Icosphere
...
+		shapes::AABB<4,Icosphere::index_t> aabb;
+		aabb.minVx[0] = 0;
+		aabb.maxVx[0] = icosphere.getPositionCount() - 1;
+
+		static_assert(sizeof(Icosphere::index_t) == 2 || sizeof(Icosphere::index_t) == 4);
+		const auto isIndex16Bit = sizeof(Icosphere::index_t) == 2;
+
+		retval->setIndexView({
+			.composed = {
+				.encodedDataRange = {.u32=aabb},
```
devshgraphicsprogramming: set the AABB via a visitor or after you set the index view, because when format is u16, you're assining to `u32` but you should assign to `u16`

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -894,27 +1001,27 @@ class Icosphere
 
 	~Icosphere() {}
 
-	unsigned int getVertexCount() const { return (unsigned int)vertices.size() / 3; }
+	unsigned int getPositionCount() const { return (unsigned int)vertices.size() / 3; }
 	unsigned int getNormalCount() const { return (unsigned int)normals.size() / 3; }
 	unsigned int getTexCoordCount() const { return (unsigned int)texCoords.size() / 2; }
```
devshgraphicsprogramming: all these conts must be identical anyway

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -894,27 +1001,27 @@ class Icosphere
...
 	const float* getTexCoords() const { return texCoords.data(); }
 	const unsigned int* getIndices() const { return indices.data(); }
 	const unsigned int* getLineIndices() const { return lineIndices.data(); }
 
 	// for interleaved vertices: V/N/T
-	unsigned int getInterleavedVertexCount() const { return getVertexCount(); }    // # of vertices
+	unsigned int getInterleavedVertexCount() const { return getPositionCount(); }    // # of vertices
 	unsigned int getInterleavedVertexSize() const { return (unsigned int)interleavedVertices.size() * sizeof(float); }    // # of bytes
 	int getInterleavedStride() const { return interleavedStride; }   // should be 32 bytes
 	const float* getInterleavedVertices() const { return interleavedVertices.data(); }
```
devshgraphicsprogramming: the icosphere can't use interleaved vertex format, must use separated

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -1838,38 +1945,109 @@ class Icosphere
...
-
-	vertexBuffer->addUsageFlags(asset::IBuffer::EUF_VERTEX_BUFFER_BIT);
-	icosphereGeometry.bindings[0] = { 0, std::move(vertexBuffer) };
-	indexBuffer->addUsageFlags(asset::IBuffer::EUF_INDEX_BUFFER_BIT);
-	icosphereGeometry.indexBuffer = { 0, std::move(indexBuffer) };
-	icosphereGeometry.indexCount = IcosphereData.getIndexCount();
-	icosphereGeometry.indexType = EIT_32BIT;
-
-	return icosphereGeometry;
+			using uv_t = uint32_t;
```
devshgraphicsprogramming: use `uint16_t2` for clarity

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -43,25 +43,26 @@ class CDirQuantCacheBase
...
 					x = val.x;
 					y = val.y;
 					z = val.z;
 					return *this;
 				}
 
-				inline core::vectorSIMDu32 getValue() const
+        hlsl::float32_t3 getValue() const
 				{
-					return core::vectorSIMDu32(x,y,z);
```
devshgraphicsprogramming: this used to return the counterpart to `uint32_t4`, it could return `vector<uint8_t,3>` I guess, but definitely not float

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -43,25 +43,26 @@ class CDirQuantCacheBase
...
 				Vector8u3() : x(0u),y(0u),z(0u) {}
 				Vector8u3(const Vector8u3&) = default;
-				explicit Vector8u3(const core::vectorSIMDu32& val)
+				explicit Vector8u3(const hlsl::float32_t3& val)
 				{
 					operator=(val);
 				}
 
 				Vector8u3& operator=(const Vector8u3&) = default;
-				Vector8u3& operator=(const core::vectorSIMDu32& val)
```
devshgraphicsprogramming: same as https://github.com/Devsh-Graphics-Programming/Nabla/pull/894#discussion_r2213305075

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -108,16 +109,17 @@ class CDirQuantCacheBase
...
 				Vector1010102& operator=(const Vector1010102&) = default;
-				Vector1010102& operator=(const core::vectorSIMDu32& val)
+				Vector1010102& operator=(const hlsl::float32_t3& val)
 				{
-					constexpr auto storageBits = quantizationBits+1u;
-					storage = val.x|(val.y<<storageBits)|(val.z<<(storageBits*2u));
+					constexpr auto storageBits = quantizationBits + 1u;
+					hlsl::uint32_t3 u32_val = { val.x, val.y, val.z };
+					storage = u32_val.x | (u32_val.y << storageBits) | (u32_val.z << (storageBits * 2u));
 					return *this;
```
devshgraphicsprogramming, line 122: this used to use the counterpart to `uint32_t4`, it could do `vector<uint16_t,3>` I guess, but definitely not float

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -130,13 +132,13 @@ class CDirQuantCacheBase
...
-				inline core::vectorSIMDu32 getValue() const
+        hlsl::float32_t3 getValue() const
 				{
-					constexpr auto storageBits = quantizationBits+1u;
-					const core::vectorSIMDu32 mask((0x1u<<storageBits)-1u);
-					return core::vectorSIMDu32(storage,storage>>storageBits,storage>>(storageBits*2u))&mask;
+					constexpr auto storageBits = quantizationBits + 1u;
+					const auto mask = (0x1u << storageBits) - 1u;
+					return { storage & mask, (storage >> storageBits) & mask, (storage >> (storageBits * 2)) & mask};
 				}
```
devshgraphicsprogramming, line 139: same as https://github.com/Devsh-Graphics-Programming/Nabla/pull/894/files#r2213314777

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -393,32 +395,31 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
...
 			}
 
-			const core::vectorSIMDu32 xorflag((0x1u<<(quantizationBits+1u))-1u);
-			auto restoredAsVec = quantized.getValue()^core::mix(core::vectorSIMDu32(0u),xorflag,negativeMask);
-			restoredAsVec += core::mix(core::vectorSIMDu32(0u),core::vectorSIMDu32(1u),negativeMask);
-			return value_type_t<CacheFormat>(restoredAsVec&xorflag);
+			//return quantized.
+			const auto negativeMulVec = hlsl::float32_t3(negativeMask.x ? -1 : 1, negativeMask.y ? -1 : 1, negativeMask.z ? -1 : 1);
+      return value_type_t<CacheFormat>(negativeMulVec * quantized.getValue());
```
devshgraphicsprogramming: the old code used to use 2s complement conciously, now you're doing a weird denormalized int- > float -> mul -> retruncate pipeline

devshgraphicsprogramming: `getValue()` should return `uint16_t`, `uint32_t`, `uint64_t` (128bit not needed because thats not compressed in any way) representation of the raw bits storing the quantized value

the `xorflag` should probably be some constexpr static inline

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -393,32 +395,31 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
...
+			const auto vectorForDots = hlsl::normalize(value);
 
 			//
-			core::vectorSIMDf fittingVector;
-			core::vectorSIMDf floorOffset;
+			hlsl::float32_t3 fittingVector;
+			hlsl::float32_t3 floorOffset;
 			constexpr uint32_t cornerCount = (0x1u<<(dimensions-1u))-1u;
-			core::vectorSIMDf corners[cornerCount] = {};
+			hlsl::float32_t3 corners[cornerCount] = {};
```
devshgraphicsprogramming: wrong type used `float32_t3` for everything, should be `vector<float32_t,dimensions>` instead!

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -377,11 +379,11 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
...
 		
 		template<uint32_t dimensions, E_FORMAT CacheFormat>
-		value_type_t<CacheFormat> quantize(const core::vectorSIMDf& value)
+		value_type_t<CacheFormat> quantize(const hlsl::float32_t3& value)
 		{
-			const auto negativeMask = value < core::vectorSIMDf(0.0f);
+			const auto negativeMask = lessThan(value, hlsl::float32_t3(0.0f));
 
-			const core::vectorSIMDf absValue = abs(value);
+			const hlsl::float32_t3 absValue = abs(value);
```
devshgraphicsprogramming: wrong type used, `vector<float32_t,dimensions>` please

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -466,18 +467,18 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
 			};
 
 			constexpr uint32_t cubeHalfSize = (0x1u << quantizationBits) - 1u;
-			const core::vectorSIMDf cubeHalfSizeND = core::vectorSIMDf(cubeHalfSize);
+			const hlsl::float32_t3 cubeHalfSizeND = hlsl::float32_t3(cubeHalfSize);
```
devshgraphicsprogramming: use `promote<vector_type>` to turn scalars into vectors of the repeated scalar

src/nbl/asset/ECommonEnums.cpp
```diff
@@ -0,0 +1,281 @@
...
+	if constexpr(std::is_constant_evaluated())
+	{
+		for (size_t ix=0ull; ix<sizeof(size_t)*8; ix++)
+			if ((0x1ull << ix) & val) return ix;
+		return ~0u;
+	} else
+	{
+		return hlsl::findLSB(val);
+	}
+}
```
devshgraphicsprogramming: why not fix up the `hlsl::findLSB` to have the `if constexpr(std::is_constant_evaluated())` under `ifndef __HLSL_VERSION`

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Removed. use encode pixels isntead

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Fixed.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Fixed.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: So we need more vertex? We need to double the vertex right? each vertex need to have two normals for each side

src/nbl/asset/ECommonEnums.cpp
kevyuu: findLSB need to support core::bitflag::enum type as well

include/nbl/asset/IAccelerationStructure.h
```diff
@@ -23,6 +23,8 @@ namespace nbl::asset
 class IAccelerationStructure : public virtual core::IReferenceCounted
 {
 	public:
+		static constexpr inline size_t TRANSFORM_DATA_MIN_ALIGNMENT = 16;
```
devshgraphicsprogramming: don't use capital snake case, we use those for preprocessor defines, `TransformDataMinAlignment` is better as a variable name

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -5,19 +5,162 @@
...
+      if constexpr(std::is_same_v<ElementT, uint8_t>)
+      {
+				return EF_R8G8_UNORM;
+      } else
+      {
+				return EF_R16G16_UNORM;
+      }
+    }
+
+}
```
devshgraphicsprogramming: don't indent just because of namespaces

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -397,29 +413,42 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
...
+      restoredAsVec += switch_vec(hlsl::uint32_t4(0u), hlsl::uint32_t4(1u), negativeMask);
+      return value_type_t<CacheFormat>(restoredAsVec & xorflag);
 		}
 
 		template<uint32_t dimensions, uint32_t quantizationBits>
-		static inline hlsl::float32_t3 findBestFit(const hlsl::float32_t3& value)
+		static inline hlsl::vector<hlsl::float32_t, dimensions> findBestFit(const hlsl::vector<hlsl::float32_t, dimensions>& value)
 		{
 			static_assert(dimensions>1u,"No point");
 			static_assert(dimensions<=4u,"High Dimensions are Hard!");
```
devshgraphicsprogramming: you can make a typedef `using float32_tN = hlsl::vector<hlsl::float32_t,dimensions>` in the function body to reduce clutter

include/nbl/asset/utils/CDirQuantCacheBase.h
devshgraphicsprogramming: bump

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -397,29 +413,42 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
...
-      return value_type_t<CacheFormat>(negativeMulVec * quantized.getValue());
+			auto switch_vec = [](hlsl::uint32_t4 val1, hlsl::uint32_t4 val2, hlsl::bool4 mask)
+      {
+					hlsl::uint32_t4 retval;
+					retval.x = mask.x ? val2.x : val1.x;
+					retval.y = mask.y ? val2.y : val1.y;
+					retval.z = mask.z ? val2.z : val1.z;
+					retval.w = mask.w ? val2.w : val1.w;
+					return retval;
+      };
```
devshgraphicsprogramming: dont we have a mix/select in our `intrinsics.hlsl ` for this ?

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -379,11 +378,28 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
...
+        } else if constexpr (dimensions == 3)
+        {
+          return {src.x, src.y, src.z, 0};
+        } else if constexpr (dimensions == 4)
+        {
+          return {src.x, src.y, src.z, src.w};
+        }
+      };
+
+			const auto negativeMask = to_float32_t4(lessThan(value, hlsl::vector<hlsl::float32_t, dimensions>(0.0f)));
```
devshgraphicsprogramming: why is negativeMask a float32_t4 and not a vector<bool,4> ?

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -397,29 +413,42 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
...
+					retval.x = mask.x ? val2.x : val1.x;
+					retval.y = mask.y ? val2.y : val1.y;
+					retval.z = mask.z ? val2.z : val1.z;
+					retval.w = mask.w ? val2.w : val1.w;
+					return retval;
+      };
+
+      const hlsl::uint32_t4 xorflag((0x1u << (quantizationBits + 1u)) - 1u);
+      auto restoredAsVec = quantized.getValue() ^ switch_vec(hlsl::uint32_t4(0u), hlsl::uint32_t4(xorflag), negativeMask);
+      restoredAsVec += switch_vec(hlsl::uint32_t4(0u), hlsl::uint32_t4(1u), negativeMask);
```
devshgraphicsprogramming: if `negativeMask` was a `uint32_t4` you could just write `B&negativeMask` instead of `switch_vec(A,B,negativeMask` because your A is always 0

include/nbl/asset/utils/CDirQuantCacheBase.h
devshgraphicsprogramming: not even a `uint32_t4`

src/nbl/asset/utils/CGeometryCreator.cpp
```diff
@@ -741,25 +674,20 @@ core::vector<core::smart_refctd_ptr<ICPUPolygonGeometry>> CGeometryCreator::crea
...
-		conePosition = {newPos.x, newPos.y, newPos.z};
-	}
-
-	return {cylinder, cone};
+	auto collection = core::make_smart_refctd_ptr<ICPUGeometryCollection>();
+	auto* geometries = collection->getGeometries();
+	geometries->push_back({
+		.geometry = cylinder
+  });
+	const auto coneTransform = hlsl::math::linalg::rotation_mat(-1.5707963268f, hlsl::float32_t3(1.f, 0.f, 0.f));
```
devshgraphicsprogramming: use `-numbers::pi*0.5`

devshgraphicsprogramming:
Aisde from my nitpicks, I assume everything works?

include/nbl/asset/IAccelerationStructure.h
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

src/nbl/asset/utils/CGeometryCreator.cpp
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Done.

kevyuu: Mix only work for floating point. there is no select

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -502,7 +505,7 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
 
 			constexpr uint32_t cubeHalfSize = (0x1u << quantizationBits) - 1u;
 			const auto test = core::vectorSIMDf(cubeHalfSize);
-			const hlsl::vector<hlsl::float32_t, dimensions> cubeHalfSizeND = hlsl::vector<hlsl::float32_t, dimensions>(cubeHalfSize);
+			const float32_tN cubeHalfSizeND = float32_tN(cubeHalfSize);
```
devshgraphicsprogramming: promote<>

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -502,7 +505,7 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
 
 			constexpr uint32_t cubeHalfSize = (0x1u << quantizationBits) - 1u;
 			const auto test = core::vectorSIMDf(cubeHalfSize);
```
devshgraphicsprogramming: whats vectorSIMDf doing, unused var?

include/nbl/asset/utils/CDirQuantCacheBase.h
devshgraphicsprogramming: @keptsecret shouldn't our `mix(vector<T,N>,vector<T,N>,vector<bool,N>)` work with any T?

include/nbl/asset/utils/CDirQuantCacheBase.h
```diff
@@ -435,25 +437,26 @@ class CDirQuantCacheBase : public virtual core::IReferenceCounted, public impl::
 
 			// for positive number xoring with 0 keep its value
 			// for negative number we xor with all one which will flip the bits, then we add one later. Flipping the bits then adding one will turn positive number into negative number
-			auto restoredAsVec = quantized.getValue() ^ switch_vec(hlsl::uint32_t4(0u), hlsl::uint32_t4(xorflag), negativeMask);
-			restoredAsVec += switch_vec(hlsl::uint32_t4(0u), hlsl::uint32_t4(1u), negativeMask);
+			auto restoredAsVec = quantized.getValue() ^ select(hlsl::uint32_t4(0u), hlsl::uint32_t4(xorflag), negativeMask);
+			restoredAsVec += (hlsl::uint32_t4(1u) & hlsl::uint32_t4(negativeMask));
```
devshgraphicsprogramming: if bool true promotes to uint 1, then the & is not needed

kevyuu:
> btw what's the companion examples/tests PR ?

https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/199

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: oh yeah. sorry, forget to remove it

kevyuu: Removed. Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: Yeah. Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
kevyuu: What is the difference from using the constructor?

kevyuu: Done.

include/nbl/asset/utils/CDirQuantCacheBase.h
keptsecret: It should. Doesn't have the new changes on the `hlsl_bxdf` branch, but `master` right now `mix(vector<T,N>,vector<T,N>,vector<bool,N>)` should be routed to the looped version that uses `mix(T, T, bool)`

PR #903, opened 2025-07-17
Fix polygon manipulator aabb computation

kevyuu:
Fix AABB computation on cpu polygon geometry


include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -61,20 +61,48 @@ class NBL_API2 CPolygonGeometryManipulator
...
+              case EIT_16BIT:
+                vertex_i = reinterpret_cast<const uint16_t*>(indexBuffer)[i];
+                break;
+              case EIT_32BIT:
+                vertex_i = reinterpret_cast<const uint32_t*>(indexBuffer)[i];
+                break;
+              default:
+                assert(false);
+                break;
+            }
```
devshgraphicsprogramming: its better to pack up the whole `for` loop into a templated lambda, and hoist the switch above (so it dispatches a separate specialized for loop per case)

this way you avoid dealing with the switch and different inted buffer decodes at every index

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -61,20 +61,48 @@ class NBL_API2 CPolygonGeometryManipulator
...
+      if (geo->getIndexView() || geo->isSkinned())
+      {
+        auto addToAABB = [&](auto& aabb)->void
+        {
+          using aabb_t = std::remove_reference_t<decltype(aabb)>;
+          for (auto i = 0; i != geo->getIndexView().getElementCount(); i++)
+          {
+            uint32_t vertex_i;
+            const auto indexBuffer = geo->getIndexView().getPointer();
+            switch (geo->getIndexType())
```
devshgraphicsprogramming: this is meant to be for Graphics Pipelines, use non-index-view  specific methods (`decodeElement` like a normal view) to decode indices (this way you'll be able to support int8 indices

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: outdated

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -66,15 +66,61 @@ class NBL_API2 CPolygonGeometryManipulator
...
+					for (const auto& jointWeight : geo->getJointWeightViews())
+					{
+						for (auto vertex_i = 0u; vertex_i < jointWeight.indices.getElementCount(); vertex_i++)
+						{
+							hlsl::vector<uint32_t, 1> index;
+							hlsl::vector<hlsl::float32_t, 1> weight;
+							jointWeight.indices.decodeElement(vertex_i, index);
+							jointWeight.weights.decodeElement(vertex_i, weight);
+							if (weight.x != 0.f)
+								jointNodeIndexes.insert(index.x);
```
devshgraphicsprogramming: the index is the index of tthe joint/bone, NOT the vertex, its enough for you to turn `if (geo->isSkinned() && jointNodeIndexes.contains(vertex_i.x)) continue;` in the `addToAABB` into a separate lambda which checks that the `jointWeight.weights` at `vertex_i` doesn't have a non-zero weight in any of the `jointWeight` views

kevyuu: You mean don't use unordered_set, instead we iterate jointWeightViews everytime even for indexed geometry?

devshgraphicsprogramming: the data views for vertices act as SoA of attributes

so element `i` in `jointWeightViews[a].indices[i][c]` tells you the `4*a+c`-th joint ID which affects the `i`-th vertex

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -68,22 +68,20 @@ class NBL_API2 CPolygonGeometryManipulator
...
 			if (geo->getIndexView() || geo->isSkinned())
 			{
-				std::unordered_set<uint32_t> jointNodeIndexes = {};
-				if (geo->isSkinned())
+				auto isVertexSkinned = [](const ICPUPolygonGeometry* geo, uint64_t vertex_i)
 				{
-					for (const auto& jointWeight : geo->getJointWeightViews())
+					if (!geo->isSkinned()) return false;
+					const auto& jointWeightView = geo->getJointWeightViews()[vertex_i];
+					for (auto weight_i = 0u; weight_i < jointWeightView.weights.getElementCount(); weight_i++)
```
devshgraphicsprogramming: you're indexing the wrong thing, its not one view per vertex

its supposed to be
```cpp
// precompute the count outside the lambda to optimize
const auto jointViewCount = geo->isSkinned() ? geo->getJointWeightViews().size():0;

for (auto j=0; j<jointViewCount; j++)
{
   hlsl::float32_t4 weight;
   jointWeightView.weights.decodeElement(vertex_i,weight);
   if (j==jointViewCount-1)
   {
      // can also precompute active channel mask on last view and use `hlsl::any(weight<promote<float32_t4>(0.f) && activeMask)`
      for (auto c=0; c<getChannelCount(jointWeightView.weights.format); c++)
      if (weight[c]>0.f)
          return true;
   }
   else if (hlsl::any(hlsl::promote<hlsl::float32_t4>(0.f)<weight))
      return true;
}
return false;
```

devshgraphicsprogramming: if you have Maximum of 1-4 bones affecting a vertex, you only have one joint view, if you have 5-8 you have two joint views (but all except last have formats with 4 components for index and weight, while the last has whatever remains to handle non multiple of 4 max simultaneous bone affections per vertex), etc.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Fixed.

include/nbl/asset/IPolygonGeometry.h
```diff
@@ -189,6 +189,8 @@ class IPolygonGeometry : public IIndexableGeometry<BufferType>, public IPolygonG
         // SoA instead of AoS, first component is the first bone influece, etc.
         struct SJointWeight
         {
+            static const size_t JOINT_COUNT_PER_VERTEX = 4;
+
```
devshgraphicsprogramming: huh ? 

1. Misleading name, the joints per vertex is an upper bound, also it comes from format channel counts across the joint weight views
2. should be `constexpr static inline` otherwise you mess up linking
3. Wrong naming convention, capital snake case is for macros only, constants are capital starting camel case

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -65,20 +65,21 @@ class NBL_API2 CPolygonGeometryManipulator
...
 					{
-						hlsl::vector<hlsl::float32_t, 1> weight;
-						jointWeightView.weights.decodeElement(vertex_i, weight);
-						auto hasWeight = false;
-						if (weight.x != 0.f) hasWeight = true;
-						if (hasWeight) return true;
+            for (auto weight_i = 0u; weight_i < jointCountPerVertex; weight_i++)
+            {
+              hlsl::vector<hlsl::float32_t, 1> weight;
+							weightView.weights.decodeElement(jointCountPerVertex * vertex_i + weight_i, weight);
```
devshgraphicsprogramming: that's not how `decodeElement` works, please read the code
https://github.com/Devsh-Graphics-Programming/Nabla/blob/2ca41814b6c1c61a8649bfff00cb857abbb03129/include/nbl/asset/IGeometry.h#L345

the view is lkike a UTB, one element is as-if a texel, it decodes to a vector

include/nbl/asset/IPolygonGeometry.h
devshgraphicsprogramming: actually just remove it

include/nbl/asset/utils/CPolygonGeometryManipulator.h
devshgraphicsprogramming: you can just do 
```cpp
   hlsl::float32_t4 weight;
   jointWeightView.weights.decodeElement(vertex_i,weight);
   for (auto c=0; c<getChannelCount(jointWeightView.weights.format); c++)
   if (weight[c]>0.f)
       return true;
 ```
 for the inner loop if you like and call it a day

File: include/nbl/asset/utils/CPolygonGeometryManipulator.h [hunk 7]
```diff
@@ -72,14 +72,20 @@ class NBL_API2 CPolygonGeometryManipulator
 				{
 					if (!geo->isSkinned()) return false;
 					constexpr auto jointCountPerVertex = ICPUPolygonGeometry::SJointWeight::JOINT_COUNT_PER_VERTEX;
-					for (auto& weightView : geo->getJointWeightViews())
+					const auto jointViewCount = geo->getJointWeightViews().size();
```
devshgraphicsprogramming: compute this at start of function outside the lambda

include/nbl/asset/IPolygonGeometry.h
devshgraphicsprogramming: still a TODO

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -72,14 +72,20 @@ class NBL_API2 CPolygonGeometryManipulator
...
-							if (weight.x != 0.f) return true;
-            }
+						const auto& weightView = geo->getJointWeightViews()[weightView_i];
+						hlsl::float32_t4 weight;
+						weightView.weights.decodeElement(vertex_i, weight);
+						if (weightView_i == jointViewCount - 1)
+						{
+							for (auto channel_i = 0; channel_i < getFormatChannelCount(weightView.weights.composed.format); channel_i++)
+							if (weight[channel_i] > 0.f)
+								return true;
```
devshgraphicsprogramming: just do this for all views, I'm not sure I put a hard requirement that all weight and index views must have 4 component format (I just put a req that component count must match for each weight-index pair)

include/nbl/asset/IPolygonGeometry.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

PR #909, opened 2025-07-24
Added length function for emulated vectors

Przemog1:


include/nbl/builtin/hlsl/emulated/float64_t.hlsl
```diff
@@ -395,6 +395,46 @@ namespace hlsl
...
+        static this_t sqrt(this_t number)
+        {
+            // so it doesn't return NaN for -0.0
+            bool isZero = !(number.data & 0x7FFFFFFFFFFFFFFFull);
+            if (isZero)
+                return number;
+
+            bool isNegative = (number.data >> 63) > 0;
+            if (isNegative)
+                return bit_cast<this_t>(ieee754::traits<this_t>::quietNaN);
```
devshgraphicsprogramming: note that floats go from 0 to MAX, then INF,, then NaN then signed 0 , then -INF, then NaN

you can test if the `number` is larger than max, then check for INF and signed 0 (if zero or inf, return number again) then return NaN for everything else

devshgraphicsprogramming: you can skip caring about `FastMath` and just make fast math behave correctly in this one case

include/nbl/builtin/hlsl/emulated/float64_t.hlsl
```diff
@@ -395,6 +395,46 @@ namespace hlsl
...
+            // find square root initial guess using the fast inverse square root algorithm
+            nbl::hlsl::emulated_float64_t<true, true> invSquareRoot = number;
+            {
+                int64_t i = 0x5fe6eb50c7b537a9ull - (number.data >> 1);
+                invSquareRoot.data = i;
+
+                nbl::hlsl::emulated_float64_t<true, true> threeHalfs = emulated_float64_t<true, true>::create(1.5);
+                nbl::hlsl::emulated_float64_t<true, true> x2 = number * emulated_float64_t<true, true>::create(0.5);
+                invSquareRoot = invSquareRoot * (threeHalfs - (x2 * invSquareRoot * invSquareRoot));
+            }
```
devshgraphicsprogramming: use fp32 square root for the initial guess, I was wrong to instruct you to use fast inverse sqrt

include/nbl/builtin/hlsl/emulated/float64_t.hlsl
```diff
@@ -395,6 +395,46 @@ namespace hlsl
...
+
+            // find sqrt approximation using the Newton-Raphson method
+            nbl::hlsl::emulated_float64_t<true, true> squareRoot = nbl::hlsl::emulated_float64_t<true, true>::create(1.0) / invSquareRoot;
+            const int Iterations = 5;
+            for (int i = 0; i < Iterations; ++i)
+            {
+                squareRoot = nbl::hlsl::emulated_float64_t<true, true>::create(0.5) * (squareRoot + number / squareRoot);
+            }
```
devshgraphicsprogramming: is there a way to get the square root via iteration without relying on divisions?

if not, then its better to call into fp64 emulated inverse sqrt, do the iterations there using newton raphson (without division they're much much cheaper) and then take a reciprocal of the accurate fp64 result at the end


And if not fast math you can do 1 or 2 iterations of newton raphson with division to improve the square root

devshgraphicsprogramming:
looks good

PR #905, opened 2025-07-18
Rt pipeline asset conversion

kevyuu:
CPU Ray Tracing pipeline conversion to GPU Ray Tracing pipeline

PR #911, opened 2025-08-11
Rt pipeline asset conversion

devshgraphicsprogramming:
## Description

Continuation of #905 905

## Testing 

Using CI

### DIFF
--- a/include/nbl/asset/ICPUAccelerationStructure.h
+++ b/include/nbl/asset/ICPUAccelerationStructure.h
@@ -16,7 +16,7 @@ namespace nbl::asset
 class ICPUBottomLevelAccelerationStructure final : public IPreHashed, public IBottomLevelAccelerationStructure
 {
 	public:
-		static inline bool validBuildFlags(const core::bitflag<BUILD_FLAGS> flags) {return validBuildFlags(flags);}
+		static inline bool validBuildFlags(const core::bitflag<BUILD_FLAGS> flags) {return IBottomLevelAccelerationStructure::validBuildFlags(flags);}
 
 		//
 		inline ICPUBottomLevelAccelerationStructure() = default;
@@ -283,7 +283,7 @@ class ICPUTopLevelAccelerationStructure final : public IAsset, public ITopLevelA
 		using blas_ref_t = core::smart_refctd_ptr<ICPUBottomLevelAccelerationStructure>;
 
 	public:
-		static inline bool validBuildFlags(const core::bitflag<BUILD_FLAGS> flags) {return validBuildFlags(flags);}
+		static inline bool validBuildFlags(const core::bitflag<BUILD_FLAGS> flags) {return ITopLevelAccelerationStructure::validBuildFlags(flags);}
 
 		//
 		ICPUTopLevelAccelerationStructure() = default;


include/nbl/asset/ICPURayTracingPipeline.h
```diff
@@ -112,18 +127,19 @@ class ICPURayTracingPipeline final : public ICPUPipeline<IRayTracingPipeline<ICP
...
-            for (const auto& closestHitInfo : self->m_hitGroups.closestHits) if (!visit(closestHitInfo.shader.get())) return;
-            for (const auto& intersectionInfo : self->m_hitGroups.intersections) if (!visit(intersectionInfo.shader.get())) return;
-            for (const auto& callableInfo : self->m_callables) if(!visit(callableInfo.shader.get())) return;
+            if (!visit(m_layout.get())) return;
+            if (!visit(m_raygen.shader.get())) return;
+            for (const auto& missInfo : m_misses) if (!visit(missInfo.shader.get())) return;
+            for (const auto& anyHitInfo : m_hitGroups.anyHits) if (!visit(anyHitInfo.shader.get())) return;
+            for (const auto& closestHitInfo : m_hitGroups.closestHits) if (!visit(closestHitInfo.shader.get())) return;
+            for (const auto& intersectionInfo : m_hitGroups.intersections) if (!visit(intersectionInfo.shader.get())) return;
+            for (const auto& callableInfo : m_callables) if(!visit(callableInfo.shader.get())) return;
```
devshgraphicsprogramming: AFAIK we can have null shaders here, only raygen is required non-null

note that we're not allowed to call `visit` on null pointers

see how descriptor set handles potentially null children assets https://github.com/Devsh-Graphics-Programming/Nabla/blob/5de6b84b35226333db92b6da17add9d045d847ca/include/nbl/asset/ICPUDescriptorSet.h#L129

devshgraphicsprogramming:
@kevyuu what's the companion PR in https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pulls ?

include/nbl/asset/ICPURayTracingPipeline.h
kevyuu: Done.

kevyuu:
> @kevyuu what's the companion PR in https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pulls ?

https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/210

include/nbl/video/IGPUComputePipeline.h
```diff
@@ -74,6 +74,10 @@ class IGPUComputePipeline : public IGPUPipeline<asset::IComputePipeline<const IG
                 return {};
             }
 
+            inline core::bitflag<FLAGS>& getFlags() { return flags; }
```
devshgraphicsprogramming, line 77: ~~why are we returning the flags as mutable on an IGPU object?~~

I only just figured out this is a member of the creation struct, just asking what's wrong with using `.cached.flags` throughout the codebase ?

include/nbl/video/IGPUGraphicsPipeline.h
```diff
@@ -87,6 +87,10 @@ class IGPUGraphicsPipeline : public IGPUPipeline<asset::IGraphicsPipeline<const
                 return stages;
             }
 
+            inline core::bitflag<FLAGS>& getFlags() { return flags; }
```
devshgraphicsprogramming, line 90: ~~again, only ICPU is allowed to do that~~

I only just figured out this is a member of the creation struct, just asking what's wrong with using `.cached.flags` throughout the codebase ?

include/nbl/video/IGPURayTracingPipeline.h
```diff
@@ -137,6 +135,10 @@ class IGPURayTracingPipeline :  public IGPUPipeline<asset::IRayTracingPipeline<c
                 return stages;
             }
 
+            inline core::bitflag<FLAGS>& getFlags() { return cached.flags; }
```
devshgraphicsprogramming, line 138: ~~again, IGPU is not allowed to do this~~

devshgraphicsprogramming: hmmm okay I see what you did, you added to `SCreationParams`, which will be returned as `const&` 

btw why add a getter when the `cached` member is just there which we can always access? seems like pointless redundant code.

Was this just to avoid writing `.cached.flags` and do `.getFlags()` instead?

src/nbl/video/utilities/CAssetConverter.cpp
```diff
@@ -1104,6 +1129,11 @@ class HashVisit : public CAssetConverter::CHashCache::hash_impl_base
 					assert(hlsl::bitCount(stage) == 1);
 					hasher << stage;
 					hasher << arg0.requiredSubgroupSize;
+					if constexpr (std::is_same_v<AssetT, ICPURayTracingPipeline>)
+					{
+						const auto groupIndex = std::get<2>(argTuple);
+						hasher << groupIndex;
+					}
```
devshgraphicsprogramming, line 1136: not sure you need this cause shaders will be hashed in the same order (from lowest index in group to highest) and you do a `nullOptional` which inserts a `\0` whenever a null shader is encountered

Also the numbers of shader groups are recorded too, so this "should" make a unique input to the hashing function

 but it can't harm you

src/nbl/video/utilities/CAssetConverter.cpp
```diff
@@ -1370,6 +1400,25 @@ bool CAssetConverter::CHashCache::hash_impl::operator()(lookup_t<ICPUComputePipe
...
+	AssetVisitor<HashVisit<ICPURayTracingPipeline>> visitor = {
+		*this,
+		{asset,static_cast<const PatchOverride*>(patchOverride)->uniqueCopyGroupID},
+		*lookup.patch
+	};
+	if (!visitor())
+		return false;
+	const auto& params = asset->getCachedCreationParams();
+	hasher << params.maxRecursionDepth;
+	hasher << params.dynamicStackSize;
```
devshgraphicsprogramming, line 1419: forgot about flags

src/nbl/video/utilities/CAssetConverter.cpp
devshgraphicsprogramming: I think you can safely comment out

PR #915, opened 2025-08-19
List indexing polygon conversion

kevyuu:
Create a method in polygon manipulator that turn any non triangle list geometry to triangle list geometry.


include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -81,6 +81,81 @@ class NBL_API2 CPolygonGeometryManipulator
 				const_cast<IGeometryBase::SAABBStorage&>(geo->getAABBStorage()) = computeAABB(geo);
 		}
 
+		static inline core::smart_refctd_ptr<ICPUPolygonGeometry> createTriangleListIndexing(ICPUPolygonGeometry* geo, bool makeNewMesh = false)
```
devshgraphicsprogramming: Take the input as const pointer, and use clone method to make the new mesh then replace the vertex and index data on that

devshgraphicsprogramming: basically make the function behave as if `makeNewMesh==true` always and take the input geometry as a const pointer

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -81,6 +81,81 @@ class NBL_API2 CPolygonGeometryManipulator
...
+			} else
+			{
+				outGeometry = core::smart_refctd_ptr<ICPUPolygonGeometry>(geo);
+			}
+
+			if (indexing && indexing->knownTopology() == EPT_TRIANGLE_LIST) 
+				return outGeometry;
+
+			auto* outGeo = outGeometry.get();
+			auto indexBuffer = ICPUBuffer::create({ primCount * 3 * indexSize, IBuffer::EUF_INDEX_BUFFER_BIT });
```
devshgraphicsprogramming: use `indexing->degree()` throughout the function instead of the constant `3`

devshgraphicsprogramming: copy the exact creation parameters of the old index buffer (if there is one) so the usages remain the same and **just change the size** before creating the new index buffer with those creation parameters, if there was no index buffer on the old mesh (IOTA) then what you have currently is the correct defaults

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -81,6 +81,81 @@ class NBL_API2 CPolygonGeometryManipulator
...
+				},
+				.src = {
+					.offset = 0,
+					.size = indexBuffer->getSize(),
+					.buffer = std::move(indexBuffer)
+				}
+			};
+
+			if (indexSize == 2)
+			{
```
devshgraphicsprogramming: being pendantic, switch with an assert in the default case would have been better

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -81,6 +81,81 @@ class NBL_API2 CPolygonGeometryManipulator
...
+
+				indexView.composed.encodedDataRange.u32.minVx[0] = 0;
+				indexView.composed.encodedDataRange.u32.maxVx[0] = maxIndex;
+				indexView.composed.format = EF_R32_UINT;
+				indexView.composed.rangeFormat = IGeometryBase::EAABBFormat::U32;
+			}
+			 
+			outGeo->setIndexing(IPolygonGeometryBase::TriangleList());
+			outGeo->setIndexView(std::move(indexView));
+			recomputeContentHashes(outGeo);
```
devshgraphicsprogramming: Only recompute the content hash of the index view

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -81,6 +81,81 @@ class NBL_API2 CPolygonGeometryManipulator
...
+
+				indexView.composed.encodedDataRange.u16.minVx[0] = 0;
+				indexView.composed.encodedDataRange.u16.maxVx[0] = maxIndex;
+				indexView.composed.format = EF_R16_UINT;
+				indexView.composed.rangeFormat = IGeometryBase::EAABBFormat::U16;
+
+			} else
+			{
+				IPolygonGeometryBase::IIndexingCallback::SContext<uint32_t> context{
+					.indexBuffer = outGeo->getIndexView().getPointer(),
```
devshgraphicsprogramming: the `indexBuffer` in the context is the INPUT, so should be `geo->` not `outGeo->`

It probably works regardless because `outGeo`'s index buffer is same as `geo` until you replace it

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -81,6 +81,81 @@ class NBL_API2 CPolygonGeometryManipulator
...
+					.indexSize = indexSize,
+					.beginPrimitive = 0,
+					.endPrimitive = primCount,
+					.out = indexBufferPtr,
```
devshgraphicsprogramming: there's a slight issue

the `SContext::indexSize` is supposed to match the input index buffer.

while the `SContext<OutT>` has `outT` as the output type

so one tiny change needed `.indexSize = geo->getIndexView().composed.getStride()` intialization instead

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -116,42 +117,49 @@ class NBL_API2 CPolygonGeometryManipulator
...
+					indexing->operator()(context);
+
+					indexView.composed.encodedDataRange.u32.minVx[0] = 0;
+					indexView.composed.encodedDataRange.u32.maxVx[0] = maxIndex;
+					indexView.composed.format = EF_R32_UINT;
+					indexView.composed.rangeFormat = IGeometryBase::EAABBFormat::U32;
+					break;
+				}
+				default:
+					assert(false);
```
devshgraphicsprogramming, line 158: needs a return `nullptr` after

devshgraphicsprogramming: cause non-Debug builds will never assert

PR #918, opened 2025-08-27
Iridescent fresnel, bxdfs

keptsecret:


include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -366,22 +367,43 @@ struct Conductor
...
+        const T etaLen2 = eta2 + etak2;
+        assert(hlsl::all(etaLen2 > hlsl::promote<T>(hlsl::exp2<scalar_type>(-numeric_limits<scalar_type>::digits))));
+        T t1 = etaLen2 * cosTheta_2;
+        const T etaCosTwice = eta * cosTheta * scalar_type(2.0);
+
+        const T rs_common = etaLen2 + cosTheta_2;
+        Rs = (rs_common - etaCosTwice) / (rs_common + etaCosTwice);
+        const T rp_common = t1 + hlsl::promote<T>(1.0);
+        Rp = (rp_common - etaCosTwice) / (rp_common + etaCosTwice);
+    }
```
devshgraphicsprogramming: make a new struct for this e.g. `ThinFilmConductor`, its separate functionality, same for the dielectric

include/nbl/builtin/hlsl/bxdf/reflection/iridescent.hlsl
```diff
@@ -0,0 +1,211 @@
...
+            f.absCosTheta = cache.getLdotH();
+            const spectral_type reflectance = f();
+            quo = reflectance * G2_over_G1;
+        }
+
+        return quotient_pdf_type::create(quo, _pdf);
+    }
+
+    SCookTorrance<Config, ndf_type, fresnel_type, measure_transform_type> __base;
+};
```
devshgraphicsprogramming: you could have just added a Delta NDF, and then used a CookTorrance with Iridescent Fresnel

include/nbl/builtin/hlsl/colorspace.hlsl
```diff
@@ -0,0 +1,198 @@
...
+namespace hlsl
+{
+namespace colorspace
+{
+
+struct colorspace_base
+{
+    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_R = 580.0f;
+    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_G = 550.0f;
+    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_B = 450.0f;
```
devshgraphicsprogramming: noooo, different colorspaces have different wavelengths of the primaries

include/nbl/builtin/hlsl/colorspace.hlsl
```diff
@@ -0,0 +1,198 @@
...
+    }
+    static float32_t3 FromXYZ(float32_t3 val) { return hlsl::mul(FromXYZ(), val); }
+
+    static float32_t3x3 ToXYZ()
+    {
+        return float32_t3x3(
+            float32_t3(0.412391f, 0.357584f, 0.180481f),
+            float32_t3(0.212639f, 0.715169f, 0.072192f),
+            float32_t3(0.019331f, 0.119195f, 0.950532f)
+        );
```
devshgraphicsprogramming: why are you copying literals ? return the encode/decode matrices we already have!

include/nbl/builtin/hlsl/colorspace/decodeCIEXYZ.hlsl
```diff
@@ -1,79 +0,0 @@
...
-namespace colorspace
-{
-namespace decode
-{
-
-NBL_CONSTEXPR float32_t3x3 XYZtoscRGB = float32_t3x3(
-    float32_t3( 3.240970f, -1.537383f, -0.498611f),
-    float32_t3(-0.969244f,  1.875968f,  0.041555f),
-    float32_t3( 0.055630f, -0.203977f,  1.056972f)
-);
```
devshgraphicsprogramming, line 24: don't delete these, they were useful as `static const` globals

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -451,6 +490,140 @@ struct DielectricFrontFaceOnly
...
+        return xyz / scalar_type(1.0685e-7);
+    }
+
+    T operator()()
+    {
+        const vector_type wavelengths = vector_type(colorspace::scRGB::wavelength_R, colorspace::scRGB::wavelength_G, colorspace::scRGB::wavelength_B);
+
+        vector_type eta12 = ior2/ior1;
+        vector_type eta23 = ior3/ior2;
+        vector_type etak23 = iork3/ior2;
```
devshgraphicsprogramming: can you precompute those and store as members?

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
keptsecret: I will note that it's the same calculation all the way until the end where the regular fresnel just returns `(Rs + Rp) * 0.5` instead of `inout Rs, Rp`

devshgraphicsprogramming: then return `complex_t<T>` and treat Rs as Real and Rp as imag, or make a struct to hold Polarized Reflectance and DRY the code

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -635,12 +645,32 @@ struct Iridescent
...
+    }
+
     scalar_type Dinc;       // thickness of thin film in nanometers, rec. 100-25000nm
-    vector_type ior1;       // usually air (1.0)
-    vector_type ior2;       // thin-film index
-    vector_type ior3;       // complex conductor index, k==0 makes dielectric
-    vector_type iork3;
+    vector_type eta12;      // outside (usually air 1.0) -> thin-film IOR
+    vector_type eta23;      // thin-film -> base material IOR
+    vector_type etak23;     // thin-film -> complex component, k==0 makes dielectric
```
devshgraphicsprogramming: what are the pros and const  of specializing and not having this member when you don't support Transmission

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -552,26 +565,23 @@ struct Iridescent
         return xyz / scalar_type(1.0685e-7);
     }
 
-    T operator()()
+    T operator()(const scalar_type clampedCosTheta /* LdotH */)
```
devshgraphicsprogramming: I'd remove the `LdotH` comment, fresnels are always evaluated with `VdotH`

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -510,13 +510,26 @@ struct Dielectric
...
-template<typename T NBL_PRIMARY_REQUIRES(concepts::FloatingPointLikeVectorial<T>)
+template<typename T, bool SupportsTransmission NBL_PRIMARY_REQUIRES(concepts::FloatingPointLikeVectorial<T>)
 struct Iridescent
 {
+    using this_t = Iridescent<T,SupportsTransmission>;
     using scalar_type = typename vector_traits<T>::scalar_type;
     using monochrome_type = vector<scalar_type, 1>;
     using vector_type = T;  // assert dim==3?
 
+    static this_t create(scalar_type Dinc, vector_type ior1, vector_type ior2, vector_type ior3, vector_type iork3)
```
devshgraphicsprogramming: make a creation struct, too many parameters to a function with same type, easy to mess up

include/nbl/builtin/hlsl/colorspace.hlsl
```diff
@@ -16,104 +18,89 @@ namespace colorspace
...
 struct colorspace_base
 {
-    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_R = 580.0f;
-    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_G = 550.0f;
-    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_B = 450.0f;
+    // default CIE RGB primaries wavelengths
+    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_R = 700.0f;
+    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_G = 546.1f;
+    NBL_CONSTEXPR_STATIC_INLINE float32_t wavelength_B = 435.8f;
 };
```
devshgraphicsprogramming: don't even, each colorspace just needs to provide them, end of

include/nbl/builtin/hlsl/bxdf/transmission/iridescent.hlsl
```diff
@@ -0,0 +1,212 @@
...
+    NBL_BXDF_CONFIG_ALIAS(anisocache_type, Config);
+    using brdf_type = reflection::SGGXIsotropic<Config>;
+
+    using ndf_type = ndf::GGX<scalar_type, false>;
+    using fresnel_type = fresnel::Iridescent<spectral_type>;
+    using measure_transform_type = ndf::SDualMeasureQuant<scalar_type,true,ndf::MTT_REFLECT_REFRACT>;
+
+    NBL_CONSTEXPR_STATIC_INLINE BxDFClampMode _clamp = BxDFClampMode::BCM_ABS;
+
+    struct SCreationParams
```
devshgraphicsprogramming: yeah special BSDF and BRDF not even needed for this

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -188,11 +188,14 @@ NBL_CONCEPT_END(
...
     ((NBL_CONCEPT_REQ_TYPE)(T::scalar_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::vector3_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::spectral_type))
     ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((iso.getV()), ::nbl::hlsl::is_same_v, typename T::ray_dir_info_type))
     ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((iso.getN()), ::nbl::hlsl::is_same_v, typename T::vector3_type))
     ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((iso.getNdotV(clampMode)), ::nbl::hlsl::is_same_v, typename T::scalar_type))
     ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((iso.getNdotV2()), ::nbl::hlsl::is_same_v, typename T::scalar_type))
     ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((iso.getPathOrigin()), ::nbl::hlsl::is_same_v, PathOrigin))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((iso.getLuminosityContributionHint()), ::nbl::hlsl::is_same_v, typename T::spectral_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((iso.getPrefixThroughputWeights()), ::nbl::hlsl::is_same_v, typename T::spectral_type))
```
devshgraphicsprogramming: I'd only have `getLuminosityContributionHint`

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -230,11 +237,20 @@ struct SIsotropic
...
+        return prefixThroughputWeights / math::lpNorm<spectral_type,1>(prefixThroughputWeights);
+    }
 
     RayDirInfo V;
     vector3_type N;
     scalar_type NdotV;
     scalar_type NdotV2;
+
+    spectral_type luminosityContributionHint;
+    spectral_type throughputWeights;    // product of all quotients so far
```
devshgraphicsprogramming: you only need `luminosityContributionHint`, so you're not doing the abs sum and div on every invocation

everything multiplied up to this point basically

include/nbl/builtin/hlsl/bxdf/ndf/beckmann.hlsl
```diff
@@ -337,8 +335,13 @@ struct Beckmann
     quant_type DG1(NBL_CONST_REF_ARG(dg1_query_type) query, NBL_CONST_REF_ARG(quant_query_type) quant_query, NBL_CONST_REF_ARG(LS) _sample, NBL_CONST_REF_ARG(Interaction) interaction, NBL_REF_ARG(bool) isInfinity)
     {
         scalar_type D = query.getNdf();
-        scalar_type dg1 = query.getNdf() / (scalar_type(1.0) + query.getLambdaV());
-        isInfinity = D == bit_cast<scalar_type>(numeric_limits<scalar_type>::infinity);
+        isInfinity = hlsl::isinf(D);
+        if (isInfinity)
+        {
+            quant_type dmq;
+            return dmq;
```
devshgraphicsprogramming, line 344: you're returning an uninitialized variable, the values will be bogus

include/nbl/builtin/hlsl/bxdf/ndf/ggx.hlsl
```diff
@@ -282,10 +277,13 @@ struct GGX
     quant_type DG1(NBL_CONST_REF_ARG(dg1_query_type) query, NBL_CONST_REF_ARG(quant_query_type) quant_query, NBL_CONST_REF_ARG(LS) _sample, NBL_CONST_REF_ARG(Interaction) interaction, NBL_REF_ARG(bool) isInfinity)
     {
         scalar_type D = query.getNdfwoNumerator();
-        scalar_type dg1_over_2NdotV = query.getNdfwoNumerator() * query.getG1over2NdotV();
-        isInfinity = D == bit_cast<scalar_type>(numeric_limits<scalar_type>::infinity);
+        isInfinity = hlsl::isinf(D);
         quant_type dmq;
-        dmq.microfacetMeasure = scalar_type(2.0) * interaction.getNdotV(_clamp) * dg1_over_2NdotV;
+        if (isInfinity)
+            return dmq;
```
devshgraphicsprogramming, line 284: again, undef var

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -155,13 +102,46 @@ struct SCookTorrance
...
+    }
+    template<class Interaction=conditional_t<IsAnisotropic,anisotropic_interaction_type,isotropic_interaction_type>,
+            class MicrofacetCache=conditional_t<IsAnisotropic,anisocache_type,isocache_type>, typename C=bool_constant<IsBSDF> >
+    static enable_if_t<C::value && IsBSDF, bool> checkValid(NBL_CONST_REF_ARG(fresnel_type) orientedFresnel, NBL_CONST_REF_ARG(sample_type) _sample, NBL_CONST_REF_ARG(Interaction) interaction, NBL_CONST_REF_ARG(MicrofacetCache) cache)
+    {
+        fresnel::OrientedEtas<monochrome_type> orientedEta = fresnel::OrientedEtas<monochrome_type>::create(scalar_type(1.0), hlsl::promote<monochrome_type>(orientedFresnel.getRefractionOrientedEta()));
+        return cache.isValid(orientedEta);
+    }
+
+    bool dotIsUnity(const vector3_type a, const vector3_type b)
```
devshgraphicsprogramming: prefix the methods with `__` to indicate privateness

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -178,7 +158,7 @@ struct SCookTorrance
 
         overwriteDG<Interaction, MicrofacetCache>(DG, ndf, gq, qq, _sample, interaction, cache, isInfinity);
 
-        if (isInfinity)
+        if (isInfinity) // after all calls setting DG, allows compiler to throw away calls to ndf.D if using overwriteDG
```
devshgraphicsprogramming: better comment, you need to specifically call out why its not combined with the `!isInfinity` above the `overwriteDG`

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming: also use `NBL_FUNC_REQUIRES` instead of `enable_if_t` on the return type

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -325,13 +308,14 @@ struct SCookTorrance
...
-        fresnel_type _f = impl::getOrientedFresnel<fresnel_type, IsBSDF>::__call(fresnel, interaction.getNdotV());
+        fresnel_type _f = getOrientedFresnel(fresnel, interaction.getNdotV());
         quant_query_type qq = impl::quant_query_helper<ndf_type, fresnel_type, IsBSDF>::template __call<Interaction, MicrofacetCache>(ndf, _f, interaction, cache);
         quant_type DG1 = ndf.template DG1<sample_type, Interaction>(dq, qq, _sample, interaction, isInfinity);
 
         NBL_IF_CONSTEXPR(IsBSDF)
         {
-            const scalar_type reflectance = _f(hlsl::abs(cache.getVdotH()))[0];
+            spectral_type prefixThroughputWeights = interaction.getPrefixThroughputWeights();
+            const scalar_type reflectance = hlsl::dot(impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(hlsl::abs(cache.getVdotH()))), prefixThroughputWeights);
```
devshgraphicsprogramming: ok you don't want to promote a 1D value to 3D to do a dot product which you know weights to 1.f

You want to do an
```cpp
fresnel_type::vector_type fresnelValue = _f(hlsl::abs(cache.getVdotH()))[0];
NBL_IF_CONSTEXPR(fresnel_type::monochrome)
{
   reflectance = fresnelValue;
}
else
{
   reflectance = hlsl::dot(interaction.getPrefixThroughputWeights(),fresnelValue);
}
```

Also something somewhere (the Config) needs to check that the Fresnel `operator()`'s return type is either monochrome or the same as Interaction's `spectral_type`

Finally you probably need this commonalized into some pseudo-private method for the BSDF cook torrance, since you need it for:
- generate to decide reflection vs refraction
- pdf to compute the pdf
- quotient_and_pdf to compute both

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -378,7 +362,13 @@ struct SCookTorrance
 
         spectral_type quo;
         NBL_IF_CONSTEXPR(IsBSDF)
-            quo = hlsl::promote<spectral_type>(G2_over_G1);
+        {
+            spectral_type prefixThroughputWeights = interaction.getPrefixThroughputWeights();
+            spectral_type reflectance = impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(hlsl::abs(cache.getVdotH())));
+            const scalar_type scaled_reflectance = hlsl::dot(reflectance, prefixThroughputWeights);
+            quo = hlsl::mix(reflectance / scaled_reflectance,
+                    (hlsl::promote<spectral_type>(1.0) - reflectance) / (scalar_type(1.0) - scaled_reflectance), cache.isTransmission()) * G2_over_G1;
```
devshgraphicsprogramming: btw fun fact, you can do `fresnelVal = mix(reflectance,hlsl::promote<fresnel_type::vector_type>(1.0)-reflectance,cache.isTransmission())` first, and then just compute `fresnelVal/dot(lumaCoeffs,fresnelVal)`

because of the way everything is a linear combo and weights add up to 1

P.S. Obivously don't do the division and dot product if Fresnel is monochrome https://github.com/Devsh-Graphics-Programming/Nabla/pull/918#discussion_r2498628665

include/nbl/builtin/hlsl/colorspace.hlsl
devshgraphicsprogramming: the base inheritance makes no sense

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -263,50 +271,25 @@ struct SCookTorrance
         {
             assert(NdotV*VdotH >= scalar_type(0.0));
         }
-        const scalar_type reflectance = _f(hlsl::abs(VdotH))[0];
+
+        spectral_type prefixThroughputWeights = interaction.getPrefixThroughputWeights();
+        const scalar_type reflectance = hlsl::dot(impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(hlsl::abs(VdotH))), prefixThroughputWeights);
```
devshgraphicsprogramming: see the other comment about avoiding dot product when fresnel is monochrome and commonalizing

include/nbl/builtin/hlsl/math/angle_adding.hlsl
```diff
@@ -21,53 +21,47 @@ struct sincos_accumulator
...
+        const T sinB = runningSum.imag();
+        const bool reverse = abs<T>(min<T>(a, cosB)) > max<T>(a, cosB);
+        const T c = a * cosB - sinA * sinB;
+        const T d = sinA * cosB + a * sinB;
 
         runningSum.real(ieee754::flipSign<T>(c, reverse));
-        runningSum.imag(hlsl::mix(bias, bias + numbers::pi<T>, reverse));
+        runningSum.imag(ieee754::flipSign<T>(d, reverse));
+
+        wraparound += hlsl::mix(0u, 1u, reverse);
```
devshgraphicsprogramming: i think `if (reverse) wraparound++;` would work more efficiently, also why not rename `reverse` into `overflow`

include/nbl/builtin/hlsl/math/angle_adding.hlsl
```diff
@@ -21,53 +21,47 @@ struct sincos_accumulator
...
-        const bool ABltminusC = cosSumAB < (-cosC);
-        const bool ABltC = cosSumAB < cosC;
-        // apply triple angle formula
-        const T absArccosSumABC = acos<T>(clamp<T>(cosSumAB * cosC - (cosA * sinB + sinA * cosB) * sinC, T(-1.0), T(1.0)));
-        return ((AltminusB ? ABltC : ABltminusC) ? (-absArccosSumABC) : absArccosSumABC) + ((AltminusB || ABltminusC) ? numbers::pi<T> : (-numbers::pi<T>));
+        return acos<T>(runningSum.real()) + wraparound * numbers::pi<T>;
     }
 
     complex_t<T> runningSum;
+    uint16_t wraparound;
```
devshgraphicsprogramming: comment that it counts in `pi` (half revolutions)

include/nbl/builtin/hlsl/math/angle_adding.hlsl
```diff
@@ -21,53 +21,47 @@ struct sincos_accumulator
...
 
-    static this_t create(T cosA)
+    static this_t create(T cosA, T sinA)
     {
         this_t retval;
-        retval.runningSum = complex_t<T>::create(cosA, T(0));
+        retval.runningSum = complex_t<T>::create(cosA, sinA);
+        // retval.runningSum.real(cosA);
+        // retval.runningSum.imag(sinA);
+        retval.wraparound = 0u;
```
devshgraphicsprogramming: fun fact, add this to your unit test
```cpp
accumulator_t::create(cos(6),sin(6)).getSumofArccos() >5.99
```

you should be setting `wraparound = 1` whenever sine here is negative

include/nbl/builtin/hlsl/math/angle_adding.hlsl
```diff
@@ -21,53 +21,47 @@ struct sincos_accumulator
...
+    void addCosine(T cosA, T sinA)
     {
-        const T bias = biasA + runningSum.imag();
         const T a = cosA;
-        const T b = runningSum.real();
-        const bool reverse = abs<T>(min<T>(a, b)) > max<T>(a, b);
-        const T c = a * b - sqrt<T>((T(1.0) - a * a) * (T(1.0) - b * b));
+        const T cosB = runningSum.real();
+        const T sinB = runningSum.imag();
+        const bool reverse = abs<T>(min<T>(a, cosB)) > max<T>(a, cosB);
```
devshgraphicsprogramming: write down the logic for how this got derived, summarize https://www.linkedin.com/pulse/spherical-triangle-sampling-only-onearccos-call-matt-kielan/?trackingId=9vN%2F1ttORoKiriQfFkC0Ug%3D%3D

include/nbl/builtin/hlsl/math/angle_adding.hlsl
devshgraphicsprogramming: or assert that `sinA>=0`

include/nbl/builtin/hlsl/math/angle_adding.hlsl
```diff
@@ -21,53 +21,47 @@ struct sincos_accumulator
...
-    static T getArccosSumofABC_minus_PI(T cosA, T cosB, T cosC, T sinA, T sinB, T sinC)
-    {
-        const bool AltminusB = cosA < (-cosB);
-        const T cosSumAB = cosA * cosB - sinA * sinB;
-        const bool ABltminusC = cosSumAB < (-cosC);
-        const bool ABltC = cosSumAB < cosC;
-        // apply triple angle formula
-        const T absArccosSumABC = acos<T>(clamp<T>(cosSumAB * cosC - (cosA * sinB + sinA * cosB) * sinC, T(-1.0), T(1.0)));
-        return ((AltminusB ? ABltC : ABltminusC) ? (-absArccosSumABC) : absArccosSumABC) + ((AltminusB || ABltminusC) ? numbers::pi<T> : (-numbers::pi<T>));
+        return acos<T>(runningSum.real()) + wraparound * numbers::pi<T>;
```
devshgraphicsprogramming: cast `wraparound` to float, or even keep it as a float I guess

include/nbl/builtin/hlsl/math/angle_adding.hlsl
devshgraphicsprogramming: I think the whole logic here depends on the angle being added being less than PI, I'm not sure this identity holds for when this angle is `[0,PI]` but angle being added is `[0,2*PI)`

include/nbl/builtin/hlsl/math/angle_adding.hlsl
```diff
@@ -21,53 +21,47 @@ struct sincos_accumulator
...
-        retval.runningSum = complex_t<T>::create(cosA, T(0));
+        retval.runningSum = complex_t<T>::create(cosA, sinA);
+        // retval.runningSum.real(cosA);
+        // retval.runningSum.imag(sinA);
+        retval.wraparound = 0u;
         return retval;
     }
 
-    void addCosine(T cosA, T biasA)
+    void addCosine(T cosA, T sinA)
```
devshgraphicsprogramming: should be called `addAngle` cause its a sine and cosine

include/nbl/builtin/hlsl/math/angle_adding.hlsl
devshgraphicsprogramming: actually I have even better idea

The `reverse` thing just simply says if the result needs to be rotated by 180 degrees because the real answer was supposed to be mode than PI and it got PI subtracted from it.

I actually have some ideas whether we can skip the `flipSign` on every `addAngle` and just do it at the end in `getSumOfArccos`, we'll investiagate them later, can write down a TODO there

include/nbl/builtin/hlsl/math/angle_adding.hlsl
devshgraphicsprogramming: btw let me break it down
```
bool overflow;
if (a<cosB) // angle B large enough to "fill the gap" in pi-A 
   overflow = cosB<abs(a);
else // angle A large enough to "fill the gap" in pi-B
   overflow = a<abs(cosB);
```
yeah I think this expects both angles are [0,PI] (`runningSum.y` and `sinA` is always positive)

I think you could get away with checking the sign of `d` below to check for overflow (if negative you have overflow)

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -187,65 +167,93 @@ struct SCookTorrance
...
-            assert(!hlsl::isnan(NdotV));
+            valid = false;
+            return sample_type::createInvalid();    // should check if sample direction is invalid
         }
 
-        const vector3_type localV = interaction.getTangentSpaceV();
-        const vector3_type localH = ndf.generateH(localV, u);
-        const scalar_type VdotH = hlsl::dot(localV, localH);
-        assert(VdotH >= scalar_type(0.0));  // VNDF sampling guarantees VdotH has same sign as NdotV (should be positive for BRDF)
+        cache = anisocache_type::createPartial(VdotH, LdotH, localH.z, transmitted, rcpEta);
```
devshgraphicsprogramming: this can be done in BSDF postlude, then you don't need to take a `cache` argument at all

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -646,11 +640,80 @@ struct Iridescent
...
+    static this_t create(scalar_type Dinc, vector_type ior1, vector_type ior2, vector_type ior3)
     {
-        OrientedEtaRcps<monochrome_type> rcpEta;
-        rcpEta.value = hlsl::promote<monochrome_type>(1.0) / eta23[0];
+        this_t retval;
+        retval.helper.Dinc = Dinc;
+        retval.helper.thinFilmIor = ior2;
+        retval.helper.eta12 = ior2/ior1;
+        retval.helper.eta23 = ior3/ior2;
+        retval.helper.etak23 = hlsl::promote<vector_type>(0.0);
```
devshgraphicsprogramming: I think you could get this https://github.com/Devsh-Graphics-Programming/Nabla/pull/918#discussion_r2486512597 done via CRTP where you feed this class into the `helper` class and it has `getEta12()` etc getters

or the other way round, a base class with getters and it goes into helper class without Curious Recurrence

devshgraphicsprogramming: Actually base class with just getters and mambers could also solve https://github.com/Devsh-Graphics-Programming/Nabla/pull/918#discussion_r2486516337

include/nbl/builtin/hlsl/bxdf/reflection/iridescent.hlsl
```diff
@@ -15,191 +15,16 @@ namespace bxdf
...
 }
 
 template<typename C>
 struct traits<bxdf::reflection::SIridescent<C> >
 {
     NBL_CONSTEXPR_STATIC_INLINE BxDFType type = BT_BRDF;
+    NBL_CONSTEXPR_STATIC_INLINE bool IsMicrofacet = true;
     NBL_CONSTEXPR_STATIC_INLINE bool clampNdotV = true;
     NBL_CONSTEXPR_STATIC_INLINE bool clampNdotL = true;
 };
```
devshgraphicsprogramming, line 30: can you not make a partial spec which catches `SCookTorrance` aliases?

include/nbl/builtin/hlsl/bxdf/ndf/beckmann.hlsl
keptsecret: Isn't it enough that we check infinity bool and not use the return value? Or just for safety

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
keptsecret: For `complex_t<T>` it seems T is only scalars, so I kept it as `inout`/ref

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -88,50 +88,64 @@ struct SCookTorrance
...
         return cache.isValid(orientedEta);
     }
 
-    bool dotIsUnity(const vector3_type a, const vector3_type b)
+    template<class Interaction=conditional_t<IsAnisotropic,anisotropic_interaction_type,isotropic_interaction_type>,
+            typename C=bool_constant<!fresnel_type::ReturnsMonochrome> NBL_FUNC_REQUIRES(C::value && !fresnel_type::ReturnsMonochrome)
+    static scalar_type __getScaledReflectance(NBL_CONST_REF_ARG(fresnel_type) orientedFresnel, NBL_CONST_REF_ARG(Interaction) interaction, scalar_type clampedVdotH)
+    {
+        spectral_type throughputWeights = interaction.getLuminosityContributionHint();
+        return hlsl::dot<spectral_type>(impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(orientedFresnel(clampedVdotH)), throughputWeights);
```
devshgraphicsprogramming, line 136: there shouldn't be an `impl::__implicit_promote` here at all, the `spectral_type` and `fresnel_type::vector_type` should match 100%

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -88,50 +88,64 @@ struct SCookTorrance
...
+    static scalar_type __getScaledReflectance(NBL_CONST_REF_ARG(fresnel_type) orientedFresnel, NBL_CONST_REF_ARG(Interaction) interaction, scalar_type clampedVdotH)
+    {
+        return orientedFresnel(clampedVdotH)[0];
+    }
+
+    bool __dotIsUnity(const vector3_type a, const vector3_type b, const scalar_type value)
     {
         const scalar_type ab = hlsl::dot(a, b);
-        return hlsl::max(ab, scalar_type(1.0) / ab) <= scalar_type(1.0 + 1e-3);
+        return hlsl::max(ab, value / ab) <= scalar_type(value + 1e-3);
```
devshgraphicsprogramming, line 148: if you take `value` and you check for that instead of `1.0` then `IsUnity` in the name makes no sense

include/nbl/builtin/hlsl/bxdf/ndf/beckmann.hlsl
devshgraphicsprogramming: its mostly so we don't get warnings when compiling and code still works if we forget to check the `inifnity` bool outside the function

include/nbl/builtin/hlsl/bxdf/ndf/beckmann.hlsl
```diff
@@ -339,6 +339,8 @@ struct Beckmann
         if (isInfinity)
         {
             quant_type dmq;
+            dmq.microfacetMeasure = scalar_type(0.0);
+            dmq.projectedLightMeasure = scalar_type(0.0);
```
devshgraphicsprogramming, line 343: both should be set to `bit_cast<scale_type>(numeric_limits<scale_type>::inf)` not 0s

include/nbl/builtin/hlsl/bxdf/ndf/ggx.hlsl
```diff
@@ -278,7 +278,11 @@ struct GGX
         isInfinity = hlsl::isinf(D);
         quant_type dmq;
         if (isInfinity)
+        {
+            dmq.microfacetMeasure = scalar_type(0.0);
+            dmq.projectedLightMeasure = scalar_type(0.0);
```
devshgraphicsprogramming, line 283: both should be set to `bit_cast<scale_type>(numeric_limits<scale_type>::inf)` not 0s

include/nbl/builtin/hlsl/bxdf/reflection/iridescent.hlsl
devshgraphicsprogramming: why doesn't 
```cpp
template<class Config, class N, class F>
struct traits<SCookTorrance<Config,N,F> >
{
   using __type = SCookTorrance<Config,N,F>;

    NBL_CONSTEXPR_STATIC_INLINE BxDFType type =  __type::IsBSDF ? BT_BSDF:BT_BRDF;
    NBL_CONSTEXPR_STATIC_INLINE bool IsMicrofacet = true;
    NBL_CONSTEXPR_STATIC_INLINE bool clampNdotV = !__type::IsBSDF;
    NBL_CONSTEXPR_STATIC_INLINE bool clampNdotL = !__type::IsBSDF;
};
```

defined in `base/cook_torrance_base.hlsl` function as a catch all ?

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
devshgraphicsprogramming: ah yeah you're right, about `complex_t`

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -397,40 +401,25 @@ struct Conductor
         return retval;
     }
 
-    // TODO: will probably merge with __call at some point
-    static void __polarized(const T orientedEta, const T orientedEtak, const T cosTheta, NBL_REF_ARG(T) Rp, NBL_REF_ARG(T) Rs)
+    static void __polarized(const T orientedEta, const T etaLen2, const T clampedCosTheta, NBL_REF_ARG(T) Rp, NBL_REF_ARG(T) Rs)
```
devshgraphicsprogramming: cool but this method/function should not be part of `Conductor` or `Dielectric` class, can make it free floating in `fresnel` namespace if you want

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -397,40 +401,25 @@ struct Conductor
...
         const T eta = orientedEta;
-        const T eta2 = eta*eta;
-        const T etak = orientedEtak;
-        const T etak2 = etak*etak;
 
-        const T etaLen2 = eta2 + etak2;
         assert(hlsl::all(etaLen2 > hlsl::promote<T>(hlsl::exp2<scalar_type>(-numeric_limits<scalar_type>::digits))));
         T t1 = etaLen2 * cosTheta_2;
-        const T etaCosTwice = eta * cosTheta * scalar_type(2.0);
+        const T etaCosTwice = eta * clampedCosTheta * hlsl::promote<T>(2.0);
```
devshgraphicsprogramming, line 411: whenever you do `vector*scalar` the scalar does not need to be promoted, its actually somewhat harmful for perf

include/nbl/builtin/hlsl/math/angle_adding.hlsl
```diff
@@ -28,31 +28,35 @@ struct sincos_accumulator
 
     static this_t create(T cosA, T sinA)
     {
+        assert(sinA >= T(0.0));
         this_t retval;
         retval.runningSum = complex_t<T>::create(cosA, sinA);
-        // retval.runningSum.real(cosA);
-        // retval.runningSum.imag(sinA);
-        retval.wraparound = 0u;
+        retval.wraparound = hlsl::mix(T(0.0), T(1.0), cosA == T(-1.0));
```
devshgraphicsprogramming: this is currenly wrong because `retval.runningSum` will be set to `-1,0` and your wraparound will be set to `1` too

and your `getSumofArccos` will return `acos(-1)+PI` which is 2PI, so wraparound at initialization always need to be 0

devshgraphicsprogramming: fixed myself

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -560,37 +543,47 @@ struct iridescent_helper
...
+        const vector_type eta12 = params.getEta12();
+        const vector_type eta23 = params.getEta23();
+        const vector_type etak23 = params.getEtak23();
+        const scalar_type cosTheta_1 = clampedCosTheta;
         vector_type cosTheta_2;
 
         vector_type R12p, R23p, R12s, R23s;
         const vector_type scale = scalar_type(1.0)/eta12;
-        const vector_type cosTheta2_2 = hlsl::promote<vector_type>(1.0) - hlsl::promote<vector_type>(1-cosTheta_1*cosTheta_1) * scale * scale;
+        const vector_type cosTheta2_2 = hlsl::promote<vector_type>(1.0) - hlsl::promote<vector_type>(1.0-cosTheta_1*cosTheta_1) * scale * scale;
```
devshgraphicsprogramming, line 559: this `1.0` should be wrapped in `scalar_type(1.0)` before `-` with cos theta

you can declare a `const scalar_type unity = 1.0;` somwhere in the function to save yourself a headache

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
devshgraphicsprogramming: nvm now I see they're actually different

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -560,37 +543,47 @@ struct iridescent_helper
...
+        R23s = hlsl::mix(R23s, hlsl::promote<vector_type>(0.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
+        R23p = hlsl::mix(R23p, hlsl::promote<vector_type>(0.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
+
         // Compute the transmission coefficients
         vector_type T121p = hlsl::promote<vector_type>(1.0) - R12p;
         vector_type T121s = hlsl::promote<vector_type>(1.0) - R12s;
 
         // Optical Path Difference
-        const vector_type D = hlsl::promote<vector_type>(2.0 * Dinc) * thinFilmIor * cosTheta_2;
+        const vector_type D = hlsl::promote<vector_type>(2.0 * params.getDinc()) * params.getThinFilmIor() * cosTheta_2;
```
devshgraphicsprogramming, line 586: `hlsl::promote<vector_type>(2.0 * params.getDinc()) * params.getThinFilmIor()` is a single vector which could be precomputed instead of `getThinFilmIor()`

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -560,37 +543,47 @@ struct iridescent_helper
...
+            vector_type etaLen2 = eta23 * eta23 + etak23 * etak23;
+            Conductor<vector_type>::__polarized(eta23, etaLen2, cosTheta_2, R23p, R23s);
+        }
 
         // Check for total internal reflection
         R12s = hlsl::mix(R12s, hlsl::promote<vector_type>(1.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
         R12p = hlsl::mix(R12p, hlsl::promote<vector_type>(1.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
 
+        R23s = hlsl::mix(R23s, hlsl::promote<vector_type>(0.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
+        R23p = hlsl::mix(R23p, hlsl::promote<vector_type>(0.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
```
devshgraphicsprogramming: you can do `const vector<bool,> tir = cosTheta2_2 <= hlsl::promote<vector_type>(0.0);`

btw since you compute it as ` cosTheta_2 = hlsl::sqrt(hlsl::max(cosTheta2_2, hlsl::promote<vector_type>(0.0)));` you can actually compute this condition as `cosTheta2_2<=0.0`

which in turns is `hlsl::promote<vector_type>(1.0-cosTheta_1*cosTheta_1) * scale * scale >= 1.0`

and in turn is `hlsl::promote<>(1.0-cosTheta_1*cosTheta_1) >= eta12`

So you can know the TIR waaay in advance

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -639,6 +632,25 @@ struct iridescent_helper
...
+    vector_type getEtak23() NBL_CONST_MEMBER_FUNC
+    {
+        NBL_IF_CONSTEXPR(SupportsTransmission)
+            return hlsl::promote<vector_type>(0.0);
+        else
+            return etak23;
+    }
 
     scalar_type Dinc;       // thickness of thin film in nanometers, rec. 100-25000nm
     vector_type thinFilmIor;
```
devshgraphicsprogramming: still you are storking `etak23` for dielectric iridescent.

don't take `SupportsTransmission` have `etak23` and `getEtak23()`  in your `iridescent_base`, define them in your `Iridescent` specializations and actually make them inherit instead of compose

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -656,32 +668,24 @@ struct Iridescent<T, false NBL_PARTIAL_REQ_BOT(concepts::FloatingPointLikeVector
...
-    static this_t create(scalar_type Dinc, vector_type ior1, vector_type ior2, vector_type ior3, vector_type iork3)
-    {
-        this_t retval;
-        retval.helper.Dinc = Dinc;
-        retval.helper.thinFilmIor = ior2;
-        retval.helper.eta12 = ior2/ior1;
-        retval.helper.eta23 = ior3/ior2;
-        retval.helper.etak23 = iork3/ior2;
-        return retval;
-    }
```
devshgraphicsprogramming: where's our `create` from a creation parameters struct ?

devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/918#discussion_r2486516337

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
devshgraphicsprogramming: `Dinc` and `thinFilmIoR` can be combined into a single member and getter

if you want more semantically clear initialization make creation parameter structs for `Iridescent::create`

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -560,37 +543,47 @@ struct iridescent_helper
...
         // Check for total internal reflection
         R12s = hlsl::mix(R12s, hlsl::promote<vector_type>(1.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
         R12p = hlsl::mix(R12p, hlsl::promote<vector_type>(1.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
 
+        R23s = hlsl::mix(R23s, hlsl::promote<vector_type>(0.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
+        R23p = hlsl::mix(R23p, hlsl::promote<vector_type>(0.0), cosTheta2_2 <= hlsl::promote<vector_type>(0.0));
+
         // Compute the transmission coefficients
         vector_type T121p = hlsl::promote<vector_type>(1.0) - R12p;
         vector_type T121s = hlsl::promote<vector_type>(1.0) - R12s;
```
devshgraphicsprogramming: actually it would make sense to compute a `vector<bool,> tir` then do `if (all(tir))`  path that skips a lot of the calculations (since `T121p`, `T121s`, `R23s`, `R23p` and `cosTheta2` are all zero then)

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -560,37 +543,47 @@ struct iridescent_helper
...
+        const scalar_type cosTheta_1 = clampedCosTheta;
         vector_type cosTheta_2;
 
         vector_type R12p, R23p, R12s, R23s;
         const vector_type scale = scalar_type(1.0)/eta12;
-        const vector_type cosTheta2_2 = hlsl::promote<vector_type>(1.0) - hlsl::promote<vector_type>(1-cosTheta_1*cosTheta_1) * scale * scale;
+        const vector_type cosTheta2_2 = hlsl::promote<vector_type>(1.0) - hlsl::promote<vector_type>(1.0-cosTheta_1*cosTheta_1) * scale * scale;
 
-        cosTheta_2 = hlsl::sqrt(cosTheta2_2);
+        cosTheta_2 = hlsl::sqrt(hlsl::max(cosTheta2_2, hlsl::promote<vector_type>(0.0)));
```
devshgraphicsprogramming: a little code-style correctness note, after you clamp `cosTheta_2` to not have imaginary numbers due to `sqrt` you loose the property that `cosTheta2_2 == cosTheta_2*cosTheta_2 `  so you should "hide" `cosTheta_2_2` in its own `{}` scope so its not accessible after the `cosTheta_2` assignment

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
devshgraphicsprogramming: Also it would make sense to have a `notTIR` instead of `tir` boolean (and `if (!any(notTIR))` )

because you can make a `const vector_type killFactor = vector_type(notTIR); // 0 when TIR, 1 otherwise` 

and then you can compute `R23p`, `R23s`, `T121p` and `T121s` as `originalMixLHS*killFactor`

then obviously you compute `R12` from `1.0-T121`

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -560,37 +543,47 @@ struct iridescent_helper
         return xyz / scalar_type(1.0685e-7);
     }
 
-    T __call(const scalar_type clampedCosTheta)
+    template<typename Params>
+    static T __call(NBL_CONST_REF_ARG(Params) params, const scalar_type clampedCosTheta)
     {
         const vector_type wavelengths = vector_type(colorspace::scRGB::wavelength_R, colorspace::scRGB::wavelength_G, colorspace::scRGB::wavelength_B);
```
devshgraphicsprogramming, line 549: suppy colorspace from `iridescent_base` last template param (you can default to `scSRGB`)

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -363,11 +385,15 @@ struct SCookTorrance
...
-            spectral_type reflectance = impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(hlsl::abs(cache.getVdotH())));
-            const scalar_type scaled_reflectance = hlsl::dot(reflectance, prefixThroughputWeights);
-            quo = hlsl::mix(reflectance / scaled_reflectance,
-                    (hlsl::promote<spectral_type>(1.0) - reflectance) / (scalar_type(1.0) - scaled_reflectance), cache.isTransmission()) * G2_over_G1;
+            NBL_IF_CONSTEXPR(fresnel_type::ReturnsMonochrome)
+                quo = hlsl::promote<spectral_type>(G2_over_G1);
+            else
+            {
+                const scalar_type scaled_reflectance = __getScaledReflectance(_f, interaction, hlsl::abs(cache.getVdotH()));
+                spectral_type reflectance = impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(hlsl::abs(cache.getVdotH())));
```
devshgraphicsprogramming, line 393: you're calling `_f()` twice, make `__getScaledReflectance` return the `_f()` value in a `NBL_REF_ARG(fresnel_type::vector_type) outFresnelValue` side-channel

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -363,11 +385,15 @@ struct SCookTorrance
...
+                quo = hlsl::mix(reflectance / scaled_reflectance,
+                        (hlsl::promote<spectral_type>(1.0) - reflectance) / (scalar_type(1.0) - scaled_reflectance), cache.isTransmission()) * G2_over_G1;
```
devshgraphicsprogramming, line 395: see https://github.com/Devsh-Graphics-Programming/Nabla/pull/918#discussion_r2498641613

you can simplify this`mix` to `mix(reflectance,hlsl::promote<fresnel_type::vector_type>(1.0)-reflectance,cache.isTransmission())/scaled_reflectance`

PR #920, opened 2025-09-02
Install Nabla builds

AnastaZIuk:
TODO: CI for install builds


CMakeLists.txt
```diff
@@ -25,7 +25,14 @@ set(NBL_BUILD_ANDROID OFF)
 include(ExternalProject)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/toolchains/android/build.cmake)
 
-project(Nabla LANGUAGES CXX C)
+include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
+
+project(Nabla
+	VERSION 0.7.2.0
```
AnastaZIuk, line 32: @devshgraphicsprogramming we will need to update it on at least release tags or have CI do it automatically after we fire releases

devshgraphicsprogramming: so each time I do a release tag on GH, I need to bump this ?

AnastaZIuk: > so each time I do a release tag on GH, I need to bump this ?

after convo on dc, not really: we put `0.x.y.z` numbers with desc before we do Release on Github thorugh UI, then CI does it automatically and Release is published only if we pass Smoke tests to make sure our Release actually works in install environment

PR #921, opened 2025-09-02
Smooth normal calculation

kevyuu:
- Implement createUnweldedList
- Implement createSmoothNormals with option to weld the results


include/nbl/asset/ICPUPolygonGeometry.h
```diff
@@ -96,6 +97,14 @@ class NBL_API2 ICPUPolygonGeometry final : public IPolygonGeometry<ICPUBuffer>
...
             return false;
         }
+        inline void* getNormalPtr()
+        {
+            if (isMutable())
+            {
+              return m_normalView.getPointer();
+            }
+            return nullptr;
+        }
```
devshgraphicsprogramming, line 107: I'll undo this myself

src/nbl/asset/utils/CPolygonGeometryManipulator.cpp
```diff
@@ -5,1727 +5,149 @@
...
-		else
-			((uint16_t*)indices)[i] = remap;
-	}
-
-    _NBL_DELETE_ARRAY(remapBuffer,vertexCount);
-
-	_NBL_DEBUG_BREAK_IF(nextVert > vertexCount)
-
-	return outbuffer;
-}
```
devshgraphicsprogramming, line 258: this shouldn't have disappeared

src/nbl/asset/utils/CSmoothNormalGenerator.h
```diff
@@ -18,55 +15,58 @@ class CSmoothNormalGenerator
...
-			inline BucketBounds getBucketBoundsById(uint32_t index) { return { buckets[index], buckets[index + 1] }; }
+			inline uint32_t getBucketCount() { return m_buckets.size(); }
+			inline BucketBounds getBucketBoundsById(uint32_t index) const { return { m_buckets[index], m_buckets[index + 1] }; }
 			BucketBounds getBucketBoundsByHash(uint32_t hash);
 
 		private:
 			static constexpr uint32_t invalidHash = 0xFFFFFFFF;
+      static constexpr uint32_t primeNumber1 = 73856093;
+      static constexpr uint32_t primeNumber2 = 19349663;
+      static constexpr uint32_t primeNumber3 = 83492791;
```
devshgraphicsprogramming, line 53: static inline constexpr

include/nbl/asset/ICPUPolygonGeometry.h
devshgraphicsprogramming: done

src/nbl/asset/utils/CPolygonGeometryManipulator.cpp
devshgraphicsprogramming: brought back

src/nbl/asset/utils/CSmoothNormalGenerator.h
devshgraphicsprogramming: done

include/nbl/asset/utils/CPolygonGeometryManipulator.h
```diff
@@ -9,16 +9,26 @@
...
 	public:
+		//vertex data needed for CSmoothNormalGenerator
+		struct SSNGVertexData
+		{
+			uint32_t index;									     //offset of the vertex into index buffer
+			uint32_t hash;											       //
+			float wage;												         //angle wage of the vertex
+			hlsl::float32_t3 position;							   //position of the vertex in 3D space
+			hlsl::float32_t3 parentTriangleFaceNormal; //
+		};
```
devshgraphicsprogramming, line 28: change `wage` to `weight`, also if we support 64bit indices, then the vertex index should be 64bit.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -123,88 +165,102 @@ struct KeyAccessor
...
+CSmoothNormalGenerator::VertexHashMap CSmoothNormalGenerator::setupData(const asset::ICPUPolygonGeometry* polygon, float epsilon)
 {
-	const size_t idxCount = buffer->getIndexCount();
-	_NBL_DEBUG_BREAK_IF((idxCount % 3));
-
-	VertexHashMap vertices(idxCount, std::min(16u * 1024u, core::roundUpToPoT<unsigned int>(idxCount * 1.0f / 32.0f)), epsilon == 0.0f ? 0.00001f : epsilon * 1.00001f);
+	const size_t idxCount = polygon->getPrimitiveCount() * 3;
 
-	core::vector3df_SIMD faceNormal;
+	VertexHashMap vertices(idxCount, std::min(16u * 1024u, core::roundUpToPoT<unsigned int>(idxCount * 1.0f / 32.0f)), epsilon == 0.0f ? 0.00001f : epsilon * 2.f);
```
devshgraphicsprogramming, line 207: the floating point multiplication of index coult is a bit disgusting, just do `roundUpToPoT<uint32_t>((idxCount+31)>>5)`

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -123,88 +165,102 @@ struct KeyAccessor
...
-		//set data for vertices
-		core::vector3df_SIMD angleWages = getAngleWeight(v1, v2, v3);
+		//set data for m_vertices
+		const auto angleWages = getAngleWeight(v1, v2, v3);
 
-		vertices.add({ i,		0,	angleWages.x,	v1,		faceNormal });
-		vertices.add({ i + 1,	0,	angleWages.y,	v2,		faceNormal });
-		vertices.add({ i + 2,	0,	angleWages.z,	v3,		faceNormal });
+		vertices.add({ i,	0,	angleWages.x,	v1,		faceNormal});
+		vertices.add({ i + 1,	0,	angleWages.y,	v2,		faceNormal});
```
devshgraphicsprogramming, line 223: I forgot that we could go through the index buffer instead with the indexing callback and not require an unwelded mesh as input 🤦

devshgraphicsprogramming, line 223: slight question, is there any harm/downside to keeping the `faceNormal*weight` instead of face normal and weight separately ?

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -123,88 +165,102 @@ struct KeyAccessor
...
-	const size_t idxCount = buffer->getIndexCount();
-	_NBL_DEBUG_BREAK_IF((idxCount % 3));
-
-	VertexHashMap vertices(idxCount, std::min(16u * 1024u, core::roundUpToPoT<unsigned int>(idxCount * 1.0f / 32.0f)), epsilon == 0.0f ? 0.00001f : epsilon * 1.00001f);
+	const size_t idxCount = polygon->getPrimitiveCount() * 3;
 
-	core::vector3df_SIMD faceNormal;
+	VertexHashMap vertices(idxCount, std::min(16u * 1024u, core::roundUpToPoT<unsigned int>(idxCount * 1.0f / 32.0f)), epsilon == 0.0f ? 0.00001f : epsilon * 2.f);
 
 	for (uint32_t i = 0; i < idxCount; i += 3)
```
devshgraphicsprogramming, line 209: type mismatch, index count is 64bit, your iterator is 32

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -123,88 +165,102 @@ struct KeyAccessor
...
+		polygon->getPositionView().decodeElement<hlsl::float32_t3>(i + 2, v3);
 
-		faceNormal = core::cross(v2 - v1, v3 - v1);
-		faceNormal = core::normalize(faceNormal);
+		const auto faceNormal = normalize(cross(v2 - v1, v3 - v1));
 
-		//set data for vertices
-		core::vector3df_SIMD angleWages = getAngleWeight(v1, v2, v3);
+		//set data for m_vertices
+		const auto angleWages = getAngleWeight(v1, v2, v3);
```
devshgraphicsprogramming, line 220: because we "technically" support FP64 precision versives, this `for` loop needs to be turned into a template lambda and we need to dispatch into a 32bit or 64bit version depending on the precision of the position view (64bit SFLOAT gets 64bit version, everything else gets 32)

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -2,118 +2,160 @@
...
-	const float asqrt = core::sqrt(a);
-	const float b = core::distancesquared(v1,v3)[0];
-	const float bsqrt = core::sqrt(b);
-	const float c = core::distancesquared(v1,v2)[0];
-	const float csqrt = core::sqrt(c);
+	const float a = distancesquared(v2, v3);
+	const float asqrt = sqrt(a);
+	const float b = distancesquared(v1,v3);
+	const float bsqrt = sqrt(b);
+	const float c = distancesquared(v1,v2);
```
devshgraphicsprogramming, line 90: I'd rework `getAngleWeight` to work in terms of edges, not vertices 

Also move it to HLSL somewhere in `shapes/triangle`

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
devshgraphicsprogramming: also means that `SSNGVertexData` and HashGrid need to exist in float32_t and float64_t flavours

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -123,88 +165,102 @@ struct KeyAccessor
...
+	// TODO: maybe use counting sort (or big radix) and use the histogram directly for the m_buckets
+	auto finalSortedOutput = core::radix_sort(m_vertices.data(),m_vertices.data()+oldSize,oldSize,KeyAccessor());
 	// TODO: optimize out the erase
-	if (finalSortedOutput!=vertices.data())
-		vertices.erase(vertices.begin(),vertices.begin()+oldSize);
+	if (finalSortedOutput!=m_vertices.data())
+		m_vertices.erase(m_vertices.begin(),m_vertices.begin()+oldSize);
 	else
-		vertices.erase(vertices.begin()+oldSize,vertices.end());
+		m_vertices.erase(m_vertices.begin()+oldSize,m_vertices.end());
```
devshgraphicsprogramming, line 183: rutn the erase into a resize in the `else` case

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -123,88 +165,102 @@ struct KeyAccessor
...
 void CSmoothNormalGenerator::VertexHashMap::validate()
 {
-	const auto oldSize = vertices.size();
-	vertices.resize(oldSize*2u);
-	// TODO: maybe use counting sort (or big radix) and use the histogram directly for the buckets
-	auto finalSortedOutput = core::radix_sort(vertices.data(),vertices.data()+oldSize,oldSize,KeyAccessor());
+	const auto oldSize = m_vertices.size();
+	m_vertices.resize(oldSize*2u);
+	// TODO: maybe use counting sort (or big radix) and use the histogram directly for the m_buckets
+	auto finalSortedOutput = core::radix_sort(m_vertices.data(),m_vertices.data()+oldSize,oldSize,KeyAccessor());
```
devshgraphicsprogramming, line 178: I ran the debugger and your `hash` doesn't get computed at all, everything has a hash of 0

And you also only ever get 2 buckets made, it means your smooth normal generator is actually O(N^2) due to a bug

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -214,59 +270,59 @@ void CSmoothNormalGenerator::processConnectedVertices(asset::ICPUMeshBuffer * bu
...
 
-std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCellHashes(const IMeshManipulator::SSNGVertexData & vertex)
+std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCellHashes(const CPolygonGeometryManipulator::SSNGVertexData & vertex)
 {
 	std::array<uint32_t, 8> neighbourhood;
 
-	core::vectorSIMDf cellFloatCoord = vertex.position / cellSize - core::vectorSIMDf(0.5f);
-	core::vector3du32_SIMD neighbor = core::vector3du32_SIMD(static_cast<uint32_t>(cellFloatCoord.x), static_cast<uint32_t>(cellFloatCoord.y), static_cast<uint32_t>(cellFloatCoord.z));
+	hlsl::float32_t3 cellFloatCoord = vertex.position / m_cellSize - hlsl::float32_t3(0.5f);
+	hlsl::uint32_t3 neighbor = hlsl::uint32_t3(static_cast<uint32_t>(cellFloatCoord.x), static_cast<uint32_t>(cellFloatCoord.y), static_cast<uint32_t>(cellFloatCoord.z));
```
devshgraphicsprogramming, line 295: `hlsl::_static_cast<hlsl::uint32_t3>(cellFloatCoord.x)` should work,

Btw there might be a problem that float->int conversions round towards zero, so [0.5,1.5) maps to 0 but also does (-0.5,0.5). Its probably not a great deal and doesn't impact correctness of the algo (it just changes the hash functions to give more hash-collisions on the x,y,z=0 planes)

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -214,59 +270,59 @@ void CSmoothNormalGenerator::processConnectedVertices(asset::ICPUMeshBuffer * bu
...
-	neighbor = neighbor - core::vector3du32_SIMD(0, 0, 1);
+	neighbor = neighbor - hlsl::uint32_t3(0, 0, 1);
 	neighbourhood[6] = hash(neighbor);
 
 	//left top near
-	neighbor = neighbor - core::vector3du32_SIMD(1, 0, 0);
+	neighbor = neighbor - hlsl::uint32_t3(1, 0, 0);
 	neighbourhood[7] = hash(neighbor);
 
 	//erase duplicated hashes
```
devshgraphicsprogramming, line 328: its a better to make a `uint8_t getNeighboringCellHashes(uint32_t* outNeighbours, const IMeshManipulator::SSNGVertexData&)` and return the number of unique hashes.

right now its 8*7 iterations to check for duplicates and then you're going over 8 output values.

Whereas we could do far less work if every time a neighbour is pushed into the output we check for duplicates against what already has been added.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -123,88 +165,102 @@ struct KeyAccessor
...
+		core::vector<CPolygonGeometryManipulator::SSNGVertexData>::iterator next = std::upper_bound(prevBegin, m_vertices.end(), prevHash);
+		m_buckets.push_back(next);
 
-		if (next == vertices.end())
+		if (next == m_vertices.end())
 			break;
 
 		prevBegin = next;
 		prevHash = next->hash;
 	}
```
devshgraphicsprogramming, line 200: you could extend the `core::radix_sort` a bit, and allow it to use an externally provided memory as the histogramming array.

Basically more this struct out of `impl` and make it clear that its an LSB sorter (change the typename)
https://github.com/Devsh-Graphics-Programming/Nabla/blob/ede1794490bde99296d8ac0fe9e8886e93644b7d/include/nbl/core/algorithm/radix_sort.h#L53

Then the left-over histogram could actually be a skiplist, which you can use instead of `buckets` to help you with your range binary searches.

So instead of doing lookups bu upper and lower bound over the whole `m_vertices` in `getBucketBoundsByHash`

You'd look up `cumulativeHistogram[hash>>(32-radixBitCount)]` and that would give you a better `begin` for your uuper and lower bounds, while `cumulativeHistogram[(hash>>(32-radixBitCount))+1]` a better `end`

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
-#endif
+
+core::smart_refctd_ptr<ICPUPolygonGeometry> CSmoothNormalGenerator::weldVertices(const ICPUPolygonGeometry* polygon, VertexHashMap& vertices, float epsilon)
+{
+	struct Group
+	{
+		uint32_t vertex_reference_index; // index to referenced vertex in the original polygon
+	};
+	core::vector<Group> groups; 
+	groups.reserve(vertices.getVertexCount());
```
devshgraphicsprogramming, line 348: improve the nomeclature, because when someone sees "Groups" they might think "SmoothGroups", so call it Duplicate Groups or just Duplicates

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
 	}
 	return neighbourhood;
 }
-#endif
+
+core::smart_refctd_ptr<ICPUPolygonGeometry> CSmoothNormalGenerator::weldVertices(const ICPUPolygonGeometry* polygon, VertexHashMap& vertices, float epsilon)
```
devshgraphicsprogramming, line 341: I'd re-architect this a bit.

Make a class which welds polygons, which can use an acceleration structure to find vertices with similar attributes, and make the `VertexHashMap` one of these acceleration structures.

That way the smooth normal calc can return the `VertexHashMap` accel structure, and the vanilla weld function can not take it and make its own hash while you can feed the left over smooth normal to the `weldVertices`

The reasons is that weld-vertices needs a lambda per attribute (you want to use different epsilons and also different ways to measure discrepancy) for equality comparison, see the old code.

So TL;DR its a bit annoying to pass these lists of lambdas to a smooth normal recompute function.

devshgraphicsprogramming: For example you'd want to compare directional attributes using dot products or some other weird metrics

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
+{
+	struct Group
+	{
+		uint32_t vertex_reference_index; // index to referenced vertex in the original polygon
```
devshgraphicsprogramming, line 345: if you allow 64bit indices, then this needs to be 64bit (or you need templated lambda as the function body and dispatch to the correct one per index size)

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
+	};
+	core::vector<Group> groups; 
+	groups.reserve(vertices.getVertexCount());
+
+	core::vector<std::optional<uint32_t>> groupIndexes(vertices.getVertexCount());
```
devshgraphicsprogramming, line 350: 1. `sizeof(std::optional<T>)>sizeof(T)` you should just use the max value as the invalid value
2. if you allow 64bit indices, then thus needs to be 64bit too
3. it has a really bad name, I had to read the code to find what it does, could be called `indexToFirstRepresentativeVertex`

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
devshgraphicsprogramming: we also have what are called "custom attributes" (with invalid strides, there's a bool method to check) which are not of any particular size, and as such they're impossible to compare because what gets used from them doesn't neatly match `vertexID*stride+offset` byte range of `stride` length

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
+    for (auto auxView_i = 0u; auxView_i < polygon->getAuxAttributeViews().size(); auxView_i++)
+    {
+      auto& inView = polygon->getAuxAttributeViews()[auxView_i];
+      auto& outView = outPolygon->getAuxAttributeViews()->operator[](auxView_i);
+      const auto attrSize = inView.composed.stride;
+      const std::byte* const inAuxs = reinterpret_cast<const std::byte*>(inView.getPointer());
+      std::byte* const outAuxs = reinterpret_cast<std::byte*>(outView.getPointer());
+      memcpy(outAuxs + group_i * attrSize, inAuxs + srcIndex * attrSize, attrSize);
+    }
+	}
```
devshgraphicsprogramming, line 537: because of https://github.com/Devsh-Graphics-Programming/Nabla/pull/921/files?diff=split&w=1#r2346607011

you need to refuse welding and unwelding on meshes with such attributes, or ask for a lambda to handle the copies (instead of `memcpy`, but memcpy can be the default one)

devshgraphicsprogramming: also I see a loot of duplicate code here

devshgraphicsprogramming: Also separation of concerns.

**Welding is simply the act of producing an new index buffer, so why  the `memcpy` and creation of new data views?**

If I want to "compact" the gaps in the vertex data, I may as well reorder the index buffer and vertex data to make efficient use of the Shader rea cache and the Vertex Post Transform cache.

This is what the "createFetchOptimized" function was for.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
devshgraphicsprogramming: tbf this could just be a dynamically sized bitset which is jjust a `vertexVisited` ballot, size should be from 0 to MaxVertex as reported by the input mesh

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
+	for (uint32_t cell = 0; cell < vertices.getBucketCount() - 1; cell++)
+	{
+		VertexHashMap::BucketBounds processedBucket = vertices.getBucketBoundsById(cell);
+
+		for (core::vector<CPolygonGeometryManipulator::SSNGVertexData>::iterator processedVertex = processedBucket.begin; processedVertex != processedBucket.end; processedVertex++)
+		{
+			std::array<uint32_t, 8> neighboringCells = vertices.getNeighboringCellHashes(*processedVertex);
+			hlsl::float32_t3 normal = processedVertex->parentTriangleFaceNormal * processedVertex->wage;
+
+			auto& groupIndex = groupIndexes[processedVertex->index];
```
devshgraphicsprogramming, line 378: the check for `!groupIndex` should be here, and vertices already visited should be skipped.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
devshgraphicsprogramming: tbf you could just create the new index buffer & view here right away instead of this vector

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
+						break;
+					}
+				}
+			}
+			if (!groupIndex)
+			{
+        // create new group if no group nearby that is compatible with this vertex
+        groupIndex = groups.size();
+        groups.push_back({ processedVertex->index});
+			}
```
devshgraphicsprogramming, line 409: you might as well have a temporary called `remappedVertexID` initialized to `processedVertex->index` and construct the new index buffer right away -  you just need to iterate over the input index list (BUT NOT like when you unweld a geometry - don't use the indexing, just go over raw index values as-if it was a list of points) instead of the vertex references in your Acceleration Structure

You'll still have to look up 2x2x2 the same cells in the hash-grid, complexity doesn't change.

devshgraphicsprogramming: Also if you iterate over the index list, then your weld function will work with any Indexing and Polygon, **and you will be able to keep the same Indexing (e.g. if its a strip it stays a strip and doesn't turn into a list)**

Why? Because then the new index buffer, is just a substitution map (oldIndex->newIndex), so you can loop over the raw index buffer values and replace them without paying attention to the particular indexing used.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
+		}
+	}
+
+  auto outPolygon = core::move_and_static_cast<ICPUPolygonGeometry>(polygon->clone(0u));
+	outPolygon->setIndexing(IPolygonGeometryBase::TriangleList());
+
+	const uint32_t indexSize = (groups.size() < std::numeric_limits<uint16_t>::max()) ? sizeof(uint16_t) : sizeof(uint32_t);
```
devshgraphicsprogramming, line 416: wrong thing to decide, while you iterate above keep track of the maxVertex value you actually use (so in new refactor `remappedVertexID` AFTER neighbours are visited) 

Then use that to determine if you need uint16, or uint32 index buffer on the output. 

**!!! Also right now you don't set the new and correct `maxVertex` on the `outPolygon` !!!**

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
+	if (indexSize == 2)
+	{
+		indexView.composed.encodedDataRange.u16.minVx[0] = 0;
+		indexView.composed.encodedDataRange.u16.maxVx[0] = groups.size() - 1;
+		indexView.composed.format = EF_R16_UINT;
+		indexView.composed.rangeFormat = IGeometryBase::EAABBFormat::U16;
+	} else if (indexSize == 4)
+	{
+		indexView.composed.encodedDataRange.u32.minVx[0] = 0;
+		indexView.composed.encodedDataRange.u32.maxVx[0] = groups.size() - 1;
```
devshgraphicsprogramming, line 438: not the correct max value to set

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
devshgraphicsprogramming: if you want to demote the format, then you need to make a view with matching format as input, use that during the iteration, and then copy over to a newer view with narrower format

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
```diff
@@ -281,6 +337,208 @@ std::array<uint32_t, 8> CSmoothNormalGenerator::VertexHashMap::getNeighboringCel
...
+		indexView.composed.rangeFormat = IGeometryBase::EAABBFormat::U32;
+	}
+
+	for (auto index_i = 0u; index_i < groupIndexes.size(); index_i++)
+	{
+		if (indexSize == 2)
+		{
+			uint16_t index = *groupIndexes[index_i];
+			memcpy(indexBufferPtr + indexSize * index_i, &index, sizeof(index));
+		} else if (indexSize == 4)
```
devshgraphicsprogramming, line 449: don't put things you can quickly template-lambda as if-else branches in hot loops

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Currently one cell can contain up to 32 vertices. I will try to reduce this number.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Yeah, but need to assume it is triangle

kevyuu: So should we add an option for welded mesh?

devshgraphicsprogramming: > So should we add an option for welded mesh?

yes

devshgraphicsprogramming: > Yeah, but need to assume it is triangle

ofc thats fine

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
devshgraphicsprogramming: the problem was that `hash` never got computed, and thats how we get the cells

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: still need parentTriangleFaceNormal, since the lambda to smotth might want the parent normal to determine whether to smooth it or not.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Sometimes we want to use the faceNormal to determine whether to smooth the normal or not.

include/nbl/asset/utils/CPolygonGeometryManipulator.h
devshgraphicsprogramming: > still need parentTriangleFaceNormal, since the lambda to smotth might want the parent normal to determine whether to smooth it or not.

make the lambda re-normalize it during compatison, not fetching the weight is a 4 byte saving in memory probably worth the ALUs

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: I cannot find createFetchOptimized. Should I implement this function?

devshgraphicsprogramming: you removed the old source code, I added the commented out stuff back

include/nbl/asset/utils/CPolygonGeometryManipulator.h
kevyuu: Done.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Is it okay If I just use 64 bit? So we always work with 64 bit normal and position even for 32 bit views? Since the format of position view is known at runtime, there will be a lot of branch down the line if we keep 32 bit and 64 bit hash grid. I think it is reasonable, we store index with 64 bit to reduce branches, so why not do the same for position?

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: > the problem was that `hash` never got computed, and thats how we get the cells

hash is computed at VertexHashMap::add. The prism that I use as the test case in example 67 have not more than 32 vertices. So it will be assign to the same cell. Since only one cell is created.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: The compiler won't be able to hoist the brances outside?

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: I add floor to compute the cellFloat. Done.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done.

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
devshgraphicsprogramming: No i wouldn't count on it hoisting and instantiating two loops

kevyuu: Done

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
kevyuu: Done

src/nbl/asset/utils/CSmoothNormalGenerator.cpp
devshgraphicsprogramming: continuing in #941

PR #922, opened 2025-09-03
Fix rotation mat

kevyuu:
Hotfix rotation matrix. Previously it works because AS build trimmed the last column which makes this function looks correct. I probably thought the default constructor of a matrix is identity matrix, which is not the case.


include/nbl/builtin/hlsl/math/linalg/transform.hlsl
```diff
@@ -50,6 +50,21 @@ matrix<T, 4, 4> rotation_mat(T angle, vector<T, 3> const& axis)
   return rotation;
 }
 
+template <uint16_t NOut, uint16_t MOut, uint16_t NIn, uint16_t MIn, typename T>
+requires(NOut >= NIn && MOut >= MIn)
+matrix <T, NOut, MOut> zero_fill(const matrix<T, NIn, MIn> inMatrix)
+{
+  matrix<T, NOut, MOut> retval;
```
devshgraphicsprogramming: that will just give you uninitialized variables in the components you don't assign to

include/nbl/builtin/hlsl/math/linalg/transform.hlsl
```diff
@@ -50,6 +50,21 @@ matrix<T, 4, 4> rotation_mat(T angle, vector<T, 3> const& axis)
...
+  for (auto row_i = 0u; row_i < NIn; row_i++)
+  {
+    for (auto col_i = 0u; col_i < MIn; col_i++)
+    {
+      retval[row_i][col_i] = inMatrix[row_i][col_i];
+    }
```
devshgraphicsprogramming: you can just do
```cpp
using out_row_t = vector<T,MOut>;
for (auto row_i = 0u; row_i < NOut; row_i++)
   retval[row_i] = row_i<NIn ? promote<out_row_t>(inMatrix[row_i]):promote<out_row_t>(0);
```

btw are you sure that zero-fill is what you want and not a `promote_affine` ? with a
```
if (row_i>=NIn || row_i>=MIn)
   retval[row_i][row_i] = T(1);
```

kevyuu: Yeah, we can do that. Currently I just need to convert 3x3 to 3x4

include/nbl/builtin/hlsl/math/linalg/transform.hlsl
```diff
@@ -52,15 +52,22 @@ matrix<T, 3, 3> rotation_mat(T angle, vector<T, 3> const& axis)
...
 {
   matrix<T, NOut, MOut> retval;
-  for (auto row_i = 0u; row_i < NIn; row_i++)
+
+  using out_row_t = hlsl::vector<T, MOut>;
+  auto expandVec = [](vector<T, MIn> inVec) -> vector<T, MOut>
+  {
+    if constexpr (MIn == MOut) return inVec;
+    return vector<T, MOut>(inVec, vector<T, MOut - MIn>(0));
+  };
```
devshgraphicsprogramming: HLSL doesn't support lambdas or `if constexpr` this won't compile with DXC

devshgraphicsprogramming: you can make `zero_expand` a general funtion for vectors

PR #943, opened 2025-10-29
Bitonic_Sort

CrabExtra:

### DIFF
--- a/include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
+++ b/include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
@@ -2,6 +2,7 @@
 #define _NBL_BUILTIN_HLSL_WORKGROUP_BITONIC_SORT_INCLUDED_
 
 #include "nbl/builtin/hlsl/bitonic_sort/common.hlsl"
+#include "nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl"
 #include "nbl/builtin/hlsl/workgroup/basic.hlsl"
 #include "nbl/builtin/hlsl/memory_accessor.hlsl"
 
@@ -78,14 +79,14 @@ inline void inThreadShuffleSortPairs(
     WorkgroupType<typename config::key_t> elems[config::E],
     NBL_CONST_REF_ARG(Comp) comp)
 {
-    uint32_t partner = stride;
+    uint32_t partner = stride >> 1;
 
     [unroll]
-    for (uint32_t i = 0u; i < config::E; i += stride * 2u)
+    for (uint32_t i = 0u; i < config::E; i += stride)
     {
         uint32_t j = i + partner;
         bool valid = j < config::E;
-        bool swap = valid && (ascending ? comp(elems[j], elems[i]) : comp(elems[i], elems[j]));
+        bool swap = valid && (comp(elems[j].key, elems[i].key) == ascending);
 
         WorkgroupType<typename config::key_t> tmp = elems[i];
 
@@ -104,26 +105,33 @@ inline void subgroupShuffleSortPairs(
     uint32_t subgroupInvocationID,
     NBL_CONST_REF_ARG(Comp) comp)
 {
-    bool isUpper = (subgroupInvocationID & stride) != 0u;
+    uint32_t partner = stride >> 1;
+    bool isUpper = (subgroupInvocationID & partner) != 0u;
 
     [unroll]
-    for (uint32_t i = 0u; i < config::E; ++i)
+    for (uint32_t i = 0u; i < config::E; i += 2u)
     {
-        typename config::key_t partnerKey = glsl::subgroupShuffleXor(elems[i].key, stride);
-        uint32_t partnerIdx = glsl::subgroupShuffleXor(elems[i].workgroupRelativeIndex, stride);
+        uint32_t j = i + 1u;
 
-        WorkgroupType<typename config::key_t> partnerElem;
-        partnerElem.key = partnerKey;
-        partnerElem.workgroupRelativeIndex = partnerIdx;
+        typename config::key_t tradingKey = isUpper ? elems[j].key : elems[i].key;
+        uint32_t tradingIdx = isUpper ? elems[j].workgroupRelativeIndex : elems[i].workgroupRelativeIndex;
 
-     
-        bool keepPartner = (ascending == isUpper) ? comp(partnerElem, elems[i]) : comp(elems[i], partnerElem);
-        if (keepPartner) {
-            elems[i] = partnerElem;
-        }
+        tradingKey = glsl::subgroupShuffleXor(tradingKey, partner);
+        tradingIdx = glsl::subgroupShuffleXor(tradingIdx, partner);
+
+        elems[i].key = isUpper ? elems[i].key : tradingKey;
+        elems[i].workgroupRelativeIndex = isUpper ? elems[i].workgroupRelativeIndex : tradingIdx;
+        elems[j].key = isUpper ? tradingKey : elems[j].key;
+        elems[j].workgroupRelativeIndex = isUpper ? tradingIdx : elems[j].workgroupRelativeIndex;
+
+        bool swap = comp(elems[j].key, elems[i].key) == ascending;
+        WorkgroupType<typename config::key_t> tmp = elems[i];
+
+        elems[i].key = swap ? elems[j].key : elems[i].key;
+        elems[i].workgroupRelativeIndex = swap ? elems[j].workgroupRelativeIndex : elems[i].workgroupRelativeIndex;
+        elems[j].key = swap ? tmp.key : elems[j].key;
+        elems[j].workgroupRelativeIndex = swap ? tmp.workgroupRelativeIndex : elems[j].workgroupRelativeIndex;
     }
-
-    atomicOpPairs<config>(ascending, elems, comp);
 }
 
 template<typename config, typename Comp, typename SMem>
@@ -155,36 +163,30 @@ inline void workgroupShuffleSortPairs(
         bool active = (tid >= start) && (tid < end);
 
         [unroll]
-        for (uint32_t i = 0u; i < config::E; ++i)
+        for (uint32_t i = 0u; i < config::E/2; ++i)
         {
-            if (active)
+            uint32_t localIdx = isUpper ? (i + config::E/2) : i;
+            if (active && (localIdx < config::E))
             {
-                k.set(tid*config::E + i, elems[i].key);
-                idx.set(tid*config::E + i, elems[i].workgroupRelativeIndex);
+                k.set(tid*config::E + localIdx, elems[localIdx].key);
+                idx.set(tid*config::E + localIdx, elems[localIdx].workgroupRelativeIndex);
             }
         }
         smem.workgroupExecutionAndMemoryBarrier();
 
         [unroll]
-        for (uint32_t i = 0u; i < config::E; ++i)
+        for (uint32_t i = 0u; i < config::E/2; ++i)
         {
-            if (active)
+            uint32_t localIdx = isUpper ? i : (i + config::E/2);
+            if (active && (localIdx < config::E))
             {
-                uint32_t myElemIdx = tid * config::E + i;
+                uint32_t myElemIdx = tid * config::E + (isUpper ? (i + config::E/2) : i);
                 uint32_t partnerElemIdx = myElemIdx ^ stride;
                 uint32_t pTid = partnerElemIdx / config::E;
                 uint32_t partnerLocalIdx = partnerElemIdx % config::E;
 
-                WorkgroupType<typename config::key_t> partnerElem;
-                k.get(pTid*config::E + partnerLocalIdx, partnerElem.key);
-                idx.get(pTid*config::E + partnerLocalIdx, partnerElem.workgroupRelativeIndex);
-
-                bool isUpper = (myElemIdx & stride) != 0u;
-
-                bool keepPartner = (ascending == isUpper) ? comp(partnerElem, elems[i]) : comp(elems[i], partnerElem);
-                if (keepPartner) {
-                    elems[i] = partnerElem;
-                }
+                k.get(pTid*config::E + partnerLocalIdx, elems[localIdx].key);
+                idx.get(pTid*config::E + partnerLocalIdx, elems[localIdx].workgroupRelativeIndex);
             }
         }
         smem.workgroupExecutionAndMemoryBarrier();
@@ -246,8 +248,8 @@ struct BitonicSort
             uint32_t idx = tid * config::E + i;
             KVPair kvpair;
             acc.template get<KVPair>(idx, kvpair);
-            elems[i].key = kvpair.first;                      // The key to sort by (random number)
-            elems[i].workgroupRelativeIndex = kvpair.second;  // The original index to track
+            elems[i].key = kvpair.first;                      
+            elems[i].workgroupRelativeIndex = kvpair.second;  
         }
 
         typename config::comparator_t comp;
@@ -268,13 +270,12 @@ struct BitonicSort
             }
         }
 
-        // Write back sorted (sortedKey, originalIndex) pairs
         [unroll]
         for (uint32_t i = 0u; i < config::E; ++i)
         {
             KVPair output;
-            output.first = elems[i].key;                      // Sorted key
-            output.second = elems[i].workgroupRelativeIndex;  // Original index
+            output.first = elems[i].key;                     
+            output.second = elems[i].workgroupRelativeIndex;  
             acc.template set<KVPair>(tid * config::E + i, output);
         }
     }


include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,132 @@
...
+    const bool doSwap = (shouldSwap == ascending);
+
+    KeyType tempKey = loKey;
+    loKey = doSwap ? hiKey : loKey;
+    hiKey = doSwap ? tempKey : hiKey;
+
+    ValueType tempVal = loVal;
+    loVal = doSwap ? hiVal : loVal;
+    hiVal = doSwap ? tempVal : hiVal;
+}
```
Fletterio: Remove this one and do everything with pairs so it's more consistent

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,132 @@
...
+    const bool loSelfSmaller = comp(loPair.first, partnerLoPair.first);
+    const bool takePartnerLo = takeLarger ? loSelfSmaller : !loSelfSmaller;
+    loPair.first = takePartnerLo ? partnerLoPair.first : loPair.first;
+    loPair.second = takePartnerLo ? partnerLoPair.second : loPair.second;
+
+    const bool hiSelfSmaller = comp(hiPair.first, partnerHiPair.first);
+    const bool takePartnerHi = takeLarger ? hiSelfSmaller : !hiSelfSmaller;
+    hiPair.first = takePartnerHi ? partnerHiPair.first : hiPair.first;
+    hiPair.second = takePartnerHi ? partnerHiPair.second : hiPair.second;
+}
```
Fletterio: Define an `operator=` for pair, then this becomes
```cpp
if(takePartnerLo)
    loPair = pLoPair;
if(takePartnerHi)
    hiPair = pHiPair;
```
don't worry about branching, since everything's an assignment it'll just be `OpSelect`s under the hood

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,132 @@
...
+NBL_REF_ARG(ValueType) hiVal)
+{
+    KeyType tempKey = loKey;
+    loKey = hiKey;
+    hiKey = tempKey;
+
+    ValueType tempVal = loVal;
+    loVal = hiVal;
+    hiVal = tempVal;
+}
```
Fletterio: delete all of these and only use the versions that use pairs. It's more consistent

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,132 @@
...
+
+template<typename KeyType, typename ValueType>
+void swap(
+NBL_REF_ARG(pair<KeyType, ValueType>) loPair,
+NBL_REF_ARG(pair<KeyType, ValueType>) hiPair)
+{
+    pair<KeyType, ValueType> temp = loPair;
+    loPair = hiPair;
+    hiPair = temp;
+}
```
Fletterio: This won't work without a definition for `operator=` for pair. Your code compiles because you're not using it rn. We want to keep this version and rewrite all the `swaps` to use this version using pairs.

The definition for pair, the overload for `operator=` and this `swap` method all belong in https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/hlsl/utility.hlsl, mimicking how `std::pair` is in the `<utility>` header in cpp. Move them over there.

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,132 @@
...
+    const bool shouldSwap = comp(hiPair.first, loPair.first);
+    const bool doSwap = (shouldSwap == ascending);
+
+    KeyType tempKey = loPair.first;
+    ValueType tempVal = loPair.second;
+    loPair.first = doSwap ? hiPair.first : loPair.first;
+    loPair.second = doSwap ? hiPair.second : loPair.second;
+    hiPair.first = doSwap ? tempKey : hiPair.first;
+    hiPair.second = doSwap ? tempVal : hiPair.second;
+}
```
Fletterio: With the definition for the `swap` below, this just becomes
```cpp
if (doSwap)
    swap(loPair, hiPair);
```

include/nbl/builtin/hlsl/memory_accessor.hlsl
```diff
@@ -24,15 +24,15 @@ namespace hlsl
...
-};
+//template<typename T1, typename T2>
+//struct pair
+//{
+//    using first_type = T1;
+//    using second_type = T2;
+//
+//    first_type first;
+//    second_type second;
+//};
```
Fletterio: Remove this

include/nbl/builtin/hlsl/pair.hlsl
```diff
@@ -0,0 +1,38 @@
...
+
+// Helper to make a pair (similar to std::make_pair)
+template<typename T1, typename T2>
+pair<T1, T2> make_pair(T1 f, T2 s)
+{
+    pair<T1, T2> p;
+    p.first = f;
+    p.second = s;
+    return p;
+}
```
Fletterio: All of this belongs in https://github.com/Devsh-Graphics-Programming/Nabla/blob/master/include/nbl/builtin/hlsl/utility.hlsl. Move the code over there and remove this file.

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,81 @@
...
+template<typename KeyType, typename ValueType, typename Comparator, class device_capabilities>
+struct bitonic_sort<bitonic_sort_config<KeyType, ValueType, Comparator>, device_capabilities>
+{
+	using config_t = bitonic_sort_config<KeyType, ValueType, Comparator>;
+	using key_t = typename config_t::key_t;
+	using value_t = typename config_t::value_t;
+	using comparator_t = typename config_t::comparator_t;
+
+	static void mergeStage(uint32_t stage, bool bitonicAscending, uint32_t invocationID, NBL_REF_ARG(key_t) loKey, NBL_REF_ARG(key_t) hiKey,
+		NBL_REF_ARG(value_t) loVal, NBL_REF_ARG(value_t) hiVal)
```
Fletterio:
```cpp
mergeStage(..., loPair, hiPair)
```

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,81 @@
...
+				// Local compare and swap for stage 0
+				nbl::hlsl::bitonic_sort::compareSwap(bitonicAscending, loKey, hiKey, loVal, hiVal, comp);
+			}
+			else
+			{
+				// Shuffle from partner using XOR
+				const key_t pLoKey = glsl::subgroupShuffleXor<key_t>(loKey, threadStride);
+				const key_t pHiKey = glsl::subgroupShuffleXor<key_t>(hiKey, threadStride);
+				const value_t pLoVal = glsl::subgroupShuffleXor<value_t>(loVal, threadStride);
+				const value_t pHiVal = glsl::subgroupShuffleXor<value_t>(hiVal, threadStride);
```
Fletterio: When you make this code work with pairs, you might want to do it like
```cpp
pair pLoPair, pHiPair;
pLoPair.key = glsl::subgroupShuffleXor<key_t>(loPair.key, threadStride);
pHiPair.key = glsl::subgroupShuffleXor<key_t>(hiPair.key, threadStride);
pLoPair.value = glsl::subgroupShuffleXor<value_t>(loPair.value, threadStride);
pHiPair.value = glsl::subgroupShuffleXor<value_t>(hiPair.value, threadStride);
```

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,81 @@
...
+
+				const bool isUpper = bool(invocationID & threadStride);
+				const bool takeLarger = isUpper == bitonicAscending;
+
+				nbl::hlsl::bitonic_sort::compareExchangeWithPartner(takeLarger, loKey, pLoKey, hiKey, pHiKey, loVal, pLoVal, hiVal, pHiVal, comp);
+			}
+		}
+	}
+
+	static void __call(bool ascending, NBL_REF_ARG(key_t) loKey, NBL_REF_ARG(key_t) hiKey,
```
Fletterio:
```cpp
__call(ascending, loPair, hiPair)
```

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,231 @@
...
+            shuffleXor(pLoPair, threadStride, sharedmemAccessor);
+            sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
+
+            nbl::hlsl::pair<key_t, value_t> pHiPair = hiPair;
+            shuffleXor(pHiPair, threadStride, sharedmemAccessor);
+
+            const bool isUpper = (invocationID & threadStride) != 0;
+            const bool takeLarger = isUpper == bitonicAscending;
+
+            nbl::hlsl::bitonic_sort::compareExchangeWithPartner(takeLarger, loPair, pLoPair, hiPair, pHiPair, comp);
```
Fletterio: you're using the shared memory accessor here, when you should be using an adaptor. The user only has to pass a generic sharedmem accessor, then the adaptor ensures its accesses are optimal. If changing accessor for adaptors is what breaks your code, we can look into making a specialized sharedmem adaptor for pairs.

Fletterio: Otherwise, we're passing the burden of writing an optimal accessor to the user, which requires the user knowing the underlying impl

Fletterio: Hold this one until Matt replies because the adaptor needs to change here

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
devshgraphicsprogramming: you can't define an `operator=` in HLSL none of the operators which should return references can be defined in HLSL so assignment and array indexing (as well as compound assignment)

However all structs in HLSL are trivial so can be assigned with `=`

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+                    accessor.template get<nbl::hlsl::pair<key_t, value_t> >(hiIx, hipair);
+
+                    swap(lopair, hipair);
+
+                    accessor.template set<nbl::hlsl::pair<key_t, value_t> >(loIx, lopair);
+                    accessor.template set<nbl::hlsl::pair<key_t, value_t> >(hiIx, hipair);
+                }
+                sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
+                }
+            }
```
Fletterio: Feels overkill to do this. Instead of having to go over half the array to swap it around just make the `E=2` struct specialization take an `ascending` parameter like you did for the subgroup sort. Set it to true by default, and in the calls above you need in this `E > 2` struct pass `ascending = !(WorkgroupID & 1)` to ensure every even workgroup is sorted ascendingly and every odd workgroup is sorted descendingly. It's the same you did for the subgroup case.

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,83 @@
...
+namespace subgroup
+{
+
+template<typename KeyType, typename ValueType, typename Comparator = less<KeyType> >
+struct bitonic_sort_config
+{
+	using key_t = KeyType;
+	using value_t = ValueType;
+	using comparator_t = Comparator;
+};
```
devshgraphicsprogramming: where's the SubgroupSizeLog2 ?

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,83 @@
...
+
+				nbl::hlsl::bitonic_sort::compareExchangeWithPartner(takeLarger, loPair, partnerLoPair, hiPair, partnerHiPair, comp);
+			}
+		}
+	}
+
+	static void __call(bool ascending, NBL_REF_ARG(pair<key_t, value_t>) loPair, NBL_REF_ARG(pair<key_t, value_t>) hiPair)
+	{
+		const uint32_t invocationID = glsl::gl_SubgroupInvocationID();
+		const uint32_t subgroupSizeLog2 = glsl::gl_SubgroupSizeLog2();
```
devshgraphicsprogramming: needs to be compile time constant from config like the FFT

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,83 @@
...
+namespace nbl
+{
+namespace hlsl
+{
+namespace subgroup
+{
+
+template<typename KeyType, typename ValueType, typename Comparator = less<KeyType> >
+struct bitonic_sort_config
+{
```
devshgraphicsprogramming: you also want to handle "multiple items per thread" in the subgroup sort

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,53 @@
...
+{
+	const bool shouldSwap = comp(hiPair.first, loPair.first);
+	const bool doSwap = (shouldSwap == ascending);
+
+	if (doSwap)
+		swap(loPair, hiPair);
+}
+}
+}
+}
```
devshgraphicsprogramming: write a TODO to make a struct which can be specialized by declaring a PoT sorting network within a thread

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,53 @@
...
+namespace bitonic_sort
+{
+
+template<typename KeyType, typename ValueType, typename Comparator>
+void compareExchangeWithPartner(
+	bool takeLarger,
+	NBL_REF_ARG(pair<KeyType, ValueType>) loPair,
+	NBL_CONST_REF_ARG(pair<KeyType, ValueType>) partnerLoPair,
+	NBL_REF_ARG(pair<KeyType, ValueType>) hiPair,
+	NBL_CONST_REF_ARG(pair<KeyType, ValueType>) partnerHiPair,
```
devshgraphicsprogramming: don't do everything for `pair<Key,Value>` but a `KeyValue`  on which `get<0,KeyValue>(kv)` and `get<0,KeyValue>(kv)` can be called (instead of `.first` and `.second`).

And then you _can trivially allow_ `pair<K,V>` be used as `KeyValue`

Why? You will have the option of "packing" the key or value into the would-be padding bits of each other (when SoA isn't possible or beneficial) and doing all sorts of other tricks

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
devshgraphicsprogramming: example
```cpp
template<typename KeyValue, uint32_t Log2N>
struct SortingNetwork
{
   void operator()(const bool ascending, NBL_REF_ARG(array<KeyValue,0x1u<<Log2N>) data, NBL_CONST_REF_ARG(Comparator) comp);
};
```
and then you can partially specialize it for all `Log2N` from 1 to 3 or 4

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,53 @@
...
+	NBL_CONST_REF_ARG(Comparator) comp)
+{
+	const bool loSelfSmaller = comp(loPair.first, partnerLoPair.first);
+	const bool takePartnerLo = takeLarger ? loSelfSmaller : !loSelfSmaller;
```
devshgraphicsprogramming: there's a faster way to write this `loSelfSmaller==takeLarger`, look at a T/F table of your current expression

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,53 @@
...
+	if (takePartnerLo)
+		loPair = partnerLoPair;
+
+	const bool hiSelfSmaller = comp(hiPair.first, partnerHiPair.first);
+	const bool takePartnerHi = takeLarger ? hiSelfSmaller : !hiSelfSmaller;
```
devshgraphicsprogramming: there's a faster way to write this `hiSelfSmaller==takeLarger`, look at a T/F table of your current expression

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,83 @@
...
+
+	static void mergeStage(uint32_t stage, bool bitonicAscending, uint32_t invocationID,
+		NBL_REF_ARG(pair<key_t, value_t>) loPair, NBL_REF_ARG(pair<key_t, value_t>) hiPair)
+	{
+		comparator_t comp;
+
+		[unroll]
+		for (uint32_t pass = 0; pass <= stage; pass++)
+		{
+			const uint32_t stride = 1u << (stage - pass); // Element stride
```
devshgraphicsprogramming: why not write the loop as `for (uint32_t stride=1u<<stage; stride; stride=stride>>1)`  and forget the `pass` ?

include/nbl/builtin/hlsl/concepts/accessors/bitonic_sort.hlsl
```diff
@@ -0,0 +1,31 @@
...
+//  * void set(uint32_t index, in uint32_t value);
+//  * void workgroupExecutionAndMemoryBarrier();
+template<typename T, typename V = uint32_t, typename I = uint32_t>
+NBL_BOOL_CONCEPT BitonicSortSharedMemoryAccessor = concepts::accessors::GenericSharedMemoryAccessor<T, V, I>;
+
+// The Accessor MUST provide the following methods:
+//  * void get(uint32_t index, NBL_REF_ARG(pair<KeyType, ValueType>) value);
+//  * void set(uint32_t index, in pair<KeyType, ValueType> value);
+template<typename T, typename KeyType, typename ValueType, typename I = uint32_t>
+NBL_BOOL_CONCEPT BitonicSortAccessor = concepts::accessors::GenericDataAccessor<T, pair<KeyType, ValueType>, I>;
```
devshgraphicsprogramming, line 25: `bitonic_sort::BitonicSort...` is a tautology, drop the `BitonicSort` prefix from the names

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
devshgraphicsprogramming: furthemore to allow working on more than 2 elements per thread the signature should really just be `bool, NBL_REF_ARG(KeyValue) data[ElementsPerThread])` 

can do `NBL_REF_ARG(array<U,ElementsPerThread>)` if DXC gives you pain on array types

devshgraphicsprogramming: btw its fine to not rewrite the function and `static_assert(ElementsPErThread==2)` for now


or just provide the partial spec for `ElementsPerThread==2` (since static assert gives trouble in DXC SPIR-V backend) but then you need to work it into a struct functor

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
```diff
@@ -0,0 +1,53 @@
...
+		hiPair = partnerHiPair;
+}
+
+template<typename KeyType, typename ValueType, typename Comparator>
+void compareSwap(
+	bool ascending,
+	NBL_REF_ARG(pair<KeyType, ValueType>) loPair,
+	NBL_REF_ARG(pair<KeyType, ValueType>) hiPair,
+	NBL_CONST_REF_ARG(Comparator) comp)
+{
```
devshgraphicsprogramming: transform into
```cpp
template<typename KeyValue, uint32_t Log2N>
struct LocalPasses
{
   void operator()(const bool ascending, NBL_REF_ARG(array<KeyValue,0x1u<<Log2N>) data, NBL_CONST_REF_ARG(Comparator) comp);
};
```

and make your current implementation a partial spec for `LocalPasses<KeyValue,2>`

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+	}
+
+	static void __call(bool ascending, NBL_REF_ARG(pair<key_t, value_t>) loPair, NBL_REF_ARG(pair<key_t, value_t>) hiPair)
+	{
+		const uint32_t invocationID = glsl::gl_SubgroupInvocationID();
+		const uint32_t subgroupSizeLog2 = glsl::gl_SubgroupSizeLog2();
+		[unroll]
+		for (uint32_t stage = 0; stage <= subgroupSizeLog2; stage++)
```
devshgraphicsprogramming: actually I'd ask for an even cooler feature, would be nice if `stage<=sortSizeLog2` and `__call` had a last argument with a default `const uint32_t sortSizeLog2=Config::SubgroupSizeLog2`

Why? Because it would be super useful for things like sorting arrays which are smaller than what a subgroup can process (so we can pack multiple arrays to be sorted independently into a single subgroup)

devshgraphicsprogramming: the `const bool bitonicAscending = (stage == subgroupSizeLog2) ?` would have to change to `const bool bitonicAscending = (stage == sortSizeLog2 ) ?`

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,83 @@
...
+				// Local compare and swap for stage 0
+				nbl::hlsl::bitonic_sort::compareSwap(bitonicAscending, loPair, hiPair, comp);
+			}
+			else
+			{
+				// Shuffle from partner using XOR
+				const key_t pLoKey = glsl::subgroupShuffleXor<key_t>(loPair.first, threadStride);
+				const value_t pLoVal = glsl::subgroupShuffleXor<value_t>(loPair.second, threadStride);
+				const key_t pHiKey = glsl::subgroupShuffleXor<key_t>(hiPair.first, threadStride);
+				const value_t pHiVal = glsl::subgroupShuffleXor<value_t>(hiPair.second, threadStride);
```
devshgraphicsprogramming: here's a problem with `subgroupShuffle` it only works on scalars and vectors.

Now why would we use a Comparison Based Sort over a Counting Sort?

Because the Key is really long in relation to number of elements to sort OR only supports custom comparison and can't be represented as a ordered number.

Therefore its most useful whenever the `Key` is a `struct`.

You need to document and require (conceptualize) that your `KeyValue` needs to have a `shuffle_type` typedef which is a `vector<fundamental_type,N>` and the `KeyValue` can be `static_cast`-ed to and from `shuffle_type`, you won't be able to just `subgroupShuffle` a `pair<Key,Value>` or anything similar

devshgraphicsprogramming: Actually drop the `static_cast` if we had a stateful converter then we could avoid doing silly things

Why are we even trading/shuffling a full value type? when all that we need to do is give each Kev a surrogate Value which is just `subgroupInvocationIndex` which tells us how to shuffle at the very end.

include/nbl/builtin/hlsl/utility.hlsl
```diff
@@ -1,40 +1,70 @@
...
+
+template<typename T1, typename T2>
+struct pair
+{
+	using first_type = T1;
+	using second_type = T2;
+
+	first_type first;
+	second_type second;
+};
```
devshgraphicsprogramming, line 24: want a `get` and a `get_helper` struct in this header, plus a particla spec for `pair` in this header

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+            [unroll]
+            for (uint32_t blockLog = simpleLog + 1u; blockLog <= totalLog; blockLog++)
+            {
+                const uint32_t k = 1u << blockLog;
+                [unroll]
+                for (uint32_t strideLog = blockLog - 1u; strideLog + 1u > 0u; strideLog--)
+                {
+                    const uint32_t stride = 1u << strideLog;
+                    [unroll]
+                    for (uint32_t virtualThreadID = threadID; virtualThreadID < TotalElements / 2; virtualThreadID += WorkgroupSize)
```
devshgraphicsprogramming: this won't unroll, but this will
```cpp
for (uint32_t baseVirtualThread=0; baseVirtualThread<HalfTotalElements; baseVirtualThread+=WorkgroupSize)
{
  const uint32_t virtualThreadID = baseVirtualThread+threadID;
```

the compiler now see that the loop condition is invariant (otherwise you have a dep on `threadID` and netiher DXC or SPIR-V opt can do the long and ardous analysis to see if your workgroup count is a multiple and the loop condition is actually uniform for whole group always)

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+namespace nbl
+{
+namespace hlsl
+{
+namespace workgroup
+{
+namespace bitonic_sort
+{
+
+template<uint16_t _ElementsPerInvocationLog2, uint16_t _WorkgroupSizeLog2, typename KeyType, typename ValueType, typename Comparator = less<KeyType> NBL_PRIMARY_REQUIRES(_ElementsPerInvocationLog2 >= 1 && _WorkgroupSizeLog2 >= 5)
```
devshgraphicsprogramming: you should just require that `_WorkgroupSizeLog2>SubgroupSizeLog2` I see you've hard coded an assumption that Subgroups are Size 32, the subgroup size needs to come into the config too

devshgraphicsprogramming: There should also be a general `TotalElementsLog2>SubgroupSizeLog2+ElementsPerInvocationLog2` then this tells us our `VirtualWorkgroupSizeLog2`

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+{
+    using key_t = KeyType;
+    using value_t = ValueType;
+    using comparator_t = Comparator;
+    NBL_CONSTEXPR_STATIC_INLINE uint16_t ElementsPerInvocationLog2 = _ElementsPerInvocationLog2;
+    NBL_CONSTEXPR_STATIC_INLINE uint16_t WorkgroupSizeLog2 = _WorkgroupSizeLog2;
+
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t ElementsPerInvocation = 1u << ElementsPerInvocationLog2;
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t WorkgroupSize = 1u << WorkgroupSizeLog2;
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t SharedmemDWORDs = sizeof(pair<key_t, value_t>) / sizeof(uint32_t) * WorkgroupSize;
```
devshgraphicsprogramming: `SharedMemorySize` should come from outside, and it must be as big as `sizeof(KeyValue::workgroup_type)<<WorkgroupSizeLog2`

the idea is that if we have more shared memory, then we issue less barriers by:
1. exchanging more items per barrier (up to `0x1u<<(WorkgroupSizeLog2+ElementPerInvocationLog2)`)
2. cycling through offsets within the shared memory to barrier every so often (every time you wrap around)

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
devshgraphicsprogramming: already pointed it out https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/209#discussion_r2318294704

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+};
+}
+
+template<typename Config, class device_capabilities = void>
+struct BitonicSort;
+
+// ==================== ElementsPerThreadLog2 = 1 Specialization (No Virtual Threading) ====================
+// This handles arrays of size WorkgroupSize * 2 using subgroup + workgroup operations
+template<uint16_t WorkgroupSizeLog2, typename KeyType, typename ValueType, typename Comparator, class device_capabilities>
+struct BitonicSort<bitonic_sort::bitonic_sort_config<1, WorkgroupSizeLog2, KeyType, ValueType, Comparator>, device_capabilities>
```
devshgraphicsprogramming: @Fletterio I think `device_capabilities` aren't really needed for anything here or the FFT because we baiscally require Shuffle operations, unlike where for the Prefix sum we use them to decide between native `subgroupAdd` and a thing made with subgroup shuffles

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+    {
+        keyAdaptor.template set<Key>(ownedIdx, p.first);
+        valueAdaptor.template set<Value>(ownedIdx, p.second);
+
+        // Wait until all writes are done before reading - only barrier on one adaptor here
+        keyAdaptor.workgroupExecutionAndMemoryBarrier();
+
+        keyAdaptor.template get<Key>(ownedIdx ^ mask, p.first);
+        valueAdaptor.template get<Value>(ownedIdx ^ mask, p.second);
+    }
```
devshgraphicsprogramming: @Fletterio an XOR shuffle of a struct (single accessor) within a workgroup should be commonalized somewhere cause its used in FFT and here

Fletterio: I'll try making a generic version but memory layouts fuck me up a bit. I'll iterate on a version.

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+
+                    accessor.template set<nbl::hlsl::pair<key_t, value_t> >(loIx, lopair);
+                    accessor.template set<nbl::hlsl::pair<key_t, value_t> >(hiIx, hipair);
+                }
+                sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
+                }
+            }
+
+            // PHASE 3: Global memory bitonic merge
+            const uint32_t totalLog = hlsl::findMSB(TotalElements - 1) + 1u;
```
devshgraphicsprogramming: `TotalElements` should already be passed as `Log2`

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+        {
+            if (sub)
+                sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
+
+                offsetAccessor.offset = sub * ElementsPerSimpleSort;
+
+                // Call E=1 workgroup sort
+                BitonicSort<simple_config_t, device_capabilities>::template __call(offsetAccessor, sharedmemAccessor);
+        }
+        sharedmemAccessor.workgroupExecutionAndMemoryBarrier();
```
devshgraphicsprogramming: put the barrier at the end of the loop then you don't need `if(sub)` and the barrier after the loop

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+            nbl::hlsl::bitonic_sort::compareExchangeWithPartner(takeLarger, lopair, plopair, hipair, phipair, comp);
+        }
+    }
+
+    template<typename Accessor, typename SharedMemoryAccessor NBL_FUNC_REQUIRES(bitonic_sort::BitonicSortAccessor<Accessor, key_t, value_t>&& bitonic_sort::BitonicSortSharedMemoryAccessor<SharedMemoryAccessor>)
+    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
+    {
+        const uint32_t WorkgroupSize = config_t::WorkgroupSize;
+
+        const uint32_t invocationID = glsl::gl_LocalInvocationID().x;
```
devshgraphicsprogramming: oh btw the Spec doesn't say that `LocalInvocationIndex = SubgroupID*SubgroupSize+SubgroupInvocationID`

This is why in such shaders you need to always use https://github.com/Devsh-Graphics-Programming/Nabla/blob/b7f3dfb7ed197b90e5e1877b97a5cfe5e7b62d6f/include/nbl/builtin/hlsl/workgroup/basic.hlsl#L26

instead of `gl_LocalInvocationID` or `gl_LocalInvocationIndex`

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+        const uint32_t subgroupSizeLog2 = glsl::gl_SubgroupSizeLog2();
+        const uint32_t subgroupSize = 1u << subgroupSizeLog2;
+        const uint32_t subgroupID = glsl::gl_SubgroupID();
+        const uint32_t numSubgroups = WorkgroupSize / subgroupSize;
```
devshgraphicsprogramming: there's actually a GLSL builtin for this

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+    using value_t = ValueType;
+    using comparator_t = Comparator;
+
+    template<typename Accessor, typename SharedMemoryAccessor>
+    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
+    {
+        const uint32_t WorkgroupSize = config_t::WorkgroupSize;
+        const uint32_t ElementsPerThread = config_t::ElementsPerInvocation;
+        const uint32_t TotalElements = WorkgroupSize * ElementsPerThread;
+        const uint32_t ElementsPerSimpleSort = WorkgroupSize * 2;
```
devshgraphicsprogramming: `ElementsPerThread` should be the number of elements each thread operates on, the multiplier for Virtual Threading (what your current `ElementsPerThread` acts as) should be derived from `TotalElementsLog2-WorkgroupSizeLog2-ElementsPerThreadLog2`

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+    static void __call(NBL_REF_ARG(Accessor) accessor, NBL_REF_ARG(SharedMemoryAccessor) sharedmemAccessor)
+    {
+        const uint32_t WorkgroupSize = config_t::WorkgroupSize;
+
+        const uint32_t invocationID = glsl::gl_LocalInvocationID().x;
+        const uint32_t subgroupSizeLog2 = glsl::gl_SubgroupSizeLog2();
+        const uint32_t subgroupSize = 1u << subgroupSizeLog2;
+        const uint32_t subgroupID = glsl::gl_SubgroupID();
+        const uint32_t numSubgroups = WorkgroupSize / subgroupSize;
+        const uint32_t numSubgroupsLog2 = findMSB(numSubgroups);
```
devshgraphicsprogramming: it would be cool derive this as `TotalElementsLog2-SubgroupSizeLog2-ElementsPerThreadLog2` this would allow us to do workgroup partitioned sorts, eg:
- sort 4x256 with a 512 workgroup elements per thread 2

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
devshgraphicsprogramming: in view of https://github.com/Devsh-Graphics-Programming/Nabla/pull/943#discussion_r2513151011 you could actually count VirtualSubgroups

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
devshgraphicsprogramming: are you sure this is correct?
https://github.com/Devsh-Graphics-Programming/Nabla/pull/940#discussion_r2411943639

if I myself am smaller, then `comp` which is `<` will return true, but  I will swap with partner when `takeLarger` is true

did you mess up variable naming here?

include/nbl/builtin/hlsl/subgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,83 @@
...
+
+				const pair<key_t, value_t> partnerLoPair = make_pair(pLoKey, pLoVal);
+				const pair<key_t, value_t> partnerHiPair = make_pair(pHiKey, pHiVal);
+
+				const bool isUpper = bool(invocationID & threadStride);
+				const bool takeLarger = isUpper == bitonicAscending;
+
+				nbl::hlsl::bitonic_sort::compareExchangeWithPartner(takeLarger, loPair, partnerLoPair, hiPair, partnerHiPair, comp);
```
devshgraphicsprogramming: comments throughout code needed about this stuff https://github.com/Devsh-Graphics-Programming/Nabla/pull/940#discussion_r2411941883

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
devshgraphicsprogramming: oh yeah with shuffling `KeyValue` and leaving the SoA/AoS bank-conflict decomposition to userspace Accessor allows us to use the workgroup shuffleXoR https://github.com/Devsh-Graphics-Programming/Nabla/pull/940#discussion_r2443602236

its probably worth extending that thing to handle Virtual Workgroups (when you have more shared memory and can do multiple shuffles without a barrier in the middle

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t ElementsPerInvocation = 1u << ElementsPerInvocationLog2;
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t WorkgroupSize = 1u << WorkgroupSizeLog2;
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t SharedmemDWORDs = sizeof(pair<key_t, value_t>) / sizeof(uint32_t) * WorkgroupSize;
+
+};
+}
+
+template<typename Config, class device_capabilities = void>
+struct BitonicSort;
```
devshgraphicsprogramming: its declaraed in a tautologic namespace, as @Fletterio mentioned we don't want `bitonic_sort::BitonicSort`

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+            nbl::hlsl::pair<key_t, value_t> plopair = lopair;
+            shuffleXor(plopair, invocationID, threadStride, keyAdaptor, valueAdaptor);
+
+            nbl::hlsl::pair<key_t, value_t> phipair = hipair;
+            shuffleXor(phipair, invocationID ^ threadStride, threadStride, keyAdaptor, valueAdaptor);
+
+            const bool isUpper = (invocationID & threadStride) != 0;
+            const bool takeLarger = isUpper == bitonicAscending;
+
+            nbl::hlsl::bitonic_sort::compareExchangeWithPartner(takeLarger, lopair, plopair, hipair, phipair, comp);
```
devshgraphicsprogramming: you're doing double the work... unlike subgroups where an invocation can't write into its  partner's register. here you can definitely have one thread read both items, and write both items

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
devshgraphicsprogramming: doesn't seem to be done and is important, but maybe lets try the Subgroup Transpose trick

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
devshgraphicsprogramming: > Otherwise, we're passing the burden of writing an optimal accessor to the user, which requires the user knowing the underlying impl

yep

include/nbl/builtin/hlsl/workgroup/bitonic_sort.hlsl
```diff
@@ -0,0 +1,253 @@
...
+        comparator_t comp;
+
+
+        using key_adaptor_t = accessor_adaptors::StructureOfArrays<SharedMemoryAccessor, uint32_t, uint32_t, 1, WorkgroupSize>;
+        using value_adaptor_t = accessor_adaptors::StructureOfArrays<SharedMemoryAccessor, uint32_t, uint32_t, 1, WorkgroupSize, integral_constant<uint32_t, WorkgroupSize * sizeof(key_t) / sizeof(uint32_t)> >;
+
+        key_adaptor_t keyAdaptor;
+        keyAdaptor.accessor = sharedmemAccessor;
+        value_adaptor_t valueAdaptor;
+        valueAdaptor.accessor = sharedmemAccessor;
```
devshgraphicsprogramming: can skip the StructureOfArrays in this code, more useful to use it in userspace , see https://github.com/Devsh-Graphics-Programming/Nabla/pull/943#discussion_r2514393464

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
devshgraphicsprogramming: `operator=` can only be defined in a `#ifndef __HLSL_VERSION` macro block, for DXC reasons

include/nbl/builtin/hlsl/bitonic_sort/common.hlsl
devshgraphicsprogramming: I've just talked with @Fletterio and we figured out how to trade subgroup variables to not have to perform redundant comparisons, this means 4-way Key-Value exchanges per thread won't be necessary and this function can disappear

PR #944, opened 2025-10-29
Precompiled shaders

Przemog1:


cmake/common.cmake
```diff
@@ -1178,9 +1184,34 @@ struct DeviceConfigCaps
 		-enable-16bit-types 
 		-Zpr 
 		-spirv 
-		-fspv-target-env=vulkan1.3
+		-fspv-target-env=vulkan1.3 
+		-WShadow 
+		-WConversion 
+		$<$<CONFIG:Debug>:-O0> 
+		$<$<CONFIG:Release>:-O3>
```
AnastaZIuk, line 1191: you can turn it into `$<$<CONFIG:Release,RelWithDebInfo>:-O3>`

PR #960, opened 2025-12-02
New quaternion in HLSL

keptsecret:


include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+        q.data = data_type(0.0, 0.0, 0.0, 1.0);
+        return q;
+    }
+    
+    static this_t create(scalar_type x, scalar_type y, scalar_type z, scalar_type w)
+    {
+        this_t q;
+        q.data = data_type(x, y, z, w);
+        return q;
+    }
```
devshgraphicsprogramming: imho this is more painful than just 
```cpp
math::quaternion q;
q.data = vector_of_xyzw;
```
delete this pseudo-constructor

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+
+    static this_t create(NBL_CONST_REF_ARG(this_t) other)
+    {
+        return other;
+    }
+
+    // angle: Rotation angle expressed in radians.
+    // axis: Rotation axis, must be normalized.
+    static this_t create(scalar_type angle, const vector3_type axis)
```
devshgraphicsprogramming: can we have axis first, then angle ?

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+    {
+        this_t q;
+        q.data = data_type(x, y, z, w);
+        return q;
+    }
+
+    static this_t create(NBL_CONST_REF_ARG(this_t) other)
+    {
+        return other;
+    }
```
devshgraphicsprogramming: we don't need this

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+    using matrix_type = matrix<T, 3, 3>;
+
+    using AsUint = typename unsigned_integer_of_size<sizeof(scalar_type)>::type;
+
+    static this_t create()
+    {
+        this_t q;
+        q.data = data_type(0.0, 0.0, 0.0, 1.0);
+        return q;
+    }
```
devshgraphicsprogramming, line 51: i think the `create` overloads need to have their argument types templated with `NBL_FUNC_REQUIRES` because DXC will screw us over with implicit conversions

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: btw you can also throw in a `uniformScale = 1.f` into the mix :P

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+        const scalar_type cy = hlsl::cos(yawDiv2);
+
+        this_t output;
+        output.data[0] = cr * sp * cy + sr * cp * sy; // x
+        output.data[1] = cr * cp * sy - sr * sp * cy; // y
+        output.data[2] = sr * cp * cy - cr * sp * sy; // z
+        output.data[3] = cr * cp * cy + sr * sp * sy; // w
+
+        return output;
+    }
```
devshgraphicsprogramming: I'd break this down into two functions
```cpp
template<typename T=vector<scalar_type,2> > requires is_same_v<vector<scalar_type,2>,T>
static this_t create(const T halfPitchCosSin, const T halfYawCosSin, const T halfRollCosSin);

template<typename T=scalar_type> requires is_same_v<scalar_type,T>
static this_t create(const T pitch, const T yaw, const T roll);
```

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+                retval.data.x = (m[0][2] + m[2][0]) * invscales.w;
+                retval.data.y = (m[1][2] + m[2][1]) * invscales.w;
+                retval.data.z = scales.w;
+                retval.data.w = (m[1][0] - m[0][1]) * invscales.w;
+            }
+        }
+
+        retval.data = hlsl::normalize(retval.data);
+        return retval;
+    }
```
devshgraphicsprogramming: only an orthogonal (no skew) and uniform scale matrix can be converted to a quaternion.

so dot products between the rows and columns should be 0, and the dot product of each column with itself should have the same length

You can write the whole algorithm assuming this, so invscales would be a scalar.

You could assert the orthogonality (no skew) and uniform scaling property OR take an extra `const bool dontAssertValidMatrix=false` where instead of asserting you return a NaN filled quaternion if the matrix is not valid

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+
+    static this_t create(NBL_CONST_REF_ARG(truncated_quaternion<T>) first3Components)
+    {
+        this_t retval;
+        retval.data.xyz = first3Components.data;
+        retval.data.w = hlsl::sqrt(scalar_type(1.0) - hlsl::dot(first3Components.data, first3Components.data));
+        return retval;
+    }
```
devshgraphicsprogramming: put this wholly inside the `static_cast` from truncated to full quaternion

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+        );
+    }
+
+    static this_t lerp(const this_t start, const this_t end, const scalar_type fraction, const scalar_type totalPseudoAngle)
+    {
+        const AsUint negationMask = hlsl::bit_cast<AsUint>(totalPseudoAngle) & AsUint(0x80000000u);
+        const data_type adjEnd = hlsl::bit_cast<scalar_type>(hlsl::bit_cast<AsUint>(end.data) ^ negationMask);
+
+        this_t retval;
+        retval.data = hlsl::mix(start.data, adjEnd, fraction);
```
devshgraphicsprogramming: Please unit test/check it somehow (re-derive) I'm not sure that when `totalPseudoAngle<0` the result is meant to be `S+(-E-S)*fraction` and not `S+(E-S)*(-fraction)

P.S. btw my recent learnings about GPU architectures have made me aware that there's more FP32 cores than INT32 on Nvidia so write a comment to benchmark the sign xor flip against just `*sign(totalPseudoAngle)`

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+    {
+        const scalar_type pseudoAngle = hlsl::dot(start.data,end.data);
+        const scalar_type interpolantPrecalcTerm = fraction - scalar_type(0.5);
+        const scalar_type interpolantPrecalcTerm3 = fraction * interpolantPrecalcTerm * (fraction - scalar_type(1.0));
+        const scalar_type adjFrac = __adj_interpolant(hlsl::abs(pseudoAngle),fraction,interpolantPrecalcTerm*interpolantPrecalcTerm,interpolantPrecalcTerm3);
+        
+        this_t retval = lerp(start,end,adjFrac,pseudoAngle);
+        retval.data = hlsl::normalize(retval.data);
+        return retval;
+    }
```
devshgraphicsprogramming, line 258: write a comment that lerp does not resepect scales, can stick an assert in `lerp` that `start` and `end` are both unit length

also make behaviour consistent, either lerp and flerp both normalize or both don't normalize the result (in case you go with normalization, make `unnormLerp` and `unnormFlerp`)

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+
+    vector3_type transformVector(const vector3_type v)
+    {
+        scalar_type scale = hlsl::length(data);
+        vector3_type direction = data.xyz;
+        return v * scale + hlsl::cross(direction, v * data.w + hlsl::cross(direction, v)) * scalar_type(2.0);
+    }
```
devshgraphicsprogramming: add a `const bool assumeNoScale=false` argument

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+        return precompPart * cosAngle + hlsl::cross(planeNormal, precompPart);
+    }
+
+    this_t inverse()
+    {
+        this_t retval;
+        retval.data.x = bit_cast<scalar_type>(bit_cast<AsUint>(data.x)^0x80000000u);
+        retval.data.y = bit_cast<scalar_type>(bit_cast<AsUint>(data.y)^0x80000000u);
+        retval.data.z = bit_cast<scalar_type>(bit_cast<AsUint>(data.z)^0x80000000u);
+        retval.data.w = data.w;
```
devshgraphicsprogramming: `retval.data.xyz = -retval.data.xyz` is more readable and compiler should get it

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+        return retval;
+    }
+
+    static this_t normalize(NBL_CONST_REF_ARG(this_t) q)
+    {
+        this_t retval;
+        retval.data = hlsl::normalize(q.data);
+        return retval;
+    }
```
devshgraphicsprogramming: specialize `normalize_helper` from intrinsics instead

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+    {
+        matrix_type mat;
+        mat[0] = data.yzx * data.ywz + data.zxy * data.zyw * vector3_type( 1.0, 1.0,-1.0);
+        mat[1] = data.yzx * data.xzw + data.zxy * data.wxz * vector3_type(-1.0, 1.0, 1.0);
+        mat[2] = data.yzx * data.wyx + data.zxy * data.xwy * vector3_type( 1.0,-1.0, 1.0);
+        mat[0][0] = scalar_type(0.5) - mat[0][0];
+        mat[1][1] = scalar_type(0.5) - mat[1][1];
+        mat[2][2] = scalar_type(0.5) - mat[2][2];
+        mat *= scalar_type(2.0);
+        return hlsl::transpose(mat);    // TODO: double check transpose?
```
devshgraphicsprogramming: yes please double check the tranpose, best add some unit tests to example 22 where random vectors get rotated by random quaternions, and you check the result is same as making a mat3 out of the quaternion and `mul(matrix,inputVector)`

devshgraphicsprogramming: @Przemog1 can advise/help

Przemog1: sure, also example 22 doesn't compile on master currently, for now just comment out the code that doesn't work

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    vector3_type transformVector(const vector3_type v)
+    {
+        scalar_type scale = hlsl::length(data);
+        vector3_type direction = data.xyz;
+        return v * scale + hlsl::cross(direction, v * data.w + hlsl::cross(direction, v)) * scalar_type(2.0);
+    }
+
+    matrix_type constructMatrix()
```
Przemog1: some functions in this class could be const member functions, just add NBL_CONST_MEMBER_FUNC

include/nbl/builtin/hlsl/math/quaternions.hlsl
keptsecret: keep this `create`?

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -48,22 +48,10 @@ struct quaternion
...
-
-    static this_t create(NBL_CONST_REF_ARG(this_t) other)
-    {
-        return other;
-    }
 
     // angle: Rotation angle expressed in radians.
     // axis: Rotation axis, must be normalized.
-    static this_t create(scalar_type angle, const vector3_type axis)
+    static this_t create(const vector3_type axis, scalar_type angle)
```
devshgraphicsprogramming: add a `const float uniformScale = 1.f` argument which basically multiples the whole `q.data`

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: ping

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -72,32 +60,78 @@ struct quaternion
...
+
+    static bool __isEqual(const scalar_type a, const scalar_type b)
+    {
+        return hlsl::max(a/b, b/a) <= scalar_type(1e-4);
+    }
+    static bool __dotIsZero(const vector3_type a, const vector3_type b)
+    {
+        const scalar_type ab = hlsl::dot(a, b);
+        return hlsl::abs(ab) <= scalar_type(1e-4);
     }
```
devshgraphicsprogramming: this needs go somewhere in a common header, can't have it in every class

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -72,32 +60,78 @@ struct quaternion
...
+            valid = __dotIsZero(m[1], m[2]) && valid;
+            valid = __dotIsZero(m[0], m[2]) && valid;
+
+            const matrix_type m_T = hlsl::transpose(m);
+            const scalar_type dotCol0 = hlsl::dot(m_T[0],m_T[0]);
+            const scalar_type dotCol1 = hlsl::dot(m_T[1],m_T[1]);
+            const scalar_type dotCol2 = hlsl::dot(m_T[2],m_T[2]);
+            valid = __isEqual(dotCol0, dotCol1) && valid;
+            valid = __isEqual(dotCol1, dotCol2) && valid;
+            valid = __isEqual(dotCol0, dotCol2) && valid;
```
devshgraphicsprogramming: you need to pack this up into `nbl::hlsl::math::linalg` `struct RuntimeTraits<SquareMatrix>` which has bool members for:
- invertible/nonZeroDeterminant
- orthogonal
- uniform scale
- orthonormal

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -72,32 +60,78 @@ struct quaternion
...
+                    retval.data = hlsl::promote<data_type>(bit_cast<scalar_type>(numeric_limits<scalar_type>::quiet_NaN));
+                    return retval;
+                }
+            else
+                assert(valid);
+        }
+
         const scalar_type m00 = m[0][0], m11 = m[1][1], m22 = m[2][2];
         const scalar_type neg_m00 = bit_cast<scalar_type>(bit_cast<AsUint>(m00)^0x80000000u);
         const scalar_type neg_m11 = bit_cast<scalar_type>(bit_cast<AsUint>(m11)^0x80000000u);
```
devshgraphicsprogramming: you can use unary `-`  no problem here, will be just as fast, but more legible

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -107,55 +141,49 @@ struct quaternion
...
-                retval.data.w = (m[1][0] - m[0][1]) * invscales.w;
+                const scalar_type invscales = scalar_type(0.5) / hlsl::sqrt(tmp.w);
+                retval.data.x = (m[0][2] + m[2][0]) * invscales;
+                retval.data.y = (m[1][2] + m[2][1]) * invscales;
+                retval.data.z = tmp.w * invscales * scalar_type(0.5);
+                retval.data.w = (m[1][0] - m[0][1]) * invscales;
             }
         }
 
         retval.data = hlsl::normalize(retval.data);
```
devshgraphicsprogramming: uniform scale needs to be restored from the matrix, so its the `inversesqrt(dot(m.anyColumn,m.anyColumn))`

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: no, I see you already did good.

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: @keptsecret did you re-derive ?

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -165,27 +193,36 @@ struct quaternion
...
             data.w * other.data.w - data.x * other.x - data.y * other.data.y - data.z * other.data.z,
             data.w * other.data.x + data.x * other.w + data.y * other.data.z - data.z * other.data.y,
             data.w * other.data.y - data.x * other.z + data.y * other.data.w + data.z * other.data.x,
             data.w * other.data.z + data.x * other.y - data.y * other.data.x + data.z * other.data.w
         );
+        return retval;
     }
 
-    static this_t lerp(const this_t start, const this_t end, const scalar_type fraction, const scalar_type totalPseudoAngle)
+    static this_t unnormLerp(const this_t start, const this_t end, const scalar_type fraction, const scalar_type totalPseudoAngle)
```
devshgraphicsprogramming, line 208: you want to `assert` both `start` and `end` have unit length, otherwise algo doesn't work

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -196,26 +233,32 @@ struct quaternion
         return fraction + interpolantPrecalcTerm3 * k;
     }
 
-    static this_t flerp(const this_t start, const this_t end, const scalar_type fraction)
+    static this_t unnormFlerp(const this_t start, const this_t end, const scalar_type fraction)
     {
         const scalar_type pseudoAngle = hlsl::dot(start.data,end.data);
```
devshgraphicsprogramming, line 244: you want to `assert` both `start` and `end` have unit length, otherwise algo doesn't work

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -196,26 +233,32 @@ struct quaternion
...
         return retval;
     }
 
-    vector3_type transformVector(const vector3_type v)
+    vector3_type transformVector(const vector3_type v, const bool assumeNoScale=false) NBL_CONST_MEMBER_FUNC
     {
-        scalar_type scale = hlsl::length(data);
+        scalar_type scale = hlsl::mix(hlsl::length(data), scalar_type(1.0), assumeNoScale);
         vector3_type direction = data.xyz;
         return v * scale + hlsl::cross(direction, v * data.w + hlsl::cross(direction, v)) * scalar_type(2.0);
```
devshgraphicsprogramming: almost correct, if you  have a scale:
```
| Ds x (Vsw + Ds x V)| = s^2 |D x (Vw + D x V)| 
```

you should compute the scale as `scaleRcp = inverseSqrt(dot(data,data))`, and then do:
```
vector3_type modV = v*2.f*scaleRcp;
return v/scaleRcp + hlsl::cross(direction, modV * data.w + hlsl::cross(direction,modV));
```

devshgraphicsprogramming: the choice of the location of the `2.f` multiplier is up to you really.

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -242,44 +287,63 @@ struct quaternion
...
+namespace cpp_compat_intrinsics_impl
+{
+template<typename T>
+struct normalize_helper<math::truncated_quaternion<T> >
+{
+    static inline math::truncated_quaternion<T> __call(const math::truncated_quaternion<T> q)
     {
-        this_t retval;
+        math::truncated_quaternion<T> retval;
         retval.data = hlsl::normalize(q.data);
```
devshgraphicsprogramming: truncated quaternion is normalized by definition (last component is dropped), this should be a NOOP

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: if you took this from old C++ code involving `matrixSIMD`, the `operator[]` gave rows, so if the tranpose was there in the old C++ it needs to be here cause HLSL gives rows too

But you can ofc rewrite this function (keep the old one for reference in the unit test) to work on a row basis.

or you can get rid of the whole `* vector3_type( 1.0,-1.0, 1.0);` thing, and just write out the swizzle element by element

include/nbl/builtin/hlsl/math/quaternions.hlsl
keptsecret: Yeah, should flip quaternion, not factor

include/nbl/builtin/hlsl/math/linalg/matrix_runtime_traits.hlsl
```diff
@@ -0,0 +1,66 @@
...
+            bool orthogonal = true;
+            NBL_UNROLL for (uint16_t i = 0; i < N; i++)
+                orthogonal = testing::relativeApproxCompare(hlsl::dot(m[i], m[(i+1)%N]), scalar_t(0.0), 1e-4) && orthogonal;
+            retval.orthogonal = orthogonal;
+        }
+        {
+            const matrix_t m_T = hlsl::transpose(m);
+            scalar_t dots[N];
+            NBL_UNROLL for (uint16_t i = 0; i < N; i++)
+                dots[i] = hlsl::dot(m[i], m[i]);
```
devshgraphicsprogramming: typo, you should be dotting `m[i]` with `m_T[i]`

include/nbl/builtin/hlsl/math/linalg/matrix_runtime_traits.hlsl
```diff
@@ -0,0 +1,66 @@
...
+    NBL_CONSTEXPR_STATIC_INLINE uint16_t N = matrix_traits<T>::RowCount;
+
+    static RuntimeTraits<matrix_t> create(const matrix_t m)
+    {
+        RuntimeTraits<matrix_t> retval;
+        retval.invertible = !testing::relativeApproxCompare(hlsl::determinant(m), scalar_t(0.0), 1e-5);
+        {
+            bool orthogonal = true;
+            NBL_UNROLL for (uint16_t i = 0; i < N; i++)
+                orthogonal = testing::relativeApproxCompare(hlsl::dot(m[i], m[(i+1)%N]), scalar_t(0.0), 1e-4) && orthogonal;
```
devshgraphicsprogramming: better to write `orthogonal = orthogonal && ...` due to short circuiting

include/nbl/builtin/hlsl/math/linalg/matrix_runtime_traits.hlsl
```diff
@@ -0,0 +1,66 @@
...
+        }
+        {
+            const matrix_t m_T = hlsl::transpose(m);
+            scalar_t dots[N];
+            NBL_UNROLL for (uint16_t i = 0; i < N; i++)
+                dots[i] = hlsl::dot(m[i], m[i]);
+
+            bool uniformScale = true;
+            NBL_UNROLL for (uint16_t i = 0; i < N-1; i++)
+                uniformScale = testing::relativeApproxCompare(dots[i], dots[i+1], 1e-4) && uniformScale;
```
devshgraphicsprogramming: same thing here

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -23,28 +41,178 @@ struct quaternion
...
+    static this_t create()
+    {
+        this_t q;
+        q.data = data_type(0.0, 0.0, 0.0, 1.0);
+        return q;
+    }
+
+    // angle: Rotation angle expressed in radians.
+    // axis: Rotation axis, must be normalized.
+    template<typename U=vector3_type, typename F=scalar_type NBL_FUNC_REQUIRES(is_same_v<vector3_type,U> && is_same_v<scalar_type,F>)
```
devshgraphicsprogramming: instead of allowing a free `F` I'd use `vector_traits<U>::scalar_type` instead

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -23,28 +41,178 @@ struct quaternion
...
+    static this_t create(const U axis, const F angle, const F uniformScale = scalar_type(1.0))
+    {
+        this_t q;
+        const scalar_type sinTheta = hlsl::sin(angle * 0.5);
+        const scalar_type cosTheta = hlsl::cos(angle * 0.5);
```
devshgraphicsprogramming: you need to define a `const scalar_type half = 0.5;` to avoid C++ making `angle * 0.5` a `double*double` multiplication and HLSL making it a `F*float32`

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: after you get the arguments you can make a `using scalar_type = ...` in the function body

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: or better `const scalar_type halfAngle = angle * scalar_type(0.5);`

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -23,28 +41,178 @@ struct quaternion
...
+        q.data = data_type(axis * sinTheta, cosTheta) * uniformScale;
+        return q;
+    }
+
+    template<typename U=vector<scalar_type,2> NBL_FUNC_REQUIRES(is_same_v<vector<scalar_type,2>,U>)
+    static this_t create(const U halfPitchCosSin, const U halfYawCosSin, const U halfRollCosSin)
+    {
```
devshgraphicsprogramming, line 70: maybe document the order of rotations and what axes theyre on (there's no single convention for pitch, yaw roll  order and what axes they correspond to)

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -55,19 +223,36 @@ struct quaternion
         return fraction + interpolantPrecalcTerm3 * k;
     }
 
-    static this_t flerp(const this_t start, const this_t end, const scalar_type fraction)
+    static this_t unnormFlerp(const this_t start, const this_t end, const scalar_type fraction)
     {
+        assert(hlsl::length(start.data) == scalar_type(1.0));
+        assert(hlsl::length(end.data) == scalar_type(1.0));
```
devshgraphicsprogramming: use `testing::relativeApproxCompare` against `1.0` for some tolerance

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -23,28 +41,178 @@ struct quaternion
...
+            data.w * other.data.z + data.x * other.data.y - data.y * other.data.x + data.z * other.data.w,
+            data.w * other.data.w - data.x * other.data.x - data.y * other.data.y - data.z * other.data.z
+        );
         return retval;
     }
 
+    static this_t unnormLerp(const this_t start, const this_t end, const scalar_type fraction, const scalar_type totalPseudoAngle)
+    {
+        assert(hlsl::length(start.data) == scalar_type(1.0));
+        assert(hlsl::length(end.data) == scalar_type(1.0));
```
devshgraphicsprogramming: same here

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -94,10 +281,106 @@ struct quaternion
...
+        return t;
+    }
+};
+
+template<typename T>
+struct static_cast_helper<matrix<T,3,3>, math::quaternion<T> >
+{
+    static inline matrix<T,3,3> cast(const math::quaternion<T> q)
+    {
+        return q.constructMatrix();
```
devshgraphicsprogramming: make the `constructMatrix` into `__constructMatrix` to indicate its supposed to be private, and `static_cast` is meant to be used

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -94,10 +281,106 @@ struct quaternion
...
+        math::truncated_quaternion<T> t;
+        t.data.x = t.data.x;
+        t.data.y = t.data.y;
+        t.data.z = t.data.z;
+        return t;
+    }
+};
+
+template<typename T>
+struct static_cast_helper<matrix<T,3,3>, math::quaternion<T> >
```
devshgraphicsprogramming, line 378: also add a `static_cast_helper<math::quaternion<T>,matrix<T,3,3> >` specialization which calls `create(mat,true)`

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -94,10 +281,106 @@ struct quaternion
...
+struct static_cast_helper<math::truncated_quaternion<T>, math::quaternion<T> >
+{
+    static inline math::truncated_quaternion<T> cast(const math::quaternion<T> q)
+    {
+        math::truncated_quaternion<T> t;
+        t.data.x = t.data.x;
+        t.data.y = t.data.y;
+        t.data.z = t.data.z;
+        return t;
+    }
```
devshgraphicsprogramming, line 374: please assert the input quaternion has uniform scale

include/nbl/builtin/hlsl/math/linalg/matrix_runtime_traits.hlsl
devshgraphicsprogramming: also make `uniformScale` into a `scalar_t uniformScaleSq`, make it NaN when its not uniform, but initialize with the dot product
```cpp
uniformScaleSq = hlsl::dot(m[0],m_T[0]);
NBL_UNROLL for (uint16_t i = 1; i < N; i++)
if (testing::relativeApproxCompare(hlsl::dot(m[i],m_T[i]),uniformScaleSq))
    uniformScaleSq = bit_cast<scalar_type>(hlsl::numeric_limits<scalar_type>::quiet_nan);
else
   break;
...

retval.orthonormal = retval.orthogonal && testing::relativeApproxCompare(uniformScaleSq, scalar_t(1.0), 1e-5)
```

this way you can get it out and use it later

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -23,28 +41,178 @@ struct quaternion
...
+                const scalar_type scales = hlsl::sqrt(tmp.w + scalar_type(1.0));
+                const scalar_type invscales = scalar_type(0.5) / scales;
+                retval.data.x = (m[0][2] + m[2][0]) * invscales;
+                retval.data.y = (m[1][2] + m[2][1]) * invscales;
+                retval.data.z = scales * scalar_type(0.5);
+                retval.data.w = (m[1][0] - m[0][1]) * invscales;
+            }
+        }
+
+        retval.data = hlsl::normalize(retval.data) / hlsl::sqrt(hlsl::dot(m[0], m[0])); // restore uniform scale
```
devshgraphicsprogramming: use the `uniformScaleSq` from the `traits`

also `hlsl::rqsrt` for bonus points

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -23,28 +41,178 @@ struct quaternion
...
 
+    static this_t unnormLerp(const this_t start, const this_t end, const scalar_type fraction, const scalar_type totalPseudoAngle)
+    {
+        assert(hlsl::length(start.data) == scalar_type(1.0));
+        assert(hlsl::length(end.data) == scalar_type(1.0));
+        // TODO: benchmark uint sign flip vs just *sign(totalPseudoAngle)
+        const data_type adjEnd = ieee754::flipSignIfRHSNegative<data_type>(end.data, hlsl::promote<data_type>(totalPseudoAngle));
+
+        this_t retval;
+        retval.data = hlsl::mix(start.data, adjEnd, hlsl::promote<data_type>(fraction));
```
devshgraphicsprogramming, line 215: don't promote `fraction` we have a scalar mix AFAIK

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -23,28 +41,178 @@ struct quaternion
...
+        );
         return retval;
     }
 
+    static this_t unnormLerp(const this_t start, const this_t end, const scalar_type fraction, const scalar_type totalPseudoAngle)
+    {
+        assert(hlsl::length(start.data) == scalar_type(1.0));
+        assert(hlsl::length(end.data) == scalar_type(1.0));
+        // TODO: benchmark uint sign flip vs just *sign(totalPseudoAngle)
+        const data_type adjEnd = ieee754::flipSignIfRHSNegative<data_type>(end.data, hlsl::promote<data_type>(totalPseudoAngle));
```
devshgraphicsprogramming: okay, the promotion makes this expensive, you need to change the signature to
```cpp
template <typename T, typename U=T>
NBL_CONSTEXPR_FUNC T flipSignIfRHSNegative(T val, U flip)
```

and then add a partial specialization of `flipSignIfRHSNegative_helper<Vectorial,typename vertor_traits<Vectorial>::scalar_type>`

which does
```cpp
		using traits_v = hlsl::vector_traits<Vectorial>;
		using scalar_t = typename traits_v::scalar_type;

		using AsUint = typename unsigned_integer_of_size<sizeof(FloatingPoint)>::type;
		const AsUint signFlipBit = ieee754::traits<scalar_t>::signMask & ieee754::impl::bitCastToUintType(flip);
		
		Vectorial output;
		array_get<Vectorial, scalar_t> getter_v;
		for (uint32_t i = 0; i < traits_v::Dimension; ++i)
		    setter(output,i, ieee754::impl::bitCastToUintType(getter_v(val,i))^signFlipBit );
```

devshgraphicsprogramming: btw the comment about benchmark is no longer applicable, I realized that when `totalPseudoAngle = 0` you have two different quaternions you can still interpolate between by a `fraction` which when normalized will produce a valid quaternion 

(there's two 4D directions which are orthogonal, the resulting quaternion's data member is interpolated between them on shortest path / within a plane spanned by them, then after normalization its like interpolating on a 4D sphere great circle)

So we can't make the `adjEnd = 0` in that case and code needs to stay as is, and all the above explanation needs to go in the comment instead of the TODO.

Btw there's a bit of a funny situation where we interpolate towards a different `adjEnd` when it dips slightly more than 90 degrees away from `start` in 4D space, thats correct but I don't yet have an understanding of why. My intuition tells me its because quaternions can represent 720 degree rotations and not just 360 due to half-angles being used in the definition, so its something to make the final rotation be performed "along the shortest path"

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -55,19 +223,36 @@ struct quaternion
...
     }
 
-    matrix_type constructMatrix()
+    vector3_type transformVector(const vector3_type v, const bool assumeNoScale=false) NBL_CONST_MEMBER_FUNC
+    {
+        const scalar_type scaleRcp = scalar_type(1.0) / hlsl::sqrt(hlsl::dot(data, data));
+        const vector3_type modV = v * scalar_type(2.0) * scaleRcp;
+        const vector3_type direction = data.xyz;
+        return v / scaleRcp + hlsl::cross(direction, modV * data.w + hlsl::cross(direction, modV));
+    }
```
devshgraphicsprogramming, line 273: adhow are you handling `assumeNoScale` ?

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -55,19 +223,36 @@ struct quaternion
...
+    {       
+        this_t retval = unnormFlerp(start,end,fraction);
         retval.data = hlsl::normalize(retval.data);
         return retval;
     }
 
-    matrix_type constructMatrix()
+    vector3_type transformVector(const vector3_type v, const bool assumeNoScale=false) NBL_CONST_MEMBER_FUNC
+    {
+        const scalar_type scaleRcp = scalar_type(1.0) / hlsl::sqrt(hlsl::dot(data, data));
```
devshgraphicsprogramming: use `inversesqrt`

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: I suggest
```cpp
vector3_type retval = v;
float modVScale = 2.f;
if (!assumeNoScale)
{
   retval /= scaleRcp;
   modVScale *= scaleRcp;
}
modV = v * modVScale;
return retval + hlsl::cross(...);
```

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -94,10 +281,106 @@ struct quaternion
...
+            return unnormLerp(start, end, fraction, totalPseudoAngle);
+    }
+
+    this_t inverse() NBL_CONST_MEMBER_FUNC
+    {
+        this_t retval;
+        retval.data.xyz = -data.xyz;
+        retval.data.w = data.w;
+        return retval;
+    }
```
devshgraphicsprogramming: does HLSL not like the unary minus `operator` ?

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -94,10 +281,106 @@ struct quaternion
...
 
+    static this_t slerp(const this_t start, const this_t end, const scalar_type fraction, const scalar_type threshold = numeric_limits<scalar_type>::epsilon)
+    {
+        const scalar_type totalPseudoAngle = hlsl::dot(start.data, end.data);
+
+        // make sure we use the short rotation
+        const scalar_type cosA = ieee754::flipSignIfRHSNegative<scalar_type>(totalPseudoAngle, totalPseudoAngle);
+        if (cosA <= (scalar_type(1.0) - threshold)) // spherical interpolation
+        {
+            this_t retval;
```
devshgraphicsprogramming, line 315: because you don't always call `unnormLerp` you need to assert that start and end are both unit length without scales

include/nbl/builtin/hlsl/math/quaternions.hlsl
keptsecret: That is the unary operator, isn't it? Only the scalar component of the inverse isn't negated for a quaternion.

devshgraphicsprogramming: I mean to have a `this_t operator-()  NBL_CONST_METHOD {` instead of `inverse()`

even though HLSL can't overload resolve these operators for a `-q` call, you can still call them as `q.operator-()`

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -324,8 +338,7 @@ struct normalize_helper<math::truncated_quaternion<T> >
 {
     static inline math::truncated_quaternion<T> __call(const math::truncated_quaternion<T> q)
     {
-        assert(hlsl::length(q.data) == scalar_type(1.0));
-
+        assert(testing::relativeApproxCompare(hlsl::length(q.data), scalar_type(1.0), scalar_type(1e-4)));
```
devshgraphicsprogramming: wrong check, truncated should never be 1.0 (thats kinda the point, because it reconstructs `w = 1-length(xyz)`, I have no clue how your tests passed in a Debug build (where asserts work)

Anyway I think that it makes no sense to have a `normalize` specialization for truncated quaternion. so can just delete the struct

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -52,16 +52,19 @@ struct quaternion
 
     // angle: Rotation angle expressed in radians.
     // axis: Rotation axis, must be normalized.
-    template<typename U=vector3_type, typename F=scalar_type NBL_FUNC_REQUIRES(is_same_v<vector3_type,U> && is_same_v<scalar_type,F>)
-    static this_t create(const U axis, const F angle, const F uniformScale = scalar_type(1.0))
+    template<typename U=vector3_type, typename F=scalar_type NBL_FUNC_REQUIRES(is_same_v<vector3_type,U> && is_same_v<typename vector_traits<U>::scalar_type,F>)
+    static this_t create(const U axis, const F angle, const F uniformScale = F(1.0))
```
devshgraphicsprogramming: just get rid of the `F` template argument, it will be much easier

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -52,16 +52,19 @@ struct quaternion
...
-        const scalar_type sinTheta = hlsl::sin(angle * 0.5);
-        const scalar_type cosTheta = hlsl::cos(angle * 0.5);
+        const scalar_t halfAngle = angle * scalar_t(0.5);
+        const scalar_t sinTheta = hlsl::sin(halfAngle);
+        const scalar_t cosTheta = hlsl::cos(halfAngle);
         q.data = data_type(axis * sinTheta, cosTheta) * uniformScale;
         return q;
     }
 
+    // applies rotation equivalent to 3x3 matrix in order of pitch * yaw * roll
```
devshgraphicsprogramming: 1. do you mean `mul(pitch,mul(yaw,mul(roll,v)))` or `mul(roll,yaw,mul(pitch,v)))`
2. around what axes https://math.stackexchange.com/questions/3526500/relating-pitch-roll-and-yaw-to-rotations-about-the-x-y-z-axis

The question is that "for our imaginary aircraft" which axis is pitch yaw roll

include/nbl/builtin/hlsl/math/quaternions.hlsl
devshgraphicsprogramming: my bad, seems we don't

include/nbl/builtin/hlsl/math/quaternions.hlsl
keptsecret: follows the glm convention, so according to [this](https://glm.g-truc.net/0.9.1/api/a00251.html#gab21847188e03d21037f895642252dcc7) it'll be X-Y-Z for pitch-yaw-roll

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -94,10 +292,106 @@ struct quaternion
...
+        return retval;
+    }
+};
+
+template<typename T>
+struct static_cast_helper<math::truncated_quaternion<T>, math::quaternion<T> >
+{
+    static inline math::truncated_quaternion<T> cast(const math::quaternion<T> q)
+    {
+        assert(testing::relativeApproxCompare(hlsl::length(q.data), scalar_type(1.0), scalar_type(1e-4)));
```
AnastaZIuk: it will not compile on debug, your `scalar_type` is not defined in this scope

you can try this snippet to see

```hlsl
#define _NBL_DEBUG 1
#include "nbl/builtin/hlsl/math/quaternions.hlsl"

using namespace nbl::hlsl;

[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 tid : SV_DispatchThreadID)
{
    impl::static_cast_helper<
        math::truncated_quaternion<float>,
        math::quaternion<float>
    >::cast(math::quaternion<float>::create());
}
```

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -94,10 +292,106 @@ struct quaternion
...
+        math::truncated_quaternion<T> t;
+        t.data.x = t.data.x;
+        t.data.y = t.data.y;
+        t.data.z = t.data.z;
```
AnastaZIuk: you should be copying from `q` and now we self assign and return garbage

include/nbl/builtin/hlsl/math/linalg/matrix_runtime_traits.hlsl
```diff
@@ -0,0 +1,72 @@
...
+                orthogonal = orthogonal && testing::relativeApproxCompare(hlsl::dot(m[i], m[(i+1)%N]), scalar_t(0.0), 1e-4);
+            retval.orthogonal = orthogonal;
+        }
+        {
+            const matrix_t m_T = hlsl::transpose(m);
+            scalar_t dots[N];
+            NBL_UNROLL for (uint16_t i = 0; i < N; i++)
+                dots[i] = hlsl::dot(m[i], m_T[i]);
+
+            scalar_t uniformScaleSq = hlsl::dot(m[0], m_T[0]);
```
AnastaZIuk: this is not scale^2, you want dot of the same axis `dot(m[i], m[i])` (or `dot(m_T[i], m_T[i])`) while making the test. Counterexample

```hlsl
#include "nbl/builtin/hlsl/math/quaternions.hlsl"
using namespace nbl::hlsl;

// take 90 degress rotation matrix
// it's true that R^T * R = I => s^2 = 1
static const float3x3 R = float3x3(
    0, -1, 0,
    1,  0, 0,
    0,  0, 1
);

RWByteAddressBuffer outBuf : register(u0);

[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 tid : SV_DispatchThreadID)
{
    math::quaternion<float> q = math::quaternion<float>::create(R, true);

    // but dot between column and row will make s = -1, it will go to sqrt and give you NaN!
    outBuf.Store(0, asuint(q.data.w)); // ctrl + f for "0x1.8p+128" in dissasembly
}
```

For an [orthogonal matrix](https://en.wikipedia.org/wiki/Orthogonal_matrix) $Q$:

$$
Q^T Q = I
$$

with uniform scale $s$:

$$
M = sQ \Rightarrow M^T M = (sQ)^T(sQ) = s^2 I
$$

devshgraphicsprogramming: aaah crap my bad I told him to change it, it should be dot product of columns, so `m_T[i],m_T[i]`

AnastaZIuk: what's more, you need to set quiet nan when this is 0

$$
s^2=0 \Rightarrow s=0 \Rightarrow M=sQ=0 \Rightarrow \det(M)=0\ (\text{singular}) \Rightarrow M^{-1}\ \text{does not exist}
$$
$$
\Rightarrow \frac{M}{s}\ \text{undefined (0/0)} \Rightarrow \text{division by 0} \Rightarrow \text{NaN}.
$$

otherwise we divide by 0 right here https://github.com/Devsh-Graphics-Programming/Nabla/blob/5f02325d67e768e7f724b19fe9e0ec0a2ceaf312/include/nbl/builtin/hlsl/math/quaternions.hlsl#L125

include/nbl/builtin/hlsl/testing/orientation_compare.hlsl
```diff
@@ -0,0 +1,44 @@
...
+template<typename FloatingPointVector NBL_PRIMARY_REQUIRES(concepts::FloatingPointLikeVectorial<FloatingPointVector>)
+struct OrientationCompareHelper
+{
+    static bool __call(NBL_CONST_REF_ARG(FloatingPointVector) lhs, NBL_CONST_REF_ARG(FloatingPointVector) rhs, const float64_t maxAllowedDifference)
+    {
+        using traits = nbl::hlsl::vector_traits<FloatingPointVector>;
+        using scalar_t = typename traits::scalar_type;
+
+        const scalar_t dotLR = hlsl::dot(lhs, rhs);
+        if (dotLR < scalar_t(0.0))
```
AnastaZIuk: `q` and `-q` is the same orientation, add `abs` on your dot

```hlsl
#include "nbl/builtin/hlsl/math/quaternions.hlsl"
#include "nbl/builtin/hlsl/testing/orientation_compare.hlsl"
using namespace nbl::hlsl;
RWByteAddressBuffer outBuf : register(u0);

[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 tid : SV_DispatchThreadID)
{
    math::quaternion<float> q = math::quaternion<float>::create(); // (0,0,0,1)
    math::quaternion<float> qn = q;
    qn.data = -qn.data;

    // comparator rejects q vs -q wheras its the same orientation
    bool ok = testing::orientationCompare(q.data, qn.data, 1e-4);
    outBuf.Store(0, ok ? 1u : 0u); // ctrl + f for %uint_0
}
```

include/nbl/builtin/hlsl/testing/orientation_compare.hlsl
```diff
@@ -0,0 +1,44 @@
...
+            return false;
+
+        const scalar_t scale = hlsl::sqrt(hlsl::dot(lhs,lhs) * hlsl::dot(rhs,rhs));
```
AnastaZIuk: we need 0 scale guard otherwise invalid quats will pass

```hlsl
#include "nbl/builtin/hlsl/math/quaternions.hlsl"
#include "nbl/builtin/hlsl/testing/orientation_compare.hlsl"
using namespace nbl::hlsl;
RWByteAddressBuffer outBuf : register(u0);

[shader("compute")]
[numthreads(1, 1, 1)]
void main(uint3 tid : SV_DispatchThreadID)
{
    math::quaternion<float> q = math::quaternion<float>::create(); // (0,0,0,1)
    math::quaternion<float> q0 = q;
    q0.data *= 0.0f; // zero length, its invalid orientation

    // comparator accepts 0-length as OK
    bool ok = testing::orientationCompare(q.data, q0.data, 1e-4);
    outBuf.Store(0, ok ? 1u : 0u); // ctrl + f for %uint_1
}
```

AnastaZIuk: might be +-

```hlsl
const scalar_t lenSqL = hlsl::dot(lhs, lhs);
const scalar_t lenSqR = hlsl::dot(rhs, rhs);
if (lenSqL <= numeric_limits<scalar_t>::min || lenSqR <= numeric_limits<scalar_t>::min)
    return false;
```

include/nbl/builtin/hlsl/math/linalg/matrix_runtime_traits.hlsl
keptsecret: resolved

include/nbl/builtin/hlsl/testing/orientation_compare.hlsl
keptsecret: resolved

include/nbl/builtin/hlsl/testing/orientation_compare.hlsl
keptsecret: resolved

include/nbl/builtin/hlsl/math/quaternions.hlsl
keptsecret: glm doesn't like unary `-` on a swizzle component, so it's been split into its components

include/nbl/builtin/hlsl/math/quaternions.hlsl
AnastaZIuk: resolved

include/nbl/builtin/hlsl/math/quaternions.hlsl
AnastaZIuk: resolved

PR #970, opened 2025-12-22
Add NSC flag to allow for preprocessing without compilation

Fletterio:


tools/nsc/main.cpp
```diff
@@ -153,6 +153,7 @@ class ShaderCompiler final : public system::IApplicationFramework
 			});
 		};
 		
+		auto preprocessOnly = findOutputFlag("-P") != m_arguments.end();
```
devshgraphicsprogramming, line 156: is `-P` used for anything already used in Clang or DXC ?

Fletterio: Clang has `-p` https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-p

DXC doesn't seem to have anything like it https://github.com/microsoft/DirectXShaderCompiler/wiki/Using-dxc.exe-and-dxcompiler.dll#dxc-command-line-interface

AnastaZIuk: `-P` is good

https://github.com/microsoft/DirectXShaderCompiler/blob/60a1c8568b405bbe9f47ab0431f627ef64988cc4/include/dxc/Support/HLSLOptions.td#L515

AnastaZIuk:
official https://godbolt.org & https://hlsl.godbolt.org doesn't support displaying preprocessed output

AnastaZIuk:
but ours does
image

PR #995, opened 2026-02-02
Cameraz

AnastaZIuk:
Introduces the new camera stack as Nabla::ext::Cameras.


Erfan-Ahmadi:
btw I found a small bug in a camera related hlsl header (not part of this pr )
usage of float instead of template in `const float w =`
```c++
template<typename FloatingPoint NBL_FUNC_REQUIRES(concepts::FloatingPoint<FloatingPoint>)
inline matrix<FloatingPoint, 4, 4> rhPerspectiveFovMatrix(FloatingPoint fieldOfViewRadians, FloatingPoint aspectRatio, FloatingPoint zNear, FloatingPoint zFar)
{
	const FloatingPoint h = core::reciprocal<FloatingPoint>(tan(fieldOfViewRadians * 0.5f));
	_NBL_DEBUG_BREAK_IF(aspectRatio == 0.f); //division by zero
	const float w = h / aspectRatio;

	_NBL_DEBUG_BREAK_IF(zNear == zFar); //division by zero

	matrix<FloatingPoint, 4, 4> m;
	m[0] = vector<FloatingPoint, 4>(w, 0.f, 0.f, 0.f);
	m[1] = vector<FloatingPoint, 4>(0.f, -h, 0.f, 0.f);
	m[2] = vector<FloatingPoint, 4>(0.f, 0.f, -zFar / (zFar - zNear), -zNear * zFar / (zFar - zNear));
	m[3] = vector<FloatingPoint, 4>(0.f, 0.f, -1.f, 0.f);

	return m;
}
```

include/nbl/ext/Cameras/IProjection.hpp
```diff
@@ -0,0 +1,71 @@
...
+///
+/// A projection transforms vectors between some input space and the projection
+/// space understood by a concrete viewport or projection consumer. Specialized
+/// interfaces such as `ILinearProjection`, `IPlanarProjection`, and
+/// `IPerspectiveProjection` refine this abstraction with additional structure.
+class IProjection
+{
+public:
+    /// @brief Common vector type used by projection and unprojection operations.
+    using projection_vector_t = hlsl::float64_t4;
```
Erfan-Ahmadi, line 19: I think the decision was made to always have doubles here, but if you want to abstract it to a type, then maybe template IProjection based on float_t?

include/nbl/ext/Cameras/ICamera.hpp
```diff
@@ -0,0 +1,509 @@
...
+            };
+        }
+    };
+
+    /// @brief Gimbal that stores the runtime camera pose and cached world-to-view transform.
+    ///
+    /// Camera implementations own one `CGimbal` instance and update it after
+    /// applying their internal state model. The gimbal stores world-space
+    /// position, orientation, and the cached view matrix derived from them.
+    class CGimbal : public IGimbal<hlsl::float64_t>
```
Erfan-Ahmadi, line 191: Do we want to template interfaces on float type or not after all?

include/nbl/ext/Cameras/CArcballCamera.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+#ifndef _C_ARCBALL_CAMERA_HPP_
+#define _C_ARCBALL_CAMERA_HPP_
+
+#include <algorithm>
+#include <cmath>
+
+#include "CSphericalTargetCamera.hpp"
+
+namespace nbl::core
```
Erfan-Ahmadi, line 13: wrong namespaces allover

include/nbl/ext/Cameras/CCameraFileUtilities.hpp
```diff
@@ -0,0 +1,94 @@
...
+/// The helpers keep camera-facing persistence code independent from ad-hoc file
+/// handling and provide one place for consistent error propagation.
+struct CCameraFileUtilities final
+{
+public:
+    /// @brief Read a whole file into a byte buffer.
+    static inline bool readBinaryFile(
+        ISystem& system,
+        const path& filePath,
+        std::vector<uint8_t>& outPayload,
```
Erfan-Ahmadi, line 25: I'd use a CPUBuffer here instead of `vector<uint>`

include/nbl/ext/Cameras/CCameraFileUtilities.hpp
```diff
@@ -0,0 +1,94 @@
...
+#include "nbl/system/ISystem.h"
+
+namespace nbl::system
+{
+
+/// @brief Shared file I/O helpers used by camera persistence and scripted-runtime loaders.
+///
+/// The helpers keep camera-facing persistence code independent from ad-hoc file
+/// handling and provide one place for consistent error propagation.
+struct CCameraFileUtilities final
```
Erfan-Ahmadi, line 18: To me this doesn't have anything to do with a Camera. it's just reading/writing file into utf8 string.
hence the camera in the name is redundant

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+
+    /// @brief Generic world-space gimbal used by runtime cameras and tracked targets.
+    ///
+    /// The gimbal stores position, orientation, scale, and an orthonormal local
+    /// basis. It also exposes `accumulate(...)`, which converts one batch of
+    /// semantic `CVirtualGimbalEvent` values into translation, rotation, and
+    /// scale impulses for a single manipulation step.
+    template<typename T>
+    requires is_any_of_v<T, hlsl::float32_t, hlsl::float64_t>
+    class IGimbal
```
Erfan-Ahmadi, line 34: I see two paths:
- either template everything on floating point type (should not be able to see float32 or 64 explicitly in the camera folder, maybe some places)
- don't try to template on anything and just use float64_t and that should be good enough on CPU

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+    {
+    public:
+        using precision_t = T;
```
Erfan-Ahmadi, line 37: that's not precision, that's float_t

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+        using quaternion_t = hlsl::camera_quaternion_t<precision_t>;
```
Erfan-Ahmadi, line 38: wait, what is `camera_quaternion_t`? what's the difference with regular quaternion? `math::quaternion`

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+
+/// @brief Camera-oriented math aliases and helpers built on top of Nabla `nbl::hlsl` types.
+template<typename T, uint32_t N>
+using camera_vector_t = vector<T, N>;
+
+template<typename T, uint32_t N, uint32_t M>
+using camera_matrix_t = matrix<T, N, M>;
+
+template<typename T>
+using camera_quaternion_t = math::quaternion<T>;
```
Erfan-Ahmadi, line 25: ok these are just aliases to hlsl counterpart.
I don't see any point in having them. Is there a distinction between a camera matirx and a regular matrix? is there a distinction between camera quaternion and regular quaternion? this is confusing. I'd drop them. and if you want type safety, you are not going to get that with a simple typedef/using.

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+
+template<typename T>
+struct SRigidTransformComponents
```
Erfan-Ahmadi, line 28: Again, decide whether you want to template on float_t or not. a lot of places are assuming float64_t

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+template<typename T>
+struct SCameraPoseDelta
+{
+    T position = T(0);
+    T rotationDeg = T(0);
+};
+
+struct SCameraViewRigDefaults final
+{
+    static constexpr double DegreesToRadians = numbers::pi<double> / 180.0;
```
Erfan-Ahmadi, line 44: small nitpicking: float64 instead of `double`

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
```diff
@@ -0,0 +1,131 @@
...
+/// to convert raw producer values into `CVirtualGimbalEvent::magnitude`.
+struct IGimbalBindingLayout
+{
+    IGimbalBindingLayout() {}
+    virtual ~IGimbalBindingLayout() {}
+
+    using gimbal_event_t = core::CVirtualGimbalEvent;
+    using encode_keyboard_code_t = ui::E_KEY_CODE;
+    using encode_mouse_code_t = ui::E_MOUSE_CODE;
+    using encode_imguizmo_code_t = gimbal_event_t::VirtualEventType;
```
Erfan-Ahmadi, line 26: has nothing to do with imguizmo directly, I'd rename it to avoid confusion. 
these are just scripted gimbal events types that don't necessarily come from inputs

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
```diff
@@ -0,0 +1,131 @@
...
+    inline static Map sanitizeMapping(const Map& source)
+    {
+        Map result;
+        for (const auto& [code, hash] : source)
+            result.emplace(code, typename Map::mapped_type(hash.event.type, hash.magnitudeScale));
+        return result;
+    }
+};
+
+class CGimbalBindingLayoutStorage : public IGimbalBindingLayout
```
Erfan-Ahmadi, line 107: what are you gaining with this inheritance? parent has decided unordered map and all function signatures and there is only one reasonable impl for the storage. I'd merge these two and not overcompilicate things

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
```diff
@@ -0,0 +1,131 @@
...
+            encode_mouse_code_t mouseCode;
+            encode_imguizmo_code_t imguizmoCode;
+        };
+
+        CKeyInfo(encode_keyboard_code_t code) : keyboardCode(code), type(Keyboard) {}
+        CKeyInfo(encode_mouse_code_t code) : mouseCode(code), type(Mouse) {}
+        CKeyInfo(encode_imguizmo_code_t code) : imguizmoCode(code), type(Imguizmo) {}
+
+        BindingDomain type;
+    };
```
Erfan-Ahmadi, line 51: unused struct

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+    struct CHashInfo
```
Erfan-Ahmadi, line 53: CHashInfo? what are you hashing, I don't get the name.
I get that you want a map from input -> event.

also it's a bit unclear, if we have a RIGHT event, that could map to a LEFT event? I think the virtual events should map 1:1? or maybe we should treat them differently.

Also it looks like dynamic data (active) is being stored next to the bindnig which causes you to "sanitize" the map. also sanitizeMapping is duplicated verbatim somewhere else.

imo the mapping from input event -> camera event should be pretty static and the state like "active" should be managed separately.

src/nbl/ext/Cameras/CGimbalInputBinder.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+    layout.updateKeyboardMapping([&](auto& map) { map = sanitizeMapping(getKeyboardVirtualEventMap()); });
+    layout.updateMouseMapping([&](auto& map) { map = sanitizeMapping(getMouseVirtualEventMap()); });
+    layout.updateImguizmoMapping([&](auto& map) { map = sanitizeMapping(getImguizmoVirtualEventMap()); });
+}
+
+void CGimbalInputBinder::copyBindingLayoutTo(IGimbalBindingLayout& layout) const
+{
+    copyActiveBindingsToLayout(layout);
+}
```
Erfan-Ahmadi, line 54: wait what?! these are just duplicated functions with the same signature.
also `sanitizeMapping` is weirdly duplicated here, and would not be needed if we seperated state from bindingconfig (I think)

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
```diff
@@ -0,0 +1,131 @@
...
+    using mouse_to_virtual_events_t = std::unordered_map<encode_mouse_code_t, CHashInfo>;
+    using imguizmo_to_virtual_events_t = std::unordered_map<encode_imguizmo_code_t, CHashInfo>;
+
+    virtual const keyboard_to_virtual_events_t& getKeyboardVirtualEventMap() const = 0;
+    virtual const mouse_to_virtual_events_t& getMouseVirtualEventMap() const = 0;
+    virtual const imguizmo_to_virtual_events_t& getImguizmoVirtualEventMap() const = 0;
+
+    virtual void updateKeyboardMapping(const std::function<void(keyboard_to_virtual_events_t&)>& mapKeys) = 0;
+    virtual void updateMouseMapping(const std::function<void(mouse_to_virtual_events_t&)>& mapKeys) = 0;
+    virtual void updateImguizmoMapping(const std::function<void(imguizmo_to_virtual_events_t&)>& mapKeys) = 0;
```
Erfan-Ahmadi, line 80: I don't get the std::functions, this is virtual, but the implementation just calls the function?!
and calling the function is simply a lambda. this is super unnecessary, the members are public.

why would you want to modify them through a std::function? just access the member, assign it, clear it.

I'd get rid of the whole updateXXXMapping through function and modify the data directly

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+        CHashInfo() {}
+        CHashInfo(gimbal_event_t::VirtualEventType _type, const double _magnitudeScale = DefaultMagnitudeScale)
+            : event({ .type = _type }), magnitudeScale(_magnitudeScale) {}
+        ~CHashInfo() = default;
+
+        /// @brief Virtual event emitted by this binding.
+        gimbal_event_t event = {};
+        /// @brief Per-binding gain applied when raw input is converted into one virtual-event magnitude.
+        double magnitudeScale = DefaultMagnitudeScale;
```
Erfan-Ahmadi, line 65: float is enough

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
```diff
@@ -0,0 +1,131 @@
...
+        /// @brief Runtime latch used by held keyboard and mouse-button bindings.
+        bool active = false;
```
Erfan-Ahmadi, line 67: CHashInfo is two unrelated things in one struct. same for magnitude inside the event
- get rid of sanitize 
- get rid of useless std::functions accessors that is equivalent to returning map&

then I would look for a solution like this:
```
struct SBinding {                 // pure config — what lives in every stored layout
    uint8_t eventIndex;           // index into CVirtualGimbalEvent::VirtualEventsTypeTable
    float   magnitudeScale;                 // sensitivity; float is ample for a multiplier
};                                // = 8 bytes
```
Binding layout would just become a struct holding three maps and that's it ;)

State of active could be:
```
std::bitset<EKC_COUNT>  m_heldKeys;      //  16 B
std::bitset<EMC_COUNT>  m_heldButtons;   //   8 B
std::array<double, CVirtualGimbalEvent::EventsCount> m_magnitude;  // 144 B
                                         // = 168 B, fixed
```

also active+magnitude should not be something that "InputBinding" should care. that's why I suggest moving it up the hierarchy when used in `IGimbalInputProcessor`


this will not only be a net reduction in memory usage, but cleans things up a lot better. and you have data with different frequncies living seperately and in a better place

include/nbl/ext/Cameras/IGimbalInputProcessor.hpp
```diff
@@ -0,0 +1,301 @@
...
+namespace nbl::ui
+{
+
+/// @brief Runtime processor that turns keyboard, mouse, and ImGuizmo input into virtual events.
+///
+/// Held keyboard and mouse-button bindings emit `frameDeltaSeconds * magnitudeScale`.
+/// Relative mouse movement, mouse scroll, and ImGuizmo deltas emit
+/// `abs(rawDelta) * magnitudeScale` per bound axis. The result is written into
+/// `CVirtualGimbalEvent::magnitude`.
+class IGimbalInputProcessor : public CGimbalBindingLayoutStorage
```
Erfan-Ahmadi, line 22: InputProcessor needs to store map of the bindings from input to events.
and it makes 0 sense that a "processor" inherits from a "BindingLayout"

Erfan-Ahmadi: you already needed it somewhere else and decided to store it as member function.
but for some reason this inherits?
```
// in PlanarProjection
ui::CGimbalBindingLayoutStorage m_inputBinding
``` 

based on <https://github.com/Devsh-Graphics-Programming/Nabla/pull/995/changes#r3933703955> this inheritance should be changed to just storage of the static "layout" (the maps) + dynamic "states" (the active status + magnitude)
also interface inheriting from impl was a bit sketchy

include/nbl/ext/Cameras/ILinearProjection.hpp
```diff
@@ -0,0 +1,128 @@
...
+    using model_matrix_t = typename ICamera::CGimbal::model_matrix_t;
+
+    /// @brief Matrix type used for fully concatenated linear transforms.
+    using concatenated_matrix_t = hlsl::float64_t4x4;
+
+    /// @brief Optional inverse of a concatenated transform when the matrix is not singular.
+    using inv_concatenated_matrix_t = std::optional<hlsl::float64_t4x4>;
+
+    /// @brief One concrete linear projection matrix together with cached inverse metadata.
+    struct CProjection : public IProjection
```
Erfan-Ahmadi, line 36: I don't get the choice to have ILinearProjection::CProjection inherit from IProjection and not the ILinearProjection itself
I would assume the convention IXXXProjection inherits from IProjection.

include/nbl/ext/Cameras/CCameraTraits.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace nbl::core
+{
+
+/// @brief Geometric constants used by target-relative camera families.
+///
+/// `MinDistance` prevents zero-distance target-relative states.
+/// `DefaultMaxDistance` is unbounded. Individual cameras and tools may apply
+/// their own finite limits on top of it.
+struct SCameraTargetRelativeTraits final
```
Erfan-Ahmadi, line 18: "Traits" is usually something that get specialized via template on different classes
this header looks more like a "limit" to me.
naming convention aside.
- why do we need to hardcode clamp distance to target by 0.1f? 
- why are values here float and the ones in `SCameraToolingThresholds` below in double?
- I think this `MinDistance ` value should be the same as whatever the projection paired with the camera uses as zNear plane.
- but again, I don't get the reason of existance for `MinDistance` and why we need to clamp some distance between a costexpr Min and Max that's baked into the library

include/nbl/ext/Cameras/CCameraTraits.hpp
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// @brief Smallest valid target-relative distance shared by spherical and path-style rigs.
+    /// This is a semantic lower bound used to avoid zero-distance degeneracy, not
+    /// the minimum representable `float` value.
+    static inline constexpr float MinDistance = 1e-1f;
+    /// @brief Default upper bound for target-relative distance when no camera-specific cap is requested.
+    static inline constexpr float DefaultMaxDistance = std::numeric_limits<float>::infinity();
```
Erfan-Ahmadi, line 25: again, this is float, and from what I've seen you're clamping/comparing doubles with this.
float infinity is not double infinity

include/nbl/ext/Cameras/CCameraTraits.hpp
```diff
@@ -0,0 +1,43 @@
...
+};
+
+/// @brief Comparison thresholds used by helper layers outside the runtime camera interface.
+struct SCameraToolingThresholds final
```
Erfan-Ahmadi, line 29: the limits in these files need to clarify:
- Numerical guards — "will this divide by zero?" 
- Policy floors — "should we let the user do this?" A UX decision
  - e.g. MinDistance, which I think is useless why would we need to clamp a positive number to 0.1 and +inf? it's not guarding a div and it seems to be soley UX related.
- Comparison tolerances — "are these two states the same?" and the epsilons relate to the floating point representation used. and their values need to be justified, their usages need to be justified as well. what are we protecting against?

include/nbl/ext/Cameras/CCameraTraits.hpp
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2018-2020 - DevSH Graphics Programming Sp. z O.O.
+// This file is part of the "Nabla Engine".
+// For conditions of distribution and use, see copyright notice in nabla.h
+
+#ifndef _C_CAMERA_TRAITS_HPP_
```
Erfan-Ahmadi, line 5: CCameraMathUtils has similar limits structs, maybe they should live in the same place.

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+        matrix<T, 4, 4> output;
+        output[0] = input[0];
+        output[1] = input[1];
+        output[2] = input[2];
+        output[3] = vector<T, 4>(T(0), T(0), T(0), T(1));
+        return output;
+    }
+
+    template<typename Vec, typename E = double>
+    static inline bool isOrthoBase(const Vec& x, const Vec& y, const Vec& z, const E epsilon = 1e-6)
```
Erfan-Ahmadi, line 107: the problem here is that's it's hardcoded for fp32 but templated on float type. you could get away with 1e-13 for doubles here 
also, instead of "length" which has a sqrt involved. you could just abs(dot(V,V)-1.0) < eps

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+        quaternion_t m_orientation;
+
+        /// @brief Scale transform component
+        vector_t<3u> m_scale = { 1.f, 1.f , 1.f };
+
+        /// @brief Orthonormal basis reconstructed from the current orientation.
+        hlsl::matrix<precision_t, 3, 3> m_orthonormal;
+
+        /// @brief Counter that increments for each performed manipulation, resets with each begin() call
+        size_t m_counter = {};
```
Erfan-Ahmadi, line 362: why reset every frame?
Isn't the point of the counter to avoid doing things everyframe and the manipulation to persist across frames? it's stored as uint64_t/size_t which means every frame there can be 2^64-1 manipulations?! even uint16 seems enough to me (~65K) for every frame.

Oh no! you're just doing `&=bool(m_gimbal.getManipulationCounter());` everywhere, it is literally just 1 bit/bool needed but somehow we have counter and isManipulating, both are useless and replacable by a bool at this stage.

I'm going to go an step further, remove the begin/end and counter reset everyframe and the weird bool below. store a 64-bit counter and increment on every edit. higher level code could store a gimbal manipulation counter value next to their computed view matrix and recalc when counters mismatch.
also some functions appear to be missing  manipulation_counter++ like setScale

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+
+        /// @brief Tracks whether gimbal is currently in manipulation mode
+        bool m_isManipulating = false;
```
Erfan-Ahmadi, line 365: this is just for `assert(m_isManipulating)` it seems. need to remove

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+                        impulse.dVirtualRotation.z -= static_cast<precision_t>(event.magnitude);
+
+                // scaling events
+                if constexpr (AllowedEvents & CVirtualGimbalEvent::ScaleXInc)
+                    if (event.type == CVirtualGimbalEvent::ScaleXInc)
+                        impulse.dVirtualScale.x *= static_cast<precision_t>(event.magnitude);
+
+                if constexpr (AllowedEvents & CVirtualGimbalEvent::ScaleXDec)
+                    if (event.type == CVirtualGimbalEvent::ScaleXDec)
+                        impulse.dVirtualScale.x *= static_cast<precision_t>(event.magnitude);
```
Erfan-Ahmadi, line 117: - you don't need scale for camera gimbal
- this class is not supposed to be general model matrix construction that handles scale, this class is all about cameras and virtual events affecting cameras.
- also I looked and there is literally 0 difference between `CVirtualGimbalEvent::ScaleXInc` and `CVirtualGimbalEvent::ScaleXDec`. you're treating them as the same thing everywhere
- also no camera allows scaling gimball event as it makes 0 sense

we need to nuke the scale event completely and remove m_scale from here.

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+    static inline constexpr camera_vector_t<T, 3> getCameraWorldUp()
+    {
+        return camera_vector_t<T, 3>(T(0), T(1), T(0));
+    }
+
+    template<typename T>
+    static inline constexpr camera_vector_t<T, 3> getCameraWorldForward()
+    {
+        return camera_vector_t<T, 3>(T(0), T(0), T(1));
+    }
```
Erfan-Ahmadi, line 221: very bad names and types.
In linear algebra it's just simply standard basis X Y Z, has nothing to do with the camera.
you transform these vectors using quaternions to put into matrix columns and construct a matrix that transforms these standard basis to the new ones.

need to rename to something that makes more sense or get rid of the functions alltogether. writing it directly makes more sense and is more readable

include/nbl/ext/Cameras/CCameraTraits.hpp
AnastaZIuk: this is not true

float infinity stays infinity when converted to double. see [C++ [conv.fpprom]/1](https://eel.is/c++draft/conv.fpprom#1).

> A prvalue of type float can be converted to a prvalue of type double. The value is unchanged.

[Godbolt](https://godbolt.org/z/rYn5jv4sh)

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: [other objects and scale were explicitly part of the design](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/156#discussion_r1827337727), so this is not limited to cameras.

the [input distinguishes factors below and above 1](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbalInputProcessor.hpp#L256-L264), although both events use multiplication in IGimbal

include/nbl/ext/Cameras/CCameraTraits.hpp
AnastaZIuk: this is not true

[computeDollyFov](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/CDollyZoomCamera.hpp#L42-L48) divides by `max(distance, MinDistance)`, so it does guard a division

using 0.1 as the minimum is a separate choice

include/nbl/ext/Cameras/CCameraTraits.hpp
AnastaZIuk: one camera can have [multiple projections](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ILinearProjection.hpp#L81-L90) with [independent clipping parameters](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IPlanarProjection.hpp#L49-L71), so which zNear would it use?

distance to the target is a different setting

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: [view updates are already conditional](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/CFPSCamera.hpp#L89-L96), the counter tells us whether something changed while isManipulating tells us whether begin/end is open.

a counter that never resets can work too, but scale does not change the [rigid view matrix](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ICamera.hpp#L235-L254)

include/nbl/ext/Cameras/ILinearProjection.hpp
AnastaZIuk: the wrapper has [multiple projections](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ILinearProjection.hpp#L81-L90), while [CProjection represents one](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ILinearProjection.hpp#L35-L68).

we can change the names, but making the wrapper implement IProjection still needs a choice of which projection project(...) uses

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
AnastaZIuk: casting the [input axes](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/CCameraMathUtilities.hpp#L843-L866) to double does not restore precision already lost in float32.

using the same epsilon for the [squared length check](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/CCameraMathUtilities.hpp#L107-L120) also changes which vectors pass

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
AnastaZIuk: we need remapping for things like inverted controls, where [the binding selects a different output command](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbalInputProcessor.hpp#L243-L264).

the name can change

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
AnastaZIuk: "Has nothing to do with the camera" is not true

we model the camera with the [gimbal's orthonormal frame](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ICamera.hpp#L235-L247) and these helpers define its [canonical right, up and forward axes](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/CCameraMathUtilities.hpp#L205-L221).

being standard basis vectors does not make them unrelated to cameras

include/nbl/ext/Cameras/CArcballCamera.hpp
AnastaZIuk: i don't think Matt and I settled where everything should go yet

[IGimbal is not just for cameras](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbal.hpp#L26-L34), so do we keep everything in ext or move the general parts to core/ui?

include/nbl/ext/Cameras/CGeneralPurposeGimbal.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include "IGimbal.hpp"
+
+namespace nbl::core
+{
+    /// @brief Minimal concrete gimbal wrapper for code that only needs the generic `IGimbal` behavior.
+    ///
+    /// The class exists mainly as a convenient instantiable type when no additional
+    /// camera-specific state or manipulation policy is required on top of `IGimbal`.
+    template<typename T = hlsl::float64_t>
+    class CGeneralPurposeGimbal : public IGimbal<T>
```
Erfan-Ahmadi, line 13: unused class and adds 0 value to parent class

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+
+            if (m_orientation.data != orientation.data)
+                m_counter++;
+
+            m_orientation = hlsl::CCameraMathUtilities::normalizeQuaternion(orientation);
+            updateOrthonormalOrientationBase();
+        }
+
+        /// @brief Apply a prebuilt rigid reference transform and an accumulated impulse in one step.
+        inline void transform(const CReferenceTransform& reference, const VirtualImpulse& impulse)
```
Erfan-Ahmadi, line 201: this function is not used anywhere. just keeping notes for future

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+            safeNormalizeVec3(up - canonicalForward * dot(up, canonicalForward), getCameraWorldUp<T>()));
+
+        canonicalRight = safeNormalizeVec3(cross(canonicalUp, canonicalForward), canonicalRight);
+        canonicalUp = safeNormalizeVec3(cross(canonicalForward, canonicalRight), canonicalUp);
+        static_assert(std::is_same_v<T, float> || std::is_same_v<T, double>, "Camera basis conversion is only implemented for float and double.");
+
+        if constexpr (std::is_same_v<T, float>)
+            return makeQuaternionFromBasisImpl(canonicalRight, canonicalUp, canonicalForward);
+        else
+            return makeQuaternionFromBasisImpl(canonicalRight, canonicalUp, canonicalForward);
```
Erfan-Ahmadi, line 339: these are the exact same, no need to specialize with if constexpr

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+        {
+            setOrientation(reference.orientation * hlsl::CCameraMathUtilities::makeQuaternionFromEulerRadiansYXZ(impulse.dVirtualRotation));
+            setPosition(
+                hlsl::float64_t3(reference.frame[3]) +
```
Erfan-Ahmadi, line 205: templated on T but using float64_t3 directly here?

Erfan-Ahmadi: doesn't matter, this will all change to float64 soon

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
AnastaZIuk: we can separate config and state and improve memory usage

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: we can keep the camera on double and template the math helpers, since we already have [float32 input transforms](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/src/nbl/ext/Cameras/IGimbalInputProcessor.cpp#L122-L127) and [float64 camera state](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ICamera.hpp#L185-L197).

we do not need to template everything

include/nbl/ext/Cameras/IProjection.hpp
AnastaZIuk: the [alias fixes the type to float64_t4](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IProjection.hpp#L17-L19), it is not a template parameter.

it does not mean we need to template IProjection

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+
+        /// @brief Rotate the gimbal around a world-space axis by the requested angle in radians.
+        inline void rotate(const vector_t<3u>& axis, float dRadians)
+        {
+            assert(m_isManipulating);
+
+            if(dRadians)
+                m_counter++;
+
+            const auto dRotation = hlsl::CCameraMathUtilities::makeQuaternionFromAxisAngle(axis, static_cast<precision_t>(dRadians));
```
Erfan-Ahmadi, line 218: float for radians/angle is enough precision. so this is fine. 
although that function needs to go into actual hlsl lib

include/nbl/ext/Cameras/CGeneralPurposeGimbal.hpp
AnastaZIuk: not being used inside Nabla alone does not make it unnecessary, but this wrapper is only a convenience type

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
AnastaZIuk: renaming is fine, but the input code is not the output command, the [mapping selects the output event and gain](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbalInputProcessor.hpp#L243-L264)

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+
+        /// @brief One frame of accumulated virtual translation, rotation, and scaling intent.
+        struct VirtualImpulse
+        {
+            vector_t<3u> dVirtualTranslate { 0.0f }, dVirtualRotation { 0.0f }, dVirtualScale { 1.0f };
+        };
+
+        /// @brief Accumulates one frame of virtual events into a translation/rotation/scale impulse.
+        template <uint32_t AllowedEvents>
+        VirtualImpulse accumulate(std::span<const CVirtualGimbalEvent> virtualEvents, const vector_t<3u>& gRightOverride, const vector_t<3u>& gUpOverride, const vector_t<3u>& gForwardOverride)
```
Erfan-Ahmadi, line 52: 3 params are not used at all.

include/nbl/ext/Cameras/ICamera.hpp
AnastaZIuk: keep cameras on float64, but let the math helpers handle float32 input too

include/nbl/ext/Cameras/CCameraFileUtilities.hpp
AnastaZIuk: CPUBuffer is fine

include/nbl/ext/Cameras/CCameraFileUtilities.hpp
AnastaZIuk: we can drop Camera from the name

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: i meant the precision chosen by T, but float_t is fine too

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: it is just math::quaternion<T> under another name

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
AnastaZIuk: they name the types used by the camera helpers, not different mathematical objects.

they do not add type safety, removing them is a naming choice

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
AnastaZIuk: the helpers handle float32 and float64, while cameras use double.

we should check the hardcoded types in each helper

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
AnastaZIuk: we can use float64_t here too

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
AnastaZIuk: another implementation could use maps from an existing layout instead of storing its own, [the interface allows that](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbalBindingLayout.hpp#L74-L126)

we can merge them if we do not need that

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
AnastaZIuk: hmm ok I found no uses in Nabla or exs, that could be some old leftover

src/nbl/ext/Cameras/CGimbalInputBinder.cpp
AnastaZIuk: the duplicates can go, but [sanitize clears active state and accumulated magnitude while preserving bindings and gain](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbalBindingLayout.hpp#L96-L103).

we can remove that once state is stored separately

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
AnastaZIuk: actually we could drop callbacks

i started with private maps and used callbacks to edit them

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
AnastaZIuk: float may be enough, more precision alone is not a reason to keep double

include/nbl/ext/Cameras/IGimbalInputProcessor.hpp
AnastaZIuk: inheritance lets the processor be used as IGimbalBindingLayout by the same binding editor.

composition works too, but callers would need access to its layout member

include/nbl/ext/Cameras/CCameraTraits.hpp
AnastaZIuk: we can put them together, but keep the different kinds of limits clear

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: it is also returned by [isManipulating()](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbal.hpp#L317-L321), not just used in asserts.

after [begin()](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbal.hpp#L164-L168) the flag is true and the counter is zero, so they tell us different things

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: the [ICamera::CGimbal::transform wrapper calls it](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ICamera.hpp#L204).

it applies translation and rotation relative to a reference frame, lack of a current caller alone does not make that operation unnecessary

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
AnastaZIuk: yeah my bad, no need for the if constexpr

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: this uses float64 despite being templated on T, we should make that consistent

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: we can move the math helper to hlsl

include/nbl/ext/Cameras/IGimbal.hpp
AnastaZIuk: yeah these three params are not used, could be another leftover

include/nbl/ext/Cameras/ICamera.hpp
```diff
@@ -0,0 +1,509 @@
...
+            const auto& gUp = this->getYAxis();
+            const auto& gForward = this->getZAxis();
+
+            assert(hlsl::CCameraMathUtilities::isOrthoBase(gRight, gUp, gForward));
+
+            const auto& position = this->getPosition();
+
+            m_viewMatrix[0u] = hlsl::float64_t4(gRight, -hlsl::dot(gRight, position));
+            m_viewMatrix[1u] = hlsl::float64_t4(gUp, -hlsl::dot(gUp, position));
+            m_viewMatrix[2u] = hlsl::float64_t4(gForward, -hlsl::dot(gForward, position));
```
Erfan-Ahmadi, line 247: make sure to mention that the inverse of an orthonormal matrix is it's transpose, so we caan put columns into rows.
also it's really confusing because you're treating m_orthonormal (where those getX/Y/ZAxis come from) as a column-major order matrix treating mat[0] as the first column but here you're doing it properly with nabla conventions (hlsl::matrix is row matrix)

and the conventions aren't clear, the when I look at some functions, some treat it as column-major some as row-major

AnastaZIuk: here the [basis stores the rotated axes as rows](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/CCameraMathUtilities.hpp#L755-L761), which is already the transpose of the orientation used with column vectors. [updateView](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ICamera.hpp#L235-L247) uses those rows and their negative dot products with position, so this view construction is correct.

i think we can document that convention

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+    template<typename T>
+    static inline camera_vector_t<T, 3> transformLocalVectorToWorldBasis(
+        const camera_vector_t<T, 3>& localVector,
+        const camera_vector_t<T, 3>& right,
+        const camera_vector_t<T, 3>& up,
+        const camera_vector_t<T, 3>& forward)
+    {
+        const camera_matrix_t<T, 3, 3> basis { right, up, forward };
+        return hlsl::mul(basis, localVector);
+    }
```
Erfan-Ahmadi, line 658: this is wrong and swapped because hlsl::matrix and mul treats matrix as row major matrix. 
I'm 90% sure this is causing a bug with the `CSphericalTargetCamera::applyPlanarTargetTranslation` using this function

AnastaZIuk: yeah these two are swapped, local to world needs the transpose when the axes are stored as rows

this also affects planar target movement

Erfan-Ahmadi: btw instead of mul(transpose(m), v) you can do mul(v, m)

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+        inline const quaternion_t& getOrientation() const { return m_orientation; }
+
+        /// @brief Scale transform component
+        inline const vector_t<3u>& getScale() const { return m_scale; }
+
+        /// @brief World matrix (TRS)
+        template<typename TRS = model_matrix_t>
+        requires is_any_of_v<TRS, model_matrix_t, hlsl::matrix<T, 4u, 4u>>
+        const TRS operator()() const
+        {
```
Erfan-Ahmadi, line 273: - this function is unused and is a bit funny, cause it's the same function but when you pass `hlsl::matrix<T, 4u, 4u>` it treats as row-major but otherwise column major and the function is never used btw.

Erfan-Ahmadi:
the convention in your camera ext is super unclear:
- hlsl::matrix in nabla is always treated as row major (that's why we privately inherit from glm's)
- means mat3x3 { right, up, forward } will put those vectors in rows not columns
- you're using camera_vector_t and camera_matrix_t and so on, but it's just alias of nbl::hlsl version and add no value and only confusion
- you can see you were confused here in [this comment](https://github.com/Devsh-Graphics-Programming/Nabla/pull/995/changes#r3958138213) and caused a bug because you assumed column-major matrix
- sometimes you're [using mat3x3 and treating it as a column vector](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/IGimbal.hpp#L302-L309) for storage only but sometimes [you're passing those columns separately because you know it'll bite you in the ass if you used a matrix](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/CCameraMathUtilities.hpp#L316-L318) 
- sometimes you're putting position in the last column [correctly](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/ICamera.hpp#L245-L247) which works with mul. but [sometimes you're treating last row as translation which is definitely incorrect](https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/ext/Cameras/CCameraMathUtilities.hpp#L838), it mostly works because nothing ever `mul`s with them. but the conventions are unclear and causes people to make bugs.

include/nbl/ext/Cameras/IGimbal.hpp
Erfan-Ahmadi: mostly hinting that a function is untested. and needs to be reviewed with more care later on or with unit tests

include/nbl/ext/Cameras/CGeneralPurposeGimbal.hpp
Erfan-Ahmadi: ok not used, but what value do they add to IGimball? they don't even specialize a template param. they are identical at this point. did you have any plans to add any feature to this class?

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
Erfan-Ahmadi: doesn't matter that much. I'd rename them to CameraStandardBasisX. 
when I see `World` in the name hints at the units being in worldspace. while these are just the standard basis of a camera in camera space. the camera's up vector is always 0,1,0. that's why it's local and not world

include/nbl/ext/Cameras/IGimbal.hpp
Erfan-Ahmadi: ok got it, you wrote it specifically for all types of objects. then you need to stop using `hlsl::CCameraMathUtilities` and `hlsl::camera_vector_t`  and anything with camera in it's name.

also `void transform` function doesn't even take `impulse.dVirtualScale` into account. we may want to fix that as well (not sure if intentional?)

include/nbl/ext/Cameras/IGimbal.hpp
Erfan-Ahmadi: btw, any reason this does not take scale of the "impulse" into account?

include/nbl/ext/Cameras/CCameraTraits.hpp
Erfan-Ahmadi: ok but why 0.1 as hardcoded choice? MinDistance is being used multiple places for multiple reasons.
here is used for 0 guard div but it's used somewhere else for something else.

include/nbl/ext/Cameras/CCameraTraits.hpp
Erfan-Ahmadi: interesting. this i learned today.
but again. let's be consistent with our types and limits

include/nbl/ext/Cameras/IGimbal.hpp
Erfan-Ahmadi: begin/end and isManipulating need to be replaced with global uint64 counter

include/nbl/ext/Cameras/ICamera.hpp
Erfan-Ahmadi: I think we need to fix up the code and anywhere in the ext that uses matrix first.
either store 3 seperate orthonormal column vectors if you want to strore columns. otherwise you'll risk bugs and somebody touching up the code later and assuming the matrices are safe to mul with, not knowing sometimes it's transposed and sometimes not.

for example we cannot move something like `getQuaternionBasisMatrix` to the hlsl library because it clashes with nabla's convention of row major matrices.

include/nbl/ext/Cameras/IGimbal.hpp
```diff
@@ -0,0 +1,370 @@
...
+        vector_t<3u> m_position;
+
+        /// @brief Normalized orientation of gimbal
+        quaternion_t m_orientation;
+
+        /// @brief Scale transform component
+        vector_t<3u> m_scale = { 1.f, 1.f , 1.f };
+
+        /// @brief Orthonormal basis reconstructed from the current orientation.
+        hlsl::matrix<precision_t, 3, 3> m_orthonormal;
```
Erfan-Ahmadi, line 359: if you want to keep this. then you can do it as seperate column vectors, because it looks like this matrix is only used for storage and not any matrix operations. and you're storing your columns in this matrx's rows and any `mul` using this matrix would break immediately.

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
Erfan-Ahmadi: ok, what's a good reason to keep a multiplier double here?
I usually don't care about camera classes being larger btw, there are very few of them. but since this is going into map of all possible inputs, i care a bit

src/nbl/ext/Cameras/CGimbalInputBinder.cpp
Erfan-Ahmadi: will be removed once the states are stored properly

include/nbl/ext/Cameras/IGimbal.hpp
Erfan-Ahmadi: ok, good plan to move math helpers to hlsl and template them

include/nbl/ext/Cameras/IGimbal.hpp
Erfan-Ahmadi: needs to be removed, only adds confusion. there to add "camera" in the name of something not inherently related to cameras.
It'd make sense if you wanted to add your own type that puts extra requirements on regular quaternions like always being normalized

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
Erfan-Ahmadi: https://github.com/Devsh-Graphics-Programming/Nabla/pull/995#discussion_r3959192797

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
Erfan-Ahmadi: agreed. helpers will move to actual hlsl namespace and templated

include/nbl/ext/Cameras/IGimbalBindingLayout.hpp
Erfan-Ahmadi: will  merge

include/nbl/ext/Cameras/CCameraTraits.hpp
Erfan-Ahmadi: my problem is not what it should be equal to. my problem is why is that a thing that is hardcoded with 0.1f.
it seems mostly like a UX thing, but breaks when user wants to work in small ranges.
it seems to be mostly for preventing the distances to targets becoming 0 and introducing div by 0 issues.

also I don't see any reason to clamp against infinity.
What i suggest is that if you want to have min/max distance, let the user provide it to you when building the camera classes, don't decide that for them

include/nbl/builtin/hlsl/math/quaternions.hlsl
```diff
@@ -380,18 +380,150 @@ struct static_cast_helper<math::truncated_quaternion<T>, math::quaternion<T> >
...
+        retval.data.z = scales * T(0.5);
+        retval.data.w = (m[1][0] - m[0][1]) * invscales;
+    }
+
+    retval.data *= uniformScale;
+    return retval;
+}
+
+template<typename T>
+inline math::quaternion<T> matrix_to_quaternion_cast(NBL_CONST_REF_ARG(matrix<T,3,3>) m)
```
Erfan-Ahmadi, line 490: why are we considering the trasnpose matrices for matrix_to_quaternion cast? why are we doing this 4 times 4 different ways and scoring them?

include/nbl/ext/Cameras/CArcballCamera.hpp
Erfan-Ahmadi: For now, everything will be nbl::ext and the rest of general math helpers will move to nbl::hlsl

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+
+    template<typename T>
+    static inline camera_matrix_t<T, 3, 3> getQuaternionBasisMatrix(const camera_quaternion_t<T>& orientation)
+    {
+        const auto normalizedOrientation = normalizeQuaternion(orientation);
+        return camera_matrix_t<T, 3, 3>(
+            normalizedOrientation.transformVector(getCameraWorldRight<T>(), true),
+            normalizedOrientation.transformVector(getCameraWorldUp<T>(), true),
+            normalizedOrientation.transformVector(getCameraWorldForward<T>(), true));
+    }
```
Erfan-Ahmadi, line 762: why not use `__constructMatrix`? already available in hlsl.

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+
+    template<typename T>
+    static inline bool hasPlanarDeltaXY(const camera_vector_t<T, 3>& value, const T epsilon = std::numeric_limits<T>::epsilon())
+    {
+        return !isNearlyZeroVector(camera_vector_t<T, 2>(value.x, value.y), epsilon);
+    }
+
+    template<typename VecA, typename VecB, typename T>
+    static inline bool nearlyEqualVec3(const VecA& a, const VecB& b, const T epsilon)
```
Erfan-Ahmadi, line 196: we have `nbl::hlsl::RelativeApproxCompareHelper` that should even work for matrices. this is redundant

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
```diff
@@ -0,0 +1,917 @@
...
+    }
+
+    template<typename T>
+    static inline constexpr T getCameraMathEpsilon()
+    {
+        return std::numeric_limits<T>::epsilon();
+    }
+
+    template<typename T>
+    static inline bool nearlyEqualScalar(const T a, const T b, const T epsilon)
```
Erfan-Ahmadi, line 172: we have `nbl::hlsl::RelativeApproxCompareHelper` that should even work for matrices. this is redundant

Erfan-Ahmadi:
not included in this PR but needs to be fixed:

https://github.com/Devsh-Graphics-Programming/Nabla/blame/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/builtin/hlsl/testing/relative_approx_compare.hlsl#L25

compare with 0.0 or any other subnormal even with a generous tolerance with return false. that's causing some quaternion stuff to be nan when they shouldn't

e.g here: https://github.com/Devsh-Graphics-Programming/Nabla/blob/89cfa0b58c4d719c3dd3a245f6b46f73c6caa638/include/nbl/builtin/hlsl/matrix_utils/matrix_runtime_traits.hlsl#L36

include/nbl/ext/Cameras/CCameraMathUtilities.hpp
devshgraphicsprogramming: probably because first this PR got started and THEN the quaternions got cleaned up by Przemog

include/nbl/ext/Cameras/CCameraTargetRelativeUtilities.hpp
```diff
@@ -0,0 +1,270 @@
...
+    hlsl::float64_t appliedDistance = static_cast<hlsl::float64_t>(SCameraTargetRelativeTraits::MinDistance);
+};
+
+/// @brief Derived basis for target-relative orbit rigs.
+struct SCameraTargetRelativeBasis final
+{
+    hlsl::float64_t3 localOffset = hlsl::float64_t3(0.0);
+    hlsl::float64_t3 right = hlsl::float64_t3(1.0, 0.0, 0.0);
+    hlsl::float64_t3 up = hlsl::float64_t3(0.0, 0.0, 1.0);
+    hlsl::float64_t3 forward = hlsl::float64_t3(0.0, 1.0, 0.0);
```
Erfan-Ahmadi, line 32: a bit contradicting with what you have in the MathUtils. why?

PR #997, opened 2026-02-06
GUI for the new Path Tracer

karimsayedre:
Added Scene and Session GUI windows:
- Every interaction with session and renderer is done through callbacks
- CUIManager orchestrates the whole gui
- Most callbacks are to be implemented, all in onAppInitialized in main.cpp
- Thumbnails are not tested
- Used suballocated descriptor set from example 61


devshgraphicsprogramming:
this is Nabla sumbmodule pointer change only, closing PR doesn't ppass CI anyway

PR #1027, opened 2026-03-20
Resampled Importance Sampling bxdfs

devshgraphicsprogramming:


include/nbl/builtin/hlsl/bxdf/base/lambertian.hlsl
```diff
@@ -62,19 +63,20 @@ struct SLambertianBase
         return generate(anisotropic_interaction_type::create(interaction), u);
     }
 
-    scalar_type pdf(NBL_CONST_REF_ARG(sample_type) _sample, NBL_CONST_REF_ARG(isotropic_interaction_type) interaction) NBL_CONST_MEMBER_FUNC
+    // pdf function, because this BxDF has a tractable pdf, indicates that eval weight is also pdf
+    scalar_type denominator(NBL_CONST_REF_ARG(sample_type) _sample, NBL_CONST_REF_ARG(isotropic_interaction_type) interaction) NBL_CONST_MEMBER_FUNC
```
devshgraphicsprogramming: you can keep existing `pdf` functions called `pdf`, not sure we need a `denominator` function, and even if our future concept requires a separate `denominator` we can make an Adaptor Class with CRTP you can launder BXDFTractablePDF matching classes though

devshgraphicsprogramming: actually they would need to be called `forwardPdf`

btw anything PDF or Weight without `eval_and_` and `quotient_and_` needs to clearly indicate with a prefix if its "forward" (takes `_sample` and `interaction` plus a cache spat out by generate) or "backward" (doesn't take a generator cache)

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: actually make this configurable, just use `vector_traits` to check dimensions are >=2 for a BRDF and >=3 for a BSDF

Reason: RIS sampling from lobe mixtures consumes more than one 2D random variables

BRDF vs BSDF should be a trait now that needs explicit specialization, its impossible to tell apart just by the dimensionality of `u`

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -1054,9 +1104,9 @@ NBL_CONCEPT_BEGIN(4)
 NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(bxdf_common_typdefs, T))
     ((NBL_CONCEPT_REQ_TYPE)(T::anisocache_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.eval(_sample, aniso, anisocache)), ::nbl::hlsl::is_same_v, typename T::spectral_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.eval_and_weight(_sample, aniso, anisocache)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
     ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.pdf(_sample, aniso, anisocache)), ::nbl::hlsl::is_same_v, typename T::scalar_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.quotient_and_pdf(_sample, aniso, anisocache)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.quotient_and_weight(_sample, aniso, anisocache)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
```
devshgraphicsprogramming: lets make all BxDFs have a cache type which goes from `generate` to `quotient_and_weight` but then `microfacet_bxdf_common` would be the weird one that puts a further constraint that the anisocache_type needs to fit some concept for Cook Torrance to be useful

and allow a 2nd cache type just for `eval_and_weight` (the two can obviously be identical in practice, e.g. our base impl for now)

tools/nsc/main.cpp
```diff
@@ -152,6 +150,61 @@ class ShaderLogger final : public IThreadsafeLogger
...
+
+    system->deleteFile(outputPath);
+    const auto moveError = system->moveFileOrDirectory(tempPath, outputPath);
+    if (moveError)
+    {
+        system->deleteFile(tempPath);
+        return false;
+    }
+
+    return true;
```
devshgraphicsprogramming, line 205: @AnastaZIuk do you think you could add an option to `createFile` flags (or pack the `path`+ IFileBase flaggs into a strcut) to create all the directories needed on the way to the path ?


cold be useful

AnastaZIuk: ay

include/nbl/builtin/hlsl/path_tracing/concepts.hlsl
```diff
@@ -261,10 +260,10 @@ NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE)(T::tolerance_method_type))
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(NextEventEstimatorSampleQuotientReturn, typename T::sample_quotient_return_type))
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(Ray, typename T::ray_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((nee.deferred_pdf(scene, id, ray)), ::nbl::hlsl::is_same_v, typename T::scalar_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((nee.template generate_and_quotient_and_pdf<impl::DummyMaterialSystem>(scene, matSys, v/*origin*/, interaction, v/*xi*/, depth)), ::nbl::hlsl::is_same_v, typename T::sample_quotient_return_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((nee.get_env_light_id()), ::nbl::hlsl::is_same_v, typename T::light_id_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((nee.get_environment_radiance(ray)), ::nbl::hlsl::is_same_v, typename T::spectral_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((nee.deferredPdf(scene, id, ray)), ::nbl::hlsl::is_same_v, typename T::scalar_type))
```
devshgraphicsprogramming: this should be called `deferredWeight`

include/nbl/builtin/hlsl/path_tracing/concepts.hlsl
```diff
@@ -182,10 +182,9 @@ NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE)(T::bxdfnode_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::create_params_t))
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(BxdfNode, typename T::bxdfnode_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((matsys.eval(matid, _sample, aniso_inter)), ::nbl::hlsl::is_same_v, typename T::measure_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((matsys.evalAndWeight(matid, _sample, aniso_inter)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
     ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((matsys.generate(matid, aniso_inter, u, cache_)), ::nbl::hlsl::is_same_v, typename T::sample_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((matsys.pdf(matid, _sample, aniso_inter)), ::nbl::hlsl::is_same_v, typename T::scalar_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((matsys.quotient_and_pdf(matid, _sample, aniso_inter, cache_)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((matsys.quotientAndWeight(matid, _sample, aniso_inter, cache_)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
```
devshgraphicsprogramming: the typedefs returned should be the new `value_and_weight` types 

Need a new file called `value_and_weight` which is essentially the same as `value_and_pdf` with similar typedefs
https://github.com/Devsh-Graphics-Programming/Nabla/blob/38f73af10a6047ede44d7822930c04ce2a395f9d/include/nbl/builtin/hlsl/sampling/value_and_pdf.hlsl

**With the exception of providing the `value()` method!**

Then implement the `value_and_pdf` (and `rcpPdf`) via composition (cause in HLSL inheritance is bound to fuck us over with templated functions) from `_and_weight` and only have it differ by providing a `value()` method

include/nbl/builtin/hlsl/path_tracing/unidirectional.hlsl
```diff
@@ -69,7 +69,7 @@ struct Unidirectional
             if (ray.shouldDoMIS() && matLightID.isLight())
             {
                 emissive *= ray.getPayloadThroughput();
-                const scalar_type pdf = nee.deferred_pdf(scene, lightID, ray);
+                const scalar_type pdf = nee.deferredPdf(scene, lightID, ray);
```
devshgraphicsprogramming: replace all mentions of `pdf` used just for MIS, with `weight` in this file

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -156,19 +157,22 @@ struct SCookTorrance
...
+    quotient_pdf_type evalAndWeight(NBL_CONST_REF_ARG(sample_type) _sample, NBL_CONST_REF_ARG(Interaction) interaction, NBL_CONST_REF_ARG(MicrofacetCache) cache) NBL_CONST_MEMBER_FUNC
     {
         fresnel_type _f = __getOrientedFresnel(fresnel, interaction.getNdotV());
         if (!__checkValid<Interaction, MicrofacetCache>(_f, _sample, interaction, cache))
-            return hlsl::promote<spectral_type>(0.0);
+            return quotient_pdf_type::create(scalar_type(0.0), scalar_type(0.0));
+
+        bool isInfinity;
+        scalar_type _pdf = __forwardPdf<Interaction, MicrofacetCache>(_sample, interaction, cache, isInfinity);
+        _pdf = hlsl::mix(_pdf, scalar_type(0.0), isInfinity);
```
devshgraphicsprogramming: pointless you already handle it on line 186

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -156,19 +157,22 @@ struct SCookTorrance
...
 
         using quant_query_type = typename ndf_type::quant_query_type;
         quant_query_type qq = impl::quant_query_helper<ndf_type, fresnel_type, IsBSDF>::template __call<Interaction, MicrofacetCache>(ndf, _f, interaction, cache);
```
devshgraphicsprogramming: you've already computed this in `__forwardPdf`

devshgraphicsprogramming: any way to making `__forwardPdf` spit this out for you?
Make make `__forwardPdf<bool FromGenerator>` returna struct
```
scalar_type pdf;
frensel_type orientedFresenel;
quant_query_type quantQuery;
[optional only when `IsBSDF && not fresnel_type::ReturnsMonochrome`] spectral_type reflectance;
[optional only when `IsBSDF && not fresnel_type::ReturnsMonochrome`] scalar_type scaled_reflectance ;
```
and doesn't abuse `NBL_REF_ARG` to return extra things

Also this way this whole block of code
```
retval.pdf = 0.f;
retval.orientedFresenel = __getOrientedFresnel(fresnel, interaction.getNdotV());
const bool cacheIsValid = __checkValid<Interaction, MicrofacetCache>(retval.orientedFresenel, _sample, interaction, cache);
assert(!FromGenerator || cacheIsValid);
if (FromGenerator ? _sample.isValid():cacheIsValid)
{
... fill the members with the old function body
}
return retval;
```
can replace the preludes in quotientAndWeight and evalAndWeight  with just
```
if (pdfQuery.pdf==0.f)
   return ...;
```
up until the G2_over_G1 computation

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming, line 440: hmm we still get a pointless mul with 1.0 if the PDF is infinity, if you move this computation to after `quo` is constructed, we can turn the conditional `G2_over_G1 = ` into `quo *=`

It does require though that you `quo = promote<spectral_type>(1.0)` for Monochrome BSDF, thats not a problem because you go contexpr expressions first, so they'll optimize out and the mul with unity will optimize out too

Just the way you have it now, won't

devshgraphicsprogramming: `__forwardPDF` already computes these would be cool to return them

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -366,27 +373,27 @@ struct SCookTorrance
...
+    scalar_type forwardPdf(NBL_CONST_REF_ARG(sample_type) _sample, NBL_CONST_REF_ARG(Interaction) interaction, NBL_CONST_REF_ARG(MicrofacetCache) cache) NBL_CONST_MEMBER_FUNC
     {
         fresnel_type _f = __getOrientedFresnel(fresnel, interaction.getNdotV());
         if (!__checkValid<Interaction, MicrofacetCache>(_f, _sample, interaction, cache))
             return scalar_type(0.0);
 
         bool isInfinity;
-        scalar_type _pdf = __pdf<Interaction, MicrofacetCache>(_sample, interaction, cache, isInfinity);
+        scalar_type _pdf = __forwardPdf<Interaction, MicrofacetCache>(_sample, interaction, cache, isInfinity);
         return hlsl::mix(_pdf, scalar_type(0.0), isInfinity);
```
devshgraphicsprogramming: this is incorrect,  it should return INF for forward, it was only backward when it was not meant to (hence why I asked for the distinction)

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming: or the `__overwriteDG`

devshgraphicsprogramming: the `quotient...` function can just not use `quantQuery`its the same as using a dummy on a `NBL_REF_ARG`

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -179,19 +183,22 @@ struct SCookTorrance
         // immediately return only after all calls setting DG
         // allows compiler to throw away calls to ndf.D if using __overwriteDG, before that we only avoid computation for G2(correlated)
         if (isInfinity)
-            return hlsl::promote<spectral_type>(0.0);
+            return quotient_pdf_type::create(scalar_type(0.0), scalar_type(0.0));
 
         scalar_type clampedVdotH = cache.getVdotH();
         NBL_IF_CONSTEXPR(IsBSDF)
             clampedVdotH = hlsl::abs(clampedVdotH);
```
devshgraphicsprogramming: you only need this to call `const spectral_type reflectance = impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(clampedVdotH));`

which we already would have had `__forwardPdf` return to us because it needed to compute it

devshgraphicsprogramming: it doesn't compute it for you in the BRDF case, but you don't ened to have this whole tempvar and clamp it then

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -179,19 +183,22 @@ struct SCookTorrance
...
             const spectral_type reflectance = impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(clampedVdotH));
-            return hlsl::mix(reflectance, hlsl::promote<spectral_type>(1.0) - reflectance, cache.isTransmission()) * DG;
+            quo = hlsl::mix(reflectance, hlsl::promote<spectral_type>(1.0) - reflectance, cache.isTransmission()) * DG;
         }
         else
-            return impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(clampedVdotH)) * DG;
+            quo = impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(_f(clampedVdotH)) * DG;
+
+        return quotient_pdf_type::create(quo, _pdf);
     }
```
devshgraphicsprogramming, line 196: in combination with previous comment it would simplify to
```cpp
NBL_IF_CONSTEXPR(fresnel_type::IsBSDF)
{
   NBL_IF_CONSTEXPR(fresnel_type::ReturnsMonochrome)
   {
      quo =  pdfRetval.reflectance;
   }
   else
   {
      quo = mix(pdfRetval.reflectance,hlsl::promote<spectral_type>(1.0) - pdfRetval.reflectance, cache.isTransmission();
   }
}
else
   quo =impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call( _f(cache.getVdotH()));
quo *= DG;
```
with `DG` mul done outside

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -902,6 +902,7 @@ NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE)(T::sample_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::spectral_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::quotient_pdf_type))
```
devshgraphicsprogramming: this now needs to be `value_and_weight_type`

devshgraphicsprogramming: also require a `cache_type` which will be passed from `generate` to `quotientAndWeight`

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: you already check these elsewhere I think

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: you could actually get rid of `microfacex_bdf_common` and the `iso_microfacet` concept (basically the entire block in `namespace impl` ehere)

and just check that `{Isotropic}MicrofacetB{R|S}DF = {Isotropic}B{R|S}DF<T> && (Isotropic ? CreatableIsotropicMicrofacetCache<typedef T::cache_type>:AnisotropicMicrofacetCache<typedef T::cache_type>);`

include/nbl/builtin/hlsl/bxdf/base/lambertian.hlsl
```diff
@@ -24,39 +24,43 @@ struct SLambertianBase
     BXDF_CONFIG_TYPE_ALIASES(Config);
 
     using random_type = conditional_t<IsBSDF, vector3_type, vector2_type>;
+    struct Cache {};
+    using isocache_type = Cache;
+    using anisocache_type = Cache;
+    using evalcache_type = Cache;
```
devshgraphicsprogramming: there should be no `evalcache_type`, its still same cache as for quotient and generate

But having an `evalAndWeight` with an overload that takes it should be a Cook Torrance / Microfacet thing only

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -900,212 +899,82 @@ NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE)(T::scalar_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::anisotropic_interaction_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::sample_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::anisocache_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::evalcache_type))
```
devshgraphicsprogramming: don't require a separate evaluation cache type

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: > also require a `cache_type` which will be passed from `generate` to `quotientAndWeight`

done

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -900,212 +899,82 @@ NBL_CONCEPT_END(
...
 #define iso NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
 #define isocache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
+#define evalcache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
 NBL_CONCEPT_END(
-    ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(microfacet_bxdf_common, T))
+    ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(bxdf_common, T))
     ((NBL_CONCEPT_REQ_TYPE)(T::isotropic_interaction_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::isocache_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.eval_and_weight(_sample, iso, isocache)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.evalAndWeight(_sample, iso, evalcache)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
```
devshgraphicsprogramming: `evalAndWeight` taking the `anisocache` or `isocache` should be an extra overload in the imcrofacet BxDF concepts

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: > this now needs to be `value_and_weight_type`

my bad, `quotient_and_weight_type`

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -160,19 +172,20 @@ struct SCookTorrance
...
-        quant_query_type qq = impl::quant_query_helper<ndf_type, fresnel_type, IsBSDF>::template __call<Interaction, MicrofacetCache>(ndf, _f, interaction, cache);
+        quant_query_type qq = pdfQuery.quantQuery;
 
         using g2g1_query_type = typename ndf_type::g2g1_query_type;
         g2g1_query_type gq = ndf.template createG2G1Query<sample_type, Interaction>(_sample, interaction);
 
+        bool isInfinity;
         quant_type D = ndf.template D<sample_type, Interaction, MicrofacetCache>(qq, _sample, interaction, cache, isInfinity);
         scalar_type DG = D.projectedLightMeasure;
         if (!isInfinity)
```
devshgraphicsprogramming, line 184: why would be not use `_pdf < INF` instead of what `template D` returns in a side channel, is there a historic reason for that like some GGX vs Beckmann optimizations?

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming: in the BRDF case you can do `__call(_f(cache.getVdotH()))` because you can pretty much assert it must be non-negative at this point (otherwise pdf is 0 and you never get here

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming: ok my bad, its because the PDF can be anything, doesn't need to be VNDF so PDF can be non INF while real value is

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -347,77 +358,76 @@ struct SCookTorrance
...
-            return DG1.projectedLightMeasure;
+            NBL_IF_CONSTEXPR(IsBSDF)
+            {
+                query.scaled_reflectance = __getScaledReflectance(query.orientedFresnel, interaction, hlsl::abs(cache.getVdotH()), cache.isTransmission(), query.reflectance);
+                query.pdf = query.scaled_reflectance * DG1.projectedLightMeasure;
+            }
+            else
+            {
+                query.pdf = DG1.projectedLightMeasure;
+            }
```
devshgraphicsprogramming, line 393: how about
```cpp
query.pdf = DG1.projectedLightMeasure;
NBL_IF_CONSTEXPR (IsBSDF)
{
   query.scaled_reflectance = __getScaledReflectance(query.orientedFresnel, interaction, hlsl::abs(cache.getVdotH()), cache.isTransmission(), query.reflectance);
   query.pdf *= query.scaled_reflectance;
}
```

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -347,77 +358,76 @@ struct SCookTorrance
...
-        scalar_type G2_over_G1 = scalar_type(1.0);
-        if (!isInfinity)
-        {
-            using g2g1_query_type = typename N::g2g1_query_type;
-            g2g1_query_type gq = ndf.template createG2G1Query<sample_type, Interaction>(_sample, interaction);
-            G2_over_G1 = ndf.template G2_over_G1<sample_type, Interaction, MicrofacetCache>(gq, _sample, interaction, cache);
-        }
+        using g2g1_query_type = typename N::g2g1_query_type;
+        g2g1_query_type gq = ndf.template createG2G1Query<sample_type, Interaction>(_sample, interaction);
+        scalar_type G2_over_G1 = ndf.template G2_over_G1<sample_type, Interaction, MicrofacetCache>(gq, _sample, interaction, cache);
```
devshgraphicsprogramming: check how well this behaves as `roughness ->0` we don't want to have NaNs

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming, line 439: why do you use ` impl::__implicit_promote<spectral_type, typename fresnel_type::vector_type>::__call(` in `eval` but not here ?

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
keptsecret: We have a check in ndfs for `a<float_min`

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
keptsecret: None of the brdf-only fresnels output monochrome so they don't need `implicit_promote`, so it's not needed anymore and I've removed it as well as usage in eval

keptsecret: Unless we still might need it?

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming: we're bypassing a bunch of things , so there's no check
https://github.com/Devsh-Graphics-Programming/Nabla/blob/25395228f8eb0b03dac7c79c5846aa8b2f343ca8/include/nbl/builtin/hlsl/bxdf/ndf/ggx.hlsl#L364

you will in-fact assert in the constructor 
https://github.com/Devsh-Graphics-Programming/Nabla/blob/25395228f8eb0b03dac7c79c5846aa8b2f343ca8/include/nbl/builtin/hlsl/bxdf/ndf/ggx.hlsl#L236
of the query
https://github.com/Devsh-Graphics-Programming/Nabla/blob/25395228f8eb0b03dac7c79c5846aa8b2f343ca8/include/nbl/builtin/hlsl/bxdf/ndf/ggx.hlsl#L104

I mean the HLSL compiles to C++ and we have a unit test so you can go through it with a debugger I guess 🙃 

I think if you know/check the PDF not being INF before setting `G2_over_G1` you can guard against that

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming: > Unless we still might need it?

no all good.

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -951,27 +947,73 @@ NBL_CONCEPT_END(
...
+NBL_CONCEPT_BEGIN(4)
+#define bxdf NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_0
+#define _sample NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_1
+#define aniso NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+#define evalcache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
+NBL_CONCEPT_END(
+    ((NBL_CONCEPT_REQ_TYPE)(T::evalcache_type))
+    ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(bxdf_common, T))
+    ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(AnisotropicMicrofacetCache, typename T::anisocache_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.evalAndWeight(_sample, aniso, evalcache)), ::nbl::hlsl::is_same_v, typename T::value_weight_type))
```
devshgraphicsprogramming: cool but `evalcache` needs to be just `microfacetCache` and be of type `T::cache_type`

devshgraphicsprogramming: no `evalcache_type`

include/nbl/builtin/hlsl/path_tracing/concepts.hlsl
```diff
@@ -216,10 +216,10 @@ NBL_CONCEPT_BEGIN(1)
 NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE)(T::scalar_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::sample_type))
-    ((NBL_CONCEPT_REQ_TYPE)(T::quotient_pdf_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::quotient_weight_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::object_handle_type))
     ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((sqr.getSample()), ::nbl::hlsl::is_same_v, typename T::sample_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((sqr.getQuotientPdf()), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((sqr.getQuotientPdf()), ::nbl::hlsl::is_same_v, typename T::quotient_weight_type))
```
devshgraphicsprogramming: `getQuotientPdf` rename to `getQuotientAndWeight`

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
keptsecret: added some checks for ggx

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -332,85 +350,81 @@ struct SCookTorrance
...
-        if (!_sample.isValid())
-            return quotient_pdf_type::create(scalar_type(0.0), scalar_type(0.0));   // set pdf=0 when quo=0 because we don't want to give high weight to sampling strategy that yields 0 contribution
-
-        bool isInfinity;
-        scalar_type _pdf = __pdf<Interaction, MicrofacetCache>(_sample, interaction, cache, isInfinity);
-        fresnel_type _f = __getOrientedFresnel(fresnel, interaction.getNdotV());
+        PdfQuery pdfQuery = __forwardPdf<Interaction, MicrofacetCache, true>(_sample, interaction, cache);
+        scalar_type _pdf = pdfQuery.pdf;
+        if (_pdf == scalar_type(0.0) || hlsl::isinf(_pdf))
+            return quotient_weight_type::create(scalar_type(0.0), scalar_type(0.0));
```
devshgraphicsprogramming: nope, this is the quotient function, so if you have ifn pdf you just don't apply the `G2_over_G1` (fastest will be to `hlsl::select` between 1 and value computed by ndf

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -972,19 +972,20 @@ NBL_CONCEPT_END(
...
 #define aniso NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
-#define evalcache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
+#define microfacetCache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
 NBL_CONCEPT_END(
-    ((NBL_CONCEPT_REQ_TYPE)(T::evalcache_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::cache_type))
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(bxdf_common, T))
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(AnisotropicMicrofacetCache, typename T::anisocache_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.evalAndWeight(_sample, aniso, evalcache)), ::nbl::hlsl::is_same_v, typename T::value_weight_type))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.evalAndWeight(_sample, aniso)), ::nbl::hlsl::is_same_v, typename T::value_weight_type))
```
devshgraphicsprogramming: BxDF_common should already check for that once

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming, line 960: checking for the wrong signature

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: same for isotropic

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: same for anisotropic

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: https://github.com/Devsh-Graphics-Programming/Nabla/pull/1027/changes/bea47804dd8f84fd56ae8ebd31820f0597bc56be#r2999586998

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
```diff
@@ -409,14 +409,18 @@ struct SCookTorrance
...
+        if (_pdf == scalar_type(0.0))
             return quotient_weight_type::create(scalar_type(0.0), scalar_type(0.0));
 
         fresnel_type _f = pdfQuery.orientedFresnel;
 
-        using g2g1_query_type = typename N::g2g1_query_type;
-        g2g1_query_type gq = ndf.template createG2G1Query<sample_type, Interaction>(_sample, interaction);
-        scalar_type G2_over_G1 = ndf.template G2_over_G1<sample_type, Interaction, MicrofacetCache>(gq, _sample, interaction, cache);
+        scalar_type G2_over_G1 = scalar_type(1.0);
+        if (_pdf < bit_cast<scalar_type>(numeric_limits<scalar_type>::infinity))
```
devshgraphicsprogramming: can you not use `hlsl::isinf()` ?

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -998,19 +1001,23 @@ NBL_CONCEPT_END(
 #define NBL_CONCEPT_PARAM_1 (_sample, typename T::sample_type)
 #define NBL_CONCEPT_PARAM_2 (iso, typename T::isotropic_interaction_type)
 #define NBL_CONCEPT_PARAM_3 (microfacetCache, typename T::cache_type)
-NBL_CONCEPT_BEGIN(4)
+#define NBL_CONCEPT_PARAM_4 (isocache, typename T::isocache_type)
```
devshgraphicsprogramming, line 1004: whats the difference between `cache_type` and `isocache_type` ?

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -973,18 +973,21 @@ NBL_CONCEPT_END(
 #define NBL_CONCEPT_PARAM_1 (_sample, typename T::sample_type)
 #define NBL_CONCEPT_PARAM_2 (aniso, typename T::anisotropic_interaction_type)
 #define NBL_CONCEPT_PARAM_3 (microfacetCache, typename T::cache_type)
-NBL_CONCEPT_BEGIN(4)
+#define NBL_CONCEPT_PARAM_4 (anisocache, typename T::anisocache_type)
```
devshgraphicsprogramming, line 976: whats the difference between `cache_type` and `anisocache_type` ?

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
keptsecret: I thought we shouldn't use `isinf()`/checking `<inf` is better
https://github.com/Devsh-Graphics-Programming/Nabla/pull/991#discussion_r2894644506

include/nbl/builtin/hlsl/bxdf/common.hlsl
keptsecret: `cache_type` is the type for `evalAndWeight` only in microfacet bxdfs. `evalAndWeight` overloads for both aniso/iso cache and eval cache types (they're the same in our current microfacet bxdfs)
https://github.com/Devsh-Graphics-Programming/Nabla/pull/1027#discussion_r2982386946

keptsecret: Unless I'm misunderstanding that comment

devshgraphicsprogramming: `cache_type` is for `quotient` in all BxDFs

I'm asking if there's any difference between the `cache_type` you'd pass in Microfacet between `generate` and `quotient` and the microfacet cache we had before for the half-vector stuff thats also useful for the one extra microfacet `eval`

include/nbl/builtin/hlsl/bxdf/base/cook_torrance_base.hlsl
devshgraphicsprogramming: you don't need to protect against `NaN` here though

include/nbl/builtin/hlsl/bxdf/common.hlsl
keptsecret: No there's currently no difference. Only that `cache_type` will be iso or aniso depending on the BSDF (determined by the NDF).
All set by the bxdf config struct

include/nbl/builtin/hlsl/bxdf/common.hlsl
keptsecret: https://github.com/Devsh-Graphics-Programming/Nabla/pull/1027#discussion_r3008828895

include/nbl/builtin/hlsl/path_tracing/unidirectional.hlsl
```diff
@@ -139,9 +139,9 @@ struct Unidirectional
                 return false;
 
             // the value of the bsdf divided by the probability of the sample being generated
-            quotient_pdf_type bsdf_quotient_pdf = materialSystem.quotient_and_pdf(matID, bsdf_sample, interaction, _cache);
-            throughput *= bsdf_quotient_pdf.quotient();
-            bxdfPdf = bsdf_quotient_pdf.pdf();
+            quotient_weight_type bsdf_quotient_weight = materialSystem.quotientAndWeight(matID, bsdf_sample, interaction, _cache);
+            throughput *= bsdf_quotient_weight.quotient();
+            bxdfPdf = bsdf_quotient_weight.pdf();
```
devshgraphicsprogramming, line 144: why is this called Pdf and why can I get a `pdf()` from quotientAndWeight` ?

include/nbl/builtin/hlsl/path_tracing/unidirectional.hlsl
```diff
@@ -92,26 +93,25 @@ struct Unidirectional
...
             scalar_type t = ret.getT();
             sample_type nee_sample = ret.getSample();
-            quotient_pdf_type neeContrib = ret.getQuotientPdf();
+            quotient_weight_type neeContrib = ret.getQuotientWeight();
 
             // While NEE or other generators are not supposed to pick up Delta lobes by accident, we need the MIS weights to add up to 1 for the non-delta lobes.
             // So we need to weigh the Delta lobes as if the MIS weight is always 1, but other areas regularly.
             // Meaning that eval's pdf should equal quotient's pdf , this way even the diffuse contributions coming from within a specular lobe get a MIS weight near 0 for NEE.
             // This stops a discrepancy in MIS weights and NEE mistakenly trying to add non-delta lobe contributions with a MIS weight > 0 and creating energy from thin air.
             if (neeContrib.pdf() > scalar_type(0.0))
```
devshgraphicsprogramming, line 108: why is there a pdf()

include/nbl/builtin/hlsl/path_tracing/unidirectional.hlsl
```diff
@@ -92,26 +93,25 @@ struct Unidirectional
...
             {
-                // TODO: we'll need an `eval_and_mis_weight` and `quotient_and_mis_weight`
-                const scalar_type bsdf_pdf = materialSystem.pdf(matID, nee_sample, interaction);
-                neeContrib._quotient *= materialSystem.eval(matID, nee_sample, interaction) * rcpChoiceProb;
+                value_weight_type bsdfContrib = materialSystem.evalAndWeight(matID, nee_sample, interaction);
+                neeContrib._quotient *= bsdfContrib.value() * rcpChoiceProb;
                 if (neeContrib.pdf() < bit_cast<scalar_type>(numeric_limits<scalar_type>::infinity))
                 {
-                    const scalar_type otherGenOverLightAndChoice = bsdf_pdf * rcpChoiceProb / neeContrib.pdf();
+                    const scalar_type otherGenOverLightAndChoice = bsdfContrib.weight() * rcpChoiceProb / neeContrib.pdf();
```
devshgraphicsprogramming, line 114: why is `pdf()` of contrib used

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
```diff
@@ -667,7 +667,7 @@ struct iridescent_base
     template<bool SupportsTransmission, typename Colorspace>
     T __call(const vector_type _iork3, const vector_type etak23, const scalar_type clampedCosTheta) NBL_CONST_MEMBER_FUNC
     {
-        return impl::iridescent_helper<T,false>::template __call<Colorspace>(D, ior1, ior2, ior3, _iork3,
+        return impl::iridescent_helper<T,SupportsTransmission>::template __call<Colorspace>(D, ior1, ior2, ior3, iork3,
```
devshgraphicsprogramming, line 670: wrong merge, iork3 shadows

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: GH UI broke, can't see the response

include/nbl/builtin/hlsl/bxdf/reflection/delta_distribution.hlsl
```diff
@@ -47,28 +52,28 @@ struct SDeltaDistribution
...
 
-    sample_type generate(NBL_CONST_REF_ARG(isotropic_interaction_type) interaction, const vector2_type u) NBL_CONST_MEMBER_FUNC
+    sample_type generate(NBL_CONST_REF_ARG(isotropic_interaction_type) interaction, const random_type u, NBL_REF_ARG(isocache_type) _cache) NBL_CONST_MEMBER_FUNC
     {
-        return generate(anisotropic_interaction_type::create(interaction), u);
+        return generate(anisotropic_interaction_type::create(interaction), u, _cache);
     }
 
-    scalar_type pdf(NBL_CONST_REF_ARG(sample_type) _sample, NBL_CONST_REF_ARG(isotropic_interaction_type) interaction) NBL_CONST_MEMBER_FUNC
+    scalar_type forwardPdf(NBL_CONST_REF_ARG(sample_type) _sample, NBL_CONST_REF_ARG(isotropic_interaction_type) interaction) NBL_CONST_MEMBER_FUNC
```
devshgraphicsprogramming, line 60: `forwardPdf` must take the cache

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -900,31 +900,41 @@ NBL_CONCEPT_END(
...
+#define NBL_CONCEPT_PARAM_3 (anisocache, typename T::anisocache_type)
+NBL_CONCEPT_BEGIN(4)
 #define bxdf NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_0
 #define _sample NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_1
 #define aniso NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+#define anisocache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
 NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(bxdf_common_typdefs, T))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.eval(_sample, aniso)), ::nbl::hlsl::is_same_v, typename T::spectral_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.pdf(_sample, aniso)), ::nbl::hlsl::is_same_v, typename T::scalar_type))
```
devshgraphicsprogramming, line 922: still need to check for `forwardPdf`

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -900,31 +900,41 @@ NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE)(T::scalar_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::anisotropic_interaction_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::sample_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::anisocache_type))
     ((NBL_CONCEPT_REQ_TYPE)(T::spectral_type))
-    ((NBL_CONCEPT_REQ_TYPE)(T::quotient_pdf_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::quotient_weight_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::value_weight_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::random_type))
```
devshgraphicsprogramming, line 907: TODO: can we check `NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT` that these are correct:
1. fundamental scalar
2. interaction
3. sample
4. spectral type is a vector ?
5. quotient and pdf type is quotient and pdf
6. value and weight is value and weight
7. random type is a floating point vectorial

We should probably remove `quotient_pdf_type`

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -936,19 +946,79 @@ NBL_CONCEPT_END(
...
 #define iso NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+#define isocache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
 NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(bxdf_common, T))
     ((NBL_CONCEPT_REQ_TYPE)(T::isotropic_interaction_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.eval(_sample, iso)), ::nbl::hlsl::is_same_v, typename T::spectral_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.pdf(_sample, iso)), ::nbl::hlsl::is_same_v, typename T::scalar_type))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.quotient_and_pdf(_sample, iso)), ::nbl::hlsl::is_same_v, typename T::quotient_pdf_type))
-    ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(concepts::FloatingPointLikeVectorial, typename T::spectral_type))
+    ((NBL_CONCEPT_REQ_TYPE)(T::isocache_type))
```
devshgraphicsprogramming, line 958: TODO: check iso types meet concepts

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -936,19 +946,79 @@ NBL_CONCEPT_END(
...
+#define NBL_CONCEPT_PARAM_3 (microfacetCache, typename T::cache_type)
+#define NBL_CONCEPT_PARAM_4 (anisocache, typename T::anisocache_type)
+NBL_CONCEPT_BEGIN(5)
+#define bxdf NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_0
+#define _sample NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_1
+#define aniso NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+#define microfacetCache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
+#define anisocache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_4
+NBL_CONCEPT_END(
+    ((NBL_CONCEPT_REQ_TYPE)(T::cache_type))
```
devshgraphicsprogramming, line 984: why have a `cache_type` when there's a separate aniso and iso cache ?

include/nbl/builtin/hlsl/bxdf/common.hlsl
```diff
@@ -960,15 +1030,19 @@ NBL_CONCEPT_END(
...
+#define NBL_CONCEPT_PARAM_3 (anisocache, typename T::anisocache_type)
+NBL_CONCEPT_BEGIN(4)
 #define bxdf NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_0
 #define aniso NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_1
 #define u NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+#define anisocache NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_3
 NBL_CONCEPT_END(
     ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(impl::bxdf_common, T))
-    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((bxdf.generate(aniso,u)), ::nbl::hlsl::is_same_v, typename T::sample_type))
+    ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(impl::VecDim2OrMore, typename T::random_type))
```
devshgraphicsprogramming, line 1042: should be done in bxdf_common

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: I take it because they're used for two different signatures

include/nbl/builtin/hlsl/bxdf/fresnel.hlsl
devshgraphicsprogramming: fixed on my branch

include/nbl/builtin/hlsl/path_tracing/unidirectional.hlsl
devshgraphicsprogramming: done by me

include/nbl/builtin/hlsl/path_tracing/unidirectional.hlsl
devshgraphicsprogramming: done by me

include/nbl/builtin/hlsl/path_tracing/unidirectional.hlsl
devshgraphicsprogramming: done by me

include/nbl/builtin/hlsl/bxdf/reflection/delta_distribution.hlsl
devshgraphicsprogramming: fixed by me

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: fixed

include/nbl/builtin/hlsl/bxdf/common.hlsl
devshgraphicsprogramming: fixed

PR #1029, opened 2026-03-24
Speed up HLSL preprocessing and prepared SPIR-V hot paths

AnastaZIuk:
## Summary

- reduce Wave include overhead in the hot HLSL path with an explicit per-session include cache using separate read and write session caches
- classify builtin and generated include roots by provenance instead of guessing from include spelling
- teach source-built `nsc` to accept `-isystem` so toolchain include roots can be registered explicitly in builtins-off flows
- keep the prepared-SPIR-V hot-path improvements: a single-entrypoint trim fast path and validation once per unique content hash
- thread one `IGPUPipelineCache` through compute, resolve, ImGui, and fullscreen present in the paired EX31 flow
- update the Examples pointer to the paired [Devsh-Graphics-Programming/Nabla-Examples-and-Tests#262](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/262)

## Root cause

Three costs were stacking on top of each other.

First, the preprocess part comes from avoidable HLSL include debt in the hot path:

- [`path_tracing/concepts.hlsl` on the base branch](https://github.com/Devsh-Graphics-Programming/Nabla/blob/25395228f8eb0b03dac7c79c5846aa8b2f343ca8/include/nbl/builtin/hlsl/path_tracing/concepts.hlsl#L7-L8) pulls `bxdf/common.hlsl` only to synthesize a placeholder interaction for `Ray::setInteraction`; that edge comes from [`4d186db76f`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/4d186db76f)
- [`member_test_macros.hlsl` on the base branch](https://github.com/Devsh-Graphics-Programming/Nabla/blob/25395228f8eb0b03dac7c79c5846aa8b2f343ca8/include/nbl/builtin/hlsl/member_test_macros.hlsl#L7-L8) uses the umbrella `boost/preprocessor.hpp` even though this header only needs a narrow subset; that comes from [`72972a9d6e`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/72972a9d6e)
- the custom Wave include bridge on this path was introduced in [`12afd3d42d`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/12afd3d42d39398e310b7b3cea8f8b61e4753c9b), which added the custom Boost.Wave context and include-path classes for the HLSL preprocessor; `dxc_compile_flags` pragma bookkeeping was later layered on in [`ae4386064cf`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/ae4386064cf); later merges, cleanup, depfile plumbing, and backports carried the same path forward but are not the semantic origin of the extra per-include work

Second, the base include-loader path paid redundant work before preprocessing reached DXC. The current disk-backed include body load path in [`IShaderCompiler.cpp`](https://github.com/Devsh-Graphics-Programming/Nabla/blob/25395228f8eb0b03dac7c79c5846aa8b2f343ca8/src/nbl/asset/utils/IShaderCompiler.cpp#L603-L721) comes from [`5ac3b55552`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/5ac3b55552) and later loader reshapes like [`cc37325f28c`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/cc37325f28c). Per-lookup content hashing on that path was added in [`cf9a866623`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/cf9a8666233). The hot include bridge also lacked an explicit notion of builtin/generated include roots, which made toolchain headers harder to classify and cache cleanly.

Third, the pre-fast-path trimmer always validated and walked the incoming module before it could know whether the requested entrypoint set already matched the prepared shader. The old flow is visible in [`ISPIRVEntryPointTrimmer.cpp#L104-L246`](https://github.com/Devsh-Graphics-Programming/Nabla/blob/25395228f8eb0b03dac7c79c5846aa8b2f343ca8/src/nbl/asset/utils/ISPIRVEntryPointTrimmer.cpp#L104-L246). That shape comes from [`cfb4bd1da6`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/cfb4bd1da6f069b23aa784467ac576eac19e751d) and [`9f3f823124`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/9f3f82312477a9fc747bf7ae186fb491b6d6edba).

The fullscreen-present helper was introduced in [`2b08a15064`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/2b08a15064de806b024a67ecb3d743a38a809ee2). In that shape [`CFullScreenTriangle.cpp#L120`](https://github.com/Devsh-Graphics-Programming/Nabla/blob/25395228f8eb0b03dac7c79c5846aa8b2f343ca8/src/nbl/ext/FullScreenTriangle/CFullScreenTriangle.cpp#L120) did not yet thread an external pipeline cache, so compute and present could not populate the same cache blob.

## What this changes

- cache and reuse include resolution results explicitly per preprocess session through separate read and write session caches
- classify builtin and generated roots when they are registered instead of inferring special treatment from include spelling
- let `nsc` accept `-isystem` and map those roots to system-classified include search paths in source-built flows
- keep toolchain and generated headers on the fast path without changing the normal `""` versus `<>` search semantics
- trim token bookkeeping in `CWaveStringResolver`
- replace the umbrella Boost include in `member_test_macros.hlsl` with the narrow Boost headers it actually uses
- remove redundant public HLSL includes from hot headers and stop pulling `bxdf/common.hlsl` into `path_tracing/concepts.hlsl`
- short-circuit `ISPIRVEntryPointTrimmer` when the incoming module is already a prepared single-entrypoint shader
- cache successful validation per unique SPIR-V blob so hot paths keep validation without paying for it again
- thread an external pipeline cache through `FullScreenTriangle` so EX31 can share one cache object across compute and present

## Validation

Validation was run on AMD Ryzen 5 5600G with Radeon Graphics (6C/12T).

Exact local source-built `nsc` `Release -P` sweeps on the current EX31 scene rules taken from the generated build commands show:

- `9` heavy scene rules total
- min `2.424 s`
- avg `2.503 s`
- max `2.632 s`

Local source-built `nsc` preprocess profiles on the current EX31 heavy sphere rule show:

- builtins `OFF`: `include_requests=586`, `include_lookups=316`, `resolution_cache_skips=270`, `session_lookup_found=0`
- builtins `ON`: `include_requests=586`, `include_lookups=234`, `resolution_cache_skips=352`, `session_lookup_found=44`

The paired EX31 branch builds and runs in `RelWithDebInfo` with both builtins modes. Current warm-cache validation on the paired branch is:

- builtins `OFF`: `first_render_submit_ms=1533`
- builtins `ON`: `first_render_submit_ms=1850`

Prepared-shader and pipeline-cache validation on the paired EX31 branch is recorded in [Devsh-Graphics-Programming/Nabla-Examples-and-Tests#262](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/262).


src/nbl/asset/utils/IShaderCompiler.cpp
```diff
@@ -656,37 +704,56 @@ IShaderCompiler::CIncludeFinder::CIncludeFinder(core::smart_refctd_ptr<system::I
...
+        else retVal = m_defaultFileSystemLoader->getInclude(requestingSourceDir.string(), lookupName, needHash);
+    }
+    else
+    {
+        if (auto contents = m_defaultFileSystemLoader->getInclude(requestingSourceDir.string(), lookupName, needHash))
+            retVal = std::move(contents);
+        else if (auto contents = tryIncludeGenerators(lookupName))
+            retVal = std::move(contents);
+        else retVal = std::move(trySearchPaths(lookupName, needHash));
+    }
```
devshgraphicsprogramming, line 843: explain the reason for this change

devshgraphicsprogramming: you shouldn't try different include generators, the include generators should only be reachable with `#include <>` a and not `#include ""`

Also why should the precedence of a search path and default include loaders change depending on the path ?

src/nbl/asset/utils/includeResolutionCommon.h
```diff
@@ -0,0 +1,36 @@
...
+    {
+        if (includeName.rfind(prefix, 0ull) == 0ull)
+            return true;
+    }
+
+    return false;
+}
+}
+
+#endif
```
devshgraphicsprogramming: I don't get the rationale for this existing

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -65,10 +68,13 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
...
 				CFileSystemIncludeLoader(core::smart_refctd_ptr<system::ISystem>&& system);
 
-				IIncludeLoader::found_t getInclude(const system::path& searchPath, const std::string& includeName) const override;
+				IIncludeLoader::found_t getInclude(const system::path& searchPath, const std::string& includeName, bool needHash = true) const override;
 
 			protected:
 				core::smart_refctd_ptr<system::ISystem> m_system;
+				mutable std::mutex m_cacheMutex;
+				mutable std::unordered_map<std::string, IIncludeLoader::found_t> m_cache;
+				mutable std::unordered_set<std::string> m_missingCache;
```
devshgraphicsprogramming: there's numerous problems I see with this approach:
1. files get created and deleted throughout the application runtime, and the `CFileSystemIncludeLoader` is a long lived object. The include file contents will change, and because the IFIle will be closed in the meantime there's nothing to stop (on the OS level) that change occuring, you will get stale cache hits and hot-reload (for whatever example implements it) won't kick in.
2. the `IIncludeLoader::found_t::contents` can be SEVERAL kilobytes per file, this cache will basically live forever as nabla runs so its basically a multi-MB memory leak

furthermore because of concurrency you needed to add a mutex

It would make more sense if you wrapped this up into a `SSessionCache` struct in the `CIncludeFinder` and specified it as a mutable member in `SProcessorOptions` 

this way:
1. you can have threadsafe session cache and non thread safe session cache (with and without mutex)
2. you can pass the session cache EXPLICITLY from one compile to another, or share it concurrently
3. you're able to get rid of its memory

devshgraphicsprogramming: I think some of the functionality of this was already in 
https://github.com/Devsh-Graphics-Programming/Nabla/blob/f5f036e488db23b6fdb2fb71e7c198f37447ee37/include/nbl/asset/utils/IShaderCompiler.h#L197

but it obviously served a different purpose (whole IShader from includes)

but the design there is sound esp with the `readCache` and `writeCache` on the Options struct

include/nbl/asset/utils/IShaderCompiler.h
```diff
@@ -140,6 +196,7 @@ class NBL_API2 IShaderCompiler : public core::IReferenceCounted
 			std::string_view sourceIdentifier = "";
 			system::logger_opt_ptr logger = nullptr;
 			const CIncludeFinder* includeFinder = nullptr;
+			mutable CIncludeFinder::SSessionCache* includeSessionCache = nullptr;
```
devshgraphicsprogramming: why mutable? and why not do a readCache and writeCache so you can copy only relevant things from read into write  or have the pointers equal

AnastaZIuk: hmm right, on it

src/nbl/asset/utils/IShaderCompiler.cpp
AnastaZIuk: >  the include generators should only be reachable with #include <> a and not #include ""

wrong, we do now what clang does

AnastaZIuk: https://github.com/llvm/llvm-project/blob/e69c7312f31be57af5ef17ada7466b46f8519f2a/clang/include/clang/Lex/HeaderSearchOptions.h#L312-L313

https://github.com/llvm/llvm-project/blob/e69c7312f31be57af5ef17ada7466b46f8519f2a/clang/include/clang/Lex/HeaderSearchOptions.h#L320-L321

https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-in-system-headers

> Here, #include "x/a.h" is treated as including a system header, even if the header is found in foo

devshgraphicsprogramming: I saw you added `isystem` really cool, approved

PR #1037, opened 2026-03-31
Fix ImGui shutdown and update EX31 runtime fixes

AnastaZIuk:
## Summary

- guard `nbl::ext::imgui::UI` teardown so shutdown only ends an active frame on the owned ImGui context
- update `examples_tests` to the paired EX31 fixes from [Devsh-Graphics-Programming/Nabla-Examples-and-Tests#264](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/264), including the RWMC default reset used for brightness-parity comparisons
- keep the engine-side change narrow: one ImGui shutdown fix plus the submodule update the branch needs

## Root cause

The engine-side runtime break here is separate from the EX31 routing fixes in `examples_tests`.

`UI::~UI()` was introduced in [`2de19f25cf`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/2de19f25cf) when the ImGui backend grew explicit shared-atlas ownership handling. In that shape the destructor unconditionally called `ImGui::EndFrame()` whenever the UI instance owned the atlas, without checking whether the owned context was current or whether a frame was still open. [`699e841a38`](https://github.com/Devsh-Graphics-Programming/Nabla/commit/699e841a38) later widened the practical exposure on the packaged-SPIR-V path by letting the UI consume precompiled ImGui shaders, which is the route exercised by this branch.

On the EX31 screenshot and exit path teardown can happen after the frame has already been ended, so the destructor hits ImGui's debug break even though rendering already finished correctly. The EX31 warmup, rectangle, and RWMC regressions themselves live in `examples_tests`, so the right Nabla-side fix is just the shutdown guard together with the paired submodule update. The additional RWMC-defaults tuning in that submodule update is an examples-side comparison-quality upgrade, not an engine-side regression fix.

### DIFF
--- a/src/nbl/ext/ImGui/ImGui.cpp
+++ b/src/nbl/ext/ImGui/ImGui.cpp
@@ -10,6 +10,7 @@
 #include "nbl/ext/ImGui/ImGui.h"
 #include "nbl/ext/ImGui/builtin/hlsl/common.hlsl"
 #include "imgui/imgui.h"
+#include "imgui/imgui_internal.h"
 #include "imgui/misc/cpp/imgui_stdlib.h"
 
 #ifdef NBL_EMBED_BUILTIN_RESOURCES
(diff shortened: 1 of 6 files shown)


AnastaZIuk:
https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/264#issuecomment-4169195718 ex40 assert issue https://github.com/microsoft/DirectXShaderCompiler/pull/8324 is gone now, CI green and ready for deployment

PR #1046, opened 2026-04-12
enable -fspv-debug=source and line by default, also make these flags affect RWDI

karimsayedre:


AnastaZIuk:
ok looks good to me but we have a problem, our SPIRV is too big for embed esp for actions runner, so we have 2 options

- ZSTD builtins integration
- we drop some of debug options by default for CI <- preferred, we can ZSTD in another PR

karimsayedre:
so basically use `NBL_CI_MODE` to check?

AnastaZIuk:
> so basically use `NBL_CI_MODE` to check?

I would do this in CI base preset and put all bits explicitly here

https://github.com/Devsh-Graphics-Programming/Nabla/blob/162b3dd6644a871645e25a72774b0dc608c58816/CMakePresets.json#L12

this way we go with debug options nicely and CI just declares less options as ON

PR #1048, opened 2026-04-14
Classic sincos patch

Fletterio:
It's my understanding that all big vendors run sqrt and sin/cos on the same hardware. So if running sin then cos you saturate the pipeline with sin calls, stall on that, then saturate the pipeline with cos calls. It also allows for vendor-specific optimizations for sincos if those exist.

Current implementation saturates the same pipeline for special functions when calling cos, then goes through FMA twice to compute 1 - c^2, then the same pipeline that would be used for sin but used to compute the sqrt, and then a sign inversion. Not only is it slower (if my understanding is correct) it's also less precise near multiples of pi (although that's probably unimportant near zero).


devshgraphicsprogramming:
I think @karimsayedre pointed out a similar thing to me recently

devshgraphicsprogramming:
@Fletterio we just merged a similar thing to `master` just now

PR #1061, opened 2026-04-23
Vk cuda interop

kevyuu:
## Description
Bridging CUDA and Vulkan

## Testing 
[Test Pull Request](https://github.com/Devsh-Graphics-Programming/Nabla-Examples-and-Tests/pull/274)


devshgraphicsprogramming:
@kevyuu CI checks are not passing I can't merge

include/nbl/video/CCUDAHandler.h
```diff
@@ -268,6 +295,14 @@ class CCUDAHandler : public core::IReferenceCounted
 		int m_version;
 };
 
+#define ASSERT_CUDA_SUCCESS(expr, handler) \
```
devshgraphicsprogramming: this macro needs a better name because you're leaking it into every public Nabla Header.

include/nbl/video/CCUDADevice.h
```diff
@@ -63,132 +74,44 @@ class CCUDADevice : public core::IReferenceCounted
 		};
 		inline E_VIRTUAL_ARCHITECTURE getVirtualArchitecture() {return m_virtualArchitecture;}
 
+		CCUDADevice(core::smart_refctd_ptr<CVulkanConnection>&& vulkanConnection, IPhysicalDevice* const vulkanDevice, const E_VIRTUAL_ARCHITECTURE virtualArchitecture, CUdevice device, core::smart_refctd_ptr<CCUDAHandler>&& handler);
+
+		~CCUDADevice();
+
```
devshgraphicsprogramming, line 74: 1. if its `Refcounted` it the destructor shall always be protected and not public
2. it makes no sense for the constructor to be public since it seems like the only way to create the `CCUDADevice` object is through a factory of some sort

include/nbl/video/CCUDADevice.h
devshgraphicsprogramming: there's now `8.6 | 8.7 | 8.9 | 9.0 | 10.0 | 10.3 | 11.0 | 12.0 | 12.1`

include/nbl/video/CCUDAHandler.h
```diff
@@ -137,6 +158,8 @@ class CCUDAHandler : public core::IReferenceCounted
 		);
 		const NVRTC& getNVRTCFunctionTable() const {return m_nvrtc;}
 
+		CCUDAHandler(CUDA&& _cuda, NVRTC&& _nvrtc, core::vector<core::smart_refctd_ptr<system::IFile>>&& _headers, core::smart_refctd_ptr<system::ILogger>&& _logger, int _version);
+
```
devshgraphicsprogramming: why is your constructor public if you have a `create` factory ?

include/nbl/video/CCUDAImportedMemory.h
```diff
@@ -0,0 +1,42 @@
...
+    public:
+
+      CCUDAImportedMemory(core::smart_refctd_ptr<CCUDADevice> device, core::smart_refctd_ptr<nbl::video::IDeviceMemoryAllocation> src,
+        CUexternalMemory cuExtMem) : 
+        m_device(device),
+        m_src(src),
+        m_handle(cuExtMem) {}
+
+      ~CCUDAImportedMemory() override;
+
```
devshgraphicsprogramming: again public constructors for an object which should only be creatable via a factory

public destructor for IReferenceCounted

include/nbl/video/CCUDAImportedMemory.h
```diff
@@ -0,0 +1,42 @@
...
+      CUexternalMemory getInternalObject() const { return m_handle; }
+      CUresult getMappedBuffer(CUdeviceptr* mappedBuffer);
```
devshgraphicsprogramming: a little bit of docs, what does this method do ?

include/nbl/video/CCUDAExportableMemory.h
```diff
@@ -0,0 +1,65 @@
...
+            size_t granularSize;
+            CUdeviceptr ptr;
+            external_handle_t externalHandle;
+        };
+
+        CCUDAExportableMemory(core::smart_refctd_ptr<CCUDADevice> device, SCachedCreationParams&& params)
+            : m_device(std::move(device))
+            , m_params(std::move(params))
+        {}
+        ~CCUDAExportableMemory() override;
```
devshgraphicsprogramming: public constructors for an object which should only be creatable via a factory

public destructor for IReferenceCounted

include/nbl/video/CCUDAExportableMemory.h
```diff
@@ -0,0 +1,65 @@
...
+            uint32_t          alignment;
+            CUmemLocationType location;
+        };
+
+        struct SCachedCreationParams : SCreationParams
+        {
+            size_t granularSize;
+            CUdeviceptr ptr;
+            external_handle_t externalHandle;
+        };
```
devshgraphicsprogramming: I don't get your split between the two structs, CreationPArams usually inherits from cached, because the idea is that Creation Params contains extra data, and Cached is the subset you want to keep.

In your case everything is cached creation params, so the creation params should inherit and be empty or just be an alias

include/nbl/video/CCUDAExportableMemory.h
```diff
@@ -0,0 +1,65 @@
...
+            : m_device(std::move(device))
+            , m_params(std::move(params))
+        {}
+        ~CCUDAExportableMemory() override;
+
+        CUdeviceptr getDeviceptr() const { return m_params.ptr;  }
+
+        const SCreationParams& getCreationParams() const { return m_params; }
+
+        core::smart_refctd_ptr<IDeviceMemoryAllocation> exportAsMemory(ILogicalDevice* device, IDeviceMemoryBacked* dedication = nullptr) const;
```
devshgraphicsprogramming: documentation, I find this confusing, esp the dedication parameter

include/nbl/video/EApiType.h
```diff
@@ -13,6 +13,45 @@ enum E_API_TYPE : uint32_t
...
+
+    const HANDLE cur = GetCurrentProcess();
+    if (!DuplicateHandle(cur, handle, cur, &re, GENERIC_ALL, 0, DUPLICATE_SAME_ACCESS))
+        return ExternalHandleNull;
+
+    return re;
+#else
+    return dup(handle);
+#endif
+}
```
devshgraphicsprogramming: this almost feels like it should belong in `system` and not `video`

include/nbl/video/ISemaphore.h
```diff
@@ -15,6 +15,27 @@ namespace nbl::video
...
+        };
+
+        //!
+        struct SCachedCreationParams
+        {
+            // Handle Type for external resources
+            core::bitflag<E_EXTERNAL_HANDLE_TYPE> externalHandleTypes = EHT_NONE;
+        };
+
+        struct SCreationParams : SCachedCreationParams {};
```
devshgraphicsprogramming: erm you can actually create a semaphore with an an initial value, you can leave old creation function in IDevice but mark it `[[deprecated]]`

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -56,22 +56,63 @@ CVulkanLogicalDevice::CVulkanLogicalDevice(core::smart_refctd_ptr<const IAPIConn
 }
 
 
-core::smart_refctd_ptr<ISemaphore> CVulkanLogicalDevice::createSemaphore(const uint64_t initialValue)
```
devshgraphicsprogramming, line 59: `intialValue` should go into `creationPArams` and the old signature should be implemented in terms of new and marked deprecated

src/nbl/video/CVulkanMemoryAllocation.cpp
```diff
@@ -4,14 +4,19 @@
...
+    const external_handle_t externalHandle,
+    SCreationParams&& params
+) : IDeviceMemoryAllocation(dev,std::move(params)), m_vulkanDevice(dev), m_deviceMemoryHandle(deviceMemoryHandle), m_externalHandle(externalHandle) {}
 
 CVulkanMemoryAllocation::~CVulkanMemoryAllocation()
 {
+    if (m_externalHandle != ExternalHandleNull)
+    {
+        bool re = CloseExternalHandle(m_externalHandle);
+        assert(re);
```
devshgraphicsprogramming: can log an error using the `m_vulkanDevice`'s debug callback

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -372,12 +515,13 @@ core::smart_refctd_ptr<IGPUImage> CVulkanLogicalDevice::createImage_impl(IGPUIma
     vk_createInfo.sharingMode = params.isConcurrentSharing() ? VK_SHARING_MODE_CONCURRENT:VK_SHARING_MODE_EXCLUSIVE;
     vk_createInfo.queueFamilyIndexCount = params.queueFamilyIndexCount;
     vk_createInfo.pQueueFamilyIndices = params.queueFamilyIndices;
-    vk_createInfo.initialLayout = params.preinitialized ? VK_IMAGE_LAYOUT_PREINITIALIZED:VK_IMAGE_LAYOUT_UNDEFINED;
+    // The Vulkan spec states: If the pNext chain includes a VkExternalMemoryImageCreateInfo or VkExternalMemoryImageCreateInfoNV structure whose handleTypes member is not 0, initialLayout must be VK_IMAGE_LAYOUT_UNDEFINED
+    vk_createInfo.initialLayout = external ? VK_IMAGE_LAYOUT_UNDEFINED : (params.preinitialized ? VK_IMAGE_LAYOUT_PREINITIALIZED : VK_IMAGE_LAYOUT_UNDEFINED);
```
devshgraphicsprogramming: I would put a check in the `valid()` of the params that `preinitialized` is not true if `external` is true

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -56,22 +56,63 @@ CVulkanLogicalDevice::CVulkanLogicalDevice(core::smart_refctd_ptr<const IAPIConn
...
+            .semaphore = vkSemaphore,
+            .handleType = handleType,
+        };
+        if (VK_SUCCESS != m_devf.vk.vkGetSemaphoreFdKHR(m_vkdev, &props, &externalHandle))
+        {
+            m_devf.vk.vkDestroySemaphore(m_vkdev, semaphore, nullptr);
+            return nullptr;
+        }
+#endif
+    }
```
devshgraphicsprogramming: this should be handled through a switch with a default case so we don't die when more handle types are added

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -136,26 +177,72 @@ core::smart_refctd_ptr<IDeferredOperation> CVulkanLogicalDevice::createDeferredO
...
+        if (info.allocateFlags.hasFlags(IDeviceMemoryAllocation::EMAF_DEVICE_ADDRESS_BIT))
             vk_allocateFlagsInfo.flags |= VK_MEMORY_ALLOCATE_DEVICE_ADDRESS_BIT;
         vk_allocateFlagsInfo.deviceMask = 0u; // unused: for now
     }
     VkMemoryDedicatedAllocateInfo vk_dedicatedInfo = {VK_STRUCTURE_TYPE_MEMORY_DEDICATED_ALLOCATE_INFO, nullptr};
+
+#ifdef _WIN32
+    VkImportMemoryWin32HandleInfoKHR importInfo = { 
+        .sType = VK_STRUCTURE_TYPE_IMPORT_MEMORY_WIN32_HANDLE_INFO_KHR,
+        .handleType = static_cast<VkExternalMemoryHandleTypeFlagBits>(info.externalHandleType),
```
devshgraphicsprogramming: you need to check the handle types at the start IMHO before just doing through with win32 or POSIX codepaths

include/nbl/video/ILogicalDevice.h
devshgraphicsprogramming, line 1109: why are we passing `dedicatedOnly` around ?

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -24,6 +24,8 @@ We only support persistently mapped buffers with ARB_buffer_storage.
 Please don't ask us to support Buffer Orphaning. */
 class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
 {
+    friend class IDeviceMemoryAllocator;
+    friend class ILogicalDevice;
```
devshgraphicsprogramming: write docs about why you need the friendship

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -153,29 +169,50 @@ class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
...
+            IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE externalHandleType = IDeviceMemoryAllocation::EHT_NONE;
+            //! Imports the given handle  if externalHandle != nullptr && externalHandleType != EHT_NONE
+            //! Creates exportable memory if externalHandle == nullptr && externalHandleType != EHT_NONE
+            external_handle_t externalHandle = 0;
+        };
+
+        struct SCreationParams: SInfo
+        {
+            core::bitflag<E_MEMORY_PROPERTY_FLAGS> memoryPropertyFlags = E_MEMORY_PROPERTY_FLAGS::EMPF_NONE;
+            const bool dedicated = false;
```
devshgraphicsprogramming: you dont need to make the struct member const, its better if you make the `m_params` const!

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -153,29 +169,50 @@ class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
...
+        };
+        
+        inline const SCreationParams& getCreationParams() const { return m_params; }
+
+        virtual external_handle_t getExternalHandle() const = 0;
```
devshgraphicsprogramming: I don't get the point of having this, and a `getCreationParams()` when I can go `getCreationParams().externalHandle`

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -153,29 +169,50 @@ class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
...
+            uint64_t allocationSize = 0;
+            core::bitflag<IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS> allocateFlags = IDeviceMemoryAllocation::EMAF_NONE;
+            // Handle Type for external resources
+            IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE externalHandleType = IDeviceMemoryAllocation::EHT_NONE;
+            //! Imports the given handle  if externalHandle != nullptr && externalHandleType != EHT_NONE
+            //! Creates exportable memory if externalHandle == nullptr && externalHandleType != EHT_NONE
+            external_handle_t externalHandle = 0;
+        };
+
+        struct SCreationParams: SInfo
```
devshgraphicsprogramming, line 205: any reason for the split between `SInfo` and `ScreationParams` we could shove everything into 24 bytes, by packing the allocateflags, handle types, memoryPropertyFlags and dedicated into 8 bytes

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -68,33 +70,47 @@ class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
...
-        inline size_t getAllocationSize() const {return m_allocationSize;}
+        inline size_t getAllocationSize() const {return m_params.allocationSize;}
 
         //!
-        inline core::bitflag<E_MEMORY_ALLOCATE_FLAGS> getAllocateFlags() const { return m_allocateFlags; }
+        inline core::bitflag<E_MEMORY_ALLOCATE_FLAGS> getAllocateFlags() const { return m_params.allocateFlags; }
 
         //!
-        inline core::bitflag<E_MEMORY_PROPERTY_FLAGS> getMemoryPropertyFlags() const { return m_memoryPropertyFlags; }
+        inline core::bitflag<E_MEMORY_PROPERTY_FLAGS> getMemoryPropertyFlags() const { return m_params.memoryPropertyFlags; }
```
devshgraphicsprogramming, line 124: whatever I can get from `getCreationParams()` mark with `[[deprecated]]`

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -68,33 +70,47 @@ class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
             EMHF_MULTI_INSTANCE_BIT = 0x00000002,
         };
```
devshgraphicsprogramming: you can take this enum down to `uint8_t`

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -68,33 +70,47 @@ class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
...
 
+        //! Flags for imported/exported allocation
+        enum E_EXTERNAL_HANDLE_TYPE : uint32_t
```
devshgraphicsprogramming, line 73: and this to `uint16_t`

include/nbl/video/IDeviceMemoryAllocation.h
devshgraphicsprogramming: ok I see its because you reuse the SInfo in the allocator, but stull it would help to pack the allocate flags and external handle type into 4 bytes in case the external handle is 4 bytes on linux

include/nbl/video/IDeviceMemoryAllocator.h
```diff
@@ -15,11 +15,9 @@ class NBL_API2 IDeviceMemoryAllocator
 		// right now we only support this interface handing out memory for one device or group
 		virtual ILogicalDevice* getDeviceForAllocations() const = 0;
 
-		struct SAllocateInfo
+		struct SAllocateInfo : IDeviceMemoryAllocation::SInfo
 		{
-			size_t size : 54 = 0ull;
-			size_t flags : 5 = 0u; // IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS
-			size_t memoryTypeIndex : 5 = 0u;
+			size_t memoryTypeIndex = 0u;
```
devshgraphicsprogramming: don't use `size_t for the index, use uint8_T at most

include/nbl/video/IDeviceMemoryAllocator.h
```diff
@@ -45,8 +43,15 @@ class NBL_API2 IDeviceMemoryAllocator
...
-					: m_allocateFlags(static_cast<uint32_t>(allocateFlags.value)), m_reqs(reqs) {}
+				IMemoryTypeIterator(const IDeviceMemoryBacked::SDeviceMemoryRequirements& reqs, 
+					core::bitflag<IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS> allocateFlags,
+					IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE handleType,
+					external_handle_t handle) : 
+					m_allocateFlags(static_cast<uint32_t>(allocateFlags.value)), 
+					m_reqs(reqs), 
+					m_handleType(handleType),
+					m_handle(handle)
+				{}
```
devshgraphicsprogramming, line 60: why does your memory type iterator care about the handle value ?  I can understand caring about the handle type and whether we're importing or exporting, but not the actual handle

include/nbl/video/IDeviceMemoryAllocator.h
```diff
@@ -106,10 +121,13 @@ class NBL_API2 IDeviceMemoryAllocator
 
 		template<class memory_type_iterator_t=DefaultMemoryTypeIterator>
 		inline SAllocation allocate(
-			const IDeviceMemoryBacked::SDeviceMemoryRequirements& reqs, IDeviceMemoryBacked* dedication=nullptr,
-			const core::bitflag<IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS> allocateFlags=IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS::EMAF_NONE)
+			const IDeviceMemoryBacked::SDeviceMemoryRequirements& reqs, 
+			IDeviceMemoryBacked* dedication = nullptr,
+			const core::bitflag<IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS> allocateFlags = IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS::EMAF_NONE,
+			IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE externalHandleType = IDeviceMemoryAllocation::EHT_NONE,
+			external_handle_t externalHandle = {})
```
devshgraphicsprogramming: last 4 arguments should be wrapped up in a structure, it kinda looks like half of your Sinfo

include/nbl/video/IPhysicalDevice.h
```diff
@@ -639,6 +639,58 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
...
+          EEMF_EXPORTABLE_BIT = 0x2,
+          EEMF_IMPORTABLE_BIT = 0x4,
+        };
+
+        struct SExternalMemoryProperties
+        {
+            IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE exportableTypes : 7;
+            IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE compatibleTypes : 7;
+            // TODO(kevin): This should actually be core::bitflag to be semantically correct. What should we do? Should we use bool for each flag instead of enum?
+            E_EXTERNAL_MEMORY_FEATURE_FLAGS features : 3;
```
devshgraphicsprogramming, line 697: `uint16_t` for the class

its fine as it is, can make a `core::bitflag<> get` method

include/nbl/video/IPhysicalDevice.h
```diff
@@ -639,6 +639,58 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
...
+            // {
+            //     std::shared_lock lock(m_externalBufferPropertiesMutex);
+            //     auto it = m_externalBufferProperties.find({ usage, handleType });
+            //     if (it != m_externalBufferProperties.end())
+            //         return it->second;
+            // }
+            //
+            // std::unique_lock lock(m_externalBufferPropertiesMutex);
+            // return m_externalBufferProperties[{ usage, handleType }] = getExternalBufferProperties_impl(usage, handleType);
+            return getExternalMemoryProperties_impl(usages, handleType);
```
devshgraphicsprogramming: we do it for image formats, I guess we should make it cached here too

include/nbl/video/IPhysicalDevice.h
```diff
@@ -639,6 +639,58 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
...
+        }
+
+        struct SImageFormatInfo
+        {
+            asset::E_FORMAT format;
+            IGPUImage::E_TYPE type;
+            IGPUImage::TILING tiling;
+            core::bitflag<IGPUImage::E_USAGE_FLAGS> usage;
+            core::bitflag<IGPUImage::E_CREATE_FLAGS> flags;
+        };
```
devshgraphicsprogramming, line 727: assert this is 5 bytes, if not find the padding

src/nbl/video/CVulkanLogicalDevice.cpp
devshgraphicsprogramming: ok I see you can support multiple handles at once so switch doesn't make sense but you should still gate/assert that the external handle type you're doing the code for is there (e.g. WIN32 or FD) and others are not

include/nbl/video/ILogicalDevice.h
devshgraphicsprogramming: ok I see, IMHO the solution is to have `IGPUBuffer` and `IGPUImage` SCONSTRUCTIONPArams and SCachedConstruction params (which can contain the creation and cached creation params) to pass things between the inside of the factory creation function and the CVulkan constructor

include/nbl/video/CCUDADevice.h
```diff
@@ -63,132 +74,44 @@ class CCUDADevice : public core::IReferenceCounted
...
-				iit->acquired = false;
-			return retval;
-		}
+		bool isMatchingDevice(const IPhysicalDevice* device) { return device && !memcmp(device->getProperties().deviceUUID, m_physicalDevice->getProperties().deviceUUID, 16); }
 
-		static CUresult acquireAndGetPointers(GraphicsAPIObjLink<video::IGPUBuffer>* linksBegin, GraphicsAPIObjLink<video::IGPUBuffer>* linksEnd, CUstream stream, size_t* outbufferSizes = nullptr);
-		static CUresult acquireAndGetMipmappedArray(GraphicsAPIObjLink<video::IGPUImage>* linksBegin, GraphicsAPIObjLink<video::IGPUImage>* linksEnd, CUstream stream);
-		static CUresult acquireAndGetArray(GraphicsAPIObjLink<video::IGPUImage>* linksBegin, GraphicsAPIObjLink<video::IGPUImage>* linksEnd, uint32_t* arrayIndices, uint32_t* mipLevels, CUstream stream);
-#endif
+		size_t roundToGranularity(CUmemLocationType location, size_t size) const;
```
devshgraphicsprogramming: isn't this just `core::roundUp` or just `align` ?

include/nbl/video/CCUDAImportedSemaphore.h
```diff
@@ -0,0 +1,46 @@
...
+      CUexternalSemaphore getInternalObject() const { return m_handle; }
+      CCUDAImportedSemaphore(core::smart_refctd_ptr<CCUDADevice> device, 
+        core::smart_refctd_ptr<ISemaphore> src, 
+        CUexternalSemaphore semaphore)
+          : m_device(std::move(device))
+          , m_src(std::move(src))
+          , m_handle(semaphore)
+      {}
+      ~CCUDAImportedSemaphore() override;
+
```
devshgraphicsprogramming: again these are public and should be private

include/nbl/video/CCUDAImportedSemaphore.h
```diff
@@ -0,0 +1,46 @@
...
+#endif
+
+// useful includes in the future
+//#include "cudaEGL.h"
+//#include "cudaVDPAU.h"
+
+namespace nbl::video
+{
+
+class NBL_API2 CCUDAImportedSemaphore : public core::IReferenceCounted
```
devshgraphicsprogramming: use `final` whenever possible

src/nbl/video/CCUDAHandler.cpp
```diff
@@ -468,15 +518,15 @@ core::smart_refctd_ptr<CCUDAHandler> CCUDAHandler::create(system::ISystem* syste
 	{
 		const void* contents = it.second.data();
 		headers.push_back(core::make_smart_refctd_ptr<system::CFileView<system::CNullAllocator>>(
-			core::smart_refctd_ptr<system::ISystem>(system),it.first.c_str(),
+			it.first.c_str(),
 			core::bitflag(system::IFile::ECF_READ)|system::IFile::ECF_MAPPABLE,
+			// ASK(kevin): What initial_modified_time should I use? Is this how this parameter is used?
+			std::chrono::clock_cast<system::IFile::time_point_t::clock>(std::chrono::system_clock::now()),
```
devshgraphicsprogramming: this is fine, we don't have a file watcher system yet to reload the file contents if they change

src/nbl/video/CCUDAImportedMemory.cpp
```diff
@@ -0,0 +1,32 @@
...
+CUresult CCUDAImportedMemory::getMappedBuffer(CUdeviceptr* mappedBuffer)
+{
+  CUDA_EXTERNAL_MEMORY_BUFFER_DESC bufferDesc = {};
+  bufferDesc.offset = 0;
+  bufferDesc.size = m_src->getAllocationSize();
+
+  auto& cu = m_device->getHandler()->getCUDAFunctionTable();
+  return cu.pcuExternalMemoryGetMappedBuffer(mappedBuffer, m_handle, &bufferDesc);
+  
+}
```
devshgraphicsprogramming, line 52: oh, the syntax for this hurts me, can we turn it into a method that returns `CUdeviceptr` (and null on failure) ?

If I want CUDA like API usage then I guess I can call `getInternalObject` and go to town

src/nbl/video/CCUDAExportableMemory.cpp
```diff
@@ -0,0 +1,54 @@
...
+    case CU_MEM_LOCATION_TYPE_HOST_NUMA_CURRENT:
+    case CU_MEM_LOCATION_TYPE_HOST:   memoryTypeBits &= ~vram; break;
+    default: break;
+	}
+
+	IDeviceMemoryBacked::SDeviceMemoryRequirements req = {};
+	req.size = m_params.granularSize;
+	req.memoryTypeBits = memoryTypeBits;
+	req.prefersDedicatedAllocation  = nullptr != dedication;
+	req.requiresDedicatedAllocation = nullptr != dedication;
```
devshgraphicsprogramming, line 44: I don't get this, if you pass a `dedication` it means you've already allocated the memory, so you may have allocated it on the wrong heap which is not compatible.

IMHO it wholesale doesn't make sense, because if you have a dedicate allocation made before you must have known the external handle to even make it, so its a chicken an egg problem.

HOWEVER you should have a method that returns the Device Memory Requirements (same as IGPUBuffer or IGPUImage) for the CCUDAExportableMemory so that these are known before we attempt an export

src/nbl/video/CCUDAExportableMemory.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+	return device->allocate(req, 
+		dedication, 
+		IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS::EMAF_NONE,
```
devshgraphicsprogramming: you've robbed me of specifying `EMAF_DEVICE_ADDRESS_BIT` and getting BDA for a CUDA buffer

src/nbl/video/CCUDAExportableMemory.cpp
```diff
@@ -0,0 +1,54 @@
...
+	req.memoryTypeBits = memoryTypeBits;
+	req.prefersDedicatedAllocation  = nullptr != dedication;
+	req.requiresDedicatedAllocation = nullptr != dedication;
+
+	return device->allocate(req, 
+		dedication, 
+		IDeviceMemoryAllocation::E_MEMORY_ALLOCATE_FLAGS::EMAF_NONE, 
+		CCUDADevice::EXTERNAL_MEMORY_HANDLE_TYPE, 
+		m_params.externalHandle).memory;
+}
```
devshgraphicsprogramming, line 53: you need the equivalent of `ICleanup` from `IGPUBuffer` and `IGPUImage` but for the `IDeviceMemoryAllocation` so you can keep a smart pointer back to the Imported thing which provides the external Handle

Right now you're not ensuring lifetimes when CUDA exports and Vulkan imports

src/nbl/video/CCUDADevice.cpp
```diff
@@ -3,128 +3,219 @@
...
+	m_logger(vulkanDevice->getDebugCallback()->getLogger()),
+  m_defaultCompileOptions(), 
+  m_vulkanConnection(std::move(vulkanConnection)), 
+  m_physicalDevice(vulkanDevice), 
+  m_virtualArchitecture(virtualArchitecture),
+	m_handle(device),
+	m_handler(std::move(handler)),
+	m_allocationGranularity{}
 {
 	m_defaultCompileOptions.push_back("--std=c++14");
```
devshgraphicsprogramming: can you do C++20 or will NVCC die?

src/nbl/video/CCUDADevice.cpp
```diff
@@ -3,128 +3,219 @@
...
+
+size_t CCUDADevice::roundToGranularity(CUmemLocationType location, size_t size) const
 {
-	assert(link->obj);
-	auto glbuf = static_cast<video::COpenGLBuffer*>(link->obj.get());
-	auto retval = cuda.pcuGraphicsGLRegisterBuffer(&link->cudaHandle,glbuf->getOpenGLName(),flags);
-	if (retval!=CUDA_SUCCESS)
-		link->obj = nullptr;
-	return retval;
+	return ((size - 1) / m_allocationGranularity[location] + 1) * m_allocationGranularity[location];
```
devshgraphicsprogramming: ah okay you do it based on location, still please use the hlsl or core utitliities, and inline the function, no need to be going across DLL boundaries for stuff like that

src/nbl/video/CCUDADevice.cpp
```diff
@@ -3,128 +3,219 @@
...
 }
-CUresult CCUDAHandler::registerImage(GraphicsAPIObjLink<video::IGPUImage>* link, uint32_t flags)
+
+CUresult CCUDADevice::reserveAddressAndMapMemory(CUdeviceptr* outPtr, size_t size, size_t alignment, CUmemLocationType location, CUmemGenericAllocationHandle memory) const
```
devshgraphicsprogramming: I find `STATUS func(RETURN*` really awkward, maybe make a `template<T> struct `nbl::video::SCUDAResult {T result; CUresult status;};` with a `void` specialization thats just the status

include/nbl/video/CCUDADevice.h
```diff
@@ -63,132 +74,44 @@ class CCUDADevice : public core::IReferenceCounted
...
-						CUdeviceptr pointer;
-					} asBuffer;
-					struct
-					{
-						CUmipmappedArray mipmappedArray;
-						CUarray array;
-					} asImage;
-				};
-		};
+		CUdevice getInternalObject() const { return m_handle; }
```
AnastaZIuk: This leaks CUDA Driver API into our public API surface. CUDA SDK becomes a transitive public compile-time dependency of `Nabla::Nabla`. That means consumers need CUDA headers just to parse Nabla headers. We will break our packages and downstream consumers, including smoke, Erfan and n4ce team aand CI already fails so ye.

ok imo this API should not be designed this way. Even if we already had similar leaks before, that does not justify sticking to them. CUDA can be used privately while building Nabla or exposed through a separate explicit CUDA-native API or target. It should not leak through the default public Nabla headers.

NVIDIA EULA is clear on things. CUDA distribution is limited to selected components, not the SDK as a stand-alone product:
- Distribution section: "The portions of the SDK that are distributable under the Agreement are listed in Attachment A."
  https://docs.nvidia.com/cuda/eula/#distribution
- Attachment A: "The following CUDA Toolkit files may be distributed with applications developed by you..."
  https://docs.nvidia.com/cuda/eula/#attachment-a
- Limitations section: "you may not distribute or sublicense the SDK as a stand-alone product."
  https://docs.nvidia.com/cuda/eula/#limitations

Lemme show you how big boys like Blender do it as counterexample. Their CUDA device boundary header exposes only Blender/Cycles abstractions and has no `cuda.h`, no `nvrtc.h`, no `CUdevice` in the API
- https://github.com/blender/blender/blob/794c527e8595a9f448e0143a217d0ceb648c5e7e/intern/cycles/device/cuda/device.h#L7-L27

CUDA Driver API stuff is hidden in implementation headers/sources. `device.cpp` includes the CUDA implementation only behind `WITH_CUDA` and the actual `CUdevice`/`CUcontext` fields live in `device_impl.h` not in the outward boundary header
- https://github.com/blender/blender/blob/794c527e8595a9f448e0143a217d0ceb648c5e7e/intern/cycles/device/cuda/device.cpp#L10-L12
- https://github.com/blender/blender/blob/794c527e8595a9f448e0143a217d0ceb648c5e7e/intern/cycles/device/cuda/device_impl.h#L12-L17
- https://github.com/blender/blender/blob/794c527e8595a9f448e0143a217d0ceb648c5e7e/intern/cycles/device/cuda/device_impl.h#L23-L30

They also default to dynamic CUDA loading with CUEW and link that loader privately instead of forcing CUDA as a public API dependency
- https://github.com/blender/blender/blob/794c527e8595a9f448e0143a217d0ceb648c5e7e/CMakeLists.txt#L664-L667
- https://github.com/blender/blender/blob/794c527e8595a9f448e0143a217d0ceb648c5e7e/extern/cuew/src/cuew.c#L341-L373
- https://github.com/blender/blender/blob/794c527e8595a9f448e0143a217d0ceb648c5e7e/intern/cycles/device/CMakeLists.txt#L152-L160

devshgraphicsprogramming: @AnastaZIuk is correct to a degree, it would be good not to leak, but in order to do anything in the example which does interop we must "leak" because I don't have the stomach for wrapping all of CUDA like we do with Vulkan.

> Lemme show you how big boys like Blender do it as counterexample. Their CUDA device boundary header exposes only Blender/Cycles abstractions and has no cuda.h, no nvrtc.h, no CUdevice in the API

This is not what we intend to do though.

I'd probably move the CUDA interop to an extension (or any other separate TARGET, shared or static whichever way you want it) so that only the examples and programs using CUDA actually transitively depend on CUDA

AnastaZIuk (changes requested):
assigning myself to deal with build system part https://github.com/Devsh-Graphics-Programming/Nabla/pull/1061/changes#r3189831621

include/nbl/video/CCUDADevice.h
devshgraphicsprogramming: ah different alignment depending on the heap/location

include/nbl/video/CCUDAImportedSemaphore.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: I admit it need better name. There are two externalHandle. The first one is the externalHandle in creationParams that is used for vulkan to import externalMemory. The second one that is created by vulkan when importing the externalHandle or when vulkan want to export memory.

include/nbl/video/IDeviceMemoryAllocator.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocator.h
kevyuu: There is already SAllocateInfo though. If I have another thing like SAllocateParams, the api become weird. dedication is not part of SInfo. So I cannot make that new struct part of SInfo.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

include/nbl/video/EApiType.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: We don't need this friendship. Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

include/nbl/video/ISemaphore.h
kevyuu: Done.

src/nbl/video/CVulkanLogicalDevice.cpp
kevyuu: Done.

src/nbl/video/CVulkanMemoryAllocation.cpp
kevyuu: Done.

src/nbl/video/CCUDAExportableMemory.cpp
kevyuu: When Vulkan import from CUDA. I duplicate the osHandle. Since OsHandle hold on the resource. This should be fine. [discussion](https://discordapp.com/channels/427551838099996672/1349315179979997224/1497201367020863489)

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

include/nbl/video/IPhysicalDevice.h
kevyuu: Done.

include/nbl/video/CCUDADevice.h
```diff
@@ -56,66 +45,57 @@ class NBL_API2 CCUDADevice : public core::IReferenceCounted
 			EVA_80,
 			EVA_COUNT
```
devshgraphicsprogramming, line 71: more EVAs please, we're at 10 or something now, gotta have them all its like pokemon

include/nbl/video/CCUDADevice.h
```diff
@@ -56,66 +45,57 @@ class NBL_API2 CCUDADevice : public core::IReferenceCounted
...
+		static constexpr uint32_t AllocationGranularityLocationTypeCount = 5u;
+		struct SNativeState;
+		CCUDADevice(core::smart_refctd_ptr<CVulkanConnection>&& vulkanConnection, IPhysicalDevice* const vulkanDevice, const E_VIRTUAL_ARCHITECTURE virtualArchitecture, std::unique_ptr<SNativeState>&& nativeState, core::smart_refctd_ptr<CCUDAHandler>&& handler);
+		bool isValid() const;
 
-    const system::logger_opt_ptr m_logger;
+		const system::logger_opt_ptr m_logger;
 		std::vector<const char*> m_defaultCompileOptions;
 		core::smart_refctd_ptr<CVulkanConnection> m_vulkanConnection;
-		IPhysicalDevice* const m_physicalDevice;
```
devshgraphicsprogramming: why are we not caching the info about the Vulkan Physical Device this CUDA device aliases anymore?

include/nbl/video/CCUDAImportedMemory.h
```diff
@@ -1,42 +1,34 @@
...
-      ~CCUDAImportedMemory() override;
-
-      CUexternalMemory getInternalObject() const { return m_handle; }
-      CUresult getMappedBuffer(CUdeviceptr* mappedBuffer);
+class NBL_API2 CCUDAImportedMemory final : public core::IReferenceCounted
+{
+	public:
+		~CCUDAImportedMemory() override;
+		cuda_interop::SCUexternalMemory getInternalObject() const;
+		bool getMappedBuffer(cuda_interop::SOutput<cuda_interop::SCUdeviceptr> mappedBuffer) const;
```
devshgraphicsprogramming: whats going on here, whats `SOutput` ?

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -63,15 +62,15 @@ class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
...
-        enum E_MEMORY_HEAP_FLAGS : uint32_t
+        enum E_MEMORY_HEAP_FLAGS : uint8_t
         {
             EMHF_NONE               = 0,
             EMHF_DEVICE_LOCAL_BIT   = 0x00000001,
             EMHF_MULTI_INSTANCE_BIT = 0x00000002,
         };
 
         //! Flags for imported/exported allocation
-        enum E_EXTERNAL_HANDLE_TYPE : uint32_t
```
devshgraphicsprogramming: bad idea, they add two more flats and you're out of uint16

include/nbl/video/IDeviceMemoryAllocation.h
```diff
@@ -63,15 +62,15 @@ class NBL_API2 IDeviceMemoryAllocation : public virtual core::IReferenceCounted
             //EMPF_RDMA_CAPABLE_BIT_NV = 0x00000200,
         };
         //
-        enum E_MEMORY_HEAP_FLAGS : uint32_t
+        enum E_MEMORY_HEAP_FLAGS : uint8_t
         {
             EMHF_NONE               = 0,
             EMHF_DEVICE_LOCAL_BIT   = 0x00000001,
             EMHF_MULTI_INSTANCE_BIT = 0x00000002,
```
devshgraphicsprogramming, line 68: if you're gonna change the type, then don't use literals that suggest 32bit type (cut 000000 from the prefix)

include/nbl/video/IDeviceMemoryAllocation.h
devshgraphicsprogramming: is it possible to make this member const ?

include/nbl/video/ILogicalDevice.h
```diff
@@ -162,7 +162,9 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
         IQueue::RESULT waitIdle();
 
         //! Semaphore Stuff
-        virtual core::smart_refctd_ptr<ISemaphore> createSemaphore(const uint64_t initialValue, ISemaphore::SCreationParams&& creationParams = {}) = 0;
+        [[deprecated]]
+        virtual core::smart_refctd_ptr<ISemaphore> createSemaphore(const uint64_t initialValue) = 0;
```
devshgraphicsprogramming: implement the deprecated variant inline and without virutalism

include/nbl/video/IPhysicalDevice.h
```diff
@@ -649,13 +649,12 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
...
-            IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE exportableTypes : 7;
-            IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE compatibleTypes : 7;
-            // TODO(kevin): This should actually be core::bitflag to be semantically correct. What should we do? Should we use bool for each flag instead of enum?
+            // Need 15 bit to store all possible value of E_EXTERNAL_HANDLE_TYPE. So bitfield will not save any space.
+            IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE exportableTypes;
+            IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE compatibleTypes;
             E_EXTERNAL_MEMORY_FEATURE_FLAGS features : 3;
             bool operator == (SExternalMemoryProperties const& rhs) const = default;
         };
-        static_assert(sizeof(SExternalMemoryProperties) == sizeof(uint32_t));
```
devshgraphicsprogramming: still  need a size assert

include/nbl/video/IPhysicalDevice.h
devshgraphicsprogramming: you can actually use enums + bitfields
```
IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE exportableTypes : 14;
IDeviceMemoryAllocation::E_EXTERNAL_HANDLE_TYPE compatibleTypes : 14;
E_EXTERNAL_MEMORY_FEATURE_FLAGS features : 3;
```
if the underlying enum types are the same

devshgraphicsprogramming: nvm it needs to be 8 bytes cause EXTERNAL_HANDLE_TYPE is already using the 14th bit

include/nbl/video/IPhysicalDevice.h
devshgraphicsprogramming: make it uint64_t sized

src/nbl/video/CVulkanMemoryAllocation.cpp
```diff
@@ -6,16 +6,17 @@ namespace nbl::video
...
 CVulkanMemoryAllocation::~CVulkanMemoryAllocation()
 {
-    if (m_externalHandle != ExternalHandleNull)
+    if (m_externalHandle != system::ExternalHandleNull)
     {
-        bool re = CloseExternalHandle(m_externalHandle);
-        assert(re);
+        const auto success = system::CloseExternalHandle(m_externalHandle);
+        if (!success) m_vulkanDevice->getLogger()->log("Failed to close external handle for Vulkan memory allocation", system::ILogger::ELL_ERROR);
+        assert(success);
```
devshgraphicsprogramming: dont assert, logging is enough

include/nbl/video/CCUDAImportedMemory.h
devshgraphicsprogramming: I see markdown code got messed up here https://github.com/Devsh-Graphics-Programming/Nabla/pull/1061#discussion_r3184944023

what I mean is 
```cpp
template<typename T>
struct SCUDAResult
{
   inline operator bool() const {/*however you're checking now*/}

   T value;
   CUDAErrorCode status;
};
```

src/nbl/video/CVulkanSemaphore.cpp
```diff
@@ -12,7 +12,9 @@ CVulkanSemaphore::~CVulkanSemaphore()
   vk->vk.vkDestroySemaphore(vulkanDevice->getInternalObject(), m_semaphore, nullptr);
 	if (m_creationParams.externalHandleTypes != EHT_NONE)
 	{
-		CloseExternalHandle(m_externalHandle);
+		const auto success = system::CloseExternalHandle(m_externalHandle);
+		if (!success) vulkanDevice->getLogger()->log("Failed to close external handle for Vulkan semaphore", system::ILogger::ELL_ERROR);
+		assert(success);
```
devshgraphicsprogramming: do not assert

src/nbl/video/CCUDADevice.cpp
```diff
@@ -1,92 +1,159 @@
...
+	"-arch=compute_52",
+	"-arch=compute_53",
+	"-arch=compute_60",
+	"-arch=compute_61",
+	"-arch=compute_62",
+	"-arch=compute_70",
+	"-arch=compute_72",
+	"-arch=compute_75",
+	"-arch=compute_80"
+};
```
devshgraphicsprogramming, line 75: moar!

src/nbl/video/CCUDADevice.cpp
```diff
@@ -1,92 +1,159 @@
...
+static_assert(sizeof(VirtualArchCompileOption)/sizeof(*VirtualArchCompileOption)==CCUDADevice::EVA_COUNT);
+
+static CUmemAllocationHandleType getAllocationHandleType()
+{
+#ifdef _WIN32
+	return CU_MEM_HANDLE_TYPE_WIN32;
+#else
+	return CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR;
+#endif
+}
```
devshgraphicsprogramming: static inline ? or really just make it a constexpr

src/nbl/video/CCUDADevice.cpp
devshgraphicsprogramming: bump

include/nbl/video/CCUDADevice.h
devshgraphicsprogramming: the destructor is still public.....

src/nbl/video/CCUDAExportableMemory.cpp
devshgraphicsprogramming: the problem is still there...

include/nbl/video/IDeviceMemoryAllocation.h
devshgraphicsprogramming: my bad, needs reverting

include/nbl/video/CUDAInteropHandles.h
```diff
@@ -0,0 +1,146 @@
...
+	SOutput(Native* native) : ptr(reinterpret_cast<Opaque*>(native)) {}
+
+	Opaque& operator*() const { return *ptr; }
+	operator Opaque*() const { return ptr; }
+	explicit operator bool() const { return ptr!=nullptr; }
+
+	private:
+		Opaque* ptr;
+};
+
```
devshgraphicsprogramming, line 70: dont need this struct

include/nbl/video/CUDAInteropHandles.h
```diff
@@ -0,0 +1,146 @@
...
+/*
+	Owned opaque value used in public Nabla ABI. Native reference conversions become available only after the opt-in
+	header specializes SOpaqueCUDANativeType for the selected CUDA SDK.
+*/
+template<typename Derived, typename Storage>
```
devshgraphicsprogramming, line 75: why do Storage as a type and not just a Bytesize ?

devshgraphicsprogramming: to get size and alignment at the same time?

include/nbl/video/CCUDAImportedSemaphore.h
```diff
@@ -4,43 +4,34 @@
...
-          , m_handle(semaphore)
-      {}
-      ~CCUDAImportedSemaphore() override;
-
-    private:
-      core::smart_refctd_ptr<CCUDADevice> m_device;
-      core::smart_refctd_ptr<ISemaphore> m_src;
-      CUexternalSemaphore m_handle;
+	public:
+		~CCUDAImportedSemaphore() override;
```
devshgraphicsprogramming: still public desctructor

include/nbl/video/CCUDAImportedMemory.h
```diff
@@ -1,42 +1,34 @@
...
+class CCUDADevice;
 
-      ~CCUDAImportedMemory() override;
-
-      CUexternalMemory getInternalObject() const { return m_handle; }
-      CUresult getMappedBuffer(CUdeviceptr* mappedBuffer);
+class NBL_API2 CCUDAImportedMemory final : public core::IReferenceCounted
+{
+	public:
+		~CCUDAImportedMemory() override;
```
devshgraphicsprogramming: still public desctructor

src/nbl/video/CVulkanLogicalDevice.cpp
```diff
@@ -55,8 +55,27 @@ CVulkanLogicalDevice::CVulkanLogicalDevice(core::smart_refctd_ptr<const IAPIConn
...
-core::smart_refctd_ptr<ISemaphore> CVulkanLogicalDevice::createSemaphore(const uint64_t initialValue, ISemaphore::SCreationParams&& creationParams)
+    VkSemaphore semaphore;
+    if (!m_devf.vk.vkCreateSemaphore(m_vkdev, &createInfo, nullptr, &semaphore) == VK_SUCCESS)
+        return nullptr;
+
+    ISemaphore::SCreationParams creationParams;
+    creationParams.initialValue = initialValue;
+    return core::make_smart_refctd_ptr<CVulkanSemaphore>(core::smart_refctd_ptr<CVulkanLogicalDevice>(this), std::move(creationParams), semaphore, system::ExternalHandleNull);
+}
+
```
devshgraphicsprogramming: this is unclean duplicate code, it should be a 1-liner to the other function

src/nbl/video/CVulkanLogicalDevice.cpp
devshgraphicsprogramming: still a TODO

src/nbl/video/CCUDAImportedMemory.cpp
```diff
@@ -2,31 +2,82 @@
...
+
+// CUDA OFF stub keeps the clean public API linkable and reports feature absence with nullptr instead of unresolved symbols.
+struct CCUDAImportedMemory::SNativeState {};
+
+CCUDAImportedMemory::CCUDAImportedMemory(core::smart_refctd_ptr<CCUDADevice> device, core::smart_refctd_ptr<IDeviceMemoryAllocation> src, std::unique_ptr<SNativeState>&& nativeState)
+	: m_device(std::move(device))
+	, m_src(std::move(src))
+	, m_native(std::move(nativeState))
+{
+	assert(m_native);
```
devshgraphicsprogramming: can actually assert false in every function because the object is never meant to be made, ever

include/nbl/video/IPhysicalDevice.h
devshgraphicsprogramming: ping

src/nbl/video/CVulkanLogicalDevice.cpp
kevyuu: Done.

include/nbl/video/CCUDADevice.h
kevyuu: Done.

include/nbl/video/CCUDADevice.h
kevyuu: Yeah. Currently we don't really use the physicalDevice, we only use it to get the logger. But in the future, we might need it. Okay, I will revert this change.

devshgraphicsprogramming: I was mostly confused because the uuid is now hidden behind the Native state do it's harder to do all that device matching oneself

include/nbl/video/ILogicalDevice.h
kevyuu: Done.

src/nbl/video/CCUDADevice.cpp
kevyuu: Done.

include/nbl/video/IPhysicalDevice.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

include/nbl/video/IDeviceMemoryAllocation.h
kevyuu: Done.

src/nbl/video/CVulkanMemoryAllocation.cpp
kevyuu: Done

src/nbl/video/CVulkanSemaphore.cpp
kevyuu: Done

src/nbl/video/CCUDADevice.cpp
kevyuu: Done

include/nbl/video/CCUDAImportedSemaphore.h
kevyuu: Done

include/nbl/video/CCUDAImportedMemory.h
kevyuu: Done

include/nbl/video/IPhysicalDevice.h
kevyuu: Done

src/nbl/video/CCUDADevice.cpp
kevyuu: Done

src/nbl/video/CCUDADevice.cpp
kevyuu: Done.

src/nbl/video/CVulkanLogicalDevice.cpp
kevyuu: Done.

src/nbl/video/CVulkanLogicalDevice.cpp
kevyuu: Done

src/nbl/video/CVulkanLogicalDevice.cpp
kevyuu: Done

include/nbl/video/CCUDAImportedMemory.h
kevyuu: Done

include/nbl/video/CUDAInteropHandles.h
kevyuu: Yes.

src/nbl/video/CCUDAExportableMemory.cpp
kevyuu: To create a buffer, we only need the external handle type not the handle itself and we don't need to allocate the backing memory. The allocation in vulkan is where we need the external handle. So this method will allocate virtual address in vulkan which point to the same physical address in Cuda. Passing dedication, will make the vk allocation dedicated and bind the resource to it.

src/nbl/video/CCUDAImportedMemory.cpp
kevyuu: Done

src/nbl/video/CCUDAExportableMemory.cpp
kevyuu: Done.

include/nbl/video/CCUDAHandler.h
kevyuu: Done

include/nbl/video/CCUDADevice.h
kevyuu: Done

include/nbl/video/CCUDADevice.h
kevyuu: Done.

PR #1063, opened 2026-04-28
Fix nullptr deref when providing no optimizer to the compiler

Fletterio:
Pushing the return until the end would sometimes result in `*otherPassesIt` dereferencing a nullptr. Changed code so that it returns before that happens.


AnastaZIuk:
oh nice catch, looking at blame its with us for 2 years now, merging!

PR #1067, opened 2026-05-07
Move CUDA interop behind native opt-in

AnastaZIuk:
Moves CUDA interop behind SDK-free Nabla headers with explicit `Nabla::ext::CUDAInterop` native opt-in. Keeps raw CUDA/NVRTC access available for consumers that ask for native opt-in while avoiding default public SDK requirements.


include/nbl/ext/CUDAInterop/CUDAInteropNative.h
```diff
@@ -0,0 +1,214 @@
...
+	public:
+		static CUexternalMemory getInternalObject(const CCUDAImportedMemory& memory);
+		static CUresult getMappedBuffer(const CCUDAImportedMemory& memory, CUdeviceptr* mappedBuffer);
+};
+
+class NBL_API2 CCUDAImportedSemaphoreAccessor
+{
+	public:
+		static CUexternalSemaphore getInternalObject(const CCUDAImportedSemaphore& semaphore);
+};
```
devshgraphicsprogramming: accessors make no sense just move all the `nbl/video/CCUDA*.h` to the extension

include/nbl/video/CCUDAHandler.h
```diff
@@ -169,142 +59,36 @@ class NBL_API2 CCUDAHandler : public core::IReferenceCounted
...
 };
 
-#define ASSERT_CUDA_SUCCESS(expr, handler) \
-    do { \
-        const auto cudaResult = (expr); \
-        if (!((handler)->defaultHandleResult(cudaResult))) { \
-            assert(false); \
-        } \
-    } while(0)
-
```
devshgraphicsprogramming, line 305: that macro was useful just needs a rename

include/nbl/video/EApiType.h
```diff
@@ -28,29 +28,8 @@ constexpr external_handle_t ExternalHandleNull = nullptr;
...
-    HANDLE re = ExternalHandleNull;
-
-    const HANDLE cur = GetCurrentProcess();
-    if (!DuplicateHandle(cur, handle, cur, &re, GENERIC_ALL, 0, DUPLICATE_SAME_ACCESS))
-        return ExternalHandleNull;
-
-    return re;
-#else
-    return dup(handle);
-#endif
```
devshgraphicsprogramming, line 52: you may want to keep that inline, these are OS calls, and when they're inline they'll work BEFORE Nabla.dll is delay loaded, which is useful

src/nbl/ext/CUDAInterop/smoke/clean_opt_in.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+public:
+	using base_t::base_t;
+
+	bool onAppInitialized(nbl::core::smart_refctd_ptr<nbl::system::ISystem>&&) override
+	{
+		static_assert(std::is_class_v<nbl::video::CCUDADevice>);
+		static_assert(std::is_class_v<nbl::video::CCUDAExportableMemory>);
+		static_assert(std::is_class_v<nbl::video::CCUDAImportedMemory>);
+		static_assert(std::is_class_v<nbl::video::CCUDAImportedSemaphore>);
```
devshgraphicsprogramming: it would make more sense to not have anything `CUDA` related in Nabla itself

src/nbl/video/CCUDADevice.cpp
```diff
@@ -49,44 +52,71 @@ CCUDADevice::CCUDADevice(
...
+	return SAccess::native(device).context;
+}
+
+size_t CCUDADeviceAccessor::roundToGranularity(const CCUDADevice& device, CUmemLocationType location, size_t size)
 {
-	return ((size - 1) / m_allocationGranularity[location] + 1) * m_allocationGranularity[location];
+	const auto& granularity = SAccess::native(device).allocationGranularity[location];
+	return ((size - 1) / granularity + 1) * granularity;
+}
+
```
devshgraphicsprogramming: I mentioned in the original PR, this should be inline

src/nbl/video/CCUDADevice.cpp
```diff
@@ -49,44 +52,71 @@ CCUDADevice::CCUDADevice(
...
 	};
 
 	if (auto err = cu.pcuMemSetAccess(ptr, size, &accessDesc, 1); CUDA_SUCCESS != err)
 	{
-		ASSERT_CUDA_SUCCESS(cu.pcuMemUnmap(ptr, size), m_handler);
-		ASSERT_CUDA_SUCCESS(cu.pcuMemAddressFree(ptr, size), m_handler);
+		if (!cuda_native::CCUDAHandlerAccessor::defaultHandleResult(*handler, cu.pcuMemUnmap(ptr, size)))
+			assert(false);
+		if (!cuda_native::CCUDAHandlerAccessor::defaultHandleResult(*handler, cu.pcuMemAddressFree(ptr, size)))
+			assert(false);
```
devshgraphicsprogramming: @kevyuu thinking of it we shouldn't crash an entire program vecause of failure here :s

AnastaZIuk:
# CUDA interop extension ABI

As discussed on Discord, the proposal is to move all CUDA interop code into a plugin as Nabla extension and expose CUDA directly from that target's public headers.

We can do that. It protects `Nabla::Nabla` from the CUDA SDK leak. It does not remove the boundary problem though. It moves that problem to `Nabla::ext::CUDAInterop`, and then the extension target owns the ABI and versioning consequences.

The goal is not to wrap CUDA. The goal is to keep CUDA SDK-defined layout out of the public class layout.

## Proposed Plugin Shape

Original PR: https://github.com/Devsh-Graphics-Programming/Nabla/pull/1061

Let's say we move the original PR shape into the extension target. In other words, the CUDA interop code no longer lives in `Nabla::Nabla`, but the public CUDA-native headers stay effectively the same and move from `include/nbl/video/CCUDA*.h` to `include/nbl/ext/CUDAInterop/CCUDA*.h`.

The exact code shape from the last commit currently on `origin/vk_cuda_interop` is here:

- `CCUDAHandler.h` exposes the `CUDA` and `NVRTC` loader table types plus inline getters: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/include/nbl/video/CCUDAHandler.h#L43-L159
- `CCUDAHandler.h` exposes `SCUDADeviceInfo` with `CUdevice`, `CUuuid`, and `int attributes[CU_DEVICE_ATTRIBUTE_MAX]`: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/include/nbl/video/CCUDAHandler.h#L190-L199
- `CCUDAHandler.h` stores `CUDA m_cuda`, `NVRTC m_nvrtc`, and `core::vector<SCUDADeviceInfo>` in the public class layout: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/include/nbl/video/CCUDAHandler.h#L284-L295
- `CCUDADevice.h` includes CUDA headers and exposes CUDA-native declarations and members such as `CUmemAllocationHandleType`, `CUdevice`, `CUmemLocationType`, `CUdeviceptr`, and `CUcontext`: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/include/nbl/video/CCUDADevice.h#L14-L114
- `CCUDAExportableMemory.h` exposes `CUmemLocationType`, `CUdeviceptr`, and cached creation params in the class layout: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/include/nbl/video/CCUDAExportableMemory.h#L8-L58
- `CCUDAImportedMemory.h` exposes `CUexternalMemory`, `CUdeviceptr`, and inline native handle access: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/include/nbl/video/CCUDAImportedMemory.h#L4-L36
- `CCUDAImportedSemaphore.h` exposes `CUexternalSemaphore` and inline native handle access: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/include/nbl/video/CCUDAImportedSemaphore.h#L7-L39
- `CCUDAHandler.cpp` fills `m_availableDevices` and writes `attributes` using `CU_DEVICE_ATTRIBUTE_MAX` from the SDK that built the binary: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/src/nbl/video/CCUDAHandler.cpp#L15-L55
- `CCUDAHandler.cpp` later reads those cached attributes through the same `SCUDADeviceInfo` layout: https://github.com/Devsh-Graphics-Programming/Nabla/blob/0d237c08b5e183074d6307e57e21f8e86546fcdd/src/nbl/video/CCUDAHandler.cpp#L573-L587

If we move this shape into `Nabla::ext::CUDAInterop`, the code no longer leaks through `Nabla::Nabla`, but these CUDA-native headers become the public ABI and compile-time dependency of the extension target.

A consumer then compiles those public extension headers again with its own CUDA SDK.

Inside our current Nabla examples build this may look fine. The examples are configured in the same build interface as Nabla and the CUDA interop target, so they naturally see the same `CUDA::toolkit` that built the extension.

But that is one project setup, not a general CMake guarantee. Another project can consume Nabla through `add_subdirectory`, a nested build, an external build, custom toolchain files, cache overrides, or manually provided CUDA targets and make the consumer side see a different SDK.

With an installed package this becomes normal downstream usage. Our package config does the same kind of CUDA discovery on the consumer side. It accepts `Nabla_CUDA_TOOLKIT_ROOT`, calls `find_dependency(CUDAToolkit 13.0 REQUIRED)`, and links `Nabla::ext::CUDAInterop` to the consumer-resolved `CUDA::toolkit`: https://github.com/Devsh-Graphics-Programming/Nabla/blob/ffba3d48d4ac5fd7f26ed324c310f338328572af/cmake/NablaConfig.cmake.in#L88-L100

So the extension binary may have been built with CUDA SDK A, while the downstream project resolves CUDA SDK B through its own package configure. We should not force that downstream SDK to be byte-for-byte the same SDK installed on the host that built the package. If the public extension headers contain SDK-defined layout, the downstream translation units compile that layout from SDK B while the already-built extension binary was compiled with SDK A.

## Actual Risk

Raw CUDA handles like `CUdevice`, `CUcontext`, `CUdeviceptr`, `CUexternalMemory`, and `CUexternalSemaphore` are not the main ABI problem by themselves. They are handle-like values.

The risky part is exposing SDK-defined layout and inline access through the public extension headers.

- `int attributes[CU_DEVICE_ATTRIBUTE_MAX]` makes `SCUDADeviceInfo` size depend on the CUDA SDK version.
- `CUDA m_cuda` and `NVRTC m_nvrtc` make the class layout depend on generated function table members.
- inline getters return references into those public-layout objects, so the consumer compiles member offsets and vector element stride from its own CUDA headers.

## Consequences

1. The extension ABI depends on the CUDA SDK header layout.

I decided to check this directly and installed multiple CUDA SDK versions locally. `CUuuid` is stable in the checked SDKs.

```text
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda.h:286
    typedef struct CUuuid_st {
    char bytes[16]

C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\include\cuda.h:303
    typedef struct CUuuid_st {
    char bytes[16]

C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.2\include\cuda.h:324
    typedef struct CUuuid_st {
    char bytes[16]
```

`CUdevice` is also a handle-like `int` in the checked SDKs.

```text
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda.h:258
    typedef int CUdevice_v1
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda.h:259
    typedef CUdevice_v1 CUdevice

C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\include\cuda.h:266
    typedef int CUdevice_v1
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\include\cuda.h:267
    typedef CUdevice_v1 CUdevice

C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.2\include\cuda.h:287
    typedef int CUdevice_v1
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.2\include\cuda.h:288
    typedef CUdevice_v1 CUdevice
```

But `CU_DEVICE_ATTRIBUTE_MAX` is not stable.

```text
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda.h:755
    CU_DEVICE_ATTRIBUTE_MULTICAST_SUPPORTED = 132
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda.h:756
    CU_DEVICE_ATTRIBUTE_MAX
    so CU_DEVICE_ATTRIBUTE_MAX == 133

C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\include\cuda.h:841
    CU_DEVICE_ATTRIBUTE_D3D12_CIG_SUPPORTED = 135
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\include\cuda.h:842
    CU_DEVICE_ATTRIBUTE_MAX
    so CU_DEVICE_ATTRIBUTE_MAX == 136

C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.2\include\cuda.h:965
    CU_DEVICE_ATTRIBUTE_ATOMIC_REDUCTION_SUPPORTED = 148
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.2\include\cuda.h:966
    CU_DEVICE_ATTRIBUTE_MAX
    so CU_DEVICE_ATTRIBUTE_MAX == 149
```

The point is not that we want to support CUDA 12.x for this feature. The point is that NVIDIA already changed this value across minor releases inside the same major line. CUDA 12.1 has 133 and CUDA 12.5 has 136. Requiring CUDA 13.x does not show that CUDA 13.0, 13.1, 13.2, or a later 13.x SDK will keep every public-layout macro and generated function table shape identical.

So the same public struct has different sizes.

```text
Built with CUDA 12.1

SCUDADeviceInfo
    CUdevice handle                      4 bytes
    CUuuid uuid                          16 bytes
    int attributes[133]                  532 bytes
Total                                    552 bytes with 4-byte alignment

Consumed with CUDA 13.2

SCUDADeviceInfo
    CUdevice handle                      4 bytes
    CUuuid uuid                          16 bytes
    int attributes[149]                  596 bytes
Total                                    616 bytes with 4-byte alignment
```

If the extension binary fills `core::vector<SCUDADeviceInfo>` using 552-byte elements and the consumer iterates it using 616-byte elements, the consumer walks the vector with the wrong stride. This is an ABI bug even when the extension is a DLL.

2. DLL does not remove the public-header ABI problem.

A DLL moves implementation into another binary, but the consumer still compiles inline methods, templates, member offsets, and `sizeof(SCUDADeviceInfo)` from the public header.

```text
const auto& devices = handler->getAvailableDevices()

consumer-side codegen uses sizeof(SCUDADeviceInfo) from the consumer SDK
for each device in devices
    useUuid(device.uuid)
```

The DLL cannot fix the fact that the consumer and the DLL disagree on the vector element type layout. A static library has the same issue unless it is rebuilt from source with the exact same CUDA SDK as the consumer.

3. Public loader tables are also ABI surface.

In the original header, `CUDA m_cuda` and `NVRTC m_nvrtc` were members of `CCUDAHandler`.

Nabla's loader macro expands every listed function into a typed member.

```text
include/nbl/system/DynamicLibraryFunctionPointer.h:88

#define NBL_SYSTEM_DECLARE_DYNLIB_FUNCPTR(FUNC_NAME) \
    nbl::system::DynamicLibraryFunctionPointer<decltype(FUNC_NAME),NBL_CORE_UNIQUE_STRING_LITERAL_TYPE(#FUNC_NAME)> p ## FUNC_NAME

include/nbl/system/DynamicFunctionCaller.h:69

NBL_FOREACH(NBL_SYSTEM_DECLARE_DYNLIB_FUNCPTR,__VA_ARGS__)
```

That means the function pointer type is taken directly from the CUDA/NVRTC header used by the current translation unit.

Local CUDA headers show version-sensitive API surface.

```text
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.1\include\cuda.h
    cuCtxCreate_v4 not present
    cuLogsRegisterCallback not present
    cuDeviceGetUuid_v2 present

C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.5\include\cuda.h
    cuCtxCreate_v4 present
    cuLogsRegisterCallback not present
    cuDeviceGetUuid_v2 present

C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.2\include\cuda.h
    #define cuCtxCreate cuCtxCreate_v4
    cuLogsRegisterCallback present at line 26475
    #define cuDeviceGetUuid cuDeviceGetUuid_v2
```

If the generated loader tables are public members of `CCUDAHandler`, then their member list and `decltype(...)` signatures become part of the public class layout of the extension binary.

4. Every `CCUDA*.h` user gets a compile-time SDK dependency.

```text
app.cpp
  -> nbl/ext/CUDAInterop/CCUDAHandler.h
    -> cuda.h
    -> nvrtc.h
    -> CUDA SDK headers
```

Updating the CUDA SDK path, version, or package can invalidate every translation unit that includes those headers. That is true even when the consumer only wants `CCUDAHandler`, `CCUDADevice`, or a helper inline method and does not directly call CUDA.

5. The compatibility statement becomes weaker.

With CUDA SDK layout in public extension classes, the true statement is this.

```text
The consumer can use a different CUDA SDK only if its headers remain ABI-compatible with the SDK used to build `Nabla::ext::CUDAInterop`.
```

It is not this.

```text
The consumer can freely use any newer CUDA SDK with an already-built extension binary.
```

Build-tree examples do not show the stronger statement. They show our current single-configure examples setup where Nabla, the extension, and the examples share the same CUDA SDK discovery. Other build-interface setups, `add_subdirectory` consumers, installed packages, and downstream overrides are the cases where the boundary matters.

## Accessor Model

The public Nabla CUDA object layout stays SDK-free.

```text
class CCUDAHandler
{
public:
    struct SCUDADeviceInfo
    {
        std::array<uint8_t,16> uuid = {}
    }

    const core::vector<SCUDADeviceInfo>& getAvailableDevices() const

private:
    struct SNativeState
    std::unique_ptr<SNativeState> m_native
    core::vector<SCUDADeviceInfo> m_availableDevices
}
```

CUDA-dependent layout stays private to the binary.

```text
struct CCUDAHandler::SNativeState
{
    CUDA cuda
    NVRTC nvrtc
    core::vector<cuda_native::SCUDADeviceInfo> availableDevices
}
```

Native CUDA access is still explicit opt-in.

```text
#include <nbl/ext/CUDAInterop/CUDAInteropNative.h>

const CUDA& cu = cuda_native::CCUDAHandlerAccessor::getCUDAFunctionTable(*handler)
const auto& nativeDevices = cuda_native::CCUDAHandlerAccessor::getAvailableDevices(*handler)
```

So the plugin as Nabla extension model is workable. If it exposes the original CUDA-heavy `CCUDA*.h` layout publicly, we choose raw CUDA ergonomics over ABI and rebuild isolation.

The accessor/native-state approach is not a CUDA wrapper. It is a small glue layer between Nabla objects and the CUDA world, only for the interop objects that cross that boundary. Raw CUDA is still available through the explicit opt-in header.

## TLDR

Yes, we can move all CUDA interop code into `Nabla::ext::CUDAInterop` and expose CUDA directly from the extension headers. That gives nicer raw CUDA ergonomics and avoids a few accessor calls.

Then we explicitly accept that `Nabla::ext::CUDAInterop` has CUDA SDK-dependent public ABI. To be safe, consumers should build against the same CUDA SDK that was used to build the extension binary, or rebuild the extension with their SDK. Otherwise layout mismatches like `CU_DEVICE_ATTRIBUTE_MAX` can produce real ABI bugs. Even if this is acceptable for this plugin, in my opinion it is limiting, and this is why I introduced the accessor/native-state model at the cost of slightly worse raw CUDA ergonomics.

This can pass cleanly in our current examples build because Nabla, the extension, and examples are configured together. That does not cover every build-interface setup, `add_subdirectory` consumer, or the package/install interface where downstream CMake may find or override a different CUDA SDK. Then the public header layout can disagree with the prebuilt extension binary.

The tradeoff is:

- plugin with public CUDA layout means fewer glue lines and nicer direct CUDA API, but SDK becomes a public compile-time dependency of the extension and ABI compatibility depends on CUDA header compatibility.
- accessor/native-state model means a few extra explicit calls, but stable SDK-free public Nabla layout, smaller rebuild surface, and consumer-side CUDA SDK version can move independently for native opt-in code.

We can move it into a plugin as Nabla extension if that is the preferred direction. My subjective take is that I do not like that model very much, but we can do it as long as we are explicit that this accepts the SDK-dependent public ABI and the potential mismatch, rebuild, and package issues described above.

devshgraphicsprogramming:
Excellent report, well presented.

I can see the problem now.

One thing to note that newer CUDA SDK versions are not allowed to change handle definitions like CUDeviceptr and external semaphore, otherwise app compiled against lower version SDK wouldn't run with a newer runtime.

> That means the function pointer type is taken directly from the CUDA/NVRTC header used by the current translation unit.

Same thing for function signatures like cuCtx, this is why they have v2 and even v4. They cannot change the signature because CUDA Driver (not runtime) API is dynamically linked and you must be able.to run with later driver update.

> Local CUDA headers show version-sensitive API surface.

What you've uncovered here is the reason why we require the SDK version to be >=13.0 because one of the function pointers in the table can only be loaded from that SDK.

Although I don't intend to provide a guarantee that every function pointer will be loaded and available, it's meant to be a quickly hacked together low effort Volk but for CUDA. Except with no global C functions or state.

The `CU_DEVICE_ATTRIBUTE_MAX` problem is interesting indeed, and highlights why your line of reasoning is correct.

I think there's a middle ground compromise we can come to

devshgraphicsprogramming:
Ok lets split into two targets with optional leaking headers

But instead of accessors lets abuse implicit conversions and constructors
https://godbolt.org/z/WsGqcc1f4

```cpp

struct alignas(8) OpaqueCUdeviceptr
{
    uint8_t value[8];
};

//===================================LEAK BOUNDARY========================================
// CUDA SDK DEP now
//#include CUDA
using CUdeviceptr = void*;

// trait
template<typename T>
struct opaque_cuda_type;

template<>
struct opaque_cuda_type<OpaqueCUdeviceptr>
{
    using type = CUdeviceptr;
};

template<typename Opaque>
struct Wrapped
{
    using cuda_t = typename opaque_cuda_type<Opaque>::type;
    static_assert(std::is_trivial_v<cuda_t>);
    static_assert(sizeof(Opaque)==sizeof(cuda_t));
    static_assert(alignof(Opaque)==alignof(cuda_t));

    // constucotrs
    Wrapped() = default;
    Wrapped(const Wrapped&) = default;
    Wrapped(const cuda_t& val) {operator=(val);}
    Wrapped(const Opaque& val) {operator=(val);}

    // assignment
    inline Wrapped& operator=(const Wrapped&) = default;
    inline Wrapped& operator=(const cuda_t& val) {value = val; return *this;}
    inline Wrapped& operator=(const Opaque& val) {operator Opaque&() = val; return *this;}

    // immplicit convert to original value reference
    inline operator cuda_t&() {return value;}
    inline operator const cuda_t&() const {return value;}

    // implicit conver to opaque handle reference
    inline operator Opaque&() {return reinterpret_cast<Opaque&>(value);}
    inline operator const Opaque&() const {return reinterpret_cast<const Opaque&>(value);}

    cuda_t value = {};
};

// Type your code here, or load an example.
CUdeviceptr pCudaCall(CUdeviceptr);

CUdeviceptr square(int num)
{
    OpaqueCUdeviceptr op = {};

    Wrapped<OpaqueCUdeviceptr> wp(op);
    wp = op;

    wp = pCudaCall(wp);
    op = wp;

    return wp;
}
```

PR #1068, opened 2026-05-11
True IR dot3 output

keptsecret:


include/nbl/asset/material_compiler3/CTrueIR.h
```diff
@@ -1057,6 +1077,19 @@ class CTrueIR : public CNodePool // TODO: turn into an asset!
...
 		};
 
+		inline core::string getNodeID(const typed_pointer_type<const INode> handle) const { return core::string("_") + std::to_string(handle.value); }
+		inline core::string getLabelledNodeID(const typed_pointer_type<const INode> handle) const
+		{
+			const INode* node = getObjectPool().deref(handle);
+			core::string retval = getNodeID(handle);
+			retval += " [label=\"";
+			retval += node->getTypeName();
+			// maybe also add hash?
```
devshgraphicsprogramming, line 1087: yes, hash printed in base64 would be very good on a newline

include/nbl/asset/material_compiler3/CTrueIR.h
```diff
@@ -969,12 +982,19 @@ class CTrueIR : public CNodePool // TODO: turn into an asset!
 				{
 					// should probably size it better, if I knew total node count allocated or live
 					visitedNodes.reserve(roots.size()<<4);
+					layerStack.reserve(roots.size());
+					for (const auto& m : roots)
+					{
+						layerStack.push_back(m.front.root);
+						layerStack.push_back(m.back.root);
```
devshgraphicsprogramming, line 989: my issue here is that I cant tell apart the front layers from back layers

src/nbl/asset/material_compiler3/CTrueIR.cpp
```diff
@@ -103,7 +103,156 @@ CTrueIR::SBasicNodes::SBasicNodes(CTrueIR* ir)
...
+							nodeStack.push_back(childHandle);
+							visitedNodes.insert(childHandle);
+						}
+						childIx++;
+					}
+					};
+				switch (node->getFinalType())
+				{
+				case INode::EFinalType::CFactorCombiner:
+				{
```
devshgraphicsprogramming, line 140: wouldn't a virtual function instead of a switch work better for this? You know PIMPL and stuff

PR #1074, opened 2026-05-18
Ir dot3 print v3

devshgraphicsprogramming:


include/nbl/asset/material_compiler3/CTrueIR.h
```diff
@@ -219,6 +219,8 @@ class CTrueIR : public CNodePool // TODO: turn into an asset!
 				};
 				virtual EFinalType getFinalType() const = 0;
 
+				virtual uint16_t getCapabilities() const { return 0; };
```
devshgraphicsprogramming, line 222: no return a `core::bitfield<Enum>` so this is all typesafe

PR #1079, opened 2026-05-26
copyMemoryToImage function added

CrabExtra:
- Add hostImageCopy as MOVE_TO_LIMIT in device_features.json.
- Query and enable VkPhysicalDeviceHostImageCopyFeaturesEXT through limits.hostImageCopy.
- Add EUF_HOST_TRANSFER_BIT and map it to VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT.
- Wire VK_FORMAT_FEATURE_2_HOST_IMAGE_TRANSFER_BIT_EXT into hostImageTransfer format usage.
- Add host image copy flags and SMemoryToImageCopy.
- Add ILogicalDevice::copyMemoryToImage validation and CVulkanLogicalDevice backend.
- Validate host-transfer image creation and EHICF_MEMCPY_BIT full-subresource rules.


include/nbl/video/ILogicalDevice.h
```diff
@@ -822,6 +823,8 @@ class NBL_API2 ILogicalDevice : public core::IReferenceCounted, public IDeviceMe
             return result!=DEFERRABLE_RESULT::SOME_ERROR;
         }
 
+        // https://docs.vulkan.org/refpages/latest/refpages/source/vkCopyMemoryToImage.html
+        bool copyMemoryToImage(IGPUImage* const dstImage, const IGPUImage::LAYOUT dstImageLayout, const core::bitflag<IGPUImage::E_HOST_IMAGE_COPY_FLAGS> flags, const std::span<const IGPUImage::SMemoryToImageCopy> regions);
```
devshgraphicsprogramming, line 806: please add the host-sde calls to perfrom a layout transition and query current layout

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -1332,7 +1339,9 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
         retval.storageImageLoadWithoutFormat = anyFlag(features, VK_FORMAT_FEATURE_2_STORAGE_READ_WITHOUT_FORMAT_BIT);
         retval.storageImageStoreWithoutFormat = anyFlag(features, VK_FORMAT_FEATURE_2_STORAGE_WRITE_WITHOUT_FORMAT_BIT);
         retval.depthCompareSampledImage = anyFlag(features, VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_DEPTH_COMPARISON_BIT);
-//        retval.hostImageTransfer = anyFlag(features, VK_FORMAT_FEATURE_2_HOST_IMAGE_TRANSFER_BIT);
+        //Not sure to remove this?
+        //        retval.hostImageTransfer = anyFlag(features, VK_FORMAT_FEATURE_2_HOST_IMAGE_TRANSFER_BIT);
+        retval.hostImageTransfer = anyFlag(features, VK_FORMAT_FEATURE_2_HOST_IMAGE_TRANSFER_BIT_EXT);
```
devshgraphicsprogramming: we need to report because not every format suports

src/nbl/video/utilities/CAssetConverter.cpp
```diff
@@ -262,6 +262,8 @@ bool CAssetConverter::patch_impl_t<ICPUImage>::valid(const ILogicalDevice* devic
 	const auto* physDev = device->getPhysicalDevice();
 	if (storageImageLoadWithoutFormat && !physDev->getLimits().shaderStorageImageReadWithoutFormat)
 		return false;
+	if (usageFlags.hasFlags(usage_flags_t::EUF_HOST_TRANSFER_BIT) && !physDev->getLimits().hostImageCopy)
+		return false;
```
devshgraphicsprogramming, line 268: it needs to be a patch-able, but leave all the Asset Converter changes to leverage host image copy to me

Erfan-Ahmadi: that means he should undo this change? or remove the flag upon this discovery?

devshgraphicsprogramming: dont do any more changes to CAssetConverter, I will accept this diff, and finish the work

src/nbl/video/device_capabilities/device_limits.json
```diff
@@ -1752,6 +1752,29 @@
...
+                    "name": "hostImageCopyDstLayouts",
+                    "value": 0
+                },
+                {
+                    "type": "bool",
+                    "name": "identicalMemoryTypeRequirements",
+                    "value": false
+                }
+            ]
+        },
```
Erfan-Ahmadi: Ok so this is wrong:
it's not a "limit" per se, or an extension/feature.
limits are either vulkan limits or vulkan features promoted to limits.
this fits neither. it's a physical device property.
Similar to `SMemoryProperties` you should have a `SHostImageCopyProperties` and function to query it, and member functions to fill.
it should be equivalent to https://docs.vulkan.org/refpages/latest/refpages/source/VkPhysicalDeviceHostImageCopyProperties.html

Look at how these work:
- `SMemoryProperties`
- `SFormatImageUsages`
- `SQueueFamilyProperties`
- or basically anything in `SInitData`

include/nbl/video/IPhysicalDevice.h
```diff
@@ -113,6 +113,9 @@ class NBL_API2 IPhysicalDevice : public core::Interface, public core::Unmovable
             char driverName[VK_MAX_DRIVER_NAME_SIZE];
             char driverInfo[VK_MAX_DRIVER_INFO_SIZE];
             APIVersion conformanceVersion;
+
+            /* VK_EXT_host_image_copy */
+            uint8_t optimalTilingLayoutUUID[VK_UUID_SIZE];
```
Erfan-Ahmadi: nope, should be in dedicate `SHostImageCopyProperties`

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -587,6 +590,30 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
+        {
+            properties.limits.identicalMemoryTypeRequirements = hostImageCopyProperties.identicalMemoryTypeRequirements;
+            memcpy(properties.optimalTilingLayoutUUID, hostImageCopyProperties.optimalTilingLayoutUUID, VK_UUID_SIZE);
+            core::vector<VkImageLayout> copySrcLayouts(hostImageCopyProperties.copySrcLayoutCount);
+            core::vector<VkImageLayout> copyDstLayouts(hostImageCopyProperties.copyDstLayoutCount);
+            VkPhysicalDeviceHostImageCopyPropertiesEXT layouts = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_HOST_IMAGE_COPY_PROPERTIES_EXT };
+            layouts.copySrcLayoutCount = hostImageCopyProperties.copySrcLayoutCount;
+            layouts.pCopySrcLayouts = copySrcLayouts.data();
+            layouts.copyDstLayoutCount = hostImageCopyProperties.copyDstLayoutCount;
+            layouts.pCopyDstLayouts = copyDstLayouts.data();
```
Erfan-Ahmadi: bug, 
vectors go out of scope and you're holding onto stale pointer. 
if you fix this [previous issue first](https://github.com/Devsh-Graphics-Programming/Nabla/pull/1079/changes#r3411386501) you'll realize why we used `smart_refctd_dynamic_array` for these types of things. it's basically a fixed sized dynamic allocation (no resize like vector + consistent pointers). it's also refcounted to ensure the memory doesn't get freed.

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -587,6 +590,30 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
             properties.limits.minAccelerationStructureScratchOffsetAlignment = accelerationStructureProperties.minAccelerationStructureScratchOffsetAlignment;
         }
 
+        if (isExtensionSupported(VK_EXT_HOST_IMAGE_COPY_EXTENSION_NAME))
```
Erfan-Ahmadi, line 594: this whole thing needs to fill initData's host image copy properties
(don't forget getters for this property)

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -630,6 +630,1084 @@ bool ILogicalDevice::nullifyDescriptors(const std::span<const IGPUDescriptorSet:
...
+    if (regions.empty())
+    {
+        NBL_LOG_ERROR("`regions` must not be empty");
+        return false;
+    }
+    if (regions.size() > 0xffff'ffffull)
+    {
+        NBL_LOG_ERROR("`regions.size()` exceeds `uint32_t` range");
+        return false;
+    }
```
Erfan-Ahmadi: ugh, I feel like this is the super protective stuff that AI generates, cool but we'd be having to do it everywhere if it was a real issue

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -630,6 +630,1084 @@ bool ILogicalDevice::nullifyDescriptors(const std::span<const IGPUDescriptorSet:
...
+    {
+        NBL_LOG_ERROR("`dstImageLayout` is not one of the device's supported host-copy destination layouts");
+        return false;
+    }
+    // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkCopyMemoryToImageInfo-regionCount-arraylength
+    if (regions.empty())
+    {
+        NBL_LOG_ERROR("`regions` must not be empty");
+        return false;
+    }
```
Erfan-Ahmadi: why return false?
user wants to copy 0 regions? done, you're good to go :D

Erfan-Ahmadi: move that to the first check

src/nbl/video/CVulkanPhysicalDevice.cpp
Erfan-Ahmadi: ok I mistunderstood, It's ok.
but try to use `smart_refctd_dynamic_array `

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -587,6 +590,30 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
+            vkGetPhysicalDeviceProperties2(vk_physicalDevice,&layouts2);
+            using layout_t = asset::IImage::LAYOUT;
+            for (uint32_t i=0u; i<=static_cast<uint32_t>(layout_t::SHARED_PRESENT); i++)
+            {
+                const VkImageLayout vkLayout = getVkImageLayoutFromImageLayout(static_cast<layout_t>(i));
+                for (const VkImageLayout l : copySrcLayouts)
+                    if (l==vkLayout) { properties.limits.hostImageCopySrcLayouts |= 1u<<i; break; }
+                for (const VkImageLayout l : copyDstLayouts)
+                    if (l==vkLayout) { properties.limits.hostImageCopyDstLayouts |= 1u<<i; break; }
+            }
```
Erfan-Ahmadi: ok I see what you're doing, instead of an array like vulkan's, you're creating a bitflag. that's actually a good idea for faster checks.

This double for loop seems weird af, you're iterating on all allowed src/dst layouts, per layout.

Why not do this instead:
loop over all copy[Src/Dst]Layouts, getImageLayout from Vk layout. check whether it's != UNDEFINED, and then put it in the correct place in the bitfield  `uint64_t` seems good enough for the bitfield, we won't be dealing with 64+ layouts. 

If there is no `getImageLayout from Vk layout`, implement it

src/nbl/video/CVulkanPhysicalDevice.cpp
```diff
@@ -587,6 +590,30 @@ std::unique_ptr<CVulkanPhysicalDevice> CVulkanPhysicalDevice::create(core::smart
...
+            core::vector<VkImageLayout> copyDstLayouts(hostImageCopyProperties.copyDstLayoutCount);
+            VkPhysicalDeviceHostImageCopyPropertiesEXT layouts = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_HOST_IMAGE_COPY_PROPERTIES_EXT };
+            layouts.copySrcLayoutCount = hostImageCopyProperties.copySrcLayoutCount;
+            layouts.pCopySrcLayouts = copySrcLayouts.data();
+            layouts.copyDstLayoutCount = hostImageCopyProperties.copyDstLayoutCount;
+            layouts.pCopyDstLayouts = copyDstLayouts.data();
+            VkPhysicalDeviceProperties2 layouts2 = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2,&layouts };
+            vkGetPhysicalDeviceProperties2(vk_physicalDevice,&layouts2);
+            using layout_t = asset::IImage::LAYOUT;
+            for (uint32_t i=0u; i<=static_cast<uint32_t>(layout_t::SHARED_PRESENT); i++)
```
Erfan-Ahmadi: very risky, what if someone added a new layout to image layout, then SHARED_PRESENT wouldn't be the last item. 
but anyways, this whole thing could be done with a single for loop

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -630,6 +630,1084 @@ bool ILogicalDevice::nullifyDescriptors(const std::span<const IGPUDescriptorSet:
...
+            validLayout = true;
+            break;
+        default:
+            break;
+    }
+    if (!validLayout)
+    {
+        NBL_LOG_ERROR("Invalid `dstImageLayout`");
+        return false;
+    }
```
Erfan-Ahmadi: vulkan has to make this validation because their flags are simple integers. we have explicit type for LAYOUT.
Just check if it's `!= UNDEFINED` ? no need for switch case

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -630,6 +630,1084 @@ bool ILogicalDevice::nullifyDescriptors(const std::span<const IGPUDescriptorSet:
...
+    {
+        NBL_LOG_ERROR("`dstImage` was not created by this device");
+        return false;
+    }
+    // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkCopyMemoryToImageInfo-flags-parameter
+    if ((flags.value & ~static_cast<decltype(flags.value)>(IGPUImage::EHICF_MEMCPY_BIT)) != 0u)
+    {
+        NBL_LOG_ERROR("Invalid host image copy flags");
+        return false;
+    }
```
Erfan-Ahmadi: again, no need for this, vulkan has to do these types of validations, cause their flags are just uint32_t's
```
typedef uint32_t VkFlags;
typedef VkFlags VkFormatFeatureFlags;
```
you don't have to make this validation. we have explicit enum class type for host image copy. and we shouldn't be resposnible for user doing anything weird with it. (we just translate this to vulkan flag and pass it down)
if vulkan gives validation error about this, we know it's us who did the translation incorrectly and passed incorrect flags

src/nbl/video/ILogicalDevice.cpp
```diff
@@ -630,6 +630,1084 @@ bool ILogicalDevice::nullifyDescriptors(const std::span<const IGPUDescriptorSet:
...
+        default:
+            break;
+    }
+    if (!validLayout)
+    {
+        NBL_LOG_ERROR("Invalid `dstImageLayout`");
+        return false;
+    }
+    // device `pCopyDstLayouts` membership: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkCopyMemoryToImageInfo-dstImageLayout-09060
+    if (!(getPhysicalDevice()->getLimits().hostImageCopyDstLayouts & (1u<<static_cast<uint32_t>(dstImageLayout))))
```
Erfan-Ahmadi: good check against bitflag, but it needs to change to physdev->gethostimagecopyproperties

src/nbl/video/CVulkanPhysicalDevice.cpp
Erfan-Ahmadi: also good to `assert` that whatever you're doing with the bitshifts are less than the size of the storage. (e.g. i < 64)

Erfan-Ahmadi: btw be careful about UNDEFINED because it is `0` in the enum and `1 << 0` is 1. 
so we should never see the first bit set. because that would indicate host image copy is valid to/from undefined layout?

PR #1086, opened 2026-07-30
Device-wide chained scan with lookback

keptsecret:


include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -147,6 +137,100 @@ enable_if_t<spirv::is_pointer_v<Ptr_T>, T> atomicCompSwap(Ptr_T ptr, T comparato
...
+template<typename T>
+NBL_PARTIAL_REQ_TOP(concepts::SignedIntegral<T>)
+struct atomicMin<T NBL_PARTIAL_REQ_BOT(concepts::SignedIntegral<T>) >
+{
+    static T __call(NBL_REF_ARG(T) ptr, T value)
+    {
+        return spirv::atomicSMin<T>(ptr, spv::ScopeDevice, spv::MemorySemanticsMaskNone, value);
+    }
+
+    template<typename Ptr_T> // DXC Workaround
```
Erfan-Ahmadi, line 154: It's best to say exactly what dxc issue we're working around and be more detailed in the comments. in case they fix it and we can upgrade our code.
I think it's something to do with passing groupshared or other address qualifiers here.

I have a question though.
These don't need to be cpp compatible, it's gpu specific, why are we using NBL_REF_ARG and both REQ_TOP and REQ_BOT (hlsl enable_if and c++20 requires). I think BOT is enough?

---
Side note:
Just putting it out there that I don't like TOP/BOT naming 😆 I'd prefer something along the lines of: 
- `NBL_HOST_CONCEPT` + `NBL_DEVICE_CONCEPT` or `NBL_DEVICE_SFINAE`. which is more clear
---

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t WorkgroupSize = 1u << Config::WorkgroupSizeLog2;
+    NBL_CONSTEXPR_STATIC_INLINE uint16_t ItemsPerInvoc = uint16_t(Config::VirtualWorkgroupSize / WorkgroupSize);
+    NBL_CONSTEXPR_STATIC_INLINE uint16_t MaxSpinCount = uint16_t(4u);
+
+    template<class DataAccessor, class ScratchAccessor, class ReductionAccessor, class WorkgroupCounter>
+    void __call(NBL_REF_ARG(DataAccessor) dataAccessor, NBL_REF_ARG(ScratchAccessor) scratchAccessor, NBL_REF_ARG(ReductionAccessor) workgroupReduction, NBL_REF_ARG(WorkgroupCounter) workgroupCounter)
+    {        
+        const uint16_t invocIx = workgroup::SubgroupContiguousIndex();
+        if (!invocIx)
```
Erfan-Ahmadi, line 86: For readability purposes if you want to check if something is equal to zero, it's best to just do  ==0 instead of treating as a boolean

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+        {
+            const uint32_t id = workgroupCounter.atomicAdd(0u, 1u);
+            scratchAccessor.template set<uint32_t, uint32_t>(0u, id);
+            sIsLocked = true;
+        }
+        scratchAccessor.workgroupExecutionAndMemoryBarrier();
+
+        uint16_t workgroupId;
+        scratchAccessor.template get<uint32_t, uint32_t>(0u, workgroupId);
+        scratchAccessor.workgroupExecutionAndMemoryBarrier();
```
Erfan-Ahmadi, line 96: you don't need a `workgroupExecutionAndMemoryBarrier` here.

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+            const uint32_t lastInvocIx = glsl::gl_SubgroupSize() * glsl::gl_NumSubgroups() - 1u;
+            scalar_t lastElem;
+            if (invocIx == lastInvocIx)
+                lastElem = wgDataAccessor.preloaded[wg_data_proxy_t::PreloadedDataCount-1u][Config::ItemsPerInvocation_0-1u];
+
+            if (Exclusive)
+                workgroup2::exclusive_scan<Config,BinOp,device_capabilities>::template __call<wg_data_proxy_t, ScratchAccessor>(wgDataAccessor, scratchAccessor);
+            else
+                workgroup2::inclusive_scan<Config,BinOp,device_capabilities>::template __call<wg_data_proxy_t, ScratchAccessor>(wgDataAccessor, scratchAccessor);
+            scratchAccessor.workgroupExecutionAndMemoryBarrier();
```
Erfan-Ahmadi, line 116: I don't think you need a `workgroupExecutionAndMemoryBarrier` Because you want to access the preloaded array which is thread local.

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+
+            currGroupReduction = wgDataAccessor.preloaded[wg_data_proxy_t::PreloadedDataCount-1u][Config::ItemsPerInvocation_0-1u];
+            if (Exclusive)
+                currGroupReduction = binop(currGroupReduction, lastElem);
+            if (invocIx == lastInvocIx)
+                scratchAccessor.template set<scalar_t, uint32_t>(0u, currGroupReduction);
```
Erfan-Ahmadi, line 122: You only need to do this for the last invocation. 
So maybe you can encompass all of it under `if (invocIx == lastInvocIx)`

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+            scratchAccessor.workgroupExecutionAndMemoryBarrier();
+
+            scratchAccessor.template get<scalar_t, uint32_t>(0u, currGroupReduction);
+        }
+        scratchAccessor.workgroupExecutionAndMemoryBarrier();
```
Erfan-Ahmadi, line 127: I don't think you need a `workgroupExecutionAndMemoryBarrier` here.

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+
+        if (!invocIx)
```
Erfan-Ahmadi, line 129: For readability purposes if you want to check if something is equal to zero, it's best to just do ==0 instead of treating as a boolean

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+        {
+            const scalar_t storeVal = hlsl::mix(Flag_Inclusive, Flag_Reduction, workgroupId > 0u) | currGroupReduction << Flag_Shift;
+            workgroupReduction.atomicExchange(workgroupId, storeVal);
+        }
+
+        if (workgroupId)
```
Erfan-Ahmadi, line 135: if (workgroupId > 0u)

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+    using scalar_t = typename BinOp::type_t;
+    using vector_t = vector<scalar_t, Config::ItemsPerInvocation_0>;   // data accessor needs to be this type
+    using binop_t = BinOp;
+
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t Flag_NotReady = 0;
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t Flag_Reduction = 1;    // workgroup only has local reduction ready
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t Flag_Inclusive = 2;    // workgroup has summed all preceding groups and added to own sum
+    NBL_CONSTEXPR_STATIC_INLINE uint32_t Flag_Mask = 3;
+    
+    NBL_CONSTEXPR_STATIC_INLINE uint16_t Flag_Shift = 2;
```
Erfan-Ahmadi, line 76: Add some comments or documentation that shows the binary operations scalar precision reserves two bits For packing extra flags for example user will only get 30 bit precision if they use uint32

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+            scratchAccessor.workgroupExecutionAndMemoryBarrier();
+
+            scratchAccessor.template get<scalar_t, uint32_t>(0u, currGroupReduction);
+        }
+        scratchAccessor.workgroupExecutionAndMemoryBarrier();
+
+        if (!invocIx)
+        {
+            const scalar_t storeVal = hlsl::mix(Flag_Inclusive, Flag_Reduction, workgroupId > 0u) | currGroupReduction << Flag_Shift;
+            workgroupReduction.atomicExchange(workgroupId, storeVal);
```
Erfan-Ahmadi, line 132: If you look more closer at implementation of glsl:::atomicExchange you'll that there is no memory semantics passed to the spirv instruction. it most likely works on most hardware especially Nvidia, and also depends on how much work you're doing from the `atomic exchange` to the next access to the value. (get)
but this is the place you actually need to be overprotective about. because doing atomics and reading it across different workgroups.
Yes you don't need an execution barrier. but you need a memory barrier with the scope of device did acquires and releases.  so you can either do the barrier yourself. or instead of using the GLSL version. use the spirv one. And actually pass memory semantics. (CC: @devshgraphicsprogramming double check please)

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        if (workgroupId)
+        {
+            bool locked = sIsLocked;
+            scratchAccessor.workgroupExecutionAndMemoryBarrier();
```
Erfan-Ahmadi, line 138: why do workgroups need to sync here?

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+                    }
+
+                    // broadcast id and prepare to do reduction ourselves
+                    if (spinCount == MaxSpinCount)
+                        scratchAccessor.template set<scalar_t, uint32_t>(0u, lookbackIx);
+                }
+                scratchAccessor.workgroupExecutionAndMemoryBarrier();
+
+                locked = sIsLocked;
+                scratchAccessor.workgroupExecutionAndMemoryBarrier();
```
Erfan-Ahmadi, line 181: why workgroupExecutionAndMemoryBarrier here ?

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+                if (locked)
```
Erfan-Ahmadi, line 182: Add comment:
Fall back path: we spun to MaxSpinCount But no previous workgroup had there global reduction ready (Flag_Inclusive).
So we try to do reduction for all previous work groups one by one until we reach Flag_Inclusive

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
Erfan-Ahmadi: I'm a bit confused Why are you using atomic exchange instead of atomic store?

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+                    fallbackDataAccessor.preload();
+                    scalar_t fallbackReduction = workgroup2::reduction<Config,BinOp,device_capabilities>::template __call<wg_data_proxy_t, ScratchAccessor>(fallbackDataAccessor, scratchAccessor);
+                    scratchAccessor.workgroupExecutionAndMemoryBarrier();
+
+                    if (!invocIx)
+                    {
+                        const scalar_t storeVal = hlsl::mix(Flag_Inclusive, Flag_Reduction, fallbackGroupId > 0u) | (fallbackReduction << Flag_Shift);
+                        const scalar_t fallbackPayload = workgroupReduction.atomicMax(fallbackGroupId, storeVal);
+
+                        prevReduction = binop(prevReduction, hlsl::mix(fallbackReduction, fallbackPayload >> Flag_Shift, fallbackPayload > scalar_t(0.0)));
```
Erfan-Ahmadi, line 198: I get why you're doing atomic Max here, You want to take the inclusive one if the original workgroup Finished after we calculated the reduction redundantly
But I don't get the mix here. Why not take the fall back payload all the time?

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+                        if (fallbackGroupId == 0u || (fallbackPayload & Flag_Mask) == Flag_Inclusive)
+                        {
+                            const scalar_t storeVal = Flag_Inclusive | (binop(prevReduction, currGroupReduction) << Flag_Shift);
+                            workgroupReduction.atomicExchange(workgroupId, storeVal);
```
Erfan-Ahmadi, line 202: Be careful again, you need some sort of memory semantics to ensure correct memory barriers when different work groups access the same value atomically.

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+                            scratchAccessor.template set<scalar_t, uint32_t>(0u, prevReduction);
+                            sIsLocked = false;
+                        }
+                        else
+                            lookbackIx--;
+                    }
+                    scratchAccessor.workgroupExecutionAndMemoryBarrier();
+
+                    locked = sIsLocked;
+                    scratchAccessor.workgroupExecutionAndMemoryBarrier();
```
Erfan-Ahmadi, line 212: why workgroupExecutionAndMemoryBarrier here ?

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
```diff
@@ -0,0 +1,267 @@
...
+                if (locked)
+                {
+                    // do reduction for lookbackIx workgroup
+                    uint16_t fallbackGroupId;
+                    scratchAccessor.template get<scalar_t, uint32_t>(0u, fallbackGroupId);
+
+                    wg_data_proxy_t fallbackDataAccessor = wg_data_proxy_t::create(dataAccessor.getInputBufAddr(), dataAccessor.getOutputBufAddr(), fallbackGroupId);
+                    fallbackDataAccessor.preload();
+                    scalar_t fallbackReduction = workgroup2::reduction<Config,BinOp,device_capabilities>::template __call<wg_data_proxy_t, ScratchAccessor>(fallbackDataAccessor, scratchAccessor);
+                    scratchAccessor.workgroupExecutionAndMemoryBarrier();
```
Erfan-Ahmadi, line 191: why workgroupExecutionAndMemoryBarrier here ?

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
devshgraphicsprogramming: or !=0u but either works

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
devshgraphicsprogramming: seconded

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
devshgraphicsprogramming: > These don't need to be cpp compatible, it's gpu specific, why are we using NBL_REF_ARG and both REQ_TOP and REQ_BOT (hlsl enable_if and c++20 requires). I think BOT is enough?

Because in C++20 you need the requires clause on the top, but `enable_if` workaround needs it in the bottom.

Thats the convention we already chose and if we were to change it we'd have to change a lot of other code as well.

devshgraphicsprogramming: > I have a question though.
> These don't need to be cpp compatible, it's gpu specific, why are we using NBL_REF_ARG and both REQ_TOP and REQ_BOT (hlsl enable_if and c++20 requires). I think BOT is enough?

you want to be forward compatible with C++20, because HLSL is going that way, also it keeps the code easy to port to CUDA and SYCL later on.

devshgraphicsprogramming: > It's best to say exactly what dxc issue we're working around and be more detailed in the comments. in case they fix it and we can upgrade our code.
> I think it's something to do with passing groupshared or other address qualifiers here.

Yes @keptsecret please cite/link the DXC issue in the comments for stuff like that (git commit message too)

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
devshgraphicsprogramming: seconded.

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
devshgraphicsprogramming: > Yes you don't need an execution barrier. but you need a memory barrier with the scope of device did acquires and releases. so you can either do the barrier yourself.

Atomics do availability, visibility and domain op automagically so you don't need a memory barrier. IIRC an atomic is kinda a memory barrier in itself **if it has the correct** Acquire/Release semantics.

HOWEVER you need to ensure the Availability of other writes you want to observe to happen before the atomic.

> I'm a bit confused Why are you using atomic exchange instead of atomic store?

Yes this could just be an atomic store, with at a Release and Make Available Semantic  since return value is not used

devshgraphicsprogramming: CC @Fletterio to triple check

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
devshgraphicsprogramming: seconded

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
devshgraphicsprogramming: seconded

include/nbl/builtin/hlsl/concepts/accessors/device_arithmetic.hlsl
```diff
@@ -0,0 +1,65 @@
...
+{
+
+template<typename T, typename V, typename I=uint32_t>
+NBL_BOOL_CONCEPT ArithmeticSharedMemoryAccessor = concepts::accessors::GenericSharedMemoryAccessor<T,V,I>;
+
+template<typename T, typename V, typename I=uint32_t>
+NBL_BOOL_CONCEPT ArithmeticReadOnlyDataAccessor = concepts::accessors::GenericReadAccessor<T,V,I>;
+
+template<typename T, typename V, typename I=uint32_t>
+NBL_BOOL_CONCEPT ArithmeticDataAccessor = concepts::accessors::GenericDataAccessor<T,V,I>;
```
devshgraphicsprogramming, line 20: why are we introducing new aliases?

include/nbl/builtin/hlsl/concepts/accessors/device_arithmetic.hlsl
```diff
@@ -0,0 +1,65 @@
...
+NBL_CONCEPT_END(
+    ((NBL_CONCEPT_REQ_TYPE_ALIAS_CONCEPT)(concepts::accessors::GenericDataAccessor, T, V, uint64_t))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((accessor.atomicMax(index, val)), is_same_v, V))
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((accessor.atomicExchange(index, val)), is_same_v, V))
+);
+#undef val
+#undef index
+#undef accessor
+#include <nbl/builtin/hlsl/concepts/__end.hlsl>
+
```
devshgraphicsprogramming, line 41: shall we generalize more and hoist it out into a `GenericAtomicAccessor` (no arithmetic)

include/nbl/builtin/hlsl/concepts/accessors/device_arithmetic.hlsl
```diff
@@ -0,0 +1,65 @@
...
+// TODO: as counter, maybe just increment 1 always?
```
devshgraphicsprogramming, line 42: nah don't bother

include/nbl/builtin/hlsl/concepts/accessors/device_arithmetic.hlsl
```diff
@@ -0,0 +1,65 @@
...
+#define accessor NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_0
+#define val NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_1
+#define index NBL_CONCEPT_PARAM_T NBL_CONCEPT_PARAM_2
+NBL_CONCEPT_END(
+    ((NBL_CONCEPT_REQ_EXPR_RET_TYPE)((accessor.atomicAdd(index, val)), is_same_v, uint32_t))
+);
+#undef val
+#undef index
+#undef accessor
+#include <nbl/builtin/hlsl/concepts/__end.hlsl>
```
devshgraphicsprogramming, line 59: same could generalize to `GenericAtomicIntAccessor` or `GenericAtomicArithmeticAccessor`

include/nbl/builtin/hlsl/glsl_compat/core.hlsl
```diff
@@ -147,6 +137,100 @@ enable_if_t<spirv::is_pointer_v<Ptr_T>, T> atomicCompSwap(Ptr_T ptr, T comparato
...
+        return spirv::atomicUMax<T>(ptr, spv::ScopeDevice, spv::MemorySemanticsMaskNone, value);
+    }
+
+    template<typename Ptr_T> // DXC Workaround
+    static T __call(Ptr_T ptr, T value)
+    {
+        return spirv::atomicUMax<T>(ptr, spv::ScopeDevice, spv::MemorySemanticsMaskNone, value);
+    }
+};
+}
```
devshgraphicsprogramming, line 211: we actually have a huge problem here and I think this is why I didn't pour that much effort into the GLSL compat functions

1. You don't know the address space so don't know the scope to emit by default - comment at the top of the atomic code block
2. memory semantics are always None

Ideally none of our code should be using the GLSL atomics because they're so bad and imprecise

include/nbl/builtin/hlsl/scan/chained_scan.hlsl
Fletterio: I think a relaxed store is enough, I don't see why release is necessary if there's no other accesses being made before this atomic. Also `workgroupReduction` has a templated get method, but if this store is atomic then that get has to be atomic as well I believe. Idk if this is enforced but I'd rename to `atomicGet` to drive that point through

PR #1088, opened 2026-08-16
Photon Caustics example with Raytracing shaders + minor nabla fixes

Pikachuxxxx:
## Description
The goal of this example is to implement basics caustics using photon mapping, the sample uses raytracing shaders to implement a simple path traces and extends that to building a photon map for tracing caustics in the path tracing pass during scene shading. We use 2 ray tracing passes:
1. Photons tracing for emissive triangles in the scene build from the TLAS/BLAS
2. uses the static photon map to trace radiance during non caustics path.
3. uses reflect/refract for simple metals/glass materials no fancy BxDF evaluation
4. Uses static scene and lights for path tracing the scene

### Photon map based on jensens specular concentration
Photon emission is aimed with a projection map rather than sampling the full hemisphere, photons are cone-sampled toward the bounding sphere of specular geometry. This reduces spreading of photon over the entire scene and focues them over specular emissive geometry.

### Optimizing with spatial hashmap
The gather at shading time is accelerated with a uniform spatial grid, the example build a 16x16x16 spatial hash grid and stores photon using InterlockedAdd during the photonMap build pass. We use this during the gatherPhoton in raytraces scene shading pass to gather photons from nearby cells that cover the range of each photon radius.

### Extra
- Also added debug views to visualize raw photon density.
- Toggle to disable caustics path in the simple raytracer path 
- Reset camera toggle and other photon map config in ImGui

## Nabla Bug Fixes
Fixed some types in IGPURayTracingPipeline shadergroups, this was caught when we used only rcgen/rcmiss/rchit shaders, unlike other samples that used all the shader stages available. 

image
image
image

### DIFF
--- a/include/nbl/video/IGPURayTracingPipeline.h
+++ b/include/nbl/video/IGPURayTracingPipeline.h
@@ -84,7 +84,7 @@ class IGPURayTracingPipeline :  public IGPUPipeline<asset::IRayTracingPipeline<c
                     }
 
                     // https://docs.vulkan.org/spec/latest/chapters/pipelines.html#VUID-VkRayTracingPipelineCreateInfoKHR-flags-03471
-                    if (cached.flags.hasFlags(FLAGS::NO_NULL_CLOSEST_HIT_SHADERS) && !shaderGroup.intersection.shader)
+                    if (cached.flags.hasFlags(FLAGS::NO_NULL_CLOSEST_HIT_SHADERS) && !shaderGroup.closestHit.shader)
                         return {};
                 }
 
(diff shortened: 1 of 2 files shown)


devshgraphicsprogramming:
@Pikachuxxxx thanks for finding the typos.

CC: @keptsecret this likely affects you.

devshgraphicsprogramming:
@CrabExtra I merged the Nabla-Examples-Tests, can you merge this into a different branch, update submodule pointer properly and update it to be not out-of-date (so CI can run) and merge when it passes?

CC @AnastaZIuk

Erfan-Ahmadi:
> @CrabExtra I merged the Nabla-Examples-Tests, can you merge this into a different branch, update submodule pointer properly and update it to be not out-of-date (so CI can run) and merge when it passes?
> 
> CC @AnastaZIuk

done by me

